PR #45, opened 2017-09-23
Component documentation

robgietema:


jensens:
This package is not in core right now? LGTM, just merge if you think its ready (no state label was applied).

PR #46, opened 2017-09-24
Component documentation with examples

VaysseB:
I've added one main file in markdown for each group, some examples (code showcase and live example) for some of theme's components.

Some live examples doesn't work yet, for some obscure reason related to `reduxAsyncConnect`.

The file `docs/create_example.txt` is kind of a _how to_ file about example.


VaysseB:
And normally it should compile this time, I hope :-D

PR #48, opened 2017-10-20
update plone api versions/bootstrap process

thet:
- use plone 5.1
- do not use bootstrap.py but install zc.buildout via pip (see: bootstrap.sh)


tisto:
@thet thanks for the upgrade. Two remarks from my side. I'd keep the bootstrap-buildout.py for people who prefer that approach. I'd also prefer a Makefile over a bash-script, because you do not risk to end up with a bunch of bash-scripts in your root folder.

PR #53, opened 2017-10-22
fix widget proptypes for value

cekk:
Looks like that this component sometimes receive a string (for title field for example), and sometimes an object (for rich text field)


robgietema:
Do you have an example where this happens? For a rich text field it should use the WysiwygWidget.

cekk:
yes, if i remember correctly, it was on edit/create a news item.
I cannot reproduce it right now, because the last master checkout seems broken (some problem with css-loader when webpack builds the bundle).

But, if you don't have this error (the proptypes one), maybe is already solved, and you can ignore this pr.

PR #54, opened 2017-12-12
fix buildout error explained here https://community.plone.org/t/build…

tareqalam:
…out-not-working-couldnt-find-index-page-for-couldnt-find-a-distribution-disabling-non-https-access-to-apis-on-pypi/5069/26

had to add
index=https://pypi.python.org/simple/
in buildout.cfg


tisto:
@tareqalam thanks for the fix!

PR #78, opened 2018-02-14
CI needs to be stable. 

tisto:
Pin Plone version and use plone.restapi release. Pin plone.app.event and plone.namedfile to make buildout work.

### DIFF
--- a/api/buildout.cfg
+++ b/api/buildout.cfg
@@ -1,10 +1,10 @@
 [buildout]
-extends = http://dist.plone.org/release/5.1-latest/versions.cfg
+extends = http://dist.plone.org/release/5.1rc2/versions.cfg
 parts = instance
         plonesite
 extensions = mr.developer
-auto-checkout = plone.restapi
-index=https://pypi.python.org/simple/
+# auto-checkout = plone.restapi
+index = https://pypi.python.org/simple/
 
 [instance]
 recipe = plone.recipe.zope2instance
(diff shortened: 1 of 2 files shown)


api/buildout.cfg
```diff
@@ -50,3 +50,6 @@ site-replace = True
 [versions]
 plone.tiles = 2.0.0b3
 plone.app.mosaic = 2.0.0rc8
+# Those pins are necessary, otherwise buildout will fail with "UnknownExtra: plone.namedfile 4.2.4 has no such extra feature 'blobs'"
+plone.namedfile = 4.2.3
```
tisto: @thet I have to do those pins in every single 5.1 buildout (on Ubuntu). Any idea what causes this? Brown bag release?

ale-rt:
Check which packages requires `plone.namedfile[blobs]` and change it to just `plone.namedfile`. 
E.g. this package should be updated: 
 -https://github.com/plone/plone.app.tiles/blob/1d1f4c87b3b3c86303dacfea7402f73da6d5fb03/setup.py#L72

To know which packages are causing you troubles do something like:
```
$ egrep 'plone.namedfile(.*)blobs' `locate requires.txt`
/home/ale/.buildout/eggs/plone.app.contenttypes-1.2.10-py2.7.egg/EGG-INFO/requires.txt:plone.namedfile[blobs]
...
```

tisto:
@ale-rt thank you for the hint! I did not know the trick with requires.txt yet. Strange that we have a plone.app.tiles dependency in plone.restapi and other buildout files (and that it can be fixed by pinning plone.app.event...

Issue #72, opened 2018-02-08
create- react-app like functionality


nileshgulia1:
plone-react should have  [create-react-app](https://github.com/facebook/create-react-app ) like functionality i.e using a package boilerplate inside project and extend it to zero configuration.The theme should be [pastanaga](https://github.com/plone/pastanaga) which may used as a boilerplate theme so that user can initialise a plone project with zero configuration.
I would love  to work on it.

nileshgulia1:
Here's the approach I have decided to follow:

   1. Trigger `create-react-app` to create a new app with a provided app name.
  1.  Download and install a list of additional packages.
 1.   Replace the default template files with new files as needed.

nileshgulia1:
@tisto  This is the logic I have implemented so far for executing CRA commands for a given app name
ref:https://github.com/facebook/create-react-app/tree/next/tasks
``` javascript 
let appName = process.argv[2] //it  will pull the provided app name out of the parameters of our shell command.
let appDirectory = `${process.cwd()}/${appName}` //store the path to the directory that CRA is going to create.

const run = async () => {
  let success = await createReactApp()
  if(!success){
    console.log('Something went wrong while trying to create a new React app using create-react-app'.red)
    return false;
  }
  await cdIntoNewApp()
  await installPackages()
  await updateTemplates()
  console.log("All done")
}
run()
const createReactApp = () => {
  return new Promise(resolve=>{
    if(appName){ //If appname provided then execute CRA command
      shell.exec(`create-react-app ${appName}`, () => {
        console.log("Created react app")
        resolve(true)
      })

    }else{
      console.log("\nNo app name was provided.".red)
      console.log("\nProvide an app name in the following format: ")
      console.log("\ncreate-plone-react-app ", "app-name\n".cyan)
        resolve(false)
    }
  })
}
```

nileshgulia1:
@tisto Can we use fork of `react-scripts` instead of ejecting the whole app so that we can have only those configurations  which we need without loosing any extra configs?

tisto:
@nileshgulia1 yes. I guess that is the idea. Our fork of react-scripts should contain the "library parts" of plone-react, the webpack config, the server side rendering, etc.. What do you think @robgietema?

nileshgulia1:
[...]

PR #87, opened 2018-02-25
generation of plone-react-app

nileshgulia1:
Related #72 
This is the very first implementation of generating plone-react app from CRA. Right now it just executes the script named `create-plone-app <app name>`.


robgietema:
This functionality needs to be in a separate package so you can install this global.

PR #102, opened 2018-03-12
Datakurre acceptance tests setup

tisto:


datakurre:
@tisto I only now noticed that Start Webpack -keyword accepts custom check parameter. So everything is great and I can finish this on Tuesday morning.

api/buildout.cfg
```diff
@@ -29,6 +29,7 @@ eggs =
     robotframework-debuglibrary
     robotframework-react
     robotframework-seleniumlibrary
+    robotframework-selenium2library
```
tisto: @datakurre I am happy to make a robotframework-react release that depends on robotframework-seleniumlibrary.

datakurre: @tisto At the end the issue was not a missing dependency, but ReactLibrary expected to find imported instance of "Selenium2Library". Probably it should look for an instance of SeleniumLibrary at first and then look for Selenium2Library (since it takes some time that all Selenium2Library users have moved back to SeleniumLibrary).

datakurre: @tisto Also, I did not get ReactLibary working yet. At first, it didn't do anything (I was missing reducer-argument). When I called it with reducer=userSession, it failed with ``JavascriptException: Message: TypeError: window.appStore is undefined``.

tisto: @datakurre done: https://github.com/kitconcept/robotframework-react/pull/9

datakurre:
@tisto This seems to pass now. Review would be appropriate. This includes

* pybot and robot-server for api
* Robot test setup with examples
* Docker for API
* Jupyter notebook setup

This still needs

* confirmation that Jupyter works for you and possible fixes (I have a little different setup due to NixOS)
* figure out the best practices to use these tools and document
* integrating tests to yarn run in package.json

datakurre:
@tisto Please, read README changes https://github.com/plone/plone-react/pull/102/files#diff-04c6e90faac2675aa89e2176d2eec7d8 for a summary

It's good to know that both docker image and Travis-CI setup are based on the requirements.txt generated by api/buildout.cfg

For Jenkins you could choose to use either buildout created pybot or pip installed pybot (as created in travis.yml).

For local test development it's easiest to use docker, because it has the test fixture set up. If you want to use buildout-created ``robot-server`` you need to check the environment variables for the environment from Dockerfile.

Running all tests with ... pybot ... tests/ will always manage both backend and frontend.

Running individual test case files ... pybot ... tests/test_login.robot assumes that backend and frontend are already running and backend is manually cleaned up after tests.

Issue #100, opened 2018-03-12
Adapting eslint config to prettier (was: Linting command files filter is wrong)


sneridagh:
It doesn't have the correct filter, it misses to catch the *.jsx files.

Try:
`./node_modules/eslint/bin/eslint.js -c .eslintrc src/**/*.jsx`

tisto:
We are currently using the Airbnb setting for eslint. Though, we should use an eslint config that does not conflict with prettier. There are lots of options, such as: https://github.com/prettier/eslint-config-prettier

We have to try them out and find one that works for us...

nileshgulia1:
I'm on it :)

sneridagh:
@nileshgulia1 thx, we will take a look!

nileshgulia1:
Fixed by #128

PR #107, opened 2018-03-12
configured prettier with eslint

nileshgulia1:
Related #100
 ` npm run eslint-check ` now shows
 `No rules that are unnecessary or conflict with Prettier were found.`

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,5 +1,8 @@
 {
-  "extends": "airbnb",
+  "extends": ["airbnb",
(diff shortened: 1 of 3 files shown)


robgietema:
Why is this a separate command, this can just be part of "yarn lint" right?

nileshgulia1:
@robgietema  Now we can use it with yarn lint  :+1:

robgietema:
Fixed by #128

Issue #140, opened 2018-05-16
Handle environment variables 


nileshgulia1:
Figure out how to handle environment variables for `plone-react` as package. Will define a new `env.js` file to read env variables in our app generator. Thinking of using `dotenv` for purpose. What do you think @robgietema @sneridagh ?

nileshgulia1:
closed via #142

PR #142, opened 2018-05-17
configured env. variables and paths 

nileshgulia1:
related #140 . @robgietema These are the common configs from ejected create-react-app.


configs/paths.js
```diff
@@ -0,0 +1,121 @@
...
+  ownNodeModules: resolveOwn('node_modules'), // This is empty on npm 3
+};
+
+// detect if template should be used, ie. when cwd is react-scripts itself
+const useTemplate =
+  appDirectory === fs.realpathSync(path.join(__dirname, '..'));
+
+checkForMonorepo = !useTemplate;
+
+if (useTemplate) {
```
nileshgulia1: I guess  in place there will be our Pastanaga template files.

nileshgulia1:
I guess should mere this now.

sneridagh:
@nileshgulia1 I just reviewed your PR, I haven't had the chance to talk with @robgietema yet, but I think we still have different visions on what we want to accomplish. I'll try to resume it, and again without having talked yet with Rob: We don't want a clone of CRA underneath plone-react-app, we want our own setup like plone-react has now, but with the feature set of CRA in terms of skeleton building, webpack/babel/eslint hidden configs, etc... not in terms of what it does at the app level.

So I don't get why you introduced some depencencies (the same that CRA has), like SVGR etc...

Let's talk about it as soon as possible. I've been busy in ReactEurope these days, maybe we should sync again next week.

/cc @tisto

nileshgulia1:
@sneridagh Actually, In the meeting earlier @robgietema and I discussed , First task should be making `plone-react` as library like react-scripts. And regarding the app layout we have discussed nothing much so I thought we should go with the same layout(skeleton building) which CRA has. Ofcourse the feature set and packages should be ours(plone-react).

> So I don't get why you introduced some depencencies (the same that CRA has), like SVGR etc...

Yes, this was done by mistake I accidentally installed these packages, assuming they may be required in future.

PR #151, opened 2018-05-24
Configured start script

nileshgulia1:
This is the custom start script implementation that starts our `dev` server. May be it needs some more refactoring in npm scripts. As discussed with @sneridagh, we have this script hidden in CRA, so I am figuring out ways to hide this script.


nileshgulia1:
messed up branches , will close this and make another one

PR #157, opened 2018-05-31
Move combineReducers to the store creation level. This will ease the …

sneridagh:
…extensibility of them in Plone-React apps.

### DIFF
--- a/src/reducers/index.js
+++ b/src/reducers/index.js
@@ -38,7 +38,7 @@ import workflow from './workflow/workflow';
  * @param {Object} action Action to be handled.
  * @returns {Object} New state.
  */
-const ploneReducers = {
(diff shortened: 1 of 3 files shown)


src/reducers/index.js
```diff
@@ -39,7 +38,7 @@ import workflow from './workflow/workflow';
  * @param {Object} action Action to be handled.
  * @returns {Object} New state.
  */
-export default combineReducers({
+const ploneReducers = {
```
robgietema: Better just call this reducers and not prefix with plone

sneridagh: ok

sneridagh: @robgietema done.

PR #160, opened 2018-06-05
Scripts/build

nileshgulia1:


nileshgulia1:
Tested and working after a local checkout of `plone-react` and using it on another project via `yarn link`

PR #167, opened 2018-06-16
Update .nvmrc version

cekk:
I have this error when using the previous version:

    error @plone/plone-react@0.5.0: The engine "node" is incompatible with this module. Expected version "~8.11.2".
    error Found incompatible module

We could also set _lts_ in this file, which is 8.11.3


tisto:
@cekk I already set LTS = 8.11.x in package.json. Does nvm automatically pick up the version in a local .nvmrc file? I think we can all agree that plone-react should run on the latest node LTS. We shouldn't be too picky with the minor version though (expect on CI, where we should pin exactly one version and always update manually to catch possible regressions).

pnicolli:
Yes, nvm does read the `.nvmrc` file if you run just `nvm use` inside the project, instead of writing the version number. I agree it should run on the latest LTS, the `.nvmrc` file could be `lts` as @cekk suggested, or it can be just `8` or `8.11` which is very similar to the declaration in the `package.json` file.
If we still want to be super precise for CI purposes, then I think we should definitely pin the `.nvmrc` file to the latest 8.11 so that it works if used for development.

PR #186, opened 2018-06-23
[WIP] Improve reference widget

cekk:
- with https://github.com/plone/plone.restapi/pull/539 plone.restapi request, we have the _widget_ property for relation widget
- with that property we can select the right widget in plone-react
- the widget allow to navigate through folders like gdrive widget

missing tasks
- [ ] tests for new components
- [ ] clearing the navigator component
- [ ] review the interface with an ux designer
- [ ] retrieve the title of current selected folder in the navigation component
- [x] default query when open the widget

### DIFF
--- a/api/.gitignore
+++ b/api/.gitignore
@@ -16,6 +16,7 @@ include
 lib
 lib64
 local
+local.cfg
 node_modules
 package.json
 parts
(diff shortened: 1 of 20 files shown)


sneridagh:
There is a basic reference widget and another more sophisticated it's on its way, being taking cared by the RT folks.

PR #207, opened 2018-06-30
starter page boilerplate

nileshgulia1:
screenshot from 2018-06-26 18-00-10


robgietema:
That the default page content  doesn't show up is because the backend doesn't return the correct value (yet). A fix is being made in the backend to fix this, so we shouldn't do this in the frontend.

PR #211, opened 2018-07-04
Anchor plugin

sneridagh:


tisto:
@gforcada any idea why we see those messages from mr roboto? @robgietema has the maintainer role on the Plone github org (same as me) and he signed the contributor agreement. Also this does not seem to happen before.

gforcada:
He did the commits with an email that are not known to github

tisto:
@gforcada we checked that and he is using his gmail email for both commits and github. Right @robgietema ?

gforcada:
@tisto yes and no :smile: if you look at the API data https://api.github.com/repos/plone/plone-react/pulls/211/commits you can spot (it took me a while) a difference between Rob and Victor commits: on Rob commit the ``author`` and ``committer`` top level field is empty, while on Victor commits both of them are filled.

I see that there is a ``commit`` field and inside it again ``author`` and ``committer``, so one is deprecated already? :thinking:

PR #214, opened 2018-07-09
Adjust widgets to load batched vocabularies to be compliant with the upcoming plone.restapi 4.0.0

davisagli:
I started adjusting the widgets to use the batched vocabularies from this pending plone.restapi PR: https://github.com/plone/plone.restapi/pull/550

There are a few places where I got stuck and would be happy for advice from @robgietema:
1. The Dropdown from semantic-ui doesn't appear to have a hook for requesting more batches as the user scrolls. Do we need to build our own dropdown or do you see a workaround?
2. It seems like the logic for loading from a batched API endpoint should be factored into a higher-order component or something, but I wasn't sure how to wire this up to redux in a generic way. 

So far I only worked on the ArrayWidget, so once we answer these questions we also need to work on the SelectWidget.


davisagli:
I didn't look yet, because I wasn't sure how important it was to keep using semantic-ui for this. There's also the tradeoff to consider between implementing it ourselves (more work now) vs. maintaining code using someone else's library (potentially more work later)

sneridagh:
@davisagli I made some tests with react-virtualized-select with good results and stressing it (several components with 15000 items in the same view).

However, the virtualized component expects to return all the items on the first async call, and the batching is not required, since the Virtualized component does all the job for you...

I've already talked with Rob on having non-SemanticUI React components in the default set of Plone-React. We both agreed that Plone-React should have the best breed of default components available out there, not only React SemanticUI ones. As I said, I toyed with react-select (and other components) and integrated them into the SemanticUI theming engine with good results.

I will talk to Rob about this in the Costa Brava Sprint next week, and try to reach an agreement.

This is the code of my tests:

```
import React, { Component } from "react";
import "react-select/dist/react-select.css";
import "react-virtualized-select/styles.css";
import Select from "react-virtualized-select";

const getOptions = input => {
  return fetch(
    `http://localhost:8080/Plone/@vocabularies/plone.app.vocabularies.AvailableContentLanguages?q=${input}`,
    {
      headers: {
        Accept: "application/json; charset=utf-8",
        Authorization: "Basic YWRtaW46YWRtaW4="
      }
    }
  )
    .then(response => {
      return response.json();
    })
    .then(json => {
      return {
        options: json.items.map(item => ({
          label: item.title,
          value: item.token
        })),
        complete: false
      };
    });
};

class App extends Component {
  render() {
    return (
      <div className="App">
        <Select async loadOptions={getOptions} />,
      </div>
    );
  }
}

export default App;
```

/cc @robgietema @tisto

sneridagh:
@tisto @davisagli @robgietema I think this is also ready for merge. It supports the changes done in 

https://github.com/plone/plone.restapi/pull/550
https://github.com/plone/plone.restapi/pull/714/files
https://github.com/plone/plone.restapi/pull/716

This makes ready Volto for the breaking changes of the upcoming p.restapi 3.0.0

PR #232, opened 2018-07-22
Run tests on Travis on Chrome.

tisto:


tisto:
@datakurre after debugging for ages I just found out that the problem is running the current login tests (without any amendments) on Chrome. I already debugged a lot on Travis but could not find out what's wrong. Any ideas?

tisto:
@datakurre FYI: I need chrome because the current way of testing DraftJS with robot does not work with Firefox.

tisto:
@datakurre oh, and of course the tests pass locally without any issues...

datakurre:
Have you tried logging browser’s console to see possible errors?

How crazy would it be to make gzipped tarball our of log.html and selenium-screenshots and dump it as base64 encoded into travis log?

tisto:
@datakurre good idea. I thought about the base64 encoding as well but never dared to try it. :)

datakurre:
It’s only a single png. You should not feel bad about it.

datakurre:
@tisto Did you found the issue?

PR #236, opened 2018-07-23
Added 404 page.

robgietema:


ebrehault:
The way to bubble up the error to the server layer is super simple. Cool!

In the future we can probably do the same to handle 301 and 302 status too.

PR #243, opened 2018-07-26
Add AlbumView

davilima6:
I'd like to include this example of view customization, not sure where:

```
/**
 * MyView view component.
 * @module components/theme/View/MyView
 */

import React, { Component } from 'react';
import { Card } from 'semantic-ui-react';

import AlbumView from './AlbumView';

/**
 * MyView view component class.
 * @class MyView
 * @extends Component
 */
export default class MyView extends Component {
  /**
   * Property types.
   * @property {Object} propTypes Property types.
   * @static
   */
  static propTypes = {
    content: PropTypes.shape({
      title: PropTypes.string,
      description: PropTypes.string,
      items: PropTypes.arrayOf(
        PropTypes.shape({
          title: PropTypes.string,
          description: PropTypes.string,
          url: PropTypes.string,
          image: PropTypes.object,
          '@type': PropTypes.string,
        }),
      ),
    }).isRequired,
  };

  /**
   * Template helper
   * @method renderCardDescription
   * @param {Object} item List item
   * @returns {string} Markup for list's content
   */
  renderCardDescription(item) {
    return (
      <Card.Description
        content={item.description_myview_overview || item.description}
      />
    );
  }

  /**
   * Render method.
   * @method render
   * @returns {string} Markup for the component.
   */
  render() {
    return (
      <AlbumView {...this.props} renderCardDescription={this.renderCardDescription} />
    );
  }
}
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,256 +4,257 @@
 
 ### Changes
 
-- Fix tags layout @robgietema @jaroel
-- Fix imports of views, widgets and tiles @robgietema @jaroel
+* Fix tags layout @robgietema @jaroel
+* Fix imports of views, widgets and tiles @robgietema @jaroel
 
 ## 0.9.3 (2018-08-10)
 
 ### Changes
 
-- Fix logo import path @robgietema @jaroel
+* Fix logo import path @robgietema @jaroel
 
 ## 0.9.2 (2018-08-10)
 
 ### Added
 
-- Automatic customization imports for images @robgietema @jaroel
+* Automatic customization imports for images @robgietema @jaroel
 
 ## 0.9.1 (2018-08-10)
 
 ### Added
 
-- Automatic customization imports @robgietema @jaroel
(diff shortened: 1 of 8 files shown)


sneridagh:
@davilima6 I'm still not 100% sold on render props... rather passing functions to the props, is not better to pass full stateless components? I know, at the end is the same, but having them separated in their own module, makes it more manageable (and overridable!).

What do you think?

sneridagh:
@robgietema any insight? ^^

davilima6:
@sneridagh, I see your point. I also thought of changing <AlbumView> to accept simpler props like "contentHeader", "cardHeader", "cardDescription", "cardExtra" which by their turn are subcomponents using dot notation, like <AlbumView.ContentHeader>, <AlbumView.CardHeader> etc.

It would be more similar to Semantic UI React and we wouldn't need to pass an argument (like "item", or "items") in the JSX caller, since we could fetch from props. Not sure if there would be any downsides...

PR #246, opened 2018-07-26
Add helper BodyClass for appending classes to the body tag from View …

sneridagh:
…components.

This might be handy for support different flavours on same view/s or other handy use cases. It will also be aware to the full body of the existance of the toolbar, which will lead to better CSS to handle it.

This makes obsolete the use of Helmet in Plone-React for managing body classes. Mixing both will not work at all. React Helmet was not going to implement this, because backwards compat, so I used the same approach (using react-side-effects) and composing the body tag incrementally. It also works on SSR without problems.

I also removed all uses of Helmet for the body classes. It might be handy to have it around to manage other things (like the title) for now.

I also fixed some wrong use of Helmet (calls to .peek() that only should be used in tests) which is discouraged in Helmet docs. It turns out that we weren't even using it.

I think it's time to start a "developer guide", might be handy from now on...

What do you think @robgietema ??

### DIFF
--- a/src/components/theme/View/View.jsx
+++ b/src/components/theme/View/View.jsx
@@ -5,7 +5,6 @@
 
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
-import Helmet from 'react-helmet';
 import { connect } from 'react-redux';
 import { Portal } from 'react-portal';
 import { Link } from 'react-router';
@@ -26,7 +25,7 @@ import {
   Workflow,
 } from '../../../components';
 import { listActions, getContent } from '../../../actions';
-import { getBaseUrl } from '../../../helpers';
+import { BodyClass, getBaseUrl } from '../../../helpers';
 
 @injectIntl
 @connect(
@@ -219,12 +218,12 @@ export default class View extends Component {
 
     return (
       <div id="view">
-        <Helmet
-          bodyAttributes={{
-            class: RenderedView.displayName
+        <BodyClass
+          className={
+            RenderedView.displayName
               ? `view-${RenderedView.displayName.toLowerCase()}`
-              : 'view-undefined',
-          }}
+              : 'view-undefined'
+          }
         />
 
         <RenderedView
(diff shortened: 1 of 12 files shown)


src/components/theme/View/View.jsx
```diff
@@ -219,12 +218,12 @@ export default class View extends Component {
...
-        <Helmet
-          bodyAttributes={{
-            class: RenderedView.displayName
+        <BodyClass
+          className={
+            RenderedView.displayName
               ? `view-${RenderedView.displayName.toLowerCase()}`
-              : 'view-undefined',
-          }}
+              : 'view-undefined'
```
davilima6: I'd have this as `view-default`, or similar, to avoid confusion with the js reserved value

src/helpers/Html/Html.jsx
```diff
@@ -28,11 +30,7 @@ import serialize from 'serialize-javascript';
 export const Html = ({ assets, component, store }) => {
   const content = ReactDOM.renderToString(component);
   const head = Helmet.rewind();
-  // Workaround for testing
-  // Otherwise we get TypeError: _reactHelmet2.default.renderStatic is not a function
-  const helmet =
-    process.env.NODE_ENV === 'production' ? Helmet.renderStatic() : null;
-  const bodyAttrs = helmet && helmet.bodyAttributes.toComponent();
+  const bodyClass = join(BodyClass.rewind(), ' ');
```
davilima6: :tada:

src/components/theme/View/View.jsx
sneridagh: In fact it's what it is... there's some components that are not getting the displayName well, we have to investigate this... :(

It felt wrong to me to append a class that doesn't really depict the real intention of having it there... so I ended up by removing it (set it to null) so nothing shows now.

davilima6: I think it's good to have a default class for unexpected cases, it seems to make styling fallback cases more explicit than an absent class... I don't like `undefined` because it kind of points to internals of javascript and may look like the system didn't handle some error, hence my suggestion for `view-default` or similar.

But this is minor, if you're not sure you agree we can revise this in a sprint after we collect more theming experience and also ask others. Thanks!!

PR #251, opened 2018-07-31
Testing pastanaga editor on with guillotina

bloodbare:
@robgietema it needs pulishment and on edit the formData does not work properly with guillotina as the tiles_layout and tiles are on the `'guillotina_cms.interfaces.tiles.ITiles'` property.


sneridagh:
Looks good to me.

@bloodbare I made some fixes and applied prettier, etc... take a look into my commit.

However, having to support both API has its drawbacks, of course. We need to find out a way to do the branching when needed (this whole PR is full of these) in a more elegant (and readable) way. Probably a matter of getting together (again) and think on something.

For now, I quite like the new API discriminator in the config.

The acceptance tests are failing right now, don't really know why... I will take a look later.

/cc @robgietema @tisto Ramon needs this for a project, he asks to merge right away, if you don't have any objections I will do it as soon as the tests are green.

Again, it's amazing to see how it already works in both APIs...

tisto:
@bloodbare @sneridagh seems there is a robot test failure on Travis. Do we have a robot setup for Guillotina?

bloodbare:
The error is due to removing the initial formData value on From creation on the Add Form. On the constructor of the Form there is already code to deal with the use case of empty formData. I think is an error to provide default formData on the Add component and I can't reproduce it with guillotina backend.

@vangheem is working on tests fixtures for robot framework on Guillotina so we can also test with guillotina.

vangheem:
Does this PR require guillotina robot tests before merge?

@bloodbare have you looked at the current failures?

vangheem:
It also seems to me like we should include docs and docker configs to run guillotina_cms backend instead of plone if you want directly from dev here.

Like,

```
If you want to run guillotina instead:

docker-compose -f api/docker-compose-guillotina.yml up
```

bloodbare:
@vangheem the current failure is a bug on plone-react code with plone backend on a functional test. On creation of an Form component  inside the Add component there was a code to initialize tiles formData, which does not make sense as constructor of Form component has already the case if tiles is empty. @robgietema is aware about it and is looking for a fix.

bloodbare:
@vangheem whould be great to be able to run the acceptance tests (robot) before we merge

vangheem:
@bloodbare I've done a bit of work on this if you look at my latest commit.

- guillotina branch with robot library. The library re-uses pytest fixtures(we can think about doing this in a more clever way if we want...)
- some docs on how to run the tests
- g-tests folder for robot tests(just a copy of existing tests but configured with guillotina)

what I don't know what to do with at this point:
- I don't think we can install python 2 & 3 at the same time on travis(it'll do 2 separate runs) -- maybe we need to detect python version and on python 3, just run guillotina tests.

I will work some more on it but it'd be great to yell at me if I'm going down a wrong route here.

vangheem:
@bloodbare updated with docker-compose variant to run tests. They work locally and pass but I'm having configuration issues with travis and docker compose dealing with ports. Would you be able to take a look or do you have ideas?

```
ERROR: for redis  Cannot start service redis: driver failed programming external connectivity on endpoint redis (56abd02b8b66215137848e200ee4aeeadf881f4a18138faba7947c5f0428612f): Error starting userland proxy: listen tcp 0.0.0.0:6379: bind: address already in use
```

vangheem:
Sweet, all green!

PR #258, opened 2018-09-20
One robot to rule

bloodbare:


tisto:
@datakurre same feeling here. We can merge and then optimize. If statements make the RF tests hard to follow and debug. Therefore I think we would be better of separating entire chunks of functionality into backend-specific keywords, e.g.:

```Test Setup
Run Keyword If   '${API}' == 'Plone'  Test setup plone
Run Keyword If   '${API}' == 'Guillotina'  Test setup guillotina
```

This will make things more clear and force us to not write hard-to-maintain RF tests.

bloodbare:
@tisto We can try easier once its merged to couple them again

PR #264, opened 2018-09-24
Docker fixing base url

bloodbare:


sneridagh:
@bloodbare same as the other, right? We'll have to port it / fix it in Volto.

PR #265, opened 2018-09-24
Revert "Base fix"

bloodbare:
Reverts plone/plone-react#263


robgietema:
Why do you want it reverted?

sneridagh:
@bloodbare We'll have to port it / fix it in Volto, right?

PR #266, opened 2018-09-26
Fixing guillotina tiles

bloodbare:


src/components/manage/Tiles/Text/Edit.jsx
```diff
@@ -201,6 +201,7 @@ export default class Edit extends Component {
             );
             const blockType = currentContentBlock.getType();
             if (!includes(listBlockTypes, blockType)) {
+              let ii = this.props.index;
```
robgietema: this is not needed right?

bloodbare: Yups debugging info

PR #278, opened 2018-11-07
Websockets

bloodbare:


robgietema:
I've merged the websockets functionality with #306, the rest is the related to the collaborative editing which is experimental at this stage, will close PR till we have time to work on this.

Issue #244, opened 2018-07-26
Download file links


robgietema:
Currently when creating a link to a file in the editor you go to the file object when you click on the link. The link point to /@@download/file to download the file.

robgietema:
replace(/a href=\"([^"]*\.[^"]*)\"/, `a href="${settings.apiPath}$1/download/file"`)

robgietema:
Fixed in #308

PR #294, opened 2018-11-10
Download file links #244

mikejmets:
Change file links in document text to point to the download view on the backend


robgietema:
Fixed in #308

PR #311, opened 2018-11-11
Factor out 'autologin as' RF keyword.

tisto:


tisto:
@bloodbare the guillotina tests fail, no idea why. Would you mind having a look?

PR #314, opened 2018-11-11
Rf news item

tisto:


tisto:
@bloodbare I skipped a few tests (news item, folder) for Guillotina because they fail (most likely because Guillotina does not have those content types). Though, I don't understand why the page test fails.

PR #579, opened 2018-12-02
Upgrade to Node 8.14.0

tisto:


Dockerfile
```diff
@@ -1,7 +1,7 @@
-FROM node:8.11.4-slim
+FROM node:8.14.0-slim
 
 RUN apt-get update -y
-RUN apt-get install -y libpng12-dev
+# RUN apt-get install -y libpng12-dev
```
tisto: @bloodbare any particular reason why we install libpng12-dev here? When I comment it out all tests still pass. If we want to keep that dependency we should use "RUN apt-get install -y libpng-dev" instead.

Issue #580, opened 2018-12-02
Upgrade to Babel 7


tisto:
I started to look into docz for our Volto docs and I ran into issues with Babel 6. Before I can proceed, we need to upgrade to Babel 6. Any pointers or anybody who can help with this?

nileshgulia1:
@tisto We could use `babel-upgrade` tool to resolve the issue of stage preset deprecation. We can use this complete migration [guide](https://babeljs.io/docs/en/v7-migration).

tisto:
@nileshgulia1 this is actually what I tried already. Did not work so far. Maybe you could look into this?

sneridagh:
@tisto @nileshgulia1 as we talked already, just to write it here, let's postpone it until Razzle fully supports it and it is released.

PR #584, opened 2018-12-12
Babel upgrade v7

nileshgulia1:
Addresses: #580


package.json
```diff
@@ -107,12 +119,13 @@
...
     "postcss-flexbugs-fixes": "4.1.0",
     "postcss-loader": "3.0.0",
+    "prepend-http": "2.0.0",
     "prettier": "1.14.3",
     "promise-file-reader": "1.0.2",
     "query-string": "6.1.0",
     "raven": "2.6.4",
     "raven-js": "3.27.0",
-    "razzle": "2.4.0",
+    "razzle": "3.0.0-alpha.0",
```
tisto: @nileshgulia1 we might want to wait for a final version of Razzle here. I guess this upgrade is necessary for Babel 7? Or does Razzle 2.4 work with Babel 7 as well?

src/components/manage/Sharing/__snapshots__/Sharing.test.jsx.snap
```diff
@@ -80,9 +80,7 @@ exports[`Sharing renders a sharing component 1`] = `
             </th>
             <th
               className=""
-            >
-              Contributer
-            </th>
+            />
```
tisto: @nileshgulia1 it seems there are a few unrelated changes in the PR (tests, CSS, SVG, Icons). Or am I missing something here?

package.json
nileshgulia1: @tisto We are having babel 7 presets with razzle v3 . But since we're using razzle's presets for babel, for v7 we have to use razzle v3, otherwise we may opt out using razzle's babel configs.

src/components/manage/Sharing/__snapshots__/Sharing.test.jsx.snap
nileshgulia1: @tisto  Yes there are some updated snapshots, and few svg's for default semantic theme, which added when working on `less v3.9.0`  and `semantic-ui-less` upgrades. Although some SVG's and icons from the default semantic theme can be removed.

sneridagh:
@nileshgulia1 Crap, look at #585 ... I already worked on that this week :(

I thought that you only were taking care of the Babel 7 testing... try to maintain your PRs as atomic as possible, so only one thing is accomplished. Then reverting or backtracking changes is also easier. You can tell by taking a look into your changelog entry... currently doesn't mention the update (which could be huge) and it's mixed with another biggie, which is Babel 7 upgrade.

I even already made a review on the semantic UI and theme upgrades and I spotted several caveats that I had to fix on my PR. I even made a quick review on the whole resulting live site and it's fine, so I would better use my PR and merge it already. @robgietema already reviewed it today. That doesn't mean that you made a great work, in fact, it's great that you already looked into that and know the innards of semantic and how it works. Keep up the good work!

@tisto regarding the upgrade, I am a bit reluctant if it's required to rely on a super recent Razzle alpha version, I won't merge it (yet) :( 

@nileshgulia1 might be a way to upgrade without upgrade razzle?

nileshgulia1:
@sneridagh would close this ticket?

sneridagh:
@nileshgulia1 no, not yet. Let's wait for Timo to weigh in. Please, remove/revert all the things related with the less and semantic upgrade as well (package.json, default theme etc)

tisto:
If we go with styleguidist instead of docz I don't need the babel 7 upgrade urgently. Would be nice to have though. Maybe we should keep this PR open until we merged all the stuff that it contains.

nileshgulia1:
The builds will pass after we get latest `less` and `semantic-ui` updates through #585 .

sneridagh:
Superseeded by https://github.com/plone/volto/pull/711

sneridagh:
@nileshgulia1 thanks a lot!!, it saved me a lot of time, just pushed https://github.com/plone/volto/pull/711

PR #590, opened 2019-01-09
Integration improvements

vangheem:
- work with mr.developer
- add alias to `@plone/volto-original` to be able to reimport overridden components
- add alias to `@package` for current dev package to be able to import modules in the current project from customizations folder
- be able to override error views based on status code of error message(previously only had 404 view)

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -29,7 +29,9 @@
     "import/prefer-default-export": "off",
     "react/prefer-stateless-function": 0,
     "import/no-unresolved": 0,
-    "import/extensions": 0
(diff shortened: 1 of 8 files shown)


vangheem:
I'll also make a PR for the starter project

sneridagh:
@vangheem Cool!! Nice additions! :)

@vangheem @robgietema We have to remember to update the submodule pointer in create-volto-app package, and make a new release (if required). That reminds me that we have to start the developer docs ASAP, there's too much inferred knowledge out there.

Issue #291, opened 2018-11-10
RSS Feed Block


robgietema:

nileshgulia1:
@polyester I was doing this an my internship . Can you please elaborate how the Tile should look and what function should it do.

polyester:
In my understanding:
- have one (and only one) RSS feed url. I haven't actually seen any real-life usecases for authenticated/protected RSS-feeds ; so for starters I would assume public feed.
- setting for number of items to display
- setting for refresh rate
- setting for title to display above the list (non-essential)

and then the tile should basically display whatever it gets fed. 

The Mosaic tile perfectly covers that functionality in python-based tiles, maybe that helps:
[https://github.com/plone/plone.app.standardtiles/blob/master/plone/app/standardtiles/rss.py](https://github.com/plone/plone.app.standardtiles/blob/master/plone/app/standardtiles/rss.py)

tisto:
@polyester just to be clear. As an editor, you want to be able to embed content from a remote website by adding the RSS-feed URL. The items of the RSS-feed should be displayed in a Volto page as a content tile, correct?

nzambello:
We have https://github.com/RedTurtle/volto-rss-block

tisto:
@nzambello nice. From the package description I couldn't fully figure out how this works. Is it possible to add a screenshot to the description?

sneridagh:
Closing this one, after the successful GSOC by @ziming-yuan :

https://maurits.vanrees.org/weblog/archive/2024/11/ziming-yuan-how-a-beginner-built-an-rss-feed-provider-in-volto

https://github.com/collective/volto-rss-provider
https://github.com/collective/rss-provider
https://github.com/collective/rss-receiver
https://github.com/collective/volto-rss-receiver

PR #595, opened 2019-01-20
[WIP] Rss tile

nileshgulia1:
#291 Right now parsing is done through `rss-parser`. Looking forward to render a correct UI based on the contents.


src/components/manage/Tiles/Rss/Edit.jsx
```diff
@@ -0,0 +1,288 @@
...
+            className="tile-delete-button"
+          >
+            <Icon name={trashSVG} size="18px" />
+          </Button>
+        )}
+      </div>
+    );
+  }
+}
+const mapStateToProps = state => {
```
sneridagh: Not used? See later comment as well.

src/components/manage/Tiles/Rss/Edit.jsx
```diff
@@ -0,0 +1,288 @@
...
+      url: this.state.url,
+    });
+    if (this.state.url.match('.rss')){
+    (async () => {
+      const parser = new Parser();
+      const myfeed = await parser.parseURL(this.state.url);
+      this.setState({
+        feed: myfeed,
+      });
+      console.log(myfeed, 'this is of type ' + typeof(this.state.feed.items))
```
sneridagh: Ping!

src/components/manage/Tiles/Rss/View.jsx
```diff
@@ -0,0 +1,84 @@
...
+/**
+ * Property types.
+ * @property {Object} propTypes Property types.
+ * @static
+ */
+View.propTypes = {
+  data: PropTypes.objectOf(PropTypes.any).isRequired,
+};
+
+const mapStateToProps = state => {
```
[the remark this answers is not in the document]
sneridagh: I think you got the Redux store wrong... let's talk about it when you have time.
I'm even having doubts if we should use the store at all here.
I would better make the request by hand on the edit and on the view, based on the stored url tile props, doing a single request on componentDidMount (on the view) and on url set (on edit).

sneridagh:
Fix build, add them to the code too.
Please also, rebase and put the changelog in the right place.

src/components/manage/Tiles/Rss/Edit.jsx
nileshgulia1: ah. crap! sorry, I used it for testing..

src/config/Tiles.jsx
```diff
@@ -29,6 +30,10 @@ const messagesTiles = defineMessages({
     id: 'video',
     defaultMessage: 'Video',
   },
+  rss: {
```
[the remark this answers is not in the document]
nileshgulia1: Done!

src/components/manage/Tiles/Rss/Edit.jsx
nileshgulia1: Done!

src/components/manage/Tiles/Rss/View.jsx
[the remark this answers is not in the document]
nileshgulia1: Done!

sneridagh:
@nileshgulia1 I tested the tile, and it wasn't working. Didn't you say that it was working? I give it a try, and it turns out that the componentDidMount request was not working (as it uses async/await) I still don't fully get it on the web side but in the docs, it says that that way only works on node environments:
https://www.npmjs.com/package/rss-parser#nodejs

On the web you have to use:
https://www.npmjs.com/package/rss-parser#web

So I changed it and got a nice CORS error... 

@tisto I investigated that, and it seems that it's nearly impossible to overcome that in the client... T_T, so we will need to write a server side way. Unless @robgietema has a better idea... Having that in the server side is not the best from the security point of view... so T_T again...

I'm open to suggestions.

nileshgulia1:
@sneridagh It is working on my side only when i enable CORS through an extension, though some of links are not fetched ( like this one:
``` javascript
{this.state.feed.items.forEach(item => (
                      <div>
                        <Feed.Event>{item.title}</Feed.Event>
                        <Feed.Event>{item.author}</Feed.Event>
                      </div>
                    ))}
```
it throws an `undefined` error. Don't know if its the CORS issue.
I'll pull your changes and test it again asap!
/cc @tisto

tisto:
@nileshgulia1 using a locally installed CORS extension is definitely no way to go. I fully agree with @sneridagh here. We have to find another way.

sneridagh:
We can use an SSR route for it :)

nileshgulia1:
This is a stale PR, would reopen if required.

PR #599, opened 2019-01-28
Update Logo.jsx

vinay72:
I have fixed the import path of Logo.


vinay72:
I have mailed a scanned copy of Plone Contributor Agreement.

PR #616, opened 2019-02-14
Bring back the stylelint default configs

sneridagh:


sneridagh:
@robgietema did you rip that out at some moment for some reason? I would like to bring them back if possible.

PR #640, opened 2019-02-20
Clean Pastanaga Theme variables files

sneridagh:
Cleaned the variables files from Pastanaga theme, since they are already inherited from the default theme during the build process. Only the ones that are being actually applied are left.


sneridagh:
@tisto Thx!!

PR #641, opened 2019-02-21
Add a releaser helper

sneridagh:
I created a "poor man" zest.releaser for Volto :)
I will test it in the next release. It uses the tool release-it which is very nice:
https://www.npmjs.com/package/release-it

It does almost the same as zest.releaser does, more git/github oriented, it also fills the Releases page and other goodies. I had to bake a changelog modifier since the tool uses the git log for it 😷 .

I will add some documentation to it.

Exportable to other projects as well...


CHANGELOG.md
```diff
@@ -6,6 +6,9 @@
 
 ### Changes
 
+- Not related to development, but addded a releaser tool to help cut releases.
```
tisto: @sneridagh do we want to include dev-related issues in the changelog? In plone.restapi I usually leave them out.

sneridagh:
@tisto let's leave them out then!

CHANGELOG.md
sneridagh: Removed!

PR #642, opened 2019-02-21
Cursor navigation in Volto Editor tiles

sneridagh:


tisto:
@sneridagh would it be possible to further enhance this so that a new tile shows up if the user clicks the bottom arrow on the last existing tile? This would resemble the medium editor behavior.

sneridagh:
@tisto I can try, doesn't seem complex.

PR #644, opened 2019-02-22
Add handler for adding new tiles at the bottom of non-text tiles

sneridagh:


tisto:
@sneridagh could you elaborate on how this is supposed to work? I checked out the branch and I don't see any difference to master, to be honest.

tisto:
@sneridagh I just checked how medium does things. They leave whitespace below the last tile (e.g. an image). If you click below the last tile, a new tile will show up.

tisto:
@sneridagh the second thing medium provides is that you can use the arrow up/down in the tiles to navigate. I guess this is what the other PR is about, right?

sneridagh:
@tisto It only works on non-test tiles, so create an Image/Video/Map tile, then hover at the bottom, the handler should appear.

sneridagh:
@tisto yes, https://github.com/plone/volto/pull/642 is about the up/down arrow tiles nav. The handler PR one relies on slight (mainly CSS) changes on that so first merge the cursor nav before this one, please.

sneridagh:
https://gfycat.com/ColdActualEmu

tisto:
@sneridagh ok, after watching your video I could figure out how to do it. This is definitely an improvement. I'd suggest three further improvements though:

- Increase the size of the area that shows new tile on hover to at least 200px
- Show the "new tile on hover" on all tiles, not just on the image tiles. Behavior should be consistent across tiles.
- if we do the two above amendments, we could as well just show the standard "new tile" instead of the "-------- + ----------", what do you think?

sneridagh:
@tisto There's some problems with empty text tiles then if we apply the three that you are suggesting: 

The current handler in the form that it is now won't work, since the two pluses will appear too close to each other, both with the same feature. Setting 200px of hover wont' work for empty text tiles also... Regarding your comment to the cursor nav, if we add the arrow down for create a new tile in a text tile, we already have two ways to create a new tile below a text tile. So we don't really need another, unless you want to spawn the new tile selector menu right away for saving a click...

In overall, for me we have to think more (@albertcasado should take a look) and solve all this drawbacks with a consistent solution that brings the same behavior to all tiles.

tisto:
Note: Albert and Victor will discuss this tonight.

tisto:
Note: @sneridagh, @robgietema, and @tisto discussed this today in a hangout. We came to the conclusion that we will:

- Add a handler at the bottom of a non-text tile, so that a user can click below a non-text tile and a new tile with a "+" icon will appear
- Add the functionality that if the user selects a non-text tile (e.g. an image) and hits return, that a new tile will appear below the non-text tile

In the long run, we have to see how that works with the vertical dividers that @albertcasado imagined for the 1-4 vertical blocks in one row.

sneridagh:
Decided that this is not going to happen anytime soon, since the keyboard approach is in place, and works well for clients.

PR #646, opened 2019-02-24
Set image width in Volto editor to 50% for images that float left/right.

### LINKED ISSUE #645: UX: Volto Editor: Float image left/right does not work on large images
At my Volto training I saw literally everybody running into the following issue when they uploaded their first image on the Volto editor:

1) Add a new tile
2) Upload image
3) Click on left or right "float" button
-> Nothing happens

This leaves the user puzzled and they feel the system is not working properly. What happens technically is that the image is just too big to make the float left/right visible.

I think we should follow what the medium editor does and just do not make the user worry at all about the size of the picture. 

screencapture-medium-p-85d74214978f-edit-2019-02-24-05_16_39

The user wants an image that floats, so just give them what they asked for and give the picture a size that makes sense. By default is should be something like 50% of the width of the text container.

It should be possible to amend the size and the way we display floating images. Though, this is something an integrator or designer of the layout should do, not the editor.

tisto:
This fixes #645

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Added
 
+- Set image width in Volto editor to 50% for images that float left/right @timo
 - Ability to navigate through the existing tiles with the cursors. @sneridagh
  - HTML Tile for Volto Editor with preview and code prettifier
    @ajayns @nileshgulia1 @sneridagh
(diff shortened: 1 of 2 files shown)


davisagli:
I'd suggest specifying max-width instead of width so that you don't increase the size of small images such as logos or icons.

tisto:
@davisagli thx for the suggestion! I amended my PR.

@sneridagh would it be possible to review and merge this PR and do a release that we ship to volto.kitconcept.com? If I present this today, I would like it to be on the demo instance. :)

PR #651, opened 2019-02-26
Fix link indicator on text tile

nileshgulia1:


sneridagh:
@nileshgulia1 take a look at this branch: https://github.com/plone/volto/tree/sneridaghinstantiatelinkplugin it compiles fine, but it doesn't fix the issue. :(

I think it's related with the update of the UI when the tile is instantiated on the fly, the data is fine in the store, and when you saves, everything is in place and works (when the tiles are already instantiated from the beginning. Certainly a problem of UI updating (or lack of).

Maybe a thing for @robgietema to look at closely when he's available.

avoinea:
@tisto @nileshgulia1 I think this was fixed by #1489

nileshgulia1:
Thanks @avoinea

PR #653, opened 2019-02-27
Add SSR helper to get resources (images/files) from the server using the API headers

sneridagh:


sneridagh:
I wake up this morning with this idea... I had to implement it right away.

After this, I really think that SSR is the hidden secret weapon of Volto...

The only "issue" that I see is that we force the SSR server to serve all the images. Do you think that we shouldn't do it for anonymous sessions? It won't be difficult I guess...

Also, if someone requests the image in Volto and it's private, it redirects to Plone... something that we can also polish I guess.

PR #655, opened 2019-02-27
Add error log in the SSR console. Fix edit forms with richtext fields…

sneridagh:
… coming from SSR.


src/server.jsx
```diff
@@ -115,6 +114,10 @@ server
           res.set({
             'Cache-Control': 'public, max-age=60, no-transform',
           });
+
+          // Displays error in console
+          console.error(error);
```
tisto: @sneridagh @robgietema we might want to think about a generic way about how to handle errors in Volto. Maybe factor that out into a function so that we can enhance it easily at a later point. Has nothing to do with this PR directly. Just a thought that crossed my mind.

tisto:
@sneridagh you are on fire man! :)

Issue #647, opened 2019-02-24
Medium-like image zoom on image tiles


tisto:
I'd like to have a medium-like (www.medium.com) image zoom for Volto. I already implemented that for Gatsby, so I guess it's not that hard.

https://github.com/kitconcept/gatsby-starter-kitconcept/pull/16

I used this library:

https://rpearce.github.io/react-medium-image-zoom/

Another option might be:

https://desmonding.me/zooming/

@nileshgulia1 @ajayns is that something you could imagine to look into?

nileshgulia1:
closed via #659

PR #659, opened 2019-03-01
Image-zoom

nileshgulia1:
#647 
-cc @tisto

### DIFF
--- a/src/components/manage/Tiles/Image/__snapshots__/View.test.jsx.snap
+++ b/src/components/manage/Tiles/Image/__snapshots__/View.test.jsx.snap
@@ -6,8 +6,18 @@ exports[`renders a view image component 1`] = `
 >
   <img
     alt=""
-    className="ui image"
+    onClick={[Function]}
+    onError={[Function]}
+    onKeyDown={[Function]}
+    onLoad={[Function]}
+    onMouseDown={[Function]}
     src="image.jpg"
+    style={
+      Object {
+        "cursor": "zoom-in",
+      }
+    }
+    tabIndex={0}
   />
 </p>
 `;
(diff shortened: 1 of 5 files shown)


src/components/manage/Tiles/Image/__snapshots__/View.test.jsx.snap
```diff
@@ -6,8 +6,18 @@ exports[`renders a view image component 1`] = `
 >
   <img
     alt=""
-    className="ui image"
```
tisto: @nileshgulia1 why did you remove the semantic ui class here? That breaks the image handling in Volto.

nileshgulia1: will undo the snapshots....

tisto: @nileshgulia1 no! this is no solution. You broke Volto by taking away the semantic ui classes. You have to amend the "<ImageZoom" to implement the same semantic ui classes that were there before.

nileshgulia1: @tisto <ImageZoom/> by default overrides css which we use. We have replaced our <Image/> from semantic ui which contains css classes already. The workaround could be just adding the css manually which semantic ui provides. 
https://github.com/Semantic-Org/Semantic-UI-React/blob/d6f20f90f35494a2eeeb0c1586658172fe98ea06/src/elements/Image/Image.js#L54

tisto: You have to pass those css classes to the ImageZoom component.

tisto:
@sneridagh works like a charm and looks good to me. Would you mind having a second look?

PR #673, opened 2019-03-13
Component docs

sneridagh:


sneridagh:
@nileshgulia1 I merged master and now the build is broken (I guess because of the docs structure change. Please create a styleguide folder on docs (`docs/styleguide`), then add the documentation there, and build from there, leaving alone all the other docs (in `docs/sources`).

We can look into the CSS issue and the not loading components because they need the store mock from there.

sneridagh:
@nileshgulia1 
In order to make the component work, I suggest follow 
https://github.com/styleguidist/react-styleguidist/blob/master/docs/Thirdparties.md#using-styleguidist-with-other-libraries

Then add a default wrapper with a fake store that provides all the possible values to the components. Make the wrapper also have a StaticRouter from react-router, this will also fix the error with the ones that have links and are wrapped with `withRouter`.

babel.js
```diff
@@ -2,6 +2,7 @@ module.exports = {
   presets: ['razzle/babel', 'stage-0'],
   plugins: [
     'transform-decorators-legacy',
+    ['transform-class-properties', { spec: true }],
```
sneridagh: This is completely required?

package.json
```diff
@@ -21,7 +21,7 @@
     "url": "https://github.com/plone/volto/issues",
     "email": "plone-developers@lists.sourceforge.net"
   },
-  "homepage": "https://plone.org",
+  "homepage": "http://plone.github.io/volto",
```
sneridagh: we will not use this URL. We will use styleguide.voltocms.com but I won't configure it here.

package.json
```diff
@@ -35,6 +35,9 @@
     "test:ci": "CI=true razzle test --env=jsdom",
     "start:prod": "NODE_ENV=production node build/server.js",
     "i18n": "NODE_ENV=production node src/i18n.js",
+    "styleguidist": "styleguidist server",
+    "styleguidist:build": "styleguidist build",
+    "deploy": "gh-pages -d styleguide",
```
sneridagh: No ghpages, we will use netlify.

package.json
```diff
@@ -98,16 +101,17 @@
   "dependencies": {
     "autoprefixer": "9.1.5",
     "babel-core": "6.26.3",
-    "babel-loader": "8.0.2",
+    "babel-loader": "7",
```
sneridagh: Why this version changes?

razzle.config.js
```diff
@@ -83,7 +83,7 @@ module.exports = {
                 importLoaders: 2,
                 sourceMap: true,
                 modules: false,
-                minimize: true,
+                //minimize: true, won't work in css-loader v2.0.0
```
sneridagh: Why we need to update css-loader?

src/components/theme/Anontools/Anontools.jsx
```diff
@@ -21,7 +21,7 @@ import { settings } from '~/config';
  * @class Anontools
  * @extends Component
  */
-export default class Anontools extends Component {
+class Anontools extends Component {
```
sneridagh: Why is this required?

babel.js
nileshgulia1: This is of support for static propTypes in ES 6 classes.

package.json
nileshgulia1: ok. I'll remove it. I used it for testing.

package.json
nileshgulia1: Hm. will remove it, I was testing with these packages :)

src/components/theme/Anontools/Anontools.jsx
nileshgulia1: This is for those decorators warning. I'll revert them, we can have a look at proptypes later

razzle.config.js
nileshgulia1: I have reverted it.

tiberiuichim:
Superseeded by https://github.com/plone/volto/tree/styleguide #1834

Issue #632, opened 2019-02-19
Editing an Event type, the form does not get the current start/end date


sneridagh:
Shows empty fields on edit.

PR #709, opened 2019-04-17
[WIP] - fix date widget

cekk:
Related to #632 

The problem was on date format: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime-local#Setting_timezones

We receive a date like _2019-04-17T10:00:00+00:00_ with also timezone info, but datetime-local html field doesn't like timezone anymore.

I needed to convert UTC datetime into local datetime and pass the value without timezone to the input field.
And on change i re-convert the date into UTC for plone.restapi.

I haven't touched the widget to use something better because i don't know what's your plans about date widgets. There are some nice plugins for React.

One other thing: right now a new event has empty dates. In Plone these dates are set as now.
Is it a good idea provide now as default when invoking this widget for event fields?
Also adding a button that sets the date as now like in Plone.

### DIFF
--- a/src/components/manage/Widgets/DatetimeWidget.jsx
+++ b/src/components/manage/Widgets/DatetimeWidget.jsx
@@ -5,8 +5,9 @@
 
 import React from 'react';
 import PropTypes from 'prop-types';
-import { Form, Grid, Input, Label } from 'semantic-ui-react';
+import { Form, Grid, Input, Label, Button } from 'semantic-ui-react';
 import { map } from 'lodash';
+import moment from 'moment';
 
 /**
  * DatetimeWidget component class.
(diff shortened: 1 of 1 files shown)


tisto:
There are tons of React widget options. We used the AirBnB datepicker in some projects and had good experience with it.

New events should have the date set to now. I guess this is something the Volto frontend could provide.

sneridagh:
@cekk @pnicolli what's the status here? There is also another approach by @nzambello ?

pnicolli:
@sneridagh I know @nzambello had something going on, but I'm actually not well informed on that. On the other hand, we will for sure need to make sure events are fully working in the next few weeks, so we definitely have this on the radar. We'll let you know.

sneridagh:
@pnicolli @cekk This is superseeded, right? Close it at will

PR #715, opened 2019-04-29
Show comments multiline

rodfersou:


sneridagh:
@rodfersou Please update the snapshot here too: https://travis-ci.org/plone/volto/jobs/526054758#L825

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### Added
 
+- Show comments multiline @rodfersou
```
sneridagh: Also, it's a fix more than a feature, right? Then it goes to changes. We use semantic versioning, so if it's truly a feature=> minor increase, if it's a fix/patch=> patch increase.

sneridagh:
@rodfersou this solution didn't work in our project, we have to backport the correct solution from there.

PR #731, opened 2019-05-11
Use eslint-config-react-app for linting

tisto:


tisto:
@robgietema @sneridagh I looked into our eslint config and we have tons of errors (180 on *.js and 1800 on *.js/*.jsx. Therefore I decided to give eslint-config-react-app from create-react-app a shot and it works fine as far as I can see. I added an exclude for the decorator problem with I will look into separately. 

I'd like to merge this asap because I would like to move to test using eslint for accessibility checks.

sneridagh:
@tisto please do not merge this. This is a bold one, and are several things that I want to keep.

sneridagh:
@tisto I know your desire to lower the bar in the exigency of the eslint rules. Using airbnb rules was always a pain in the ass, and I saw several of posts against it (from Kent C. Dodds, lately) saying that the cons outweigh the pros by far... but in my personal opinion, it's an endless source of knowledge, that we can't overlook.

So I would be +1-ish to remove, but knowing all what we are loosing. BTW, the a11y is already being checked with airbnb.

After saying that, I would make comments inline.

.eslintrc
```diff
@@ -1,42 +1,8 @@
 {
-  "extends": ["airbnb", "prettier"],
```
sneridagh: We need prettier config, it works together with the prettier plugin and the plugin in IDEs to match the rules everywhere.

.eslintrc
```diff
@@ -51,6 +17,5 @@
     "jest": true,
     "socket": true,
     "webpackIsomorphicTools": true
-  },
-  "plugins": ["prettier", "react-hooks"]
```
sneridagh: We need prettier plugin, it works together with the prettier config and the plugin in IDEs to match the rules everywhere.

.eslintrc
```diff
@@ -1,42 +1,8 @@
...
-    "es6": true,
-    "browser": true,
-    "node": true,
-    "mocha": true,
-    "jasmine": true,
-    "protractor": true
-  },
-  "parser": "babel-eslint",
-  "rules": {
-    "prettier/prettier": [
```
sneridagh: We need the specific rules for prettier.

.eslintrc
```diff
@@ -1,42 +1,8 @@
...
-      "error",
-      { "trailingComma": "all", "singleQuote": true }
-    ],
-    "jsx-a11y/anchor-is-valid": [ "error", { "components": [ "Link" ], "specialLink": [ "to" ] } ],
-    "import/prefer-default-export": "off",
-    "react/prefer-stateless-function": 0,
-    "import/no-unresolved": 1,
```
sneridagh: I want this, it is very useful. (import/no-unresolved)

.eslintrc
```diff
@@ -1,42 +1,8 @@
...
-    "import/extensions": 0,
-    "import/no-extraneous-dependencies": 0,
```
sneridagh: "import/no-extraneous-dependencies": 0 This is required also, it used to complain with Volto itself or Volto dependencies. Maybe, airbnb forced, and react-app doesn't, then it's not required.

.eslintrc
```diff
@@ -1,42 +1,8 @@
...
-    "react/no-unused-prop-types": 0,
-    "no-plusplus": 0,
-    "no-nested-ternary": 0,
-    "react-hooks/rules-of-hooks": "error",
-    "react-hooks/exhaustive-deps": "warn"
-  },
-  "settings": {
```
sneridagh: All the settings are required. If not, the ~/config root imports complains and the aliases also do.

package.json
```diff
@@ -138,7 +140,7 @@
     "@babel/preset-stage-0": "7.0.0",
     "autoprefixer": "9.1.5",
     "babel-7-plugin-root-import": "5.4.0",
-    "babel-eslint": "10.0.1",
```
sneridagh: Why the downgrades?

package.json
```diff
@@ -240,5 +242,8 @@
     "tlds": "1.203.1",
     "xmlrpc": "1.3.2"
   },
-  "devDependencies": {}
```
sneridagh: We can't have devDependencies in Volto, because we want them all installed on create-volto-app projects. So all should go to deps.

src/helpers/ScrollToTop/ScrollToTop.jsx
```diff
@@ -1,4 +1,4 @@
-import React, { Component } from 'react';
+import { Component } from 'react';
```
sneridagh: I don't know if you can do this in React...
Better:
```
import React from 'react';

...

class bla extends React.Component {...
```

sneridagh:
@tisto just merged a fix that was preventing eslint to parse correctly some files, sorry, probably you'll have more violations now :/

src/helpers/ScrollToTop/ScrollToTop.jsx
tisto: We need to define which way we prefer. I don't care either way, we just have to make a decision. If importing React is the way to go, I am fine with it. In this case the "React" import was not used at all.

tisto:
@sneridagh I think we only disagree on the way how to get where we both want to go.

I get "1895 problems (1872 errors, 23 warnings)" when I run eslint on the current master on Travis:

https://travis-ci.org/plone/volto/jobs/531352428

I can not reproduce the problem locally and I have no idea where this comes from. We have two options IMHO:

1) you (or someone who fully understands the current setup) fixes the eslint config/errors and makes sure it passes on Travis (I am totally fine with that)

2) we start over from the scratch with merging this PR and re-add the stuff that we find useful step by step and document it properly on the way

I can work on this if we choose option 2), if we choose 1) I don't know what to do right now. 

FYI: I am a bit under time pressure here because I need this working eslint config in order to be able to work on the accessibility checks that we need for one of our current projects (not the one you are working on right now ;)).

package.json
tisto: Fixed.

src/helpers/ScrollToTop/ScrollToTop.jsx
tisto: Fixed. We still kind of need to define that somewhere, or better find a linter that checks this.

.eslintrc
tisto: I wasn't aware that the IDEs need that. Because the CRA config does not have it I assumed it would not be necessary. I re-added it.

.eslintrc
tisto: Same as above. CRA does not have it, so I removed. Re-added it though.

tisto:
@robgietema I talked to @sneridagh and we kind of agreed to start simple and then re-add stuff that we need/want. 

Question is would you be ok fort you to move from airbnb rules to create-react-app rules? TBH I don't have a strong opinion either way, I just want a working eslint setup ASAP.

tisto:
@robgietema @sneridagh I re-added everything except the airbnb plugin and the additional configs that are based on that. We have a working eslint and Travis setup here now (if we can make the decision to move from airbnb to cra).

I am more than happy to move back to airbnb and/or re-add any of those additional checks at a later point. Though, I think a basic working setup is better than a complete setup that is untested and that does not run on Travis.

sneridagh:
@tisto @robgietema I'm sold to abandon the airbnb style rules. I've updated to the last version in 3.0.x and they are getting the point of being ridiculous and way overkill. I've also seen lately some rants from React gurus complaining hard about it. So changing to the react-app ones is fine for me.

@tisto the current config ruleset is fine for me and the IDE features are fine too.

package.json
sneridagh: @tisto why is this?
`"babel-eslint": "9.x",`
it's some requirement? Why not the last one?

tisto: @sneridagh my mistake.I fixed it.

tisto:
@robgietema @sneridagh ready for another round of review.

PR #744, opened 2019-05-21
add makefiles,backend scripts to gitignore

nileshgulia1:
Not sure, some scripts in `api/` should be in gitignore?

-cc @sneridagh


sneridagh:
@nileshgulia1 These ones are not present any more, take a look at the current api folder:
https://github.com/plone/volto/tree/master/api
they mush have been leftovers from an old branch in your machine.

PR #745, opened 2019-05-21
Fix portal root

### LINKED ISSUE #743: /login only works on the portal root
In standard Plone you can append "/login" anywhere in the site and it will show the login form. In Volto that won't work. If you call for instance:

https://volto.kitconcept.com/news/login

You will see an empty screen.

nileshgulia1:
Fixes #743 
-cc @tisto


tisto:
@nileshgulia1 nice. Thanks for the fix!

tisto:
I am wondering if the redirect is working, e.g.:

1) A user goes to a private URL
2) User is redirected to /login
3) User logs in
4) After the login the user is automatically redirected to the original (private) URL

Issue #737, opened 2019-05-13
Listing Tile


tisto:
Create a tile that lists the items in the page, similar to the folder_contents view in Plone.

Check this PR about how to create a tile that lists remote content:

https://github.com/plone/volto/pull/573

This is the first step towards implementing the smart collection tile.

Issue #287, opened 2018-11-10
Smart Listing Tile (Collection)


robgietema:

tisto:
@nileshgulia1 @sneridagh as long as we are missing the collection functionality we should remove it from the Volto UI. Can you please take care of this tomorrow?

tisto:
@sneridagh the collection tile is one of the last large and important missing pieces. Maybe someone in Sorrento would like to have a look...

tisto:
@sneridagh @albertcasado it would be great if we could get a rough draft of how the collection tile (and maybe the listing tile) should work and look like. I'd like to make this a priority for the Beethoven sprint next week.

pnicolli:
WIP PR in #747

PR #747, opened 2019-05-23
Listing block

rodfersou:
Ref.: https://github.com/plone/volto/issues/737


src/components/manage/Edit/Edit.jsx
```diff
@@ -157,7 +157,7 @@ export class EditComponent extends Component {
       }
     }
     // Hack for make the Plone site editable by Volto Editor without checkings
-    if (this.props.content['@type'] === 'Plone Site') {
+    if (!!this.props.content && this.props.content['@type'] === 'Plone Site') {
```
rodfersou: image

pnicolli:
Listing tile will also have collection powers. See also #287

pnicolli:
Updated with new sidebar, now waiting for the querystring widget. See #801

pnicolli:
Updated to latest master. It would be awesome to keep working on this block.
@tisto @sneridagh Is it ok to go on with this, or do you have something different in mind?
AFAIK, this needs a check on keyboard behavior while going through tiles in edit mode, and all the sidebar querystring widgets.

sneridagh:
@pnicolli delivered what I promised. Now it's all yours for polishing, etc... the CSS might want some love for sure.

sneridagh:
@pnicolli I started to use this convention of naming components in a Block. It turned out very well, and allow good composability/extensibility. I would like to use it in all blocks. So it would be like:

- View - entrypoint for the block, same as before, try to be superslim
- Edit - same as above
- <BlockFeat>Body - The body of the UI unit, trying to be as autocontained as possible. eg. if it's an image, it should be able to know how to upload an image by itself. For the grids this is very important. In the case of this Block, I still think it could be improved... I will think about it.
- <BlockFeat>Data - the parameterization of the Block, what goes inside the sidebar
- <BlockFeat>Sidebar - the content of the sidebar, this is important in case you want to reuse Data components from other blocks, eg. images.

What do you think?

/cc @tisto @jackahl @robdayz @robgietema @rodfersou

sneridagh:
@pnicolli just pushed more improvements.

I've also thought that a setting for parameterizing the block in case you want to show the image or not would be suitable too.

pnicolli:
@sneridagh Great news! Will look into it asap.
I like the naming convention after the first look, will also look deeper into this with other Blocks.

Especially for this block, it is probably very useful because we will need several listing blocks with different styles and this would help a lot.

sneridagh:
@pnicolli @giuliaghisini is this ready to merge? If green I will merge then.

pnicolli:
@sneridagh Last time I checked, only tests needed some work. Will take a look at failing tests asap so we can eventually merge this.

pnicolli:
@sneridagh Looked into it. There was a bug in tests due to another merged fix (sidebar default tab), and this is fixed.
Actually, at this time, the relative path criterion isn't working, so its test is correctly failing. I tried both with the provided plone buildout as-is ('api' folder) and with the same buildout but activating the latest master for plone.rest, plone.restapi and kitconcept.voltodemo.
It needs some debugging to understand where the problem is (querystring widget or restapi), I will work more on it in the next 2 days and will post another comment here. (see #1059)

pnicolli:
Found the problem with relative path and fixed it. The querystring endpoint doesn't allow calling it on a context, and there was no way of calculating the correct context for the relative path. At this time, I fixed it in the querystring action by manipulating the query in the relative path case.

pnicolli:
@sneridagh Merged latest master and resolved newer conflicts. Do you think this is mergeable after tests? It would be awesome to start working with these in master.

pnicolli:
I notice there's a new error in tests after latest merge, unrelated to this block apparently. See also: https://github.com/plone/volto/issues/1120

sneridagh:
@pnicolli bubt then how is that the PR travis build went ok? I think this is still a heisenbug that we have around :(

sneridagh:
@pnicolli @giuliaghisini  green!

pnicolli:
@sneridagh Yes, they were green for us as well

PR #750, opened 2019-05-28
Fix i18n script to update local .po files for projects

sneridagh:


sneridagh:
My bad, the script was ok.

Issue #741, opened 2019-05-13
Table Block


tisto:
Allow editors to add pre-defined tables in Volto. Not a full-fledged table editing solution but basic functionality like:

- add a table
- add a column to an existing table
- add a row to an existing table

Customizations could be integrator work.

nileshgulia1:
@tisto Do we already have a mockup for it, or just simple semantic `<Table/>` will do?

nileshgulia1:
Okay, we have it already :)

https://app.zeplin.io/project/5c76b54e0e8c876239917ea3/screen/5cbee0647b72a0b757dd5f7b

tisto:
@nileshgulia1 FYI: that is not a mock of the table tile. Don't worry too much about the style yet. A very simple HTML table with the default styling is sufficient. Would you feel comfortable working on the add row/column functionality? I think this is the core of this feature.

tisto:
@sneridagh @albertcasado this is not as high-priority as the collection/listing UX. Though, it would be nice to have a rough outline about how the user adds rows/columns (tile icons?) and removes them.

nileshgulia1:
@sneridagh Can we somehow embed an `<Editor />` onto the tables to achieve the add rows/columns functionality, Like the ones we have with tiles?

sneridagh:
@nileshgulia1 yes, that would be the idea. However, I'm concerned about the performance of such a component that might have X x Y editors instantiated at the same time... Maybe you can do some performance tests.

I've already talked with Albert, we will have mocks after the sprint, but the initial idea is to use an UX very similar to the "Apple Pages" table embed widget.

sneridagh:
Done.

PR #757, opened 2019-06-07
table tile draft

nileshgulia1:
#741  Its currently a draft , needs discussion for internal structure.


tisto:
@nileshgulia1 thank you for working on this. I would say that we should start with a very basic implementation of the table component. We have one table style. When the user adds a table tile she gets an initial table with say, 4 columns and 3 rows. The user can then add new rows and new columns (maybe via tile option icons like we use them for images and videos).

robgietema:
I've merged this branch with the table-tile one, which is now finished and merged to master. Will close this one.

PR #778, opened 2019-06-21
added version overview in controlpanel

### LINKED ISSUE #671: Show the current version of Volto in the control panel view
It would be handy to see somewhere on the volto.kitconcept.com demo site which version of Volto is running. The main page mentions the plone.restapi version but not the Volto release.

Another place some version information could be added is the site setup page below the control panel listings. (Like default Plone iirc)

For end-user/editor testing and bug reporting this would be nice as well, so people can test if things have been fixed in a certain version or use the demo site to test and report issues.

nzambello:
Fixes #671.

I tested it on a production build and it works.
Now there is Volto version number but I put it in a list like versions are displayed on Plone right now, so we can add `bulleted` prop on List component and include any other useful version numbers as List items, like backend that's in use (Plone/Guillotina).


sneridagh:
@nzambello Can you take a look at the tests?

PR #796, opened 2019-06-22
Add development mode for kitkoncept.voltodemo

fredvd:
to  /api in plone backend


robgietema:
Why do we want this by default? The release is ok for now right?

fredvd:
@robg sorry, wrong wording. It’s just adding the package to [sources] so that you can do a bin/develop checkout if you want.  Nothing is switched to development by default (auto-checkout)

PR #798, opened 2019-06-22
Added SidebarPortal component

pnicolli:
This allows for easier sidebar management, since it hides some internal details (like `__CLIENT__` and the id of the sidebar html element), and helps tile developers to only show the sidebar when the tile is selected.

Upgraded Enzyme because shallow rendering allows for simple snapshots of the portal component. `react-test-renderer` regular renderer does not allow rendering Portals.


pnicolli:
Apparently, after merging this something happened with github and @sneridagh and I can't see options to remove branch and/or close this. @robgietema Do you see those?

PR #801, opened 2019-06-23
Querystring widget

robgietema:


sneridagh:
Tested with https://github.com/plone/plone.restapi/pull/759 and it works well.

I know it's still WIP, but I identified this things that should be addressed or at least have them in mind:

- [] As stated in the https://github.com/plone/plone.restapi/pull/759 comments, the vocabularies are set "inline" as we did before the spring breaking change. We might want to change that.
- [] The dates widget is the HTML5 one, at some point we might want to improve it to show a decent one.
- [] The path criteria in plone was one showing the Object browser and you were able to choose one by navigating into it. We might want to do that as well.
- [] The sort_on field is empty, as it's only a string, I guess that controlled by a zc3.form hint that we are missing. It should show the sortable fields.

pnicolli:
> Tested with [plone/plone.restapi#759](https://github.com/plone/plone.restapi/pull/759) and it works well.
> 
> I know it's still WIP, but I identified this things that should be addressed or at least have them in mind:
> 
> * [ ]  As stated in the [plone/plone.restapi#759](https://github.com/plone/plone.restapi/pull/759) comments, the vocabularies are set "inline" as we did before the spring breaking change. We might want to change that.
> * [ ]  The dates widget is the HTML5 one, at some point we might want to improve it to show a decent one.
> * [ ]  The path criteria in plone was one showing the Object browser and you were able to choose one by navigating into it. We might want to do that as well.
> * [ ]  The sort_on field is empty, as it's only a string, I guess that controlled by a zc3.form hint that we are missing. It should show the sortable fields.
> 

@sneridagh I see both the endpoint in restapi and the widget here have been merged. Does this mean that the things quoted here are solved?
Can we consider https://github.com/plone/volto/pull/747 unlocked and try that tile with the new widget?

PR #861, opened 2019-09-17
Improve usability of login form

sneridagh:
I know that this has a11y issues, but we are wondering if there's no better option (maybe include a hidden for SR-only link inside the proper tab flow... 

/cc @tisto


polyester:
Can't fully test right now, traveling. But a hidden link might be problematic; there is high probability of SR users not getting to that link, as autofocus grabs the focus and a positive tabindex messes up the tabindex.

Plus, autofocus has more a11y issues as well, for mobile users, who will be yanked to the input field, possibly missing the context when they have limited screen realestate.

It can work, but only if the login is a *dedicated* page/form/modal with nothing else on it, whose only purpose and content is to login. I don't know by heart if this component is meant to be a single-purpose one or if it is mixed into a page with multiple other elements.

Even then, you have to provide good aria labels to explain to the user that they have to type in a username and that the purpose of that is to login, as they will have missed the title of the form being something like "login to our lovely site"...

sneridagh:
@polyester It's for just plain login form:

image

It will only sport the username/login, it's unfortunate in this case that we should have the link in the description of the field, that's because the forced "jump". Autofocus is useful in desktop web, but indeed can be problematic in mobile browsers. I can see also a big problem in the fact that the "register" link is the last one in the tabindex chain, that's why I was asking for the hidden one.

polyester:
Thinking out loud: would it be possible to let the default be accessible, and 'only' use some js to autofocus if you're pretty certain it does more good than harm? IE on a desktop without assistive technology? Don't know if there are helper functions for that. Can check later when back home

PR #880, opened 2019-09-20
update yarn i18n files

jackahl:


tisto:
@jackahl I added an i18n check and needed to fix i18n to prove my check works. Closing this PR.

PR #891, opened 2019-09-26
Rework users controlpanel

jackahl:


jackahl:
@nileshgulia1 I changed some things on the users and groups control panel, because your CSS broke the login form. Can you check if these changes are ok?

PR #895, opened 2019-09-27
Move the Tile Edit wrapper one level up to the tiles engine

sneridagh:


sneridagh:
Now I think it's fine :)

PR #914, opened 2019-10-04
Fix page jump in add form because of visual

sneridagh:


davisagli:
Oldest open PR...I assume it's no longer relevant but of course feel free to reopen if it is.

Issue #933, opened 2019-10-18
Giving a page only a space as a title leads to error


jackahl:
1. Create a new page and add a space as title
2. hit save
Error:
```
Sorry, something went wrong with your request 
this.props.content is null


    in t
    in f
    in t
    in div
    in t
    in f
    in injectIntl(Connect(t))
    in DragDropContext(injectIntl(Connect(t)))
    in t
    in t
    in main
    in div
    in _
    in n
    in f
    in f
    in t
    in t
    in t
    in r
    in R
    in f
    in withRouter(Connect(R))
    in t
    in withRouter(t)
    in t
    in n
    in i
    in f
    in IntlProvider
    in f
    in f
    in s
```
![Screenshot_2019-10-18 Document hinzufügen](https://user-images.githubusercontent.com/44289551/67090978-b0d59600-f1ab-11e9-909a-b74ac184437e.png)

wkbkhard:
I can't reproduce this any more.

PR #960, opened 2019-10-27
Trim whitespace of form fields on form submission

JeffersonBledsoe:
As shown in https://github.com/plone/volto/issues/933, it's currently possible to submit a required form field with just whitespace as the data, causing the front-end to crash. While there probably should be a way to fail more gracefully if a form was submitted with erroneous data, this PR should at least solve this situation.

As an aside, it would be great to have a method of failing gracefully when running Volto in develop mode, such as showing the backend error message to the developer, rather than just showing where React crashed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 - Add image to listing view @robgietema
 - Fix `SchemaWidget` @robgietema
 - Move styles import to a separate file @pnicolli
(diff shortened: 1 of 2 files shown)


tisto:
@JeffersonBledsoe LGTM! Would you mind adding an entry to CHANGELOG.md?

JeffersonBledsoe:
@tisto Done!

Closes #933

PR #974, opened 2019-10-30
Fix icon in TextWidget

sneridagh:


src/components/manage/Widgets/TextWidget.jsx
```diff
@@ -5,9 +5,10 @@
 
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
-import { Form, Grid, Input, Label, Icon } from 'semantic-ui-react';
+import { Form, Grid, Input, Label, Icon as IconOld } from 'semantic-ui-react';
 import { map } from 'lodash';
 import { defineMessages, injectIntl } from 'react-intl';
+import { Icon } from '@plone/volto/components';
```
tiberiuichim: I propose renaming this import as { VoltoIcon }

PR #976, opened 2019-10-30
[WIP] - Fix link redirection

cekk:
This is still a work in progress because every time that i scratch the surface of links, i found new edge cases that need to be fixed/covered.

As i said during the sprint, <Link> component doesn't work well with links outside the router routes. We also don't need the router with external links.

I fixed LinkView checking if the url has the same domain as the window.location one.
Probably in the future this check should be overriden by a simple check on how the link string is made (when we implement internal links into Link objects).
btw right now it kinda works.

The other thing that i wanted to fix, is the link generated into Navigation component.
In Plone, if the user has the permission to edit that object, the link in Navigation is the url of the object itself, otherwise, if the user can't edit it, the url is remoteUrl.

I created a new wrapper component for manage links: this check if the link is internal or external, and creates a simple <a> or a router's <Link>.
It's a proof of concept because i don't know how you want to handle links into Volto.

There are some problems:
- The first (and most important) is that right now in "isExternalLink" function i'm checking the link's domain with the one from window.location, but in SSR we don't have "window" so i cheated using the apiPath. This could probably fixed in the near future when we develop link's widget for handling internal and external links.
- Links should have different behaviors if the current user can edit it or not. Do we want to emulate all the Plone features (for example in Navigation show absolute_url or remoteUrl, in contents view, always show the absolute_url and then let the view make the redirect) or we decide that in Volto they should work in a different way?

### DIFF
--- a/src/components/index.js
+++ b/src/components/index.js
@@ -9,6 +9,7 @@ export ContactForm from '@plone/volto/components/theme/ContactForm/ContactForm';
 export Footer from '@plone/volto/components/theme/Footer/Footer';
 export Header from '@plone/volto/components/theme/Header/Header';
 export Icon from '@plone/volto/components/theme/Icon/Icon';
+export Link from '@plone/volto/components/theme/Link/Link';
 export Logo from '@plone/volto/components/theme/Logo/Logo';
 export Navigation from '@plone/volto/components/theme/Navigation/Navigation';
 export SearchWidget from '@plone/volto/components/theme/SearchWidget/SearchWidget';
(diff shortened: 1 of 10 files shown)


sneridagh:
@tisto Is this already merged and works?

tisto:
@sneridagh the very basic link redirection functionality was implemented by @robgietema. Though, this PR is about what happens when the user clicks on a link in the navigation and other "edge cases".

cekk:
yes, there are edge cases and i wanted to discuss with you about how to handle them

sneridagh:
@cekk this can be closed, right?

davisagli:
Closing since this is quite old and appears to overlap with functionality that was already implemented. Let us know if there's anything here that is still important.

Issue #977, opened 2019-10-31
document port/host backend api overrides


fredvd:
I'm trying to add more documentation in the README.md for devs so they can override Volto's frontend port/host and the path to the backend api on the fly.   I ran into this when I had multiple plone sites runnig on my dev laptop and wanted to switch the backend.

From looking at src/config/server.js I thought this should work:
```shell
HOST=localhost PORT=3001 RAZZLE_API_PATH=http://localhost:8181/OtherPlone yarn start
```

But there are two situations: a) working on Volto core when you use the the volto source repo, and b) working on a volto app created by create-volto-app and where volto is imported as an npm package. 

create-volto-app
-----------------

Situation b) , using a create-volto-app app.  In the scaffolded repo there is src/config.js whee you can add port (and host and apiPath I assume), which is documented in the header:

```export const settings = {
  ...defaultSettings,
  port: 3333,
  host: 'localhost',
  apiPath: 'http://localhost:8383/Plone',
};
````
With `yarn start', the frontend starts, but listens on port 3000 instead of my port 3333 set in config.js and I see th default Plone welcome page. 

With 'PORT=3333 yarn start' the frontend starts on port 3333, but I get an error "This page does not seem to exist…" on the first page load.  so changing the port on the command line breaks up the apiPath setting in src/config.js. 

'PORT=3333 RAZZLE_API_PATH=http://localhost:8383/Plone yarn start' doesn't help, same"This page does not seem to exist…". 

In both this and the previous case the Network tab in the browser shows that all resources are loaded fine from the backend (and using localhost:8383/Plone). 

github.com/plone/volto master
-------------------------------

When I try to use PORT and RAZZLE_API_PATH on a volto master checkout I see slightly different behavior. I didn't mess with the source code by editing a config.js (not here). 

'RAZZLE_API_PATH=http://localhost:8383/Plone yarn start' starts on port 3000 , uses the changed backedn and shows my default Plone page

'PORT=3333 RAZZLE_API_PATH=http://localhost:8383/Plone yarn start' gives the same behavior as with the create-volto-app situation: frontend volto port is changed to 3333, but the first page cannot find the Plone backend homepage anymore. 

Summary: there is a weird interaction between ENV variables and config.js:   src/config.js port doesn't, work, only using RAZZLE_API_PATH on volto master works, but adding PORT breaks the front page loading.  

Reporting port number
-----------------------

[...]

PR #978, opened 2019-10-31
document command line port overrides in the README

fredvd:
I thought it would be a simple one to add this doc to the README.md, but there are inconsistencies, see issue report #977

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,6 +95,16 @@ Go to [http://localhost:3000](http://localhost:3000) in your browser.
 
 You can try a Volto online demo in [https://volto.kitconcept.com](https://volto.kitconcept.com)
 
+## Different host, port and backend location
+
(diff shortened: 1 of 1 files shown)


tisto:
@fredvd thank you for your contribution! I am wondering if it makes sense to keep the README as minimal as possible and add that kind of information into our regular docs.

fredvd:
@tisto I thought about the same, the README is quite long already. 

On the other hand, the essential stuff is at the top and the RADME now contains also a lot of info on setting up development, I think ports is essential  enough  for developers to be able to change the port and couldn't find this myself without digging in the source.

Maybe the whole development part with the port info can be moved into a separate doc later to clean up the main README.

fredvd:
@tisto or others, please don't merge this yet:  The 'documented' env variables in this small README update are not working consistently.  See https://github.com/plone/volto/issues/977

PR #996, opened 2019-11-04
improve documentation for Icon component

jackahl:


src/components/theme/Icon/Icon.jsx
```diff
@@ -17,6 +17,17 @@ const defaultSize = '36px';
...
  * @returns {string} Markup of the component.
+ *
+ * Use:
+ * drop icon to the icons folder ("src/icons")
+ * import svg into the file
+ * import this Icon component
+ * add icon component with name = your imported svg
+ *
+ * Reasoning:
+ * add a11y title to SVGs
```
sneridagh: I would add https://kitconcept.com/blog/pastanaga-icon-system/
as reference.

sneridagh: And also mention briefly that inlining an SVG is the best:
-  Zero conf Inlined SVGs, as it is the best option when working with SVGs

PR #998, opened 2019-11-07
Resolve UID

tisto:


tisto:
@sneridagh @robgietema I have a working version of the resolveuid feature in Volto now. I also added a Cypress test for this which passes. Problem is that this does not work reliably (both locally and on Travis, in tests and manually).

The REST API part seems to work fine (I added print statements to the code to show this) but Volto sometimes seems to show the old link instead of the new one (when an item has been moved).

I first thought it was a Volto/Redux update problem. Though, a hard reload of the page with the link does not help.

Maybe this is something we can discuss on Monday...

tisto:
FYI: Guillotina tests are failing. Plone tests pass. Will look into it...

PR #1008, opened 2019-11-14
Upgrade react and react-dom to 16.12.0

pnicolli:
There are some quality of life upgrades that would be a nice addition, plus this would fix the warnings in the loadable components testing (see #967).

For example:

- Add support for editing useState state from DevTools.
- Warn when setState is called from useEffect, creating a loop.
- Report Hook calls at the top level as a violation.

See more here:
https://github.com/facebook/react/blob/master/CHANGELOG.md


pnicolli:
Upgraded directly to react 16.12.0 which was released right after opening the PR

sneridagh:
Findings while upgrading:
Helmet seems pretty unmaintained, and because of it, a warning is logged in the console because they are using an old version of react-side-effect. I opened this:
https://github.com/nfl/react-helmet/issues/490

We can fork it internally in Volto, the code is quite short.

Regarding semantic-ui-react... have you seen this?
https://github.com/microsoft/fluent-ui-react
https://github.com/microsoft/fluent-ui-react/blob/master/MANIFESTO.md
It seems that someone at MS is refactoring stardustUI which in turn is a fork on semantic UI, however the project is ongoing but it's based on React Semantic UI... However, something that we have to follow.

However, all the semantic ui theming is lost as we know it.

PR #1014, opened 2019-11-25
Fix 1011

tisto:


sneridagh:
This seems a test and it's 4 years old :)

PR #1015, opened 2019-11-26
Remove "documentDescription" class in table block

sverbois:


sneridagh:
@sverbois Thanks! Would you mind to fix the test and write a changelog?

src/components/manage/Blocks/Table/View.jsx
```diff
@@ -17,7 +17,7 @@ import { settings } from '~/config';
  * @extends Component
  */
 const View = ({ data }) => (
-  <p className="documentDescription">
+  <p>
```
mck9: Why is the <table> in a <p> tag? According to the HTML5 spec <p> will be closed automatically at the beginning of <table>. The following </p> will yield "Error: No p element in scope but a p end tag seen." on https://validator.w3.org

PR #1018, opened 2019-11-27
Remove @testing-library/cypress dep, as it breaks builds if the inter…

sneridagh:
…nal cypress release is different than the one in this package

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@
 - Fix toolbar collapsed color @sneridagh
 - Minor CSS fixes @sneridagh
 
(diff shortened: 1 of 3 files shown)


sneridagh:
@pnicolli we've found this issue... :( If you update the cypress to a version different than the one in @testing-library/cypress it breaks for new machines, as it tries to download both at the same time! ...

https://github.com/cypress-io/cypress/issues/4840

we can put it back when they fix the issue

pnicolli:
Oh wow, didn't know about it. Sure, it makes sense to remove this part of testing-library.

PR #1025, opened 2019-12-02
Forked `react-helmet` since it seems unmaintained. Now it's a Named i…

sneridagh:
…mport in helpers


sneridagh:
See https://github.com/nfl/react-helmet/issues/490

Issue #1037, opened 2019-12-04
Video Block: Allow to add Vimeo videos


tisto:
The video block in Volto only allows YouTube videos so far. It would be nice to allow Vimeo (and in the future others as well).

Issue #1021, opened 2019-11-29
video block: typing in a "wrong" URL leads to error


steffenri:
steps to reproduce:
1) go to your local volto build (or https://volto.kitconcept.com/)
2) create a new video block
3) type in an URL which does not lead to a video (can also type in anything you want which is not a correct link)
-> error message:
-> expected: some message that the link you typed in does not refer to a video or does not exist.

Sorry, something went wrong with your request 
Cannot read property '1' of null

    in t
    in InjectIntl(t)
    in div
    in t
    in DragSource(t)
    in DropTarget(DragSource(t))
    in div
    in t
    in InjectIntl(t)
    in div
    in t
    in p
    in InjectIntl(Connect(t))
    in DragDropContext(InjectIntl(Connect(t)))
    in t
    in t
    in main
    in div
    in v
    in n
    in p
    in p
    in t
    in t
    in t
    in r
    in Y
    in p
    in t
    in withRouter(Connect(Y))
    in t
    in t
    in withRouter(t)
    in t
    in n
    in r
    in p
    in IntlProvider
    in p
    in t
    in s

wkbkhard:
WIP while working on #1037

PR #1051, opened 2019-12-10
Iss1037

wkbkhard:
Worked on #1037 and #1021 

Video-Block allows Vimeo Videos and Youtube Videos now, if something else is in the data.url it won't fail, but show message

Also there are cypress tests for both video types


sneridagh:
@wkbkhard unit tests seems that are failing, please can you take a look and create the changelog entry?

tisto:
@wkbkhard works like a charm! Thanks! :)

PR #1053, opened 2019-12-11
Add new line tests

rodfersou:


cypress/integration/content.js
```diff
@@ -39,6 +39,28 @@ describe('Add Content Tests', () => {
       cy.contains('This is the text');
     }
   });
+  it('As a site administrator I can add a new line to a text block using shift + enter', function() {
```
tisto: @rodfersou I think this test should go into block-text.js. This is where we have all the Text Block related tests.

cypress/integration/content.js
```diff
@@ -39,6 +39,28 @@ describe('Add Content Tests', () => {
...
+      .type('Second line;{control}{enter}')
+      .type('Third line.')
+      .get('span[data-text]')
+      .contains(`First line;
+Second line;
+Third line.`);
+    //then the block should not be deselected
+    cy.get('.block.inner.text:nth-child(2n) .block.text.selected');
+    //then there should not be a new block created
+    cy.get('.block.inner.text:nth-child(3n)').should('not.exist');
```
tisto: @rodfersou we also have to check the status after saving. We need to make sure there is a newline in the view.

cypress/integration/block-text.js
```diff
@@ -27,6 +27,36 @@ if (Cypress.env('API') !== 'guillotina') {
         .should('have.attr', 'href')
         .and('include', 'https://google.com');
     });
+
+    it('As a site administrator I can add a new line to a text block using shift + enter', function() {
```
tisto: "site administrator" -> "editor".

cypress/integration/block-text.js
```diff
@@ -27,6 +27,36 @@ if (Cypress.env('API') !== 'guillotina') {
...
+      cy.visit('/');
+      cy.get('#toolbar-add').click();
+      cy.get('#toolbar-add-document').click();
+      cy.get('.documentFirstHeading > .public-DraftStyleDefault-block')
+        .type('This is a page')
+        .get('.documentFirstHeading span[data-text]')
+        .contains('This is a page');
```
tisto: This block shouldn't be necessary since the test setup already opens a page in edit mode.

cypress/integration/block-text.js
```diff
@@ -27,6 +27,36 @@ if (Cypress.env('API') !== 'guillotina') {
...
+        .type('First line;{shift}{enter}')
+        .type('Second line;{control}{enter}')
+        .type('Third line.')
+        .get('span[data-text]')
+        .contains(`First line;
+Second line;
+Third line.`);
+      //then the block should not be deselected
+      cy.get('.block.inner.text:nth-child(2n) .block.text.selected');
+      //then there still should not have <BR> tag
```
tisto: I don't fully get what we are checking here. Why is the absence of a br tag something we check? If there is no way to visually check DraftJS for a BR-like tag, we shouldn't check anything at all.

cypress/integration/block-text.js
rodfersou: @tisto don't understand

sneridagh:
Volto no longer uses DraftJS, and Slate handles this correctly.

PR #1064, opened 2019-12-14
Added lxml version pin to latest release

tiberiuichim:
It fixes compilation issue with Python3.8


sneridagh:
We should fix this and improve the buildout, but versions.cfg is not used.

PR #1077, opened 2019-12-22
Fix SSR rendering of of querystring widget, in case widget has value

### LINKED ISSUE #1020: Querystring widget breaks edit page of Collection if page is refreshed
How to replicate:

- Go to https://volto.kitconcept.com/add?type=Collection
- Add a new collection, set a simple criteria with Types and Document as values
- Save the collection
- Click Edit on the collection
- Force refresh of page

What happens is: componentDidMount() calls the getQuerystring action in the browser, this populates the this.props.indexes property. But the server doesn't have the ``indexes`` populated, so code lower down in the selection filters all fail. 

My fix was to add this check at the top of the render() method:
```
    if (Object.keys(indexes).length === 0)
      return <div>Loading...</div>;
```

tiberiuichim:
Fixes #1020


src/components/manage/Widgets/QuerystringWidget.jsx
```diff
@@ -263,6 +263,9 @@ class QuerystringWidget extends Component {
       intl,
     } = this.props;
 
+    // SSR fallback
+    if (Object.keys(indexes).length === 0) return '';
```
sneridagh: LGTM, but can you please move it next to the main return? Then we keep all the return together.

sneridagh:
@tiberiuichim Now that the SSR renders something, the tests fails, can you please fix them?
Thanks!

sneridagh:
Superseeded by: https://github.com/plone/volto/issues/1020#issuecomment-577096273

Issue #1076, opened 2019-12-22
Export unconnected classes


tiberiuichim:
To be able to extend component classes (for example widgets) they have to be also exported as individual classes, in addition to the already default export as redux connected classes.

tiberiuichim:
An example of how this can be useful: I want to reuse the QuerystringWidget, to filter the indexes it receives. I can't do that now because the only export of that module is already connected, but if the class was individually exported, I could write:
```
export default connect(
    state => ({
          indexes: filterIndexes(state.querystring.indexes),
    }),
    { getQuerystring },
)(QuerystringWidget);
```
Also, the QuerystringWidget needs to be exported by ``@plone/volto/components``.

sneridagh:
@robgietema I need your feedback here and in #1078

ichim-david:
We moved from class components and the pull request that provided this change was not accepted as such we can close this ticket, if ever we need something similar another issue and pull request can be opened

PR #1078, opened 2019-12-22
Export unconnected widget classes

tiberiuichim:
Allows extending and overriding the connected props. More flexibility is good. Need to do this for all connected components, right now I've only edited the widgets.


tiberiuichim:
Fixes #1076

api/plone-5.2.x.cfg
```diff
@@ -8,3 +8,4 @@ versions=versions
 
 [versions]
 plone.restapi =
+lxml = 4.4.2
```
sneridagh: Please do not mix things in PRs. If this is needed because some other reason, please create a PR for it.

sneridagh:
@tiberiuichim Jus saw the issue related, and you might have a valid point in that example. However, I'm still undecided on this one, since it adds complexity to the code that a newbie might not get right. Doing this everywhere is also a lot IMHO.

/cc @robgietema can you take a look?

tiberiuichim:
For what is worth, for a newbie (IMHO), groking redux and ``connect`` might be bigger hurdles then the ES module system.

Would it help if I write some documentation for Volto that tackle these issues? Not that I'm an expertise, but contributing to Volto is a good cause.

tiberiuichim:
Another personal impression: I prefer component inheritance or composition rather then override (using customizations, jbot behavior), Volto is still a young system with a lot of development, there are still things that are fixed in core components. Customizing those creates a high maintenance burden which we've already paid many times in the last months. I think it might even be easier to maintain a fork of Volto that can be mostly automatically merged, rather then keep a customized version of a component and keep it up to date with latest Volto master changes.

tiberiuichim:
We should export these widgets as UnconnectedSomething, just like: https://github.com/plotly/react-chart-editor/blob/87e4c5dde5f32d32ca40460fdcd88527568965d4/src/components/fields/NumericOrDate.js

tiberiuichim:
The decision was to export named objects when declaring them (such as `export class SelectWidget {}`) and to export default connected objects.

tiberiuichim:
I think this PR is outdated. We already have the widgets in storybook, so the problem must have been fixed by now.

PR #1082, opened 2019-12-25
Querystring broken if index value is missing from vocab

tiberiuichim:
Fix Querystring widget problem when content has an index value that is missing from vocabulary


sneridagh:
@tiberiuichim ups, this one also got addressed with:
https://github.com/plone/volto/pull/1123

also see:
https://github.com/plone/plone.restapi/pull/860

PR #1089, opened 2020-01-04
Avoid console warnings in AddFormLink.jsx

tiberiuichim:


sneridagh:
@tiberiuichim @tisto adding key issues in React using indexes is something that is considered bad practice, and as far as I remember, it throws a warning in the console too. #1086 was solved also the same way, just realized. If there's no element unique that identifies the object inside the array being iterated (usually there's one in Volto coming from p.restapi, e.g. the item `@id`), it's better to use a random uuid e.g.

```
import { v4 as uuid } from 'uuid';

// then use uuid() as key
```

@tiberiuichim can you amend this and #1086? Thanks!!

tisto:
@sneridagh right, now I recall that I tried to get rid of those warnings by adding keys as well and then stopped because more warnings popped up. Thanks for the clarification!

tiberiuichim:
@sneridagh I really appreciate the knowledge, I'll tweak the PR.

PR #1093, opened 2020-01-07
enable to paste links into textfields

jackahl:
This makes it possible to paste links into the normal text-blocks. All other styling is still stripped.


jackahl:
this is the package I added: https://github.com/thibaudcolas/draftjs-filters

PR #1103, opened 2020-01-10
Video block mp4 support

jackahl:


.eslintrc
```diff
@@ -21,7 +21,8 @@
     "prettier/prettier": [
       "error",
       { "trailingComma": "all", "singleQuote": true }
-    ]
+    ],
+    "jsx-a11y/media-has-caption": "off"
```
tisto: @jackahl isn't that a useful check in general?

sneridagh: Yeah, I would leave that out and add it to a per project config. So we leave it as cannon for Volto.

tisto: @sneridagh excluding a single line from eslint does not work. No idea why. I had to re-add this line to make Travis pass. I'd say let's cut a release and then re-visit this later. A linter problem should not hold back a release we need.

PR #1109, opened 2020-01-14
Fix token expiration and unusable UI, fixes #674

### LINKED ISSUE #674: Volto is in a non-working state when user keeps the browser open > 24h
Steps to reproduce:

1) Fire up Volto and log in as a user
2) Keep the browser open for > 24h (and put the computer in sleep mode)
3) Revisit Volto in the browser

-> The user will see the toolbar open but only with the three dots and the user icon. 
-> Everything else is gone.

Welcome to Volto  (3)

-> Browser console shows network calls with an unauthorized (HTTP 401) response:

Screenshot 2019-03-14 at 09 54 21

sneridagh:


sneridagh:
Merge before 4 final, but needs testing.

tisto:
@sneridagh what better time to test this than in the alpha phase? I mean we can revert at any time if that causes problems. That won't be possible once we are in beta or final.

PR #1123, opened 2020-01-21
Fixcollections

sneridagh:


src/components/manage/Widgets/QuerystringWidget.jsx
```diff
@@ -366,7 +366,7 @@ class QuerystringWidget extends Component {
                       components={{ DropdownIndicator, Option }}
                       value={{
                         value: row.i,
-                        label: indexes[row.i].title,
+                        label: indexes[row?.i]?.title,
```
pnicolli: IIRC, the `?` operator returns undefined for `row?.i` if row is null or undefined. If this is correct, then this can potentially turn into a `indexes[undefined]` which I think would throw even if there's a `?` after that.

Haven't had the time to test it now, but I wanted to drop a line because I'm not sure I'll be able to get back to my laptop until tomorrow :)

sneridagh: Mmm in fact I did that change separately from the others, and got in at the same time, but indeed it should be different, I don't like it now :)
I will try and ask you for review.

src/components/manage/Toolbar/Toolbar.jsx
```diff
@@ -418,7 +418,7 @@ class Toolbar extends Component {
                   messages.shrinkToolbar,
                 )}
                 className={cx({
-                  [this.props.content.review_state]:
+                  [this.props.content?.review_state]:
                     this.props.content && this.props.content.review_state,
```
pnicolli: Could be `this.props.content?.review_state` here as well.

PR #1134, opened 2020-01-24
Try to improve the realibility of the Cypress tests, reorganize commands

sneridagh:


cypress/integration/blocks.js
```diff
@@ -1,6 +1,8 @@
 if (Cypress.env('API') !== 'guillotina') {
   describe('Blocks Tests', () => {
     beforeEach(() => {
+      // Wait a bit to previous teardown to complete correctly because Heisenbug in this point
+      cy.wait(2000);
```
tisto: @sneridagh shouldn't that go into teardown then? We should also try to wait for the XHR request instead of adding wait.

sneridagh: I have the feeling that the error is because the beforeEach enters before the teardown has completed, so it felt more natural to make the beforeEach wait instead to add lag to the end of the teardown. We can test this for a while, and try the other approach if it doesn't work.

PR #1139, opened 2020-01-25
Fix for i18n to use defaultMessages instead to default to the id, Upd…

sneridagh:
…ate `babel-plugin-react-intl` to latest


sneridagh:
@tisto @robgietema https://github.com/formatjs/react-intl/blob/master/docs/Components.md#message-formatting-fallbacks it seems `react-intl` own fallback already provides what the modified line was trying to accomplish. Since we are setting our own fallback, the other was broken. Changed to use only `react-intl` fallback since at the end it will do the same.

theme/themes/pastanaga/extras/blocks.less
```diff
@@ -622,6 +622,10 @@ body.has-toolbar.has-sidebar-collapsed .ui.wrapper > .ui.inner.block.full {
     .icon {
       margin-bottom: 10px !important;
     }
+
+    &.table {
```
tisto: @sneridagh committed by accident?

sneridagh: all fine, it's a small CSS fix that it was not worth the hassle to create a single PR for it... it's kind of related with the blockchooser

sneridagh: I'm in get V4 final asap mode ;)

PR #1172, opened 2020-02-06
Hide Blocks fields in Layout fieldset in Add/Edit form

pnicolli:
The ``blocks`` and ``blocks_layout`` fields are visible in the Add/Edit form of content that has the Pastanaga editor enabled. These fields should not be visible there, because they are used and modified by the Pastanaga editor itself.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ### Changes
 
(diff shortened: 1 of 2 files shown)


sneridagh:
❤️ We were tempted to do so in Plone but they are extremely useful to have them there.

PR #1174, opened 2020-02-06
In the Display menu, only show views that are implemented

pnicolli:
This way we are offering a better user experience, imho.
Is this something you would be interested into?

The way it's implemented, they automatically show up when you configure a component to render that view. Also tested with sites created with ``create-volto-app``.

### DIFF
--- a/src/components/manage/Toolbar/More.jsx
+++ b/src/components/manage/Toolbar/More.jsx
@@ -11,11 +11,11 @@ import { compose } from 'redux';
 import { Link } from 'react-router-dom';
 import { find } from 'lodash';
 
-import { Icon, Display, Workflow } from '../../../components';
-import { getBaseUrl } from '../../../helpers';
+import { Icon, Display, Workflow } from '@plone/volto/components';
+import { getBaseUrl } from '@plone/volto/helpers';
 
-import rightArrowSVG from '../../../icons/right-key.svg';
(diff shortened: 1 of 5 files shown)


src/components/manage/Toolbar/More.jsx
```diff
@@ -11,11 +11,11 @@ import { compose } from 'redux';
...
 import { find } from 'lodash';
 
-import { Icon, Display, Workflow } from '../../../components';
-import { getBaseUrl } from '../../../helpers';
+import { Icon, Display, Workflow } from '@plone/volto/components';
+import { getBaseUrl } from '@plone/volto/helpers';
 
-import rightArrowSVG from '../../../icons/right-key.svg';
-import userSVG from '../../../icons/user.svg';
+import rightArrowSVG from '@plone/volto/icons/right-key.svg';
```
sneridagh: Unrelated, right? However, leave it! I'm +1 to do a sweep over all relative paths everywhere (we already did that in the lazy load) but it's a good addition before 4 final and will ease the merge for that branch.

pnicolli: Yes, it's unrelated, but I figured I'd start to change the ones I encounter on myh way in this polishing phase. :D

sneridagh:
Merging, travis is green but gh indicator is still yellow, weird.

Issue #1173, opened 2020-02-06
Blocks: The current focus management provokes a screen "jump" when selecting a block


sneridagh:
This is because the browser tries to "center" the focused element, related with the keyboard navigation management too.

Investigate if there's other approach to do this.

sneridagh:
Fixed long time ago.

PR #1202, opened 2020-02-16
Fix the 'jump' on the blocks editor on focusing blocks

sneridagh:


sneridagh:
It turns out, it was only a one liner change!
And I spent several hours on it... meh :/

sneridagh:
Fixes #1173

PR #1203, opened 2020-02-16
Add permission check to edit form

sneridagh:


tisto:
I can confirm that the Cypress tests hang locally as well on the block reindexing test (same as on Travis).

sneridagh:
WAT? ... ok, we'll have to investigate

tisto:
The block-indexing.js Cypress test hangs:

Screenshot 2020-02-16 at 21 04 03

If you re-load the page edit view manually the test pass without problems:

```
    beforeEach(() => {
      cy.autologin();
      cy.createContent('Document', 'my-page', 'My Page');
      cy.visit('/my-page/edit');
      cy.waitForResourceToLoad('@navigation');
      cy.waitForResourceToLoad('@breadcrumbs');
      cy.waitForResourceToLoad('@actions');
      cy.waitForResourceToLoad('@types');
      cy.waitForResourceToLoad('?fullobjects');
    });
```

I am not sure if this is a test only issue or a real issue. It seems the edit view somehow does not get the auth cookie.

sneridagh:
A Forbidden shows if the token is present, but the action is not there. Why could that be?

tisto:
@sneridagh no idea. I set wait statements everywhere and it does not change a thing. The really odd thing is that block-text.js uses exactly the same beforeEach and it does not fail. It fail/pass is also not a heisenbug bug, it is reproducible everywhere.

sneridagh:
@tisto, ok I think now I got it right.

sneridagh:
I messed it with git merges, so I created: https://github.com/plone/volto/pull/1215 that superseeded this one.

PR #1204, opened 2020-02-16
Disable Guillotina tests on Travis.

tisto:
The Guillotina tests started to fail on Travis without any change in the repository. I have no idea what's wrong or how to fix it. In order to being able to move on we have to disable the Guillotina tests for now.

I re-ran the last master build that was green to prove something outside the code repo must have changed to make the build fail.

After re-running

https://travis-ci.org/plone/volto/jobs/650499720

the build was red.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,9 @@ matrix:
     - name: 'Plone'
       python: 3.7
       env: TEST_SUITE=plone ZSERVER_PORT=55001
-    - name: 'Guillotina Tests'
-      python: 3.7
-      env: TEST_SUITE=guillotina
+    # - name: 'Guillotina Tests'
+    #   python: 3.7
+    #   env: TEST_SUITE=guillotina
     - name: 'Unit Tests'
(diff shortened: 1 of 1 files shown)


bloodbare:
As far as I can see the bug is introduces by volto as it fails running server.js:

```
Creating postgres_test ... 
    at ChildProcess.onClose (/home/travis/build/plone/volto/node_modules/start-server-and-test/src/index.js:68:14)
    at ChildProcess.emit (events.js:321:20)
    at maybeClose (internal/child_process.js:1021:16)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:286:5)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

tisto:
@bloodbare as said I re-run a master Travis job that was passing before. After re-running it failed with the error above. This is proof that nothing in the repo causes this problem. So this bug is definitely not introduced by Volto. The only remaining possible reason is that something changed on Travis.

PR #1209, opened 2020-02-17
Reference widget for related items

giuliaghisini:
I fixed this widget to display the title and path of object selected/selectable. 
I would ask if is possible to merge it on master until we 'll have the reference widget with object browser implemented. So, the related items field is usable with this temporary widget.
What do you think about?


tisto:
Good work @giuliaghisini and thanks for your contribution! I especially like the hover-functionality that shows the full path.

I only have one small comment. The check icons looks a little bit displaced:

Screenshot 2020-02-17 at 19 58 42

Maybe we should just remove it altogether.

giuliaghisini:
@tisto yess, I was also thinking to position the path to a new line, because in sidebar there's not enought space.. maybe today ill'fix it

PR #1212, opened 2020-02-18
Norelativeimports

sneridagh:


sneridagh:
@pnicolli I guess this might get easy mergings in the code splitting branch, and also fix this historic "bug" when using relative imports.

PR #1219, opened 2020-02-18
fix(Guillotina) Fixing travis tests on Guillotina backend

bloodbare:
The problem was that postgres docker maintainers decided to change the image forcing that now a password is needed. Just forced an old version of postgres makes it work again.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,9 @@ matrix:
     - name: 'Plone'
       python: 3.7
       env: TEST_SUITE=plone ZSERVER_PORT=55001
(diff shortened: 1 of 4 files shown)


bloodbare:
Looks like that tests fail because of content.js, nothing related with G as did not changed its image.

bloodbare:
@tisto travis fails because of content.js error. Can somebody (@sneridagh) check what happens with the content.js tests ?

bloodbare:
@sneridagh @tisto tests are green, you can merge

PR #1224, opened 2020-02-20
Paginationonlistings

sneridagh:


tisto:
@sneridagh works like a charm! Though, we currently do not batch the results when no criteria is selected (folder listing). This might be confusing for the users. Though, maybe that's something that we can do later...

sneridagh:
@tisto we do, create more than 25 items.

tisto:
@sneridagh that's good. I set the batch size to 2. Then batching works fine if I choose criteria, if not I do not see the pagination. Technically I am fine with that. It is just, as said, confusing for the user.

cekk:
yeah, i've found the same issue: i can confirm that's confusing.
Probably we should take care about pagination setting also if no query is set (and we took folder listing results).

sneridagh:
@tisto @cekk unfortunatelly, the GET content does not support batching (yet). We could trigger a search to the context (same as `items` code is doing), but with b_start and b_size support, but I think it's better to wait for it to be implemented.

What we could do is to disable the querystring fields if no query is present.

What do you think?

tisto:
@sneridagh yeah. disable the field makes sense.

Issue #1186, opened 2020-02-12
Layout navigation and user actions (login, register) overlapping on medium windows


ksuess:
![Welcome_to_Volto_](https://user-images.githubusercontent.com/1144737/74336700-17a43e00-4d9f-11ea-8874-e1ac9f9b3ebe.png)

tisto:
What about hiding login/registration for smaller viewports? The navigation could still overlap the search but then maybe we have to hide the search for smaller viewports as well. Or switch to mobile view for smaller viewports.

nileshgulia1:
@ksuess The nav links (login,register) do not overlap in when the i18 is en lang. But I think we can switch to mobile view on smaller screens on other languages as well.

ksuess:
Anontools (login and register), even if not needed on many websites, are IMHO better hidden in custom themes, not in default theme.

As you said @tisto there is still the search that could be overlapped by a long navigation.
So here is my first throw of a flexible responsive header: 
https://github.com/plone/volto/pull/1233

short navigation:

![Gartenplan_-_Mozilla_Firefox](https://user-images.githubusercontent.com/1144737/75140378-12c08200-56ef-11ea-9ca4-b96d579c992b.png)

long navigation:

![Kräuter_-_Mozilla_Firefox](https://user-images.githubusercontent.com/1144737/75140258-cecd7d00-56ee-11ea-8e7d-1b06bc020632.png)

PR #1225, opened 2020-02-20
Hide login,register when in small viewport

nileshgulia1:
related: #1186


ksuess:
default theme: In my opinion a link to "login" should be visible.
custom theme / authorization: feel free.

sneridagh:
@nileshgulia1 I think @ksuess in #1233 addressed it more thoroughfully.

nileshgulia1:
Yeah #1233  looks more specific to what we want to achieve.

Issue #1221, opened 2020-02-19
Draft-js Links on text fields of ContentType


giuliaghisini:
- Create a new News item
- open the console
- insert some text
- select a word in text and add a link on it
- write a link (f.e. http://www.google.it) or choose a internal link from the list
- an error is fired and link is not created.
This error occurs also when you try to enter edit-mode of an existing content-type wich have link inside text, and all the view is broken.

tisto:
@giuliaghisini could you imagine to write a Cypress test that shows this problem (and fails)? It bugs me that we don't catch regressions like this. You can basically just copy over this test:

https://github.com/plone/volto/blob/master/cypress/integration/block-text.js

And apply it for the news item (or any other content object that does not use the blocks layout/editor).

@robgietema any idea what causes this?

giuliaghisini:
yess. I created a new branch #draftjs-link-in-contenttype. I'm going to write test now..

giuliaghisini:
pushed failing cypress test

tisto:
@giuliaghisini thank you! Would you mind adding a link here to the PR?

cekk:
We've found the problem: [this](https://github.com/plone/volto/commit/c3e0b4aeea17ada86b68813795ed63183a79cc20) commit

Basically LinkEntity checks if the user is logged-in or not and creates a different component (router link or standard <a>).

On wysiwyg editor it breaks because the editor try to generate html with [renderToStaticMarkup](https://github.com/plone/volto/blob/master/src/components/manage/Widgets/WysiwygWidget.jsx#L210) that's now wrapped into a redux Provider.

@robgietema any ideas on how to fix/avoid this problem?

PR #1228, opened 2020-02-21
Fix problem with not wrapped element in `Provider` store in `WysiwygW…

sneridagh:
…idget` component

  due that now, the links are wrapped with a connected component


sneridagh:
Fixes #1221

Issue #1208, opened 2020-02-17
Got wrong HTML DOM in client rendering with YouTube and Image Blocks


terapyon:
I got strange HTML DOM in client rendering.

I make a page in YouTube block and two Image blocks. It's fine for first rendering. But the browser reloaded, I got wrong HTML DOM in client rendering, however the server rendering is no problem.

First rendering HTML is:

```
<p class="block image align center"><img class="" src="/demo/papa-cloud_mini.jpg/@@images/image" alt="papa-cloud_mini.jpg"></p>
<p>some text</p>
```

Reloaded rendering HTML is:

```
<p><img class="" src="/demo/papa-cloud_mini.jpg/@@images/image" alt="papa-cloud_mini.jpg"></p>
<p class="block image align center">some text</p>
```

After that, I removed YouTube block, I got fine HTML DOM.
I think this is related YouTube block or YouTube HTML.

terapyon:
I got another situation, it is Google Map block and image block.
I set aline right for image. 

I took a video of this error
https://youtu.be/_vLhJpXsGRk

PR #1229, opened 2020-02-22
Fix Video and Maps blocks hydration quirks on view mode

sneridagh:


sneridagh:
@terapyon this fixes #1208 :)

PR #1233, opened 2020-02-24
Responsive header

ksuess:
Fix for https://github.com/plone/volto/issues/1186 "navigation and user actions (login, register) overlapping on medium windows".

1. Anontools (login, register) wrapping under long navigation. 
2. Hamburger menu also on tablet.
3. Mobile: compact display of anontools and search.

tbd: changed DOM in Header.jsx: wrapper for tools and search


CHANGELOG.md
```diff
@@ -21,6 +21,7 @@
 - Fix Guillotina tests @bloodbare
 - Fix problem with not wrapped element in `Provider` store in `WysiwygWidget` component
   due that now, the links are wrapped with a connected component @sneridagh
+- Responsive header @ksuess
```
sneridagh: While doing the other changes, can you update it and put it into the right position?

src/components/theme/Header/Header.jsx
```diff
@@ -56,13 +56,15 @@ class Header extends Component {
               </div>
               <Navigation pathname={this.props.pathname} />
             </div>
-            {!this.props.token && (
-              <div className="tools">
-                <Anontools />
+            <div className="meta-wrapper">
```
sneridagh: Can you change this with a less generic class name? Like "tools-wrapper"?

ksuess:
I feel uncomfortable changing the DOM for wrapping anontools and search. But I tried seriously to avoid a DOM change while letting both flow under longer navigations. So, leaving or changing?

BTW this all about a header is just a suggestion, not meant to be merged as it is. I just felt urged to let the header look OK for navigations, that not only have three or four short items.

theme/themes/pastanaga/elements/header.overrides
```diff
@@ -1,3 +1,85 @@
...
+
+.ui.basic.segment .header {
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: space-between;
+  margin-right: -@mini;
+  margin-left: -@mini;
+
+  & > * {
+    margin-right: @mini;
```
sneridagh: @ksuess Why did you do this? I tend to like the logo sharp with the container (same at the end). 
image

The rest is fine for me!

ksuess: It's a way to give items in a flex container identical margin.
  margin-right: -@mini;
  margin-left: -@mini;
  & > * {
    margin-right: @mini;
    margin-left: @mini;
  }
But in this case it's an obsolete left over of an earlier state where three items had to be distributed, not two (logo-nav-wrapper and tools-search-wrapper). I'll do that simpler.

ksuess: BTW the logo is in both cases exactly aligned, but has a padding in image itself.

PR #1216, opened 2020-02-18
Table Content: Fix content overflow

sumukhah:
contents will get out of the box if there are many column on a table,
So allow horizontal scrolling if there are many column.
this
Fixes: #1033

Issue #1033, opened 2019-12-04
Properties of contents get displayed wrong if too many in content menu


steffenri:
1) go to your local volto build and open the content menu (or volto.kitconcept.com/contents)
2) select more properties that should be shown (like size, id, date, etc.)
-> details of content are not in the container anymore if too many properties are chosen
![displayoptions](https://user-images.githubusercontent.com/54619639/70127356-2c67a580-167b-11ea-9214-c0de947f7b4b.png)

sumukhah:
Hello @steffenri @tisto 
I would like to work on this issue, 
But I am curious why there are very few(only 2) CSS files in the entire code base?
Should we avoid using CSS styling in our codes?

jackahl:
@sumukhah we you use the less css-precompiler in Volto (http://lesscss.org/) Those files are named .less, .overrides and .variables v(https://github.com/plone/volto/tree/master/theme/themes/pastanaga).  Just add your fix to one of these files where it seems fitting.

robdayz:
I don't know if this change modifies some other table appearance, but for me, folder_contents is well-displayed whit this. I will check some other table to get if it works well..
 ```
 .ui.table {
    display: block;
    overflow: auto;
    overflow-y: hidden;
  }
```

stevepiercy:
@7twik, while you are awaiting an answer, please see https://6.docs.plone.org/volto/developer-guidelines/contributing.html#create-a-pull-request

sneridagh:
This issue was solved. Now the contents table goes from side to side, and it has a workable horizontal scrolling that shows the overflowed items.

Closing.

PR #1239, opened 2020-02-26
Revert "Table Content: Fix content overflow"

sneridagh:
Reverts plone/volto#1216 because it breaks the layout for table blocks.


sneridagh:
image

nzambello:
@tisto @sneridagh Due to this, is #1033 still unsolved? Should we reopen that?

sneridagh:
Reopened!! Unfortunately the CSS proposed was leaking everywhere a table was. We need a better solution.

Issue #1098, opened 2020-01-08
Use the ObjectBrowser for creating links in text editor


frapell:
I think the text editor needs a better UI for creating links, and I think re-using the ObjectBrowser for navigating/searching items in the site for this, could be a good idea, however I am having the hardest time in trying to make it happen.

If you think this is not a good idea, please let me know... If you think it is a good idea, I need help :)

I have added the `openObjectBrowser` prop to the `LinkButton` component and calling `openObjectBrowser({ mode: 'link' })` inside the `onAddLinkClick` event.
I've also wrapped the `LinkButton` export with `withObjectBrowser`.

I am stuck at this point, because I cannot find a way of passing the rest of props that `ObjectBrowserBody` expects, such as `data` or `block`... I have tried defining them as props for the `LinkButton` and defining some default values inside `defaultProps` but no dice... I do see the rest of the props being passed in tough, so not sure what am I missing so these new ones will be included.

I have read https://reactjs.org/docs/higher-order-components.html and while I understand the idea for HOC, I seem to be missing to fully comprehend what's going on...

I have commited what I did into branch https://github.com/plone/volto/tree/frapell-objectBrowser-links , if someone (cough, cough... @sneridagh :) ) wants to take a look and maybe give me some hints on what am I missing...

tisto:
@frapell thank you for looking into this! We discussed this in the past and we think it is a good idea. 

My only UX concern is that the user always has to reach for the sidebar to accomplish a rather simple task (adding a link). The good thing is that lots of user interaction already happens on the sidebar. If we want an object browser there is no way around re-using the sidebar object browser anyways...

frapell:
@tisto I am no UX expert, but I like the workflow when working with blocks, using the right sidebar, so I believe it would be expected for a user to not only use the sidebar for navigating/searching for an object when adding a link, I also think this is where the different settings for the link should live... I'm thinking for instance when clicking an already created link, a sidebar should show up with different settings like whether the link should open in a new window, rel attribute, etc...

It is my understanding by reading the code, that the ObjectBrowser assumes it is being called from a Block, and this is not necessarily always the case for the DraftJS editor, so there's a challenge in there...

tiberiuichim:
[...]

PR #1223, opened 2020-02-20
Draft js link uses ObjectBrowser

giuliaghisini:
Added objectBrowser to select a link in draft-js. 

Now the widget stores in draft-js structure the link without portalUrl. Do you think it may be right or do you prefer to save the portalUrl in draft-js structure on DB?

Sorry, i committed all buildout files of /api folder for a mistake

PR #1240, opened 2020-02-26
Use the ObjectBrowser for creating links in text editor

frapell:
This is coming from https://github.com/plone/volto/issues/1098


frapell:
@sneridagh Can you take a look and share your thoughts?

sneridagh:
@frapell nice! yeah, we are looking something like this. @pnicolli and the RT folks are also looking for a UX formula that works like this. 

The only thing missing is how we provide a way to specify external URLs, and if we want to maintain the "lite inline chooser" in some manner.

Do you have something in mind for that?

I have to contact @albertcasado for searching a good UX solution for it.

We have to discuss this in the next weeks, I will include you in the meetings/discussions!

frapell:
@sneridagh Ahhh... totally ignored that use case... I am no UX/UI designer, so not sure the best approach... 

One way could be maybe to keep the AddLinkForm, without the search capability, and instead add the "navTreeSVG" button to the left of it... That way you can type any URL you want, but also you can click this button to navigate using the ObjectBrowser...

For sure, let me know if you are meeting, I could ask our UX/UI guy to join me and share some ideas!

frapell:
@sneridagh Also, btw, I think that when you highlight an existing link, you should get under the block sidebar all of the different controls for it, like the target, title, WCAG, etc... Something like the Image block has...

sneridagh:
@frapell I almost forgot! This is the PR with RT folks current work: https://github.com/plone/volto/pull/1223 

We should definitely sum efforts!

frapell:
@sneridagh LOL, seems they indeed added the navTreeSVG to the AddLinkForm XD

I guess we can discard this then, and continue on that branch ;)

sneridagh:
It’s great to have different PoCs, so we can decide having different options.

Anyways, I will contact Albert, let’s see what’s his take on this. Then we can schedule a meeting together to decide the final approach.

/cc @pnicoli @tisto @cekk @giuliaghisini

PR #1244, opened 2020-03-02
translation for Japanese

terapyon:


tisto:
@terapyon @sneridagh we had the same problem with the German translations. It would be great if we could add a check for untranslated strings that at least warns us if there are missing translations. I would be more than happy to take care of the German translations and make sure that we always have 100% of all strings translated. 

@robgietema is something like this possible with our current i18n setup?

PR #1253, opened 2020-03-04
Toc cypress

steffenri:


cypress/integration/blocks.js
```diff
@@ -596,5 +596,35 @@ if (Cypress.env('API') !== 'guillotina') {
     //   //   cy.contains(expected);
     //   // }
     // });
+
+    it.only('Add Table of Contents block', () => {
+      const h2 = 'expectedH2';
```
tisto: @steffenri I prefer to be explicit here and not use a variable.

cypress/integration/blocks.js
```diff
@@ -596,5 +596,35 @@ if (Cypress.env('API') !== 'guillotina') {
...
+
+      //Add H2
+      cy.get('.block.inner.text .public-DraftEditor-content')
+        .type(` ${h2}`)
+        .setSelection(` ${h2}`);
```
tisto: @steffenri why is .setSelection necessary here?

cypress/integration/blocks.js
```diff
@@ -596,5 +596,35 @@ if (Cypress.env('API') !== 'guillotina') {
...
+        .click();
+      cy.get('.ui.basic.icon.button.toc')
+        .contains('Table of Contents')
+        .click();
+
+      //Save
+      cy.get('#toolbar-save').click();
+
+      //View
+      cy.get(' a').contains(` ${h2}`);
```
tisto: @steffenri this checks any "a" element. We should check the output of the TOC block for the h2 headline.

cypress/integration/blocks.js
steffenri: using .setSelection here allows us to style in draftJS. In this case we can make the selected text a h2.

cypress/integration/blocks.js
steffenri: changed.

cypress/integration/blocks.js
steffenri: Fixed that aswell. We now check for an a element in the list that gets created by the TOC Block. We also check for the exact header type.

Issue #1256, opened 2020-03-05
Add possibility to get initialBlocks and initialBlocksLayout from backend


avoinea:
Add possibility to get **initialBlocks** and **initialBlocksLayout** from backend when adding new content in order to let Site Administrators to dynamically define content default blocks together with DX content definition. 

**Plone Dexterity Fields**
![dx](https://user-images.githubusercontent.com/323385/75970039-76f5f980-5ed8-11ea-9172-bb84365e68bb.png)

**Volto Add Form**
![default](https://user-images.githubusercontent.com/323385/75970042-78272680-5ed8-11ea-8060-f4a62b262a74.png)

avoinea:
Ok, maybe because we didn't make it clear from the beginning what we are trying to achieve with this, the confusion.

Our main **requirement** from **Volto** or a **Volto Add-on** right now is to be able to define Dexterity Content-Types TTW, as in the real world **one page content-type with Title** as the default block doesn't fit all cases and customers need content with more metadata, they need to define content initial blocks and maybe some required ones. They need to easily change logo, labels, placeholders, default values, etc (not a developer's job, if you ask me)

For this:

1. We need Dexterity within Volto > ControlPanel ASAP
2. We are going to extend this panel with a Volto TTW Dexterity layout editor (which may end-up in the core or as a Volto Add-on, to be discussed).

> Excerpt from the **Zen of Volto**:
>
> That means this someone has to be able to run, test and use Volto (and thus Plone), and ultimately customize a simple project without any problem and without having to touch Plone by any mean or write a single line of Python **nor implement a single Plone add-on product.**

With this in mind, as Dexterity Content-Types already provide the **default values** feature, **without having to write a single line of Python code within a custom Plone Add-on**
we think it is the best place were **initialBlocks** and **requiredBlocks** should come from. Of course nobody will have to edit this values manually and there will be a nice Volto Control-Panel interface for it.

Also, within current Volto implementation of **initialBlocks** and **requiredBlocks** one can NOT provide blocks default values, custom placeholders, etc. Not mentioning the fact that if you add a new Dexterity-Type on backend you need to hire a JavaScript Developer to define the **initialBlocks** and the **requiredBlocks** for it. And how is the front-end back-end agnostic if it requires to re-deploy front-end every time you create a new Dexterity Content-Type on backend?

My feeling is that this settings should stay together with the content-type definition.

**One last thought:**

[...]

Issue #1024, opened 2019-12-02
Default blocks per content type, configurable


sneridagh:
So you can have default blocks per content type configurable via the config object.

PR #1257, opened 2020-03-05
Refs #1256 - Get default blocks from backend otherwise fallback to code defined blocks

avoinea:
See #1256


sneridagh:
@avoinea I don't get it, where is the list coming from the backend?

sneridagh:
@avoinea oh I see, but then, in p.restapi we need to add them. But you can't achieve a per content type default controlled by the backend using the behavior, right? or I am missing something?

avoinea:
@sneridagh Well right now, for testing, I disabled `Blocks behavior` and manually added `blocks` and `blocks_layout` JSON fields. This is one option.

The other would be to add within **plone.restapi** either another behavior called `Blocks (with defaults)` or update existing one and add the possibility to `Save defaults` on these fields.

sneridagh:
@avoinea I think, there's a flaw in that implementation, since the default values set in Plone will have always the same block UUIDs, as they are statically defined.

I'm a bit torn on this one TBH, while developing Volto we try to have always present not to fall in the same pitfalls that we did previously in Plone in the past, also avoiding too much coupling with the backend part. So the question here is, what value gives us the ability to define the default blocks in the backend? Does it make the development process easier, or forces you to go to the backend to do so. If we leave both approaches in place, aren't we falling in the same pitfall that we fell so many times in the past in Plone itself, where we have two different ways to do things? Then you have to document them both, and explain the user which is the best.

Two examples: that's why we opted to define the blocks in the front, and release Plone (and the devs) from having to do so. 

This philosophy has to be written and explained, so everybody knows why we took that decisions and the reasoning behind. These are a quick summary of the main ones:

- Approachability (for newbies and for people that are completely aliens to Plone)
- Developer experience simplified to the max
- Do not enable more than one way of doing things

We can talk about it in the next meeting, or schedule a meeting for this, so we are all in the same page.

Thoughts? /cc @robgietema @tisto @pnicolli @cekk

demarant:
@sneridagh 
I agree, the UUID should be made unique on content creation. The solution is a proof of concept and would need further work.

It would be great to have a meeting with you and see on how we can easily contribute to the new Plone UI, which looks promising. 
As you may already know, EEA is developing many websites based on Plone and we would like to use our experience on large websites development to develop Volto further, make it robust, flexible, secure and feature rich. 

This can of course be done either via custom addons or by contributing to the main product. I see the latter as a better option for everybody if the features are generic enough and fit the volto main goals.

I find having a dexterity editor and volto-templating editor TTW a very empweringl feature for site administrators and early adoptors. Would this be in line with the Volto roadmap? We could save lot of energy currently spent on creating templates in nodejs code that can just be automated.

avoinea:
@sneridagh 
Within my last commit I fixed the fallback on code if schema.defaults is empty (Blocks behavior enabled) and generate new uids for blocks.

datakurre:
Could this be achieved with a custom (or configurable) context aware zope.schema default value factories or or z3c.form default value adapters for blocks and blocks_layout -fields instead?

sneridagh:
As said, we are preparing "the Zen of Volto" writing, you can see the ongoing PR here:
https://github.com/plone/volto/pull/1263

Now we already fulfill a feature defined by https://github.com/plone/volto/issues/1024:
Ability to define initial blocks per content type.

And it works simply by adding this object in the config (see https://docs.voltocms.com/blocks/settings/#initialblocks-initial-blocks-per-content-type):

```
const initialBlocks = {
    Document: ['leadimage', 'title', 'text', 'listing' ],
    Event: ['title', 'description', 'event', 'text']
};
```
I can't see anything more simple than this.

This approach enables two ways of doing things in Volto. If it's not estrictly necessary or gives a fundamental advantage, I would refrain of doing it.

@datakurre for sure we can do a fancy adapter or modifying behaviors to be configurable per content type or find another solution for the backend. But hey, it's working now, and it's simple. Simple to implement, simple to teach. Look at the documentation: it's three lines to explain the concept.

@avoinea What advantage do you have by defining it in the backend, instead of doing it in the frontend with the configuration object? You had to amend what's coming from the backend to make it work properly, so the backend (in the shape that it is now) don't fulfill your requirements, so you have to amend the uuids in the front. 

Currently, the frontend is already taking care of this:
https://github.com/plone/volto/blob/master/src/components/manage/Add/Add.jsx#L108

avoinea:
I think we should continue this discussion on the dedicated ticket. 

Thus see my comment on #1256

PR #1263, opened 2020-03-07
internal: zen of Volto philosophy

sneridagh:


docs/source/contributing/index.md
```diff
@@ -1,5 +1,74 @@
...
+# Contributing guidelines
 
-!!! note
-    This documentation is a work in progress. Any help is welcome to fill in the
-    gaps!
+Since the very beginning, Volto has been developed with several directives in mind,
+brought from the years of experience developing Plone core and implementing projects on
+it.
+
+## Prime directive
```
tisto: @sneridagh Star Trek Nerd! :)

sneridagh: :)

datakurre:
While I agree with the principles, the current version looks too much like a rant to me. Would it be possible to somehow separate the principles from reasoning – to hide the unnecessary Plone rant?

What I am missing here are the limitations of the front end: what kind of features remain out of the scope of Volto and should be implemented in the back end (and possibly exposed with REST API). I have a feeling that this was not completely clear in Plone vs. ZMI, causing confusing portal management experience. (I still believe in what I said at Plone Conf: That we should considering "embracing" ZCA so that the main user interface could be get as simple as possible, but we would still have a place for [product killing checkboxes](https://limi.net/checkboxes).)

sneridagh:
@datakurre After reading it again, I agree with you. The rant should be separated from the main concepts. However, I feel that a bit of rant is necessary when explaining the reasoning behind. 

How much separated do you think they should be, like in another page, or in the same page, after the list?

datakurre:
@sneridagh Same page is good enough, but try to figure out a such subtitle for it that only Plone-developers will continue to read :)

docs/source/contributing/index.md
```diff
@@ -1,5 +1,75 @@
...
-    This documentation is a work in progress. Any help is welcome to fill in the
-    gaps!
+Since the very beginning, Volto has been developed with a set of principles in mind,
+brought from the years of experience developing Plone core and implementing projects on
+it.
+
+## Approachability
+
+Volto has to be 100% approachable by someone with zero Plone or Python knowledge. This
+is the main Votlo principle.
```
fulv: Volto

docs/source/contributing/index.md
```diff
@@ -1,5 +1,75 @@
...
+
+That means this someone has to be able to run, test and use Volto (and thus Plone), and
+ultimatelly customize a simple project without any problem and without having to touch
+Plone by any mean or write a single line of Python nor implement a single Plone add-on
```
fulv: means

docs/source/contributing/index.md
```diff
@@ -1,5 +1,75 @@
...
+product. This person has to be able of using only JavaScript to parameterize it.
+
+The rationale behind this is to be able to attract people from the React community or
+other communities and drag them to use and know Plone using Volto as a facilitator. This
+way we keep the existing learning curve coming from Plone away from Volto, so Volto can
+make (and control) from scratch its own learning curve, without any inherited burden.
+
+A number of decissions has been made based on this directive. To give an example, that's
```
fulv: decisions

fulv: have

docs/source/contributing/index.md
```diff
@@ -1,5 +1,75 @@
...
+why it was decided not to use the tiles implementation and definitions existing in the
+Plone ecosystem. We started to do it, but all of a sudden it felt like duplicating
+efforts by defining them on both sides, while clearly, having them defined on Plone
+didn't give any value.
+
+## Developer experience first
+
+Developing in Volto has to be fun, not a burden. Implementing or customizing anything in
+Volto (or in a Volto project) should be easy and straightforward, and done in a simple
+single way.
```
fulv: single, simple way.

docs/source/contributing/index.md
```diff
@@ -1,5 +1,75 @@
...
+
+Developing for Plone has become a burden, even doing the most simple thing often implies
+teaching several other things and technologies, and explain the reasoning (and the
+history!) behind why is like that, and worse, often say sorry about it. Those who teach
```
fulv: why it is like that

docs/source/contributing/index.md
```diff
@@ -1,5 +1,75 @@
...
+Plone outhere know it's like this. Almost twenty years of history, thousands of
```
fulv: out here / out there... two words either way

docs/source/contributing/index.md
```diff
@@ -1,5 +1,75 @@
...
+## Do not overengineer things
+
+In the Plone community we had a tendency to overengineer things. We have the chance to
+stop doing it. While developing Volto we live in the Web Technologies world, let's use
+them in a simple and straightforward way. Do not try to implement ZCA, Portlets or five
+way adapters for widgets in JS.
+
+## Each feature has only one way of doing it
+
+We won't introduce (at least willinglly) several ways of doing the same thing. Then you
```
fulv: willingly

docs/source/contributing/index.md
```diff
@@ -1,5 +1,75 @@
...
+have to document them both, and explain the user which is the best. Specially on the
```
fulv: explain to the user
Especially in cases where you something

docs/source/contributing/index.md
```diff
@@ -1,5 +1,75 @@
...
+cases that you can do a thing in the backend or in the frontend.
+
+## Frontend and backend are meant to be decoupled
+
+So let's them keep them decoupled.
+
+We will face in the future while developing Volto and in our Volto projects a lot of
```
fulv: In the future, while developing Volto and in our Volto projects, we will face a lot of decisions and answer questions like:

docs/source/contributing/index.md
```diff
@@ -1,5 +1,75 @@
...
+decisions and answer to questions like:
+
+- What is the value to define something in the backend?
+- Is that going to improve the (frontend) developer experience?
+- Separating this is going to give any good or will force us to repeat ourselves?
```
fulv: Is separating this going to do any good or will it force us to repeat ourselves?

docs/source/contributing/index.md
```diff
@@ -1,5 +1,75 @@
...
+- or worse, it would take two people (frontend dev and a backend dev) to accomplish a simple task?
+- Are we introducing two different ways of doing things?
+
+## Focus on the UI/UX implementation and upcoming challenges
+
+Volto is a frontend UI implementation of a CMS backends. We need to focus on this
```
fulv: backend

docs/source/contributing/index.md
```diff
@@ -1,5 +1,75 @@
...
+instead to focus on the relationship with the backend(s). JS world advances quick, and
```
fulv: instead of focusing on

docs/source/contributing/index.md
```diff
@@ -1,5 +1,75 @@
...
+we have to keep its pace. React's own async (Suspense) mode it's on its way, and that
```
fulv: keep pace
or
keep pace with it

sneridagh:
@fulv Fixed!! Thanks a lot! Any feedback?

tiberiuichim:
I'll try to throw some ideas into the discussion:

I think the extensibility story, third party addons and community is missing from this manifesto, and it's important. The perceived simplicity of "customizing a single thing in Volto should be easy" is fake, unless there are good ways offered by Volto to avoid maintenance burden and to allow code sharing (even packaging Volto customizations as addons). If Volto goes down the route of "z3c.jbot customizability is the default", I fear there won't be a community of addons and Volto's value as a CMS will not reach its well-deserved height.

The problem, as I perceive it, is that Volto doesn't have enough "insertion points" and it forces, in many cases, something that I consider an anti-pattern when it comes to maintaining codebases: creating a copy of a file and customizing it. It is possible to package volto customizations in addons, we've done that, but each add-on can potentially introduce conflicts with other addons. It also makes maintaining both the addon, by the addon author but also the final published website, more difficult then it should be: if Volto tweaks a file, then the addon author needs to watch and maintain the tweaked files and so on. Not pretty, IMHO. Of course, it's also difficult to pinpoint which areas should be improved and "opened" with insertion/extension points. See maybe this PR as an example: https://github.com/plone/volto/pull/1078

I think internally Volto should be structured so that it's easy to reuse parts of it through composition and inheritance, so that code can be shared and reused in addons. As an example I can point to React Chart Editor. Taking random files as example, see how the components are strongly divided in [presentation components](https://github.com/plotly/react-chart-editor/blob/master/src/default_panels/StyleColorbarsPanel.js), which are dumb, devoid of any heavy logic or methods and [base/basic components](https://github.com/plotly/react-chart-editor/blob/master/src/components/fields/Dropdown.js#L7) with inner logic, but that can be extended. Notice how the Dropdown is exported as both UnconnectedDropdown and its default "connected" version.

[Composition is used](https://github.com/plotly/react-chart-editor/blob/master/src/lib/connectToContainer.js):  (maybe not the best example, but I'm not very familiar with the codebase), but so is [inheritance](https://github.com/plotly/react-chart-editor/blob/master/src/components/containers/PlotlySection.js#L40).

So, basically, if composition is to be used and preferred for developing, then Volto's presentation layer needs to be properly separated from logic and both need to be reusable. I'm gonna give you an extreme example, but suppose I want to build a new type of editing form (something that I imagine, are interesting use cases). Then, ideally, Volto's [Form.jsx](https://github.com/plone/volto/blob/master/src/components/manage/Form/Form.jsx) should be packaged so that its internal logic can be reused as much as possible. The discrimination mechanism used in determining the layouts and widgets should also be used for the edit form implementation.

datakurre:
@tiberiuichim Please, take a look at https://www.gatsbyjs.org/docs/themes/what-are-gatsby-themes/

GatsbyJS chose jbot approach and are getting pretty far with it. That managed to make it possible to extend components while overriding them. While I agree with your concerns, I must also admit that with properly granular components, GatsbyJS approach works really far without complicating the the architecture.

sneridagh:
> I'll try to throw some ideas into the discussion:
> 
> I think the extensibility story, third party addons and community is missing from this manifesto, and it's important. The perceived simplicity of "customizing a single thing in Volto should be easy" is fake, unless there are good ways offered by Volto to avoid maintenance burden and to allow code sharing (even packaging Volto customizations as addons). If Volto goes down the route of "z3c.jbot customizability is the default", I fear there won't be a community of addons and Volto's value as a CMS will not reach its well-deserved height.

I agree that the extensibility story should also be taken into account and should also have its own section in this manifesto. However, and this in Plone core is true, Volto (core) should not be at the service of the add-ons, nor compromise any of the other manifesto points because of that. 

Regarding the "component shadowing" is the default... What other options do we have in JS world, and more specific, dealing with JSX? (I agree on you later on make the components smaller and separate logic from presentation).

One thing to think about it and we have to have it always present in our minds. I don't know any other framework in the JS scene that tries to do the same extensibility goal as we try to accomplish, apart from Gatsby. Even Gatsby does not have the level of starting point complexity as we have (with all CMS features in place). So the lack of our extensibility needs use case in the JS community makes even more challenging what we are trying to achieve. I'm sure that we will find the way through and get a way to excel in this point.

> 
> The problem, as I perceive it, is that Volto doesn't have enough "insertion points" and it forces, in many cases, something that I consider an anti-pattern when it comes to maintaining codebases: creating a copy of a file and customizing it. It is possible to package volto customizations in addons, we've done that, but each add-on can potentially introduce conflicts with other addons. It also makes maintaining both the addon, by the addon author but also the final published website, more difficult then it should be: if Volto tweaks a file, then the addon author needs to watch and maintain the tweaked files and so on. Not pretty, IMHO. Of course, it's also difficult to pinpoint which areas should be improved and "opened" with insertion/extension points. 

I think that the problem you are mentioning happens also in Plone (even using ZCA). Think about how do you customize things in Plone. If you use zc3.jbot or override via zcml and layers a template or the related logic, you can't customize it again from other add-on, or you'll get a conflict error. Upgrade Plone and might be that your customized code (although working) is not updated. Might be that two add-ons in Plone does not get along well. Same applies for Volto.

I see your point anyways, I agree that we don't have enough insertion points AND our components are gigantic (as you mention later with the Form component, see comment later). They are two different problems that we have to deal with them differently.

Adding insertion points upfront before the use cases arrives is terribly difficult. I tried with the Toolbar component, and after 8 months thinking about how its shape and architecture should be, I finally opted by laying down the architecture for making it extensible easily, but I had to left it half way, waiting for the use cases to arise.

I'm +1 to add those insertion points as they appear, more than prematurely add them without knowing if they are going to be used at all. I can think on a few examples in Plone codebase where very complex implementations were laid down, and they are used nowhere else.

> See maybe this PR as an example: #1078

I won't call this an insertion point but more a logic reusability improvement. As said, I'm +1 to talk about it, but I would like @robgietema and others weigh in before.

> 
> I think internally Volto should be structured so that it's easy to reuse parts of it through composition and inheritance, so that code can be shared and reused in addons. As an example I can point to React Chart Editor. Taking random files as example, see how the components are strongly divided in [presentation components](https://github.com/plotly/react-chart-editor/blob/master/src/default_panels/StyleColorbarsPanel.js), which are dumb, devoid of any heavy logic or methods and [base/basic components](https://github.com/plotly/react-chart-editor/blob/master/src/components/fields/Dropdown.js#L7) with inner logic, but that can be extended. Notice how the Dropdown is exported as both UnconnectedDropdown and its default "connected" version.
> 
> [Composition is used](https://github.com/plotly/react-chart-editor/blob/master/src/lib/connectToContainer.js): (maybe not the best example, but I'm not very familiar with the codebase), but so is [inheritance](https://github.com/plotly/react-chart-editor/blob/master/src/components/containers/PlotlySection.js#L40).
> 

I'm +1 as said before, as for now, we tried build Volto to make things work first, and deal with other secondary issues later. We have to start thinking in refactor to separate (as much as possible) logic from presentation. However, in JSX and React this might be challenging with some use cases...

> So, basically, if composition is to be used and preferred for developing, then Volto's presentation layer needs to be properly separated from logic and both need to be reusable. I'm gonna give you an extreme example, but suppose I want to build a new type of editing form (something that I imagine, are interesting use cases). Then, ideally, Volto's [Form.jsx](https://github.com/plone/volto/blob/master/src/components/manage/Form/Form.jsx) should be packaged so that its internal logic can be reused as much as possible. The discrimination mechanism used in determining the layouts and widgets should also be used for the edit form implementation.

The Form component implementation has a reason of being like it is now. I'm the first one that I would like to have (and others, including Asko and Ramon, because of Gatsby and Angular) the Pastanaga Editor stripped out as a reusable component, the simplify the form compoent. We will reach to that, but with the sufficient amount of time and resources availability.

We are reaching the point in Volto maturity when we have to polish what has been done, and refactor whenever needed.

Extensibility story is huge, and get it right from the beginning is extremely difficult. Luckily, we have the experience of being developing Plone since 2001, facing and solving the same issues we have in Volto now. I myself would like also to have full CSS framework independence at some point!

I'm +1 also to have a meeting and continue discussing all this with you at any time!

tisto:
Let's schedule a call to discuss this...

tisto:
I tend to merge this now. I think it is a good base for future discussions at a sprint or a hangout and we spend a good amount of time on discussions. We can always change things at a later point...

Issue #1175, opened 2020-02-06
Sorting does not work in folder_contents view


tisto:
0) Login as admin
1) Go to https://volto.kitconcept.com/contents
2) Sort the folder_contents view by any way
-> nothing happens
-> it seems no sort_on param is sent by react

pnicolli:
Possible duplicates or similar issues:
https://github.com/plone/volto/issues/1035
https://github.com/plone/volto/issues/940

PR #1269, opened 2020-03-10
Sorting in folder contents view /contents

ksuess:
Fix for https://github.com/plone/volto/issues/1175

Sorting was prepared well. I just filled a gap.


src/components/manage/Contents/Contents.jsx
```diff
@@ -1133,7 +1140,12 @@ class Contents extends Component {
                     <Table.Row>
                       <Table.HeaderCell>
                         <Dropdown
-                          trigger={<Icon name="sort content ascending" />}
+                          trigger={
+                            <Icon
+                              name="sort content ascending"
+                              data-cy="sort"
```
sneridagh: @tisto what's the best practice here for marking an element for acceptance testing?

sneridagh: @ksuess I remember Timo saying that using `data-` attributes is not the best approach for this, let's see what he says, not sure either that using ids is the best accepted approach either.

src/constants/Indexes.js
```diff
@@ -4,15 +4,23 @@
...
+    label: 'Publication date',
+    type: 'date',
+    sort_on: 'effective',
+  },
   review_state: { label: 'Review state', type: 'string' },
-  id: { label: 'ID', type: 'string' },
+  id: { label: 'ID', type: 'string', sort_on: 'id' },
   ExpirationDate: { label: 'Expiration date', type: 'date' },
-  CreationDate: { label: 'Created on', type: 'date' },
+  CreationDate: { label: 'Created on', type: 'date', sort_on: 'created' },
```
sneridagh: At some point, we will have to address the i18n problem with this literals...

cypress/support/commands.js
```diff
@@ -23,23 +23,28 @@ Cypress.Commands.add('autologin', () => {
 Cypress.Commands.add(
   'createContent',
   (contentType, contentId, contentTitle, path = '') => {
-    let api_url;
+    let api_url, auth;
```
tisto: @ksuess I am wondering if we shouldn't use the autologin keyword to login before calling createContent. It crossed my mind as well to add auth to createContent directly. Though, we have to take into account that we will have different username/pw combinations in the future. I am a bit undecided on this...will think about it... :)

cypress/support/commands.js
```diff
@@ -78,17 +80,14 @@ Cypress.Commands.add(
         },
       });
     }
-    if (contentType === 'Document' || contentType === 'News Item') {
+    if (['Document', 'News Item', 'CMSFolder'].includes(contentType)) {
```
tisto: @ksuess nice improvement. Where does "CMSFolder" come from?

cypress/support/guillotina.js
```diff
@@ -19,14 +19,15 @@ export function setupGuillotina() {
...
 
-  cy.request({
-    method: 'POST',
-    url: `${api_url}/container/@addons`,
-    headers,
-    body: { id: 'dbusers' },
-  }).then(response => console.log('dbusers add-on installed'));
-
   // guillotina_dbusers is no longer installed in the guillotina_cms docker image
+  //
```
tisto: @ksuess can you elaborate on why we don't need this any longer. Just wondering...

src/components/manage/Contents/Contents.jsx
tisto: @ksuess @sneridagh data-cy is recommended by the Cypress docs. I am not sure if I agree personally on this. I tried something similar with Robot Framework and it kind of felt wrong to add something for testing purposes only. We risk to test the wrong this. The right thing to do is always to rely on what an actual user would rely on, e.g. 'click button labled "Save"'. data-cy is kind of the last resort if nothing else works. I guess this is something we have to figure out for Volto in general. Nothing that would prevent this PR to being merged though. I promise to think about it any maybe I find time to write down some Cypress best practices.

src/components/manage/Contents/Contents.jsx
```diff
@@ -1150,16 +1162,17 @@ class Contents extends Component {
                                 'EffectiveDate',
                                 'CreationDate',
                                 'ModificationDate',
-                                'Type',
+                                'portal_type',
                               ],
                               index => (
-                                <Dropdown.Item key={index}>
+                                <Dropdown.Item key={index} data-cy={index}>
```
tisto: @ksuess this might be fragile. An index is a number that is subject to change if the items in the dropdown change.

src/components/manage/Contents/Contents.jsx
```diff
@@ -1150,16 +1162,17 @@ class Contents extends Component {
...
                                   <Icon name="dropdown" />
                                   <FormattedMessage id={Indexes[index].label} />
                                   <Dropdown.Menu>
                                     <Dropdown.Item
                                       onClick={this.onSortItems}
-                                      value={`${index}|ascending`}
+                                      value={`${Indexes[index].sort_on}|ascending`}
+                                      data-cy={`${Indexes[index].sort_on}|ascending`}
```
tisto: @ksuess can't we just rely on "value" here instead of "data-cy"? I mean we already have a way of getting this exact Dropdown.Item, why add a second one? I know the folks at Cypress recommend this. Though, this is the exact reason why I don't like their recommendation. It makes people think that they should always rely on data-cy attributes (which is a bad thing IMHO), not use them as a last resort.

ksuess:
I prefer to clean up and react on your comments. Give me the monday.

tisto:
@ksuess fine with me. Take as much time as you like. I just did not want to reject the merge because of issues that we haven't properly solved as a community yet. Your fix is important, so we should ship it with the next release.

src/components/manage/Contents/Contents.jsx
ksuess: data-cy: replaced by additional class name.
See "Cypress best practices" #1298

src/components/manage/Contents/Contents.jsx
ksuess: index iterates 
                              [
                                'id',
                                'sortable_title',
                                'EffectiveDate',
                                'CreationDate',
                                'ModificationDate',
                                'portal_type',
                              ]

cypress/support/guillotina.js
ksuess: unrelated to this PR: reverted

cypress/support/commands.js
ksuess: TODO but not related to this issue: Reduce Cypress command "createContent" by authorization. A use of Cypress command autologin in a test should make 
      auth = {
        user: 'root',
        pass: 'root',
      };
in cy.request() superfluous. 
In this PR I just refactored, but did not change the Cypress command "createContent".
-> New Issue
-> Add task to Issue "Cypress best practice"

cypress/support/commands.js
ksuess: CMSFolder is for Guillotina tests

src/components/manage/Contents/Contents.jsx
ksuess: index iterates 
                              [
                                'id',
                                'sortable_title',
                                'EffectiveDate',
                                'CreationDate',
                                'ModificationDate',
                                'portal_type',
                              ]

ksuess:
After your reviews, @tisto, @sneridagh, I changed what I think could and should be changed in here. As mentioned above, there are topics that are moved to other more specific issues.

Please review again, thanks.

cypress/support/guillotina.js
tisto: Your revert brought back the commted out code block.

cypress/support/commands.js
tisto: I'd prefer if we would leave this out of this PR and handle that in a separate issue.

tisto:
@ksuess if you could address the two issues I raised, we can merge.

cypress/support/commands.js
ksuess: I understand that it is not obvious. In this PR I just refactored, but **did not change** the Cypress command "createContent". Only difference to before: I added Guillotina login.

tisto:
@ksuess pls rebase and then let's merge.

tisto:
@ksuess seems to me like the multilingual PR somehow got into this PR as well. Did you rebase from that branch maybe?

Issue #1272, opened 2020-03-12
UX: + sign disappears in desktop when sidebar is open


fredvd:
If an editor tries to edit their first page in Volto by creating a new block,
* and the sidebar is open, which it is by default
* and their browser window is not in full screen but less than 85% width of their laptop screen (testing this on a MacbookPro 2019 where I had my browser not set to fullscreen)

* The (+) button to select a different block type on the empty text block is gone. With text in the text block the drag handle does move inside the text block, like the delete button on the right. 

This is a full width image from my laptop screen:

<img width="1434" alt="no-plus-desktop-size-sidebar" src="https://user-images.githubusercontent.com/394784/76538196-3883b000-647f-11ea-9c6c-90e7a276efd6.png">

There a number of other UX issues with the block selector mentionned in #682 . one of the suggestions there is to move the (+) below the empty/ephemal text block.

sneridagh:
Fixed using the new add button location below the block. This is enabled by default in Volto 18 and was optional in 17.

PR #1273, opened 2020-03-12
Move block selector into the empty text block on smaller screens

fredvd:
When sidebar is open. See #1272 

This at least makes the block selector visible again similar with how the drag handle and the delete icon are shifted when it's required.

Technically I had to add a new offset parameter (addoffset), the drag & delete icons are handled relatively, but the (+) icon is using absolute positioning. The BlockChooser is at the moment shifted with the same amount.

The current nested -> .contentWitdh() -> .contentwidthMedia() less functions are a bit problematic with fine tuning the icon positions.  If you look closely to the trash icon, the mobile shift is not exactly right, the \@offset variable is used by both the drag handle & the trashcan. For every element we'd need to pass in a new variable.  But maybe 3 variables is enough, there's only limited space for more icons :-P

I would like to shift the Draf.js Placeholder text a bit (15px?) more to the right so that the (+) can be moved completely inside the editor line, but I cannot find the place where to edit this. As soon as someone starts typing the (+) button disappears again, then the actual text can be at the beginning of the editor line again. 

Current WIP position of the (+) on mobile. Can anybody with more semantic uid experience fix this a bit more? 

block chooser icon visible wip

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 - Improve the UX of the listing block when queries are running @sneridagh
 
+- Move block selector into the empty text block on smaller screens when sidebar is open, partially fixes #1272 @fredvd
+
 ## 4.0.1 (2020-03-09)
 
 ### Changes
--- a/api/pyvenv.cfg
+++ b/api/pyvenv.cfg
@@ -1,0 +1,3 @@
+home = /usr/local/bin
+include-system-site-packages = false
+version = 3.7.6
--- a/theme/themes/pastanaga/elements/container.overrides
+++ b/theme/themes/pastanaga/elements/container.overrides
@@ -15,7 +15,7 @@
     margin-right 0.3s cubic-bezier(0.6, -0.28, 0.735, 0.045);
 }
 
-.contentWidthMedia(@width, @gutter, @offset) {
+.contentWidthMedia(@width, @gutter, @offset, @addoffset) {
   .ui.container {
     width: @width !important;
     margin-right: @gutter !important;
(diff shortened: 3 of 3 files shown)


tisto:
@fredvd This is really difficult to fix IMHO. If we move the + icon inside the field, it overlaps with the field itself. This is a UX issue. Maybe @albertcasado can weight in...

fredvd:
To clarify, this is how the empty text field currently looks (before this pull request) in volto 4.0.2+ master below a certain size of the browser window on deskop mode, the (+) is completely gone.

add block icon missing

fredvd:
> If we move the + icon inside the field, it overlaps with the field itself. This is a UX issue.

I don't know to which element the border belongs, you could put the (+) choose block inside the text border and the draft.js editor next to it, then they don't overlap. The 'draggable' is doing the same thing ATM on the left side. If you start typing the first character will hide the add block icon and the draft.js editor can fill up the whole line. 

But it's still ugly-ish, the real problem is the design to maximise editor line width and hide the left gutter margin where these icons are (block selector, draggable & trashcan on the right)

tiberiuichim:
This issue has dragged on too much and it's a major usability issue, as far as I'm concerned. Even I, as a developer, struggle with this issue whenever I develop on my laptop. The current situation now (the block chooser + button is hidden) is even worse then any overlap or ugly things showing up at that resolution. I propose that we merge this. 

Alternatively, on the low res screens we can use something along the direction of block decorations used in volto-columns-block. See https://raw.githubusercontent.com/eea/volto-columns-block/docs/docs/edit.png

rioksane:
Just bumped into this issue with our users. It does appear, that even in 1200px wide resolution the Plus icon is not visible. So that excludes tablets and some monitors in a vertical setting, some laptops too. I too find this a rather major usability issue (that could be quite easily improved). 

Quick and dirty fix could be just add massive z-index to the plus icon, like 
.ui.basic.icon.button.block-add-button {
  z-index: 100;
} 

Even that would be better than no icons visible.

Issue #1267, opened 2020-03-09
Link inside text for content-type without blocks


giuliaghisini:
This error seams related to this problem https://github.com/plone/volto/pull/1228/files. 
I think the problem is in  WysiwygWidget.
These are the steps to reproduce the problem:

- Add a new Event (or ContentType without blocks)
- in Text field, write some text
- select a word and click on link button to add a link
- type something (f.e. 'Events') to choose an internal link
- view the console: an error is fired. It seems that it cannot find a router. 
A different error is also fired if you write an external link. 

This is blocker because content-types without blocks cannot be used, and migrated contents with links inside text will generate errors on edit. 

Please fix it asap.

sneridagh:
@giuliaghisini @cekk It would be great if we write a Cypress test for normal RichText fileds.

PR #1276, opened 2020-03-14
Fix for #1267 - Link inside text for content-type without blocks brea…

sneridagh:
…ks the site

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ### Changes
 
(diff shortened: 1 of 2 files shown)


sneridagh:
Fixes #1276 

/cc @giuliaghisini @cekk @pnicolli

sneridagh:
@giuliaghisini @cekk @pnicolli I think now we are safe, tell me if there are any other issue.

giuliaghisini:
> @giuliaghisini @cekk @pnicolli I think now we are safe, tell me if there are any other issue.

i think now it's ok!

Issue #888, opened 2019-09-26
unable to uncheck checkboxes 


steffenri:
in volto if you try to uncheck checkboxes, its not working
its only not working graphically because if you reload the page the box is unchecked

go to https://volto.kitconcept.dev
create a new page
enable versioning
save
edit and try to disable versioning
-> issue occurs
-> the other way around its working so checking works just fine

steffenri:
seems like this problem occurs on every checkbox on the site

sneridagh:
@steffenri Thanks! This one was a hidden huge one! Fixed, ready to merge when green.

/cc @tisto

PR #1299, opened 2020-03-26
Multilingual

sneridagh:


api/base.cfg
```diff
@@ -1,11 +1,11 @@
 [buildout]
 index = https://pypi.org/simple/
-parts = instance plonesite robot-server
+parts = instance robot-server
 versions = versions
 
 extensions =
     mr.developer
-# auto-checkout =
+auto-checkout = plone.restapi
```
tisto: @sneridagh don't we need to check out a specific branch of plone.restapi here?

sneridagh: Yes, I already amended it, back to released.

src/components/manage/Toolbar/Types.jsx
```diff
@@ -32,6 +34,58 @@ const Types = ({ types, pathname }) => {
...
+          const translationsLeft = filter(
+            settings.supportedLanguages,
+            lang =>
+              !Boolean(
+                find(content['@components'].translations.items, {
+                  language: lang,
+                }),
+              ) && currentLanguage !== lang,
+          );
+          console.log(translationsLeft);
```
robgietema: Remove the console log

tisto: I somehow recall adding an eslint check for console and excluding the node files from that. Will re-visit. Those things should get catched by the CI.

api/base.cfg
```diff
@@ -1,6 +1,6 @@
 [buildout]
 index = https://pypi.org/simple/
-parts = instance plonesite robot-server
```
tisto: @sneridagh why don't we create a plonesite any longer?

sneridagh: fixed.

src/components/manage/Toolbar/Types.jsx
sneridagh: I was surprised as well, the config is:

```
    "no-alert": 1,
    "no-debugger": 1,
```

Catch console statements is in other directive. However, it's kind of tricky because it will catch also really wanted `console` statements (eg. in the server part, in the error log, and in the console messages) that we will have to disable via `disable-line` comment statements.

sneridagh:
@tisto ready this too, when p.restapi with #888 is released.

src/components/manage/Toolbar/Types.jsx
tisto: You can exclude console statements if you really need them (I did this in the server.jsx). Still odd why this wasn't caught by eslint.

tisto:
@sneridagh plone.restapi 6.5.0 released: https://pypi.org/project/plone.restapi/6.5.0#id1

PR #1365, opened 2020-04-11
Codesplittingreactselect

sneridagh:


cypress/integration/blocks-image.js
```diff
@@ -10,7 +10,7 @@ if (Cypress.env('API') !== 'guillotina') {
       cy.waitForResourceToLoad('@breadcrumbs');
       cy.waitForResourceToLoad('@actions');
       cy.waitForResourceToLoad('@types');
-      cy.waitForResourceToLoad('?fullobjects');
+      cy.waitForResourceToLoad('my-page?fullobjects');
```
tisto, line 13: @sneridagh doesn't that wait for "/my-page/my-page?fullobjects"? I mean we are on "mypage/edit", right?

cypress/support/commands.js
```diff
@@ -11,12 +11,14 @@ Cypress.Commands.add('autologin', () => {
     password = 'secret';
   }
 
-  cy.request({
-    method: 'POST',
-    url: `${api_url}/@login`,
-    headers: { Accept: 'application/json' },
-    body: { login: user, password: password },
-  }).then(response => cy.setCookie('auth_token', response.body.token));
+  return cy
```
tisto, line 14: @sneridagh just out of curiosity. Could you elaborate why returning cy.request here makes a difference?

sneridagh: It seems that Cypress expects you to return (the promise) so he can know when it’s resolved better.

cypress/integration/blocks-image.js
sneridagh: I left it like this to be more explicit to know what are we expecting but later on, I realised the command is using an ‘includes’ so it does not make much of a difference.

PR #1368, opened 2020-04-13
Cypress content refactoring

tisto:


cypress/integration/content.js
```diff
@@ -79,12 +90,18 @@ describe('Add Content Tests', () => {
         .contains('file.pdf');
     }
     cy.get('#toolbar-save').click();
-    cy.visit('/contents');
-    cy.contains('This is a file');
+
+    // then a new file should have been created
+    if (Cypress.env('API') === 'guillotina') {
```
tisto, line 95: @bloodbare @vangheem Guillotina seems to take the title of an image/file and normalizes that to the ID. Plone uses the image/file name itself for that. Any particular reason you chose to derive from Plone here?

bloodbare: We did not derive. Guillotina is not opinionated about it and guillotina cms has an adapter that sets the id to the file name in case its a File: https://github.com/plone/guillotina_cms/blob/946fe9fb6129f214b2e47be053f2665137741ba9/guillotina_cms/content/id_generator.py

Filename is received on payload['file']['filename']

PR #1371, opened 2020-04-14
Adding new User to User Control Panel

### LINKED ISSUE #1271: Newly created user is not displayed in the User Control Panel
After I added a new user via volto, I can not see this user in the volto user control panel. Also a search did not find the user. On the Backend (plone/@@usergroup-userprefs) the user is created correctly

iFlameing:
Fixes #1271

### DIFF
--- a/cypress/integration/user-control-panel.js
+++ b/cypress/integration/user-control-panel.js
@@ -1,0 +1,42 @@
+if (Cypress.env('API') !== 'guillotina') {
+  describe('User Control Panel Test', () => {
+    beforeEach(() => {
+      // given a logged in editor
+      // and a folder that contains a document
+      // and the folder contents view
+      cy.visit('/');
+      cy.autologin();
+      cy.visit('/controlpanel/users');
+      cy.waitForResourceToLoad('@navigation');
+      cy.waitForResourceToLoad('@breadcrumbs');
+      cy.waitForResourceToLoad('@actions');
+      cy.waitForResourceToLoad('@types');
+    });
+
+    it.only('Should add User to controlPanel', () => {
+      // when I added a user from controlPanel
+      cy.get('.addSVG').click();
+      cy.get('input[id="field-username"]')
+        .clear()
+        .type('iFlameing');
+      cy.get('input[id="field-fullname"]')
+        .clear()
+        .type('Alok Kumar');
+      cy.get('input[id ="field-email"]')
+        .clear()
+        .type('ialokkumarsingh0@gmail.com');
+      cy.get('input[id="field-password"]')
+        .clear()
+        .type('test@test');
+      cy.get('.arrow').click();
+
+      // then the user section must contains a fullname when I searched the
+      // same with the same username
+      cy.get('input[id="user-search-input"]')
+        .clear()
+        .type('i');
+      cy.get('.icon.button:first').click();
+      cy.get('.fullname').should('have.text', 'Alok Kumar');
+    });
+  });
+}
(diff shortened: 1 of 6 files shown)


cypress/integration/user-control-panel.js
```diff
@@ -0,0 +1,42 @@
+if (Cypress.env('API') !== 'guillotina') {
+  describe('User Control Panel Test', () => {
+    beforeEach(() => {
+      // given a logged in editor
```
tisto, line 4: Nice!

cypress/integration/user-control-panel.js
```diff
@@ -0,0 +1,42 @@
...
+      cy.visit('/');
+      cy.autologin();
+      cy.visit('/controlpanel/users');
+      cy.waitForResourceToLoad('@navigation');
+      cy.waitForResourceToLoad('@breadcrumbs');
+      cy.waitForResourceToLoad('@actions');
+      cy.waitForResourceToLoad('@types');
+    });
+
+    it.only('Should add User to controlPanel', () => {
```
tisto, line 16: You want to remove the it.only here. Happens to me all the time as well. :)

cypress/integration/user-control-panel.js
```diff
@@ -0,0 +1,42 @@
...
+      // when I added a user from controlPanel
+      cy.get('.addSVG').click();
```
tisto, line 18: We might want to add a more meaningful class to the code here so that we can use a more meaningful selector in the test.

cypress/integration/user-control-panel.js
```diff
@@ -0,0 +1,42 @@
...
+      cy.get('input[id="field-username"]')
+        .clear()
+        .type('iFlameing');
+      cy.get('input[id="field-fullname"]')
+        .clear()
+        .type('Alok Kumar');
+      cy.get('input[id ="field-email"]')
+        .clear()
+        .type('ialokkumarsingh0@gmail.com');
```
tisto, line 27: I would not recommend to use any real email address here and particularly not your own. :)

Just use "mail@example.com" or something similar.

cypress/integration/user-control-panel.js
```diff
@@ -0,0 +1,42 @@
...
+      cy.get('input[id="field-password"]')
+        .clear()
+        .type('test@test');
+      cy.get('.arrow').click();
```
tisto, line 31: Don't we have a better class here we can rely on? If not, we should add one.

cypress/integration/user-control-panel.js
```diff
@@ -0,0 +1,42 @@
...
+
+      // then the user section must contains a fullname when I searched the
```
tisto, line 33: Am I correct assuming that we only close the modal here? Then we do not have to wait for a location change, right?

cypress/integration/user-control-panel.js
```diff
@@ -0,0 +1,42 @@
...
+      // same with the same username
+      cy.get('input[id="user-search-input"]')
+        .clear()
+        .type('i');
+      cy.get('.icon.button:first').click();
+      cy.get('.fullname').should('have.text', 'Alok Kumar');
```
tisto, line 39: Is it possible to restrict that to the search results? Just in case there is another object with class="fullname".

src/components/manage/Controlpanels/UsersControlpanel.jsx
```diff
@@ -804,6 +804,7 @@ class UsersControlpanel extends Component {
                     messages.searchUsers,
                   )}
                   onChange={this.onChangeSearch}
+                  id="user-search-input"
```
tisto, line 807: There is this ongoing discussion of ids vs classes in CSS. Maybe @sneridagh wants to weight in or we want to have a brief discussion about it.

iFlameing:
ok!!!

cypress/integration/user-control-panel.js
iFlameing: yes

PR #1376, opened 2020-04-15
Add test for listing block sort by effective rev order

tisto:


cypress/integration/blocks-listing.js
```diff
@@ -66,6 +66,71 @@ if (Cypress.env('API') !== 'guillotina') {
...
+      cy.get('.blocks-chooser .common')
+        .contains('Listing')
+        .click();
+
+      //verify before save
+      cy.get(`.block.listing .listing-body:first-of-type`).contains('Page One');
+
+      // set effective date (reverse order)
+      cy.get('#select-listingblock-sort-on').click();
+      cy.get('#select-listingblock-sort-on #react-select-8-option-2-1').click();
```
steffenri, line 117: We are choosing the sort criteria here by the order. Would it be a good idea to just type the sorting criteria with .type('Effective date {enter} ') ?

tisto: @steffenri yeah. good idea. give it a shot! I guess modifying react-select is not really an option here.

cypress/integration/blocks-listing.js
```diff
@@ -66,6 +66,71 @@ if (Cypress.env('API') !== 'guillotina') {
...
+    it('Add Listing Block: sort by effective date', () => {
+      cy.visit('/my-page');
+      cy.waitForResourceToLoad('@navigation');
+      cy.waitForResourceToLoad('@breadcrumbs');
+      cy.waitForResourceToLoad('@actions');
+      cy.waitForResourceToLoad('@types');
+      cy.waitForResourceToLoad('my-page?fullobjects');
+
+      // given a page with two pages
+      cy.createContent('Document', 'page-one', 'Page One', 'my-page');
```
steffenri, line 78: The test is green but the functionality is not working so that means the test is not working as it should.
We either have to counter check the "Reversed Order" Button and the order of the results should be the other way around or create another page and make the second one the newest/oldest.

tisto: You are correct. I did not want to commit a failing test because I want to make sure we can merge it and it passes for now. We can then create a PR that shows the failing test.

PR #1415, opened 2020-04-23
Make label of checkboxwidget clickable 

### LINKED ISSUE #1414: Label in CheckboxWidget not clickable
E.g. when editing Events for the field `open_end`
The solution is to set the label in `Checkbox`: 

```jsx
  <Checkbox
    name={`field-${id}`}
    checked={value}
    disabled={onEdit !== null}
    onChange={(event, { checked }) => onChange(id, checked)}
    label={title}
  />
```

pbauer:
fix #1414

### DIFF
--- a/src/components/manage/Widgets/CheckboxWidget.jsx
+++ b/src/components/manage/Widgets/CheckboxWidget.jsx
@@ -131,13 +131,7 @@ const CheckboxWidget = ({
                 checked={value}
                 disabled={onEdit !== null}
                 onChange={(event, { checked }) => onChange(id, checked)}
-              />
-              <Label
-                basic
-                horizontal
-                as="label"
-                htmlFor={`field-${id}`}
-                content={title}
+                label={title}
               />
             </div>
             {map(error, message => (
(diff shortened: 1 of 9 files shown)


src/components/manage/Widgets/CheckboxWidget.jsx
```diff
@@ -131,13 +131,7 @@ const CheckboxWidget = ({
                 checked={value}
                 disabled={onEdit !== null}
                 onChange={(event, { checked }) => onChange(id, checked)}
-              />
-              <Label
-                basic
-                horizontal
-                as="label"
-                htmlFor={`field-${id}`}
```
sneridagh, line 139: We need the `for` because of accessibility :(
Unfortunately Semantic React does that wrong :( :(
Maybe copying the `onClick` prop in the original label is better?

pbauer: True. It is mentioned in https://github.com/Semantic-Org/Semantic-UI-React/issues/3745. 
I missed that.

pbauer: Done

PR #1418, opened 2020-04-23
Adding word limit into text area.

### LINKED ISSUE #1412: Show maximum number of allowed characters in a Volto form textfield

iFlameing:
Fixes #1412 This is just a dummy code that we have to implement for the real Textarea field.


sneridagh:
Well, that is not what we asked for.
We want a *new* widget: `TextAreaLimitLengthWidget`, different from the "normal" `TextAreaWidget`.
The error validation is not an issue, we said that is not important. Please, create the new widget and use it (you can test it in the summary (description) widget of any content type).

For more info: https://training.plone.org/5/volto/custom-widget.html

src/components/manage/Widgets/TextareaWidget.jsx
```diff
@@ -139,11 +138,11 @@ const TextareaWidget = ({
                 onChange(id, target.value === '' ? undefined : target.value)
               }
             />
```
iFlameing, line 140: I will revert this :)

sneridagh:
Okay, I took a look into this, and turns out that applying this to fields might be tricky. So, the API is able to define a schema text field with a constraint "max_lenght".

1. go to: https://volto.kitconcept.com/api/dexterity-types/Document/@@fields (or localhost:8080)
2. click on "Add new field"
Screenshot 2020-04-30 at 11 38 28

3. Add it.
4. Go to the field settings
image

add a value in max (eg. 200).
In volto schema now you should see:
Screenshot 2020-04-30 at 11 41 13

On the widget props:
image

sneridagh:
So @iFlameing sorry to change the specs again, can we revert to add this feature to the `TextAreaWidget`? (remove the new created one). So if the schema has the maxLenght set, then it's the max value accepted by the widget?

PR #1420, opened 2020-04-24
Adding expires Label

### LINKED ISSUE #1413: Folder contents view: show expired label for expired content
The folder contents view in classic Plone shows a nice "expired" label on expired content:

Deutsch (30)

Not sure if this is in @albertcasado's mocks but this is a nice feature.

iFlameing:
Fixes #1413 
Screenshot from 2020-04-24 18-50-31


tisto:
@iFlameing can we make the label a bit more subtle? Maybe use a pill or so. I guess @sneridagh should weight in as well.

sneridagh:
@iFlameing yes, please more subtle, make it "pastanaga" red-ish (like the one in the cancels). Also, aligned with the text height, not that big in length.

sneridagh:
Also, right aligned, I guess...
@tisto Just realised that we already have a problem we have to fix elegantly, which is the length of the title... try to set a big title, it breaks the whole thing....

iFlameing:
@tisto @sneridagh After changing now It looks like this.
Screenshot from 2020-04-30 18-58-45

ksuess:
I think you mark pages that are already expire**d**, don't you? Then a label "Expired" would fit.

tisto:
@sneridagh there is this react lib that restricts the number of characters and adds a "..." at the end if necessary. Maybe we should start using that. I discovered recently that there is also a CSS option for that (https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-line-clamp), not sure if that works for a single line as well.

@iFlameing pls change the label to "expired". I'd personally prefer a more pill-like appearance but this depends on Pastanaga UI I guess.

tisto:
@iFlameing pls fix "expired" and resolve conflicts.

theme/themes/pastanaga/extras/contents.less
```diff
@@ -73,9 +73,28 @@
     padding-top: 4px;
   }
 
+  .ui.table .iconAlignName {
```
sneridagh, line 76: Although it can be unconsistent in some places, but we are going to - separated CSS classnames everywhere. Can you amend that? Also keep that in mind for next ones.

iFlameing: ok

PR #1424, opened 2020-04-26
Html custom macros

nzambello:
*Proposal*
I added "macros" to Html.jsx to customize blocks in head and body.

Dev use case: I have to add hotjar or an external resource, but I don't want to customize the entire base Html component, so I can use the macro in that component.

The point is that I don't like to customize a basic component like Html (for maintainability) and I don't have to change anything but only add a part in that.
So, like Plone macros was meant, I added "macros"  (let me use this Plone specific term) to Html to allow this.

We'll have dumb components which won't be rendered using the default ones but I can customize those in my projects and Html will be kept untouched.


tiberiuichim:
If we're gonna introduce these slots, I'd rather this concept formalized into something like the viewlets system so that we can register components from the global configuration and add a bit of smartness by making the components dependent on route. Of course, a component like this could decide by itself if it wants to render or not, but it becomes more flexible (typically you don't want stuff on the View pages also rendered in the Edit pages).

I've worked a bit on this [viewlet concept here](https://github.com/eea/volto-addons/tree/master/src/Viewlets) but there was not "macro slot" declared anywhere, instead I've just relied on Portal nodes. Still, I think it would be nicer to have dedicated, specified, insertion points all over the page.

nzambello:
I guess we have two different use cases.
This was not meant for UI components, rather to add scripts, link and meta tags.
With viewlet components you're building I would render something for the UI which is more probably location related and you cannot add it in the head.

tiberiuichim:
@nzambello In Plone there is a special viewlet that allows insertion of custom resources in the <head>. 

A "slot or viewlet" component could do anything: it can render something visually, it can just ``require('some.css')``, it can render any meta tag, etc. I don't see them as different use cases.

sneridagh:
Let's continue the discussion in #1430. See: https://github.com/plone/volto/issues/1430#issuecomment-621034421

tiberiuichim:
I'll close this, as we have the AppExtras now and we're planning the slots as well.

PR #1425, opened 2020-04-27
No listing results

giuliaghisini:
Removed message for 'no results found' in view component of ListingBody for Listing block. 
On the public site it's not always nice to see the message 'no results found'.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Breaking
 
+- Listing block: removed message 'No results found' only in view mode on public site, in editMode is still present. @giuliaghisini
+
 ### Feature
 
 - /sitemap to view sitemap based on @navigation with depth 4 @giuliaghisini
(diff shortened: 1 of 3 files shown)


sneridagh:
Merged, in a recent talk with Albert, he said that it's confusing to have the message even in edit, since all the blocks editor seems so WYSIWYG, and that messages are not in the final result.

We will get to that later.

sneridagh:
@giuliaghisini I just released 5.2.0. BTW, I removed the `Breaking` label. The idea is that every time we introduce  breaking thing, we should release a major version. However, that applies to "I introduced something that will hard break existing projects", like in they won't start. For visual and UI changes... I'm a bit reluctant of marking them as breaking, since that will happen quite often... 

Maybe we need to come up with a definition of "breaking". What do you think?
/cc @tisto @pnicolli @tiberiuichim @avoinea @nzambello

nzambello:
@sneridagh I agree with your definition.
This was important but not breaking.
In fact we released a minor, not a major (following semver).

Breaking should be for something which won't make the app to start or to be working (or rendered) well.

If there is something we would like to make in evidence in the changelog, we could arrange like a label before the entry. 🤔

sneridagh:
@nzambello I also thought about the new entry too. Something like a watchout! there's a visual change from this one, even if there's a small one.

nzambello:
Yep, exactly.
It's markdown so it's formattable.

PR #1429, opened 2020-04-27
Adding image template for Listing

### LINKED ISSUE #1426: Gallery component
It would be nice to have a gallery component.

I put an eye on: https://www.npmjs.com/package/react-image-gallery

Can we search for some other options, then make a decission?

We will need to load it lazily, of course.

iFlameing:
Fixes #1426


iFlameing:
@sneridagh pure-react-carousel might be a good one. I enable the lazy load images and also the component.

tisto:
@iFlameing @sneridagh there must be something wrong with our bundlesize/bundlewatch. It reports zero change in the bundlesize even though we added a new lib.

sneridagh:
@tisto that might be because of the loadable. But the overall bundle might be bigger, for sure...

sneridagh:
LGTM, let me ask some questions first.

@giuliaghisini @nzambello @tiberiuichim @avoinea @pnicolli with this change, I guess we start having "default" listing templates being shown. What do you think?

Also, what do you think about using this library? Too much? Should we find a modest one?

nzambello:
I think the library is nice, the carousel seems to be accessible (or at least I was able to navigate it using the keyboard) and easy to customize.
I was wondering if we can override button components using some pastanaga icons instead of its defaults.

sneridagh:
@iFlameing can you change the default button (next/previous/play/pause) with Pastanaga ones? I hope the library allows you to do that.

iFlameing:
@sneridagh the icons are hardcoded in the library. I open an issue regarding that.

iFlameing:
Sorry!!! for the previous comment we can add custom icons. I have added Pastanaga Icons for navigation. Now it looks like 
Screenshot from 2020-05-08 17-40-27

PR #1438, opened 2020-04-29
Add babel and externals support in Webpack for add-on infrastructure

sneridagh:
So this is it.

I couldn't made it work with bare Volto, I have no clue why... but it works in a standard Volto project... I guess that's where it counts 😀😀

package.json
```json
...
"addons": [
    "volto-myaddon"
]
...
```

So it's left the config aggregator and what else?

### DIFF
--- a/razzle.config.js
+++ b/razzle.config.js
@@ -287,11 +287,14 @@ module.exports = {
     if (packageJson.name !== '@plone/volto') {
       include.push(fs.realpathSync(`${voltoPath}/src`));
     }
-    packageJson.addons.forEach(addon =>
-      include.push(
-        fs.realpathSync(`${projectRootPath}/node_modules/${addon}/src`),
-      ),
-    );
+    // Add babel support external (ie. node_modules npm published packages)
+    if (packageJson.addons) {
+      packageJson.addons.forEach(addon =>
+        include.push(
+          fs.realpathSync(`${projectRootPath}/node_modules/${addon}/src`),
+        ),
+      );
+    }
 
     config.module.rules[babelRuleIndex] = Object.assign(
       config.module.rules[babelRuleIndex],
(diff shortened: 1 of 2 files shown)


sneridagh:
okay, will check what happens later, for not defined addons.

razzle.config.js
```diff
@@ -287,12 +287,25 @@ module.exports = {
     if (packageJson.name !== '@plone/volto') {
       include.push(fs.realpathSync(`${voltoPath}/src`));
     }
+    packageJson.addons.forEach(addon =>
+      include.push(
+        fs.realpathSync(`${projectRootPath}/node_modules/${addon}/src`),
```
sneridagh: We need to decide if we want the addon to have `src` or not.

sneridagh:
Related: #1433

tiberiuichim:
I like it. This opens the door to a lot of other possibilities, including automatic addon configuration.

Not directly related to this PR, it would be nice to have some references to the rules declared in razzle.config.js, so that this sort of dance:
```
    const babelRuleIndex = config.module.rules.findIndex(
      rule =>
        rule.use &&
        rule.use[0].loader &&
        rule.use[0].loader.includes('babel-loader'),
    );
``` 
is not required in Volto projects.

nzambello:
This looks nice, I find it a good first step for an addon system likely automagical.

sneridagh:
I made some tests, it works perfectly with the current infrastructure (at least ours, based in yarn workspaces), so it's not breaking. I will merge it, but it would be great if you can test it on your own before I cut a release.

For those using yarn workspaces, with this one, the `rm -rf ./node_modules/mypackage` workaround is no longer needed, if the `addon` entry is declared in `package.json`!

So we can have full `yarn workspaces` feature working in Volto (which I think is the way to go). It works either with mrs.developer setups (to develop add-ons while developing projects) and with released packages.

/cc @ebrehault @vangheem

PR #1444, opened 2020-05-03
Move dev dependencies to devDependencies in package.json

tiberiuichim:


sneridagh:
@tiberiuichim if we do this, then we have to set them all in Volto projects...

sneridagh:
I would be +1 (this will fix the hoisting deps problems that we have right now). BUT we will need a tool that updates the dependencies like `react-codemod`that automates that, given a Volto version. Shouldn't be difficult to do...

Also, if we ever do that it will be a breaking change, because people will require to act over their projects.

tiberiuichim:
@sneridagh Yeah, I know. I've created the PR to see the result of the Travis bot, I was curious what the problems would be. volto-starter-project would have to be updated, for sure. 

Although not directly related, what triggered me to start this was me trying to improve our docker images. My ideal scenario (although I'm not sure how feasible this is) would be that I can have a multi-stage Dockerfile that builds the project in stage 1 then just copies the bundled files to stage 2. This is impossible right now because node_modules is declared as external for "node" target. Do we have any benefit from this? Is this required? Are there any binary dependencies for server.js ? I guess I'll have to try this at some point.

If the dev dependencies would be properly set, then I can set them properly in my Volto project as well and my 2-stage build can run in --production in second stage, this way the resulting docker image can be smaller, with less stuff in node_modules.

davisagli:
Closing since this is old and not likely to be merged in this form -- but I do hope we'll come back to doing this in a way that still makes it easy for projects to install the volto dev tools they need.

PR #1447, opened 2020-05-05
Removing title field from dropdown in content view

iFlameing:
Fixes #1107

PR #1448, opened 2020-05-06
Fix table tests

sneridagh:


iFlameing:
@sneridagh I just fixed this test in https://github.com/plone/volto/pull/1447

sneridagh:
ok, closing then. We have master red now. We should fix it asap.

PR #1449, opened 2020-05-06
Eea dx fields validation

rexalex:
Fix validation for number inputs, email, URL, password, minimum/maximum value, minimum/maximum length.
Show errors from Backend for corresponding fields.
Show proper input for respective property type (email field for email property).
Revalidate field on blur for better ux.


tiberiuichim:
How are the UrlWidget and EmailWidget different from TextWidget? Would it not be possible to reuse that component directly? Fill in typeTranslations in ``Widgets/TextWidget.jsx#L44``

rexalex:
> How are the UrlWidget and EmailWidget different from TextWidget? Would it not be possible to reuse that component directly? Fill in typeTranslations in `Widgets/TextWidget.jsx#L44`

I followed the existing pattern where there was already a PasswordWidget, that was a simpler more appropriate function component. In truth TextWidget could be used for all (Url, Password, Email). But that would mean, deleting the existing PasswordWidget.

PR #1463, opened 2020-05-08
Remember selected folder contents sort on and order

ksuess:
Fix for https://github.com/plone/volto/issues/1461
Still without test. Please review, then I'll add a test.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### Feature
 
 - Added Dexterity Types Controlpanel @avoinea #1285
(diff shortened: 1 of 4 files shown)


sneridagh:
@ksuess Changelog? Also, it's not used anywhere yet, right?

ksuess:
> @ksuess Changelog? Also, it's not used anywhere yet, right?

@sneridagh What is not used where?

ksuess:
Cypress test is impossible. Problem is the nested dropdown menu for sorting: Cypress struggles with clicks on elements that get visible after a previous click.

PR #1471, opened 2020-05-09
Upgrade to Razzle 3.1.1

sneridagh:
It has some nice things, upgrade direct deps, and others like the client only build.

@terapyon did you asked for it in the past, right?

I haven't the use case yet, but who knows!

We will have to add the documentation, but now it's possible:
`yarn start-spa`
`yarn build-spa`

You have to rename `public/index.html.spa` to `public/index.html`... works like a charm 😊


sneridagh:
@tiberiuichim indeed, it's the remainings from the original plugin fork. I did the replacement quick to make it work.

pnicolli:
Awesome! 🎉 

The pedantic part of me would say that removing support for CSS Modules should be referred to as a breaking change.
On the other hand, not sure if anybody uses those for Volto projects. It would be interesting to know that.

.eslintrc
```diff
@@ -20,6 +20,7 @@
     "import/no-unresolved": 1,
     "no-alert": 1,
     "no-debugger": 1,
+    "react-hooks/exhaustive-deps": 0,
```
pnicolli, line 23: I kind of liked it. I understand it might happen that you need to go against this rule, but I think I would rather have specific magic comments to disable the rule when needed.
The final decision might also depend on how many situations there are that would need this rule to be disabled, that is something I couldn't check at this time.

theme/themes/pastanaga/collections/form.overrides
```diff
@@ -96,3 +96,12 @@
...
 }
+
+// kasdjaksdjas
+.myclass {
+  color: red;
+
+  .myotherclass {
+    color: yellow;
+  }
+}
```
pnicolli, line 107: Is this maybe something left over from some tests?

.eslintrc
sneridagh: I found that almost always I have to set the magic comment because I need to remove some direct dependencies to make the component behave correctly. Unfortunately, Volto was architectured in the pre-hooks era and the component props hierarchy are not thought to work with them (yet).

One can always re-add the rule in projects. If you think it's better to leave it we can discuss it or bring it back later!

theme/themes/pastanaga/collections/form.overrides
sneridagh: It is! I will remove it

sneridagh:
> The pedantic part of me would say that removing support for CSS Modules should be referred to as a breaking change.

It will for sure! I also prepared the upgrade docs.

sneridagh:
@pnicolli reenabled! https://github.com/plone/volto/pull/1495

PR #1489, opened 2020-05-15
Fix for the long lasted issue when creating links in newly created te…

sneridagh:
…xt blocks not showing as links until you save.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Bugfix
 
 - Fix Bug in Form Component, that lead to site crash when transmitting a comment @jackahl #1475
+- Fix for the long lasted issue when creating links in newly created text blocks not showing as links until you save @avoinea
 
 ### Internal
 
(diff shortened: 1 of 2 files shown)


sneridagh:
All props for this one to @avoinea !!

sneridagh:
@avoinea why the tests are failing?

avoinea:
Because they were wrongly recorded :smile: 

Just kidding, I'll take a look.

avoinea:
@sneridagh It looks like the focus on `componentWillUnmount` was the problem. I don't know why we'd want to focus on unmount?! :thinking: Removed it completely!

sneridagh:
Yeah, strange... I also can’t remember why we left it like this... I will investigate a bit.

sneridagh:
Probably my bad... 


https://github.com/plone/volto/commit/5b43e08f056fafe62f1e3da0e11e16ea6ac1687a

I can’t see now any use for that!

PR #1491, opened 2020-05-17
Exhaustive deps eslint rule

tiberiuichim:
According to Volto's [upgrade guide](https://docs.voltocms.com/upgrade-guide/#update-your-eslint-config) the eslintrc from Volto is to be used as mixin source for the project's eslintrc.

I think it's better to provide a reasonable default (recommended by React) or to document that Volto projects should add back that rule.

I've tried to follow eslint's recommendations and add the missing dependencies for the cases where useEffect was used. If that rule has to be disabled in some cases, maybe it's a code smell? Would it be worth refactoring that code, or document inline why that's a special case?


sneridagh:
@tiberiuichim I can explain why the deps can't be there extrictly, and why I think it's better to remove the rule. Hint: take a look at the network tab on a listing block.

tiberiuichim:
@sneridagh I've examined the code, to understand it, but I couldn't see any difference in behavior of the Listing Block Edit component compared to Volto's demo on kitconcept. On clicking the block, it always does a network fetch, on changing other params (such as block size limit) it does a network fetch, it doesn't do it when other blocks are changed.

sneridagh:
It is, on view it does three requests instead of one.

sneridagh:
@tiberiuichim I will give it a try again! It might be that some of the last changes made the component tree behave in a different way!

src/components/manage/Multilingual/CreateTranslation.jsx
```diff
@@ -36,8 +36,7 @@ const CreateTranslation = (props) => {
       );
     };
     // On mount only
-    /* eslint-disable react-hooks/exhaustive-deps */
-  }, []);
+  }, [dispatch, language, translationOf]);
```
sneridagh, line 39: This one should be `[]`, because we want it happen only on componentDidMount.

src/components/manage/Blocks/Listing/Edit.jsx
```diff
@@ -26,8 +26,7 @@ const Edit = ({
         block,
       });
     }
-    /* eslint-disable react-hooks/exhaustive-deps */
-  }, []);
+  }, [block, data, onChangeBlock]);
```
sneridagh, line 29: We also want that one on `componentDidMount`.

src/components/manage/Blocks/Listing/ListingBody.jsx
```diff
@@ -17,14 +17,15 @@ const ListingBody = ({ data, properties, intl, path, isEditMode }) => {
...
 
   React.useEffect(() => {
-    if (data?.query?.length > 0) {
+    if (queryLength > 0) {
       dispatch(
         getQueryStringResults(path, { ...data, fullobjects: 1 }, data.block),
       );
     }
-  }, [dispatch, data, data.block, path]);
+  }, [data, dispatch, path, queryLength]);
```
sneridagh, line 28: `queryLenght` will be different on every rerender. We don't want the action happen every time that happen.

sneridagh:
@tiberiuichim I can still see a query going out if the component props change (selected). Ideally we don't want that to happen.

src/components/manage/Blocks/Listing/ListingBody.jsx
tiberiuichim: ``data`` was already a dependency to the useEffect. I thought I might be able to eliminate the dependency on data with a ``useCallback``, but of course, once ``data`` changes for useCallback, the callback function changes and the effect is called again.

sneridagh: @tiberiuichim yeah, I've been in the useCallback/useMemo story! No luck :( 
`data.block` it used to work well (no render when selecting the block).
I have to admit that adding Redux and `useSelector` in play, the scenario is more difficult and forces to write `useEffect` in an specific way.

sneridagh:
@tiberiuichim I'm +1 to leave the defaults for the `exhaustive-deps` rule we can revisit it in the future. What do you think?

sneridagh:
reenabled in https://github.com/plone/volto/pull/1495

tiberiuichim:
To summarize some of the stuff Victor has shown me:

The ListingBody has a ``useEffect()`` that triggers the getQuerystring results network fetch. This useEffect, according to eslint hooks plugin, should list ``data`` and ``path`` as dependencies. The problem is that the useEffect is called a second time when leaving the page that has the ListingBody component, because the path is changed by the router and then it propagates to the dependency of useEffect.

As a consequence, the useEffect dependency in ListingBody has to omit the path, to avoid this retriggering the network fetch.

Some stuff to be investigated: would getting rid of react-connected-router improve things?

tiberiuichim:
@sneridagh: this seemed to be ok for me, with useRef. My backend was old, didn't have the latest plone.restapi, but I don't see a network fetch when navigating away from a listing block

PR #1474, opened 2020-05-13
Improving Drag and Drop

iFlameing:
Fixes #1446

PR #1492, opened 2020-05-18
Update CheckboxWidget.jsx

nzambello:
Fix `htmlFor`  attribute for label.
In JSX for is a reserved keyword for JS `for` loop, so it's `htmlFor`


nzambello:
Closing this, I saw it's been fixed in #1474

PR #1515, opened 2020-05-22
fixed edit link when selection is from end of the word

giuliaghisini:
There was a bug when trying to edit a link, and selecting link-text from the end to start: the text was unlinkable and so the link was un-editable. 
Now it's fixed but, it's a bug from draft-js EditorUtils.js, so I had to use EditorUtils.js which was present in Volto but was not used.


tisto:
@giuliaghisini I recall that we had that problem as well and I was under the assumption we fixed it. It would be cool if we could write a Cypress test for this. Maybe @iFlameing can jump in and write one. We would have to check if Cypress allows selecting a string starting with the end of the word.

tisto:
@iFlameing could you check if it is possible to write a Cypress tests that does the selection of a word starting from the last letter of the word? I'd love to have a Cypress test that confirms this bugfix.

tisto:
@giuliaghisini could you please elaborate what I have to do to being able to reproduce the original issue? I tried the following:

1) create a new page
2) create a new link (from end to the beginning of the word)
3) save the page and edit again
4) select the existing link (from end to the beginning of the word)

This all works both in your PR branch an on latest Volto as far as I can see. Maybe I am missing the point here. Can you help me?

giuliaghisini:
@tisto yes, these are the steps to reproduce the issue:

1. create a new page
2. create a new link selecting a word (there's no difference here where selection starts and if it's an internal or external link)
3. save the page and edit again
4. select the existing link (from the end to the beginning of the word)
5. click on the 'chain' button to remove the link
You are not able to edit/remove the link if selection is from end to begin of the word, but you are able to do it if you select the word from the beginning to the end.

PR #1541, opened 2020-05-27
Implementation of config.addonRoutes

tiberiuichim:
Simple implementation, no tests included.

Questions: is naming ok? It's exposed as config.addonRoutes.

TODO: include more documentation?

TODO: include config.addonRoutes in create-volto-app.


tiberiuichim:
I know realize that what I've done is to "add" to the default routes, instead of taking the Volto aproach, which is to replace the config.routes and use those directly.

src/routes.js
```diff
@@ -185,7 +186,11 @@ const routes = [
   {
     path: '/',
     component: App,
-    routes: defaultRoutes,
+    routes: [
+      // addon routes have a higher priority then default routes
+      ...(addonRoutes || []),
```
sneridagh, line 191: If so, they should go the last, right?

tiberiuichim: No, the router use the first match in the list

Issue #1536, opened 2020-05-27
[Addons] Extend Redux reducers from addons


sneridagh:
meta issue: #1433

tiberiuichim:
See PR https://github.com/plone/volto/pull/1542

PR #1542, opened 2020-05-27
Addon reducers for #1536

tiberiuichim:


src/reducers/index.js
```diff
@@ -35,6 +35,7 @@ import userSession from '@plone/volto/reducers/userSession/userSession';
 import vocabularies from '@plone/volto/reducers/vocabularies/vocabularies';
 import workflow from '@plone/volto/reducers/workflow/workflow';
 import toolbar from '@plone/volto/reducers/toolbar/toolbar';
+import { addonReducers } from '~/config';
```
sneridagh, line 38: While porting this to create-volto-app I am wondering why we want this in Volto itself. Wouldn't it go better in create-volto-app (since volto itself is not intended to use addons).

sneridagh: @tiberiuichim ^^

tiberiuichim: @sneridagh Indeed, it's mostly a matter of default configuration.

There's also the "bigger" picture: a Volto project uses Volto as a library. I think it's a bit weird to have Volto running standalone and then running it as a library, it's not "natural". You can't run Django standalone, you have to create a django app, same for rails, etc. It's just us, coming from the "standalone application server" that want to have a big monolith and build around it.

Issue #1509, opened 2020-05-21
Bug in Richtext render in DefaultView.jsx


sneridagh:
https://github.com/plone/volto/blob/master/src/components/theme/View/DefaultView.jsx#L85-L88

Buggy when a mailto: can be used, then it got replaced too.

Since we are already treating it in:
https://github.com/plone/volto/blob/master/src/config/RichTextEditor/ToHTML.jsx#L173-L197

I think we can assume we can remove it.

nileshgulia1:
closed via #1545

PR #1545, opened 2020-05-27
fix: mailto replacing by regex

nileshgulia1:
#1509


src/components/theme/View/DefaultView.jsx
```diff
@@ -82,10 +82,7 @@ const DefaultView = ({ content, intl, location }) => {
       {content.text && (
         <div
           dangerouslySetInnerHTML={{
-            __html: content.text.data.replace(
-              /a href="([^"]*\.[^"]*)"/g,
```
nileshgulia1, line 86: MayBe instead of removing we can add only condition for not matching mailto:
```a href="((?!mailto:)[^"]*\.[^"]*)"```

Issue #1163, opened 2020-02-03
Image block: allow users to select a miniature?


cekk:
I don't know if this is a topic already discussed, but for me it could be useful adding a "miniature" field in image block configuration, to select what miniature to use for the image in the text.

I have several cases where i migrate pages from old sites and images are very large because they are used in different places, but in the text they use a smaller miniature..and in Volto they appear too big ;)

tisto:
@cekk can you elaborate a bit on this? I don't fully understand this TBH. :)

datakurre:
@cekk  @sneridagh So, my opinion on that was that let's not add miniature field, but use picture tag with sources with sizes. Once those are defined, browser should pick whatever size is the best for the defined area.

If you later want to give editor more options on image size, give something that makes more sense for editor than miniature option with Plone image scales vocabulary. Then it is up to the integrator to configure Plone to provide such scales that give best experience with the current image uses. The best part is that Plone image scales could be change later because they are not referred directly, but they just affect the available sizes in source-tag.

cekk:
Ok, @nzambello was supposed to reply to this thread some weeks ago ;)
My use-case was that on Plone, editors can choose which miniature include into text based on their needs in that particular content.
With Volto we can only choose alignment and use the original image. And the integrator can scale it in the theme. But editors can't choose its dimension.

Of course we could customize that widget/block to allow miniatures selection, but i wanted to know if that's a feature that we don't want to have in Volto because we don't want to give users the powers to choose a miniature by default (and let the browsers/themes choose the right one based on screen size). I'm not talking about cropping or something else, just to do what we can do in classic Plone in Tiny.

tisto:
![Add_Document (1)](https://user-images.githubusercontent.com/632105/82649808-0da19f00-9c1a-11ea-8248-4c0b7b51cf8d.png)

tisto:
After giving this some thought I think we should add two new buttons to the alignment list with smaller images, so we have:

- Image left
- Image left with a small image
- Image right
- Image right with a small image
- Image center
- Image full width

cekk:
can't agree more ;)

tisto:
Implemented and released.

PR #1552, opened 2020-05-28
add smaller aligmend options to image

jackahl:
Adds feature requested in #1163

The additional buttons use the same icon as the normal left/right align buttons. We probably need new icons for the smaller options.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ### Feature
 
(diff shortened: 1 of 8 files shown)


sneridagh:
@tisto how many options do we add? I bet more for small/medium/big (whtever the context is)

sneridagh:
@jackahl I thought more in something like: #1565 
combined with: #1553

PR #1553, opened 2020-05-28
Adding src-set in image block

### LINKED ISSUE #1428: Implement srcset and sizes for images in Volto
Todo:

- [x] Define new scale formats for Plone (we currently have 200/400/800/1200/1600)
- [x] Implement new scale formats in kitconcept.volto (preliminary until PLIP has been accepted) -> https://github.com/kitconcept/kitconcept.volto/blob/main/src/kitconcept/volto/profiles/default/registry.xml#L23
- [ ] Amend image block to use srcset/sizes
- [ ] Amend hero block to use srcset/sizes
- [x] PLIP new image sizes for Plone 6 https://github.com/plone/Products.CMFPlone/issues/3279
- [ ] PLIP accepted by FWT
- [ ] PLIP merged into Plone 6

@datakurre already did most of the work in this commit: https://github.com/datakurre/my-volto-app/commit/89c86ba93db2ad404f45a7dfc5fc9dde686390e6

Twitter found out that 3x for retina is not worth the effort: https://blog.twitter.com/engineering/en_us/topics/infrastructure/2019/capping-image-fidelity-on-ultra-high-resolution-devices.html

iFlameing:
Fixes #1428


sneridagh:
@tiberiuichim the idea is that using srcsets the browser decide which size to display, depending on the container current width, so the user don't have to set them anymore.

The proposed sizes range are thought to cover the major range of possibilities that these specific components will handle. Custom components will have to define new, if required, then use them (ideally using srcsets too).

We will go from choosing scales, to choose the size of the containers, and that's in the layout of each component. I am missing something or some use case in these components that should be configurable?

tiberiuichim:
@sneridagh thanks for the explanation, LGTM

tisto:
@iFlameing could you please rebase your branch?

iFlameing:
@tisto I rebased my pr :)

iFlameing:
@datakurre I am trying to make the image responsive in volt. But when I am checking with respImage lint it is falling. can you tell me why?
Screenshot from 2020-06-15 20-30-03

datakurre:
@iFlameing Have you also configured scales at Plone control panel as in the commit messages of https://github.com/datakurre/my-volto-app/commit/89c86ba93db2ad404f45a7dfc5fc9dde686390e6

src/components/manage/Blocks/Image/View.jsx
```diff
@@ -30,28 +31,33 @@ const View = ({ data, detached }) => (
...
-                      return `${flattenToAppURL(data.url)}/@@images/image`;
-                    })()
-                  : data.url
-              }
-              alt={data.alt || ''}
-            />
+            <picture>
+              <source
+                srcSet={`
+            ${flattenToAppURL(data.url)}/@@images/image/mini 200w,
```
tiberiuichim, line 37: Maybe save ``flattenToAppURL(data.url)`` to an intermediary var? This way code is shorter and optimized

tisto:
@datakurre @iFlameing I guess we have to add those new scalings to kitconcept.volto before we merge this PR. The question is what happens if those scales are not available in the backend? I'd guess the browser would just fail to load them which kind of sucks since we essentially break things for anybody who is not using kitconcept.volto in its latest version. Therefore we might have to consider waiting with this until we do another major Volto release. What do you think?

datakurre:
Only two of those sizes are completely missing: great and huge. But because those scales should also be available in REST API data, would it make image tiles too complex if it only renders the sizes that are really available? But sure the experience would still be bad, because the original scales sizes don’t double well as x2 or x3 sizes compared to those in my example 🤔

The original issue, for sure, are the outdated default scales in Plone :(

sneridagh:
Superseded by https://github.com/plone/volto/pull/3337

PR #1554, opened 2020-05-28
Do not store the server information of the image block in the block

sneridagh:


tiberiuichim:
Wouldn't this get fixed by the deserializer on the image block? Of course, right now we have who knows what versions of blocks, with who knows what data inside.

https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/deserializer/blocks.py#L145 

I think we're missing a serializer for the image block, though. I don't have a test website with latest code, but I'll try to test this next week.

sneridagh:
@tiberiuichim I think in this case it was a bug I introduced when I wrote the block in the first place, so it's sensible that we should fix it in the source. It's going to work anyways, since it also keeps the backwards compat. I plan to write a Plone script to migrate too, just in case.

PR #1563, opened 2020-05-29
Initial Slots implementation

sneridagh:


docs/source/slots/anatomy.md
```diff
@@ -0,0 +1,58 @@
...
+- leftAsideSlot
+- rightAsideSlot
+
+- afterApp
+- afterToolbar
+
+- htmlHead
+- htmlBeforeBody
+- htmlAfterBody
+
```
sneridagh, line 38: I initially identified these slots (and their insertion positions), did I miss something? We can add more later, for sure. @nzambello the ones referring to your "Macros" PR are there as well.

docs/source/slots/anatomy.md
```diff
@@ -0,0 +1,58 @@
...
+insertion point. You can control in which route do you want that element appear as well
+as the order (position) of the items that you add in the in the slots. Slots are named,
+so you can add in the configuration object:
+
+```js
+export const slots = {
+  aboveContentTitle: [
+    // List of components (might have config too, maybe in `data` property)
+    { path: '/', component: 'Component', data: {} },
+    // It can include blocks too (makes sense when we will be able to save them)
```
sneridagh, line 14: It's an initial thought about the future of slots, first iteration can have only "viewlet-ish", no sved config. In a second iteration we can provide a full backend support for saving those persistently, and have the ability to save slots state so we can even use blocks inside slots.

docs/source/slots/anatomy.md
tiberiuichim: I think Sidebar is a good candidate, as well.

Though I can definitely see Martjin's point for the toolbar: if extra content is needed there, then maybe we need a system of definable actions. For the Toolbar, my wishlist include new buttons in the main toolbar, but also now components available for the "More" menu.

tiberiuichim (approved):

davisagli:
@sneridagh Is there anything here worth keeping, or should we close the PR?

sneridagh:
Superseded by #5775

PR #1565, opened 2020-06-02
Add image sizer widget

sneridagh:
image


sneridagh:
In the meanwhile we don't have a full working srcset solution, I've implemented this intermediate solution (which will work also with the final one, widget included) and fills the gap of having different sizes for the image block.

tisto:
@nzambello could you elaborate? I am not sure I fully understand. The image options are defining how an image is displayed and the way an image is displayed does not depend on its actual size. You can not do both at the same time. Maybe I am entirely missing the point here though. :)

nzambello:
> @nzambello could you elaborate? I am not sure I fully understand. The image options are defining how an image is displayed and the way an image is displayed does not depend on its actual size. You can not do both at the same time. Maybe I am entirely missing the point here though. :)

Sorry, I saw now. Images are not displayed naturally but handling their width via max-width, so this is super reasonable.

Issue #1525, opened 2020-05-26
Folder contents should show the portal type as a tooltip


tiberiuichim:
I miss this small feature from Plone, showing the portal type as a tooltip for the entries in the folder listing.

nzambello:
https://github.com/plone/volto/blob/master/src/components/manage/Contents/ContentsItem.jsx#L164

Issue #1462, opened 2020-05-08
Remember folder contents additional columns to show


ksuess:

ksuess:
TODO

- [ ] make table scrollable. Current state: table with more than the default columns is overlapping the container and overlapping content is not accessible.

ksuess:
done

PR #1568, opened 2020-06-05
Folder contents view: Store additional indexes and order of indexes

ksuess:
Fix for https://github.com/plone/volto/issues/1462
- Store additional indexes and order of indexes
- New action 'indexContent' to store indexes and their order
- Sticky columns: title and actions 

Fix for https://github.com/plone/volto/issues/1525

**Change** 30.06.2020
New action INDEX_CONTENT to store selection and order of columns in folder contents view
    No layout changes. Layout changes ( overlapping table, ... ) are postponed to a later PR.

**Change** 10.10.2020
New action renamed to UPDATECOLUMNS_CONTENT


ksuess:
grafik

cypress/integration/folder-contents.js
```diff
@@ -26,14 +26,10 @@ if (Cypress.env('API') !== 'guillotina') {
 
     it('Renaming via folder contents view', () => {
       // when I rename a content object
-      cy.get('svg[class="icon unchecked"]').click();
+      cy.get('svg[class="icon unchecked"]').click({ force: true });
```
tisto, line 29: @ksuess any particular reason why we have to use force=true here?

cypress/integration/folder-contents.js
```diff
@@ -108,15 +102,33 @@ if (Cypress.env('API') !== 'guillotina') {
...
       cy.get('.item.sort_created_descending').click({ force: true });
-
-      // then the last document created must be the firs element
-      cy.get('tbody>tr:nth-child(1) .expire-align > span').should(
-        'have.text',
-        ' My Child',
+      cy.waitForResourceToLoad(
+        'my-folder/@search?path.depth=1&sort_on=created&sort_order=descending',
       );
+      cy.wait(5000);
```
tisto, line 111: @ksuess can you elaborate on why waitForResourceToLoad is not sufficient here? Wait statements slow down the test runs and should be avoided if possible.

tisto:
@ksuess impressive work! I will try to find time to do a code review this week. I think we should discuss the UX of this a bit. The separation of the flexible part and the fixed part in the table could be improved IMHO. I guess we should ask @albertcasado for his opinion on this. Maybe making all columns flexible is the way to go. I know this does not make much sense for the columns that are currently fixed but it is hard for a user to understand this separation.

ksuess:
> I will try to find time to do a code review this week. I think we should discuss the UX of this a bit. The separation of the flexible part and the fixed part in the table could be improved IMHO. I guess we should ask @albertcasado for his opinion on this. Maybe making all columns flexible is the way to go. I know this does not make much sense for the columns that are currently fixed but it is hard for a user to understand this separation.

That would be great if @albertcasado would bring in his ideas!

ksuess:
It would be nice to have _folder contents view_ remember any added columns/indices. It's a PR with just a few lines of code: please review @plone/volto-team!

tiberiuichim:
@ksuess Maybe rename the action? INDEX_CONTENT is a bit misleading. Also, as far as I can see, this is "session-based", the choice is not persisted anywhere, right?

ksuess:
Naming: Action name INDEX_CONTENT is following naming of existing events and constants like onSelectIndex. But yes it's not intuitive understandable. To be consistent it would be necessary to rename a lot of existing code. What do you suggest?  

Persistence: The selection is stored in app store, so does live until the next reload. IMHO this is enough.

ksuess:
Overlapping table in folder contents view.
Proposed are fixing name/title and actions column: https://github.com/plone/volto/pull/1568#issuecomment-639467899
Hi @tisto, any news/comments/proposal from Albert?

sneridagh (approved):
LGTM! Good work!

sneridagh:
> Hi @tisto, any news/comments/proposal from Albert?

What I remember from Albert is he saying that we will need to add some horizontal scrolling (same in mobile/small screens), so your proposal is correct, even improved, since I always imagined it not keeping the title 'sticky'. So my +1 to implement your proposal. Is not included in this PR, right?

ksuess:
> LGTM! Good work!

Thank you, @sneridagh, you made my day!

ksuess:
> > Hi @tisto, any news/comments/proposal from Albert?
> 
> What I remember from Albert is he saying that we will need to add some horizontal scrolling (same in mobile/small screens), so your proposal is correct, even improved, since I always imagined it not keeping the title 'sticky'. So my +1 to implement your proposal. Is not included in this PR, right?

Well, in an early comment @tisto mentioned he think this proposal (see screenshot above) is not well understandable. I still like it in general, just the style could be nicer, yes. So we agreed that I split this PR in 
- new action to store additional columns and order of columns
- UX: how to handle the case that the table does not fit in the container.
The PR is split, so that this PR is free from UI stuff and just adds the new action.

(I would be happy to set up a new PR with the UI changes(that were already done and can be taken out of the drawer), but please I need some input from Albert, @tisto)

PR #1584, opened 2020-06-15
selectableTypes in ObjectBrowserWidget

giuliaghisini:
if in content schema are defined some selectableTypes for a specific field (in widgetOptions.patterns_options), the objectBrowser allows to select only elements of those types


src/components/manage/Widgets/ObjectBrowserWidget.jsx
```diff
@@ -188,6 +190,8 @@ class ObjectBrowserWidget extends Component {
             onChange(id, []);
           };
 
+    console.log('props', this.props);
```
nzambello, line 193:
```suggestion

```

PR #1587, opened 2020-06-15
Used moment-timezone in DatetimeWidget; Added default timezone in config

razvanMiu:
I used moment-timezone to set a default timezone so that if the server is in a different timezone than UTC we can tweak a config setting to set the date inputs of DatetimeWidget to that specific timezone of the server.


tisto:
@razvanMiu would you mind rebasing your PR?

tisto:
@razvanMiu can you try to rebase again? Now 69 files have been changed. Maybe you rebased with an old version of master. Merging master into your branch is fine. I don't care too much about a super clean history.

razvanMiu:
I've made another branch, #1669

PR #1600, opened 2020-06-18
translated workflow state in contents, added item type in title

nzambello:
- translated workflow state in contents
- added item type in title as a tooltip


src/components/manage/Contents/Contents.jsx
```diff
@@ -454,7 +454,7 @@ class Contents extends Component {
    */
   onSelectAll() {
     this.setState({
-      selected: map(this.state.items, (item) => item['@id']),
+      selected: map(this.state.items, item => item['@id']),
```
tisto, line 457: I am wondering why this is changed back and forth I guess by prettier. Any idea?

nzambello: Outdated, but yeah it was prettier-related

nzambello:
@tisto @sneridagh, do you think this can be merged in to the master?
We would need it for some of our projects
Thank you!

PR #1607, opened 2020-06-19
added translations for select in querystring widget

nzambello:
Translations were missing here


tisto:
@nzambello @giuliaghisini you folks are on fire! :)

PR #1608, opened 2020-06-19
Fixing style in blockquote

### LINKED ISSUE #1601: Blockquote: error when text has some style
**Short:** If a text inside a blockquote has some styles (bold, italic for example), the view breaks.
**Steps to reproduce:**
- Go to edit mode of a page in Volto
- Add some text in a text block
- Add some more text to that block in a new line using shift+enter keys
- Add some style to some words (for example bold)
- Mark all the text you have just written and select the blockquote option for it
- Save the page

Result: The page breaks.
Expected: we see linebreaks and styles in blockquote

This bug is related to the changes made in #1249

iFlameing:
Fixes #1601 

@cekk can you review this?


cekk:
better, but there are still some problems. Probably i'm finding some edge-cases but my clients can do worse things:
- add a new block
- insert some text (aaa)
- shift + enter (new line)
- add some other text (bbb)
- select all the text and select blockquote
- select the first row and make it BOLD
- shift + enter (create a new line between the two already created)
- type some letters (ccc)
- save

expected:

**aaa
ccc**
bbb

got:

**aaa ccc**
bbb

btw there is the same problem with callouts too (here it seems that you haven't already managed split lines)

tisto:
@cekk @iFlameing it would be cool if we would have Cypress tests for the scenarios that Andrea describes.

iFlameing:
@cekk @tisto I investigated the test case which you mentioned above. But it is not related to code but it is related to the draftjs how It handles another tag like(strong). Draftjs handle us an array of content that is present in a blockquote, but when we introduce a strong tag then it gives us a react object of that element and we just render it.
The problem with the above test case is draftjs wrapping both sentences into a single strong tag without br tag.

I am attaching a screenshot in which you can clearly see that there is an enter tag between aaa and ccc but it is not rendered properly. @tisto we need to discuss how we can handle it.

Screenshot from 2020-06-21 20-27-18

iFlameing:
@tisto I think that we can now solve it. I will look into this tomorrow.

sneridagh:
@iFlameing see my last commit has a generic test for ToHTML structures. We now need to fill it with every imaginable use case.

tiberiuichim (approved):
I've tested the code, looks good

sneridagh:
@iFlameing adding a link is also broken, and the callout is totally broken when used with inline styles too.

tiberiuichim:
Heh, spoke too soon.

iFlameing:
@sneridagh @tiberiuichim can you review this once again.
I know that I am not following DRY but will look into it.

tiberiuichim:
@iFlameing there's 4 instances of that breakline adding code in that module... It really should be a separate function.

iFlameing:
@tiberiuichim Will look into this tomorrow. Can you please try to break this code because we really need a good test case for proof.

sneridagh:
Closing, since it's draftJS issue and it seems quite outdated.

PR #1614, opened 2020-06-22
Usage of Contettype label in Add component

giuliaghisini:


giuliaghisini:
@tisto @sneridagh, do you think this can be merged in to the master?
We would need it for some of our projects
Thank you!

PR #1546, opened 2020-05-27
Addon configuration loading

tiberiuichim:
Just some docs for now, code will come soon.

PR #1615, opened 2020-06-22
Configurable default block

tiberiuichim:
The default block type is hardcoded in Volto to be "text". This PR brings some flexibility to that, by adding a new setting, "settings.defaultBlockType", and a new ``helpers.Blocks.blockHasValue()`` function.

The PR is based on the "load_addons_1531" PR so it can't be cleanly merged unless the other one is merged as well.

There is a choice that I have made: text blocks will continue to add "text" blocks on split paragraph, while Title, Description, etc will use the ``settings.defaultBlockType`` block.

There is some pending work at the moment, for the TOC View.jsx component, which I'd like to adjust in a separate PR and the Table's Cell.jsx which continues to use draftjs.


avoinea:
Depends on #1546

PR #1620, opened 2020-06-23
Allow Volto addons to tweak razzle configuration

tiberiuichim:
This is a continuation of the load_addons_1531 PR, separated to keep things somewhat organized and simpler to follow.

This PR allows Volto addons to plug into the razzle.config.js modify() process in a way that's very similar to how we do the configuration (actually, the way razzle.config works was the inspiration for the Volto addon configuration method).

As a convention, we're looking for a ``razzle.extend.js`` file in an addon root folder (so parent folder of src, right next to the addon's package.json). I'm not very happy about this, I think I'll change it to add a new key in the addon's package.json.

I've tested this setup with a new addon, where I have this razzle.config.js file:

```
const plugins = (defaultPlugins) => {
  return defaultPlugins.concat(['scss']);
};
const modify = (config, { target, dev }, webpack) => {
  return config;
};

module.exports = {
  plugins,
  modify,
};
```

This, in effect, loads the razzle-plugin-scss plugin, making the scss loader work. I think, as a convention, that such small "helpers" should sit in standalone packages. So I would create an addon called "volto-scss", with the above code, a minimal index.js and nothing else. 

Of course, for the scss loader, it would have been simpler to just add it to Volto. In my case, I've wanted to test the razzle extension mechanism.

See https://github.com/tiberiuichim/volto-searchkit/blob/master/razzle.config.js.


tiberiuichim:
Recreated in #1649

Issue #1631, opened 2020-06-30
Allow endpoint API calls to work on URLs, in addition to paths


tiberiuichim:
I'd like to use the ``api.js`` middleware to make JSON api calls to a third-party server (not backend, not apiPath).

At this moment, the api requests are done like:

```
request = superagent[method](formatUrl(path));
```

and formatUrl is https://github.com/plone/volto/blob/master/src/helpers/Api/Api.js#L19 :

```
return `${apiPath}${adjustedPath}`;
```

As you can see, it hardcodes apiPath as prefix to the path (URL in my case).

I'm proposing that we determine if the ``path`` argument is a URL and only add the apiPath prefix if we're dealing with a path.

PR #1632, opened 2020-06-30
Allow JSON api middleware to use external servers

tiberiuichim:
Solves #1631:

I'd like to use the ``api.js`` middleware to make JSON api calls to a third-party server (not backend, not apiPath).

At this moment, the api requests are done like:

```
request = superagent[method](formatUrl(path));
```

and formatUrl is https://github.com/plone/volto/blob/master/src/helpers/Api/Api.js#L19 :

```
return `${apiPath}${adjustedPath}`;
```

As you can see, it hardcodes apiPath as prefix to the path (URL in my case).

I'm proposing that we determine if the ``path`` argument is a URL and only add the apiPath prefix if we're dealing with a path.


tisto:
@tiberiuichim pls rebase. When you create a branch instead of a fork it is easier for me to rebase myself. :)

PR #1634, opened 2020-06-30
Fix use of form/segment classes in InlineForm.jsx

tiberiuichim:
Just removed a non-needed CSS class from InlineForm.jsx.


tiberiuichim:
### Before

Annotation 2020-06-30 164414

### After
Annotation 2020-06-30 164621

tisto:
@tiberiuichim pls rebase

PR #1645, opened 2020-07-02
fix text overflow in pastanaga-menu header if title is too long

giuliaghisini:
If page title was too long, it overflowed from the menu.
Fixed: 
Schermata 2020-07-02 alle 11 29 38


tiberiuichim:
@giuliaghisini Add a Changelog entry so that it passes tests with green?

PR #1649, opened 2020-07-03
 Allow Volto addons to tweak razzle configuration

tiberiuichim:
Description copied/tweaked from #1620 and PR recreated because Github wouldn't let me change the base branch for a PR

This PR allows Volto addons to plug into the razzle.config.js modify() process in a way that's very similar to how we do the configuration (actually, the way razzle.config works was the inspiration for the Volto addon configuration method).

As a convention, we're looking for a ``razzle.extend.js`` file in an addon root folder (so parent folder of src, right next to the addon's package.json). If you have a better proposal, give me a shout

I've tested this setup with a new addon, where I have this razzle.config.js file:

```
const plugins = (defaultPlugins) => {
  return defaultPlugins.concat(['scss']);
};
const modify = (config, { target, dev }, webpack) => {
  return config;
};

module.exports = {
  plugins,
  modify,
};
```

This, in effect, loads the razzle-plugin-scss plugin, making the scss loader work. I think, as a convention, that such small "helpers" should sit in standalone packages. So I would create an addon called "volto-scss", with the above code, a minimal index.js and nothing else. 

Of course, for the scss loader, it would have been simpler to just add it to Volto. In my case, I've wanted to test the razzle extension mechanism.

See https://github.com/tiberiuichim/volto-searchkit/blob/master/razzle.extend.js.

One additional (biggish) thing: I've restructured razzle.config.js to split some of the code, it was getting hard to understand the code flow. I've created an `AddonConfigurationRegistry` class, which makes it easier to reason about the addon process. But I've added tests, so at least a part of the old razzle.config.js is now tested. This is also to make it easier for the next step, addon customization story (addons can customize Volto, Volto project can customize addons).


CHANGELOG.md
```diff
@@ -55,6 +64,8 @@
 - Added item type as a tooltip in contents @nzambello
 - Added italian translations and translated array, token and select widget. @giuliaghisini
 - Added uploading image preview in FileWidget @iFlameing
+- Allow custom express middleware declared with ``settings.expressMiddleware``. See [Customizing Express](docs/customizing/express.md) @tiberiuichim
```
tisto, line 67: @tiberiuichim seems this is a duplicate line...

CHANGELOG.md
```diff
@@ -71,7 +82,6 @@
 
 ### Feature
 
-- Added Italian translations and translated array, token and select widget. @giuliaghisini
```
tisto, line 74: Seems there are some problems with the changeling. I can look into this before merging thg.

tiberiuichim:
@tisto many commits later (sorry!) the CHANGELOG is finished. Probably best if you squash this when merging.

tisto:
@tiberiuichim no worries. I can take care of cleaning up the changelog when I actually do the merge. This is the next PR that we should merge after the 7.0.0 release from yesterday, right?

@robgietema would be great if you could have a quick look here as well. Same for @nzambello @cekk @giuliaghisini @pnicolli.

tiberiuichim:
@tisto Yeah, this is the biggest jump. While it looks big, this PR is actually quite tame. As a feature, it only allows addons to provide a `razzle.extend.js` file, nothing else. But to make it easy for the next PR (which will be about customizations), I've restructured razzle.config.js. The good thing is now that the stuff that was refactored out of razzle.config.js is tested can be used in many other places, for example resurecting the styleguide branches.

PR #1658, opened 2020-07-08
managed hidden fields

giuliaghisini:
if field has property mode setted to 'hidden' (from plone), the field is not rendered.


giuliaghisini:
> @giuliaghisini please add a changelog entry.

sorry again! added

tisto:
@giuliaghisini np. thanks for the swift reply! :)

PR #1662, opened 2020-07-09
Upgrade razzle to latest release

tiberiuichim:
Just a bump, minor fixes.

I've pinned razzle to `^3.1.5`, from `3.1.2`. It's not clear to me if Volto's dependencies should be pinned to absolute versions or we can use semantic versioning. See also https://github.com/plone/volto/pull/1444

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,8 @@
 
 ### Internal
 
+- Upgrade razzle to `^3.1.5`. @tiberiuichim
+
 ## 7.0.1 (2020-07-07)
 
 ### Bugfix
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -310,7 +311,7 @@
     "query-string": "4.1.0",
     "raven": "2.6.4",
     "raven-js": "3.27.2",
-    "razzle": "3.1.2",
+    "razzle": "^3.1.5",
```
tisto, line 314: @tiberiuichim can we pin this specifically to 3.1.5? I prefer to have exactly pinned versions to have a repeatable CI run.

CHANGELOG.md
```diff
@@ -20,6 +20,8 @@
 
 ### Internal
 
+- Upgrade razzle to `^3.1.5`. @tiberiuichim
```
tisto, line 23: Just wondering here if that's "internal". Need to think about it and maybe do some research...

PR #1664, opened 2020-07-09
Provide a new webpack alias that makes it easier to reference Volto themes folder

tiberiuichim:
This is just a convenience alias to make some bits nicer in `theme.config`.

So instead of having:

```
/* Path to theme packages */
@themesFolder : '../../node_modules/@plone/volto/theme/themes';

/* Path to site override folder */
@siteFolder  : "../../theme";
```

We can have:

```
/* Path to theme packages */
@themesFolder  : '~volto-themes'

/* Path to site override folder */
@siteFolder  : '~@package/../theme';
```

For the first one, themesFolder, it's a matter of cosmetics. The second one will make it easier for people new to Volto to understand which path it's about.

I'm also recommending to move themes to their own folder, to prepare for the addition of addon theming, so you could override an addon theme.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@
 
 ### Feature
 
+- Provide a new webpack alias, `volto-themes`, which points to Volto's theme
+  folder. See details in the https://docs.voltocms.com/upgrade-guide/
+
 ### Bugfix
 
 ### Internal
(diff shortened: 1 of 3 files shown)


tisto:
@tiberiuichim nice. Maybe we should add this to our regular docs as well. This could be useful in other circumstances as well...

PR #1673, opened 2020-07-15
Automatically select next/previous block on block deleted

### LINKED ISSUE #1671: Automatically select next block (or previous, if next is not available) when a block is deleted
A small usability enhancement, easy to achieve with current Volto code.

To the implementer: be aware that it may be needed to force React to serialize the order of operations with setTimeout().

tiberiuichim:
This is already implemented when using backspace in draftjs editor, but not if using delete button in Block/Edit. This "hardcodes" selectPrev as true and enhances the implementation to also automatically select next block if previous block is not possible (for example when deleting the first block).

Closes #1671

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ### Feature
 
+- On block deletion, automatically select previous block (or next block, if the first block is deleted). @tiberiuichim
+
 ### Bugfix
 
 ### Internal
--- a/src/components/manage/Blocks/Block/Edit.jsx
+++ b/src/components/manage/Blocks/Block/Edit.jsx
@@ -156,7 +156,7 @@ class Edit extends Component {
           <Button
             icon
             basic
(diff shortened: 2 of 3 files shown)


tiberiuichim:
I'm a little bit unsure of how this thing should work. Right now, with current Volto master, the "select previous block on delete" is only used when deleting a draft block with backspace, so it makes sense to "delete this and select backward, the previous block", but when deleting a block with the delete button, it might make sense to "delete this and select forward, the next block"

@giuliaghisini @tisto  What do you think?

tiberiuichim:
I've tested again, and I think it makes sense to "delete and select forward", this way, if I'm under the Title block and I delete blocks, I don't need to reselect blocks.

I'll reimplement the PR

nileshgulia1:
Just an opinion, I think we should make it consistant across all scenarios, selecting a previous block in all conditions (the backspace or trash button) still make sense to me. 
Nonetheless the PR is not up to date now.

PR #1679, opened 2020-07-16
Adding dropzone in Filewidget

### LINKED ISSUE #1675: Allow to drag & drop images to the file widget
It would be nice to being able to drag & drop an image to the file widget (e.g. on a page with an image field).

iFlameing:
Fixes #1675


tisto:
@plone/volto-team please review this PR and let us know what you think about this functionality.

tisto:
@iFlameing some nitpicking and UX improvements:

Add_News_Item (2)

One other thing I noticed: when I delete an existing image by clicking on the trash icon, the image is deleted but not the preview.

iFlameing:
@tisto I have fixed the spacing between the image and label, aligned the image. I only not added the line regarding "You may also drag and drop" because I think we can substitute the preview image to some more meaningful image.

theme/themes/pastanaga/extras/objectbrowser-widget.less
```diff
@@ -30,4 +30,14 @@
...
+  margin-bottom: 20px;
+}
+
+.ui.input.file-widget-input {
+  display: none;
+}
+
+.label-file-widget-input {
+  padding: 5px;
+  border: 2px solid #ccc;
```
pnicolli, line 42: Not a great fan of adding this in this file, since it should be dedicated to the object browser.

pnicolli:
Interestingly, we had the very same need last week and implemented our own widget. We planned to share it this week in order to gather ideas, so here we are :)
The good thing about this is that we also used ``react-dropzone`` to implement this 😬 

This is our current implementation:
https://github.com/RedTurtle/design-volto-theme/pull/66/files

Empty widget, it says "Drop a new file to upload"
image

File dropped, it says "Drop a file to replace the uploaded one"
image

Editing an already-uploaded file
image

Uploaded a file to replace the current one
image

We're definitely missing the image preview feature, that was still on the list. Also, it would be useful to show the file type and size, like Plone does.

We're not planning to update our widget this week, that's why I wanted to share it, in order to make some ideas and code available for you to merge, if you find it useful.

A file widget like this is a must have in Volto, imho.

theme/themes/pastanaga/extras/objectbrowser-widget.less
iFlameing: Should I create a new file? what's your suggestion?

pnicolli: Yes, I would create a separate file in order to separate concerns.

iFlameing:
@pnicolli  I am trying to replace the current image Svg to this. It will do the trick. In this way, we will able to give the idea to the user with a preview image feature.
what do you think?????
Dropzone

iFlameing:
@tisto It is ready for review

tisto:
@pnicolli would you mind having a second look at this PR?

pnicolli:
> @pnicolli I am trying to replace the current image Svg to the below one. It will do the trick. In this way, we will able to give the idea to the user with a preview image feature.
> what do you think?????
> block-image
> 
> Dropzone

It would be awesome to have something like a visible dropzone like you suggested here. You can use simple html+css to achieve it, instead of using an svg.

A couple extra feedback:
1. There is no visual feedback on the dropzone, except for when you are already dragging something onto it.
2. There is no visual feedback on hover of the explicit "Add" button. Some graphical hover effect would help users (background and/or border change). Also, changing the cursor to a ``cursor: pointer`` on hover would help.
3. When removing an image, the preview correctly disappears but the default doesn't come back:

image

sneridagh:
@iFlameing I think we should go for the RedTurtle solution, use HTML and CSS for the dropzone. Take a look at their PR too to take some ideas.

Then follow @pnicolli advices in his last comment. 

Having the preview would be awesome, but maybe fells out of the scope. /cc @tisto

iFlameing:
@sneridagh @tisto If we use SVG instead of html/css then we are able to get the preview feature without doing anything. The current implementation will handle that.

iFlameing:
@sneridagh @pnicolli can both of you review this again because I fixed all your comments in recent changes.

tisto:
@sneridagh would be nice if we could look into this in the next days.

sneridagh:
image

I improved the CSS a bit, based on the text buttons originals from Pastanaga UI

sneridagh:
I would change the literal "choose a file/image". In the case of the image it's misleading, since if you upload a file, it might break. Nothing new, I'm +1 to unify both images and files objects too.

tiberiuichim (approved):

PR #1683, opened 2020-07-20
Making Content browser aware of context 

### LINKED ISSUE #1674: Content Browser should be context aware
When an editor adds an image block, and chooses and existing image with the content browser (right sidebar), the content browser always starts at the portal root. It would be more convenient if the content browser would be context sensitive and start at the current object.

1) Create a new page in an existing folder structure (e.g. "cars/tesla/model-s")
2) Add a new image block
3) Click on the content browser icon to choose an image
-> the content browser should show ("cars/tesla/model-s") or maybe "cars/tesla" (we have to user test this).

iFlameing:
Fixes #1674


tisto:
@plone/volto-team this is something that our clients asked for a lot. What is your experience here? Are you ok with merging this and starting the content browser in the current location/context instead of on the root?

tiberiuichim:
I'm very much ok with it.

Issue #1558, opened 2020-05-29
Error while deleting portrait in Personal Info


nileshgulia1:
Steps to reproduce:
1. Click on Profile in user settings.
2. Upload a portrait.
3. try to remove the portrait or existing portrait.

<img width="1173" alt="Screenshot 2020-05-29 at 1 26 30 PM" src="https://user-images.githubusercontent.com/22280901/83235761-03e6e100-a1b0-11ea-8ca7-9c1bc4c69266.png">

Though we can change the existing portrait.

sneridagh:
@nileshgulia1 is this still happening?

nileshgulia1:
@sneridagh No, I'm still not able to remove the portrait.

PR #1684, opened 2020-07-20
fix upload avatar in personal preferences.

nileshgulia1:
Related #1558


nileshgulia1:
This no longer reproducible.

PR #1691, opened 2020-07-23
Avoid console warning in Toolbar.jsx

### LINKED ISSUE #1685: Warning in developer console generated by Toolbar.jsx
There's an annoying warning generated by Toolbar.jsx:

```
index.js:1 Warning: Expected server HTML to contain a matching <div> in <div>.
    in div (at Toolbar.jsx:290)
    in Toolbar (created by ConnectFunction)
    in ConnectFunction (created by Context.Consumer)
    in injectIntl(Connect(Toolbar)) (at View.jsx:257)
    in Portal (at View.jsx:256)
    in div (at View.jsx:220)
    in View (created by ConnectFunction)
    in ConnectFunction (created by Context.Consumer)
    in injectIntl(Connect(View)) (created by Context.Consumer)
    in Route (created by App)
    in Switch (created by App)
    in main (at App.jsx:143)
...
```

I've tracked the difference in output. On the client side:

```
<div role="navigation" aria-label="Toolbar" id="toolbar"><div class="toolbar-content"><div class="pusher-puller"></div></div><div class="toolbar expanded"><div class="toolbar-body"><div class="toolbar-actions">......(shortened)....<div class="pastanagalogo"><img src="http://localhost:3001/static/media/pastanaga.08f24832.svg" alt=""></div></div></div><div class="toolbar-handler"><button aria-label="Shrink toolbar" class=""></button></div></div><div class="pusher"></div></div>
```

While output from the server (generated with curl with __ac cookie set), looks like:

```
    <div role="navigation" aria-label="Toolbar" id="toolbar"></div>
```

I've tried adding the listActions to the App.jsx asyncConnect, but it didn't work to get rid of the warning.

tiberiuichim:
Closes #1685


tiberiuichim:
Thanks @giuliaghisini @nzambello for the review!

tiberiuichim:
Not ready yet, I've noticed that the /contents view also needs the treatment. Probably everywhere where `<Toolbar />` is referenced

tisto:
@tiberiuichim nice catch! I remember complaining about this for quite a while. :)

tiberiuichim:
Yeah, I was wondering if this is the right approach, but there's this bit in the official [hydrate ](https://reactjs.org/docs/react-dom.html#hydrate) docs:

> If you intentionally need to render something different on the server and the client, you can do a two-pass rendering. Components that render something different on the client can read a state variable like this.state.isClient, which you can set to true in componentDidMount(). This way the initial render pass will render the same content as the server, avoiding mismatches, but an additional pass will happen synchronously right after hydration. Note that this approach will make your components slower because they have to render twice, so use it with caution.

PR #1692, opened 2020-07-23
Sidebar toggle highlight animation

tiberiuichim:
ezgif com-video-to-gif


tiberiuichim:
Not the best preview gif in the above description, unfortunately. 

Ideally I'd like to see an animation on the sidebar-container, a smart decrease in size.

Issue #1690, opened 2020-07-22
Missing next & previous icons in contents view


robdayz:
In Volto contents view, with a big list of items (to force pagination), you can see 'next' and 'previous' icons to move through pages, but icons are not displayed/loading  :( 

<img width="1184" alt="image" src="https://user-images.githubusercontent.com/506444/88193531-11639900-cc3e-11ea-806d-e0aee7fd0d47.png">

nileshgulia1:
@robdayz Just noticed that, maybe semantic-ui icons are not loading inside the Menu. We can render using our Icon component.

nileshgulia1:
@robdayz see #1693

robdayz:
Merged! Thanks @nileshgulia1

PR #1693, opened 2020-07-23
fix: pagination-icon

nileshgulia1:
#1690 
This is how it looks with the pastanaga icon.
Screenshot 2020-07-24 at 12 22 14 AM


robdayz:
LGTM! Thanks

PR #1699, opened 2020-07-25
Wrap form in FormContext

tiberiuichim:
**What is this?**

At its base, this PR separates the internal state of Form.jsx into its own component and exposes it as a context to any of its children or grandchildren.

**What are the benefits?**

- Performance: the `properties` doesn't need to be passed to the EditBlocks anymore, so each edit block won't re-render when other blocks are changed.
- It enables intra-form communication (via the form global state) for edit widgets. Blocks are being passed `properties={formData}` but widgets don't have that
- It enables form state mutation from any of the form components

**What can we do with it?**

I have multiple uses and it enables some advanced scenarios. For example:

- Inter-widgets communication. I have a widget where I can set "ElasticSearch Hostname" and the next widget is a dropdown with a list with the indexes on that ES Server. That means that the second widget needs to be able to read the values from the first widget (so that it knows which server to query for index values). I've used [a similar idea](https://github.com/eea/volto-searchkit/blob/599deced62f69bd041c488cfa90c03d937b840de/src/components/Widgets/SelectField.jsx#L27) in the volto-searchkit prototype.
- "Batch" editing of blocks. Right now the `on*Block`  methods don't use promises, so it's not easily possible to ensure order of execution or state integrity, so I'm forced to wrap mutation calls in setTimeout, which seems to ensure a sort of "synchronicity" in multiple setState calls. See [this](https://github.com/eea/volto-slate/blob/387129e652740da45d424b590e0b64ca6b864cb1/src/utils/volto-blocks.js#L70) as a counter-example of the workarounds that are needed.

**Is this a breaking change?**

Not really. I haven't removed yet the [properties being passed to EditBlock](https://github.com/plone/volto/blob/0542851a68ff73f519a9e97cc06e2c41420d76a2/src/components/manage/Form/Form.jsx#L777), but I'd like to do that. Removing it would improve the performance of the edit form, as blocks wouldn't have to re-render when one of them is changed (unless they specifically watch the context formData). I've changed all the Volto blocks that needed any change, so that we are prepared to remove that passed `properties` parameter. I don't know if there's any other blocks "out there" that depend on the "properties" passed to them. 

All of this is quite cleanly implemented and the advanced behavior is "opt-in". Child components need to use `useFormStateContext()` or `static contextType = FormStateContext` to get access to the ancestor form's internal state/setState.

My intention is to further build, as next PRs::

- promisify the form state change
- wrap the other form implementation (ModalForm, InlineForm) in the FormStateContext provider
- Explain all this in documentation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### Feature
 
 - Added Placeholder attribute to Textwidget and TextAreaWidget @iFlameing
+- Separate the Form.jsx state in its own component exposed as a context. This enables clean multi-block editing from other blocks, exposes the form state to widgets and many other advanced scenarios. @tiberiuichim
 
 ### Bugfix
 
--- a/src/components/manage/Blocks/Description/Edit.jsx
+++ b/src/components/manage/Blocks/Description/Edit.jsx
@@ -10,6 +10,7 @@ import { stateFromHTML } from 'draft-js-import-html';
 import { Editor, DefaultDraftBlockRenderMap, EditorState } from 'draft-js';
 import { defineMessages, injectIntl } from 'react-intl';
 import cx from 'classnames';
+import { FormStateContext } from '@plone/volto/components/manage/Form/FormContext';
 
 const messages = defineMessages({
   description: {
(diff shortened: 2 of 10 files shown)


tiberiuichim:
As a teaser, what this PR enables: proper transformation of copy/paste to Volto blocks for rich text content (via volto-slate).

vokoscreen-2020-07-27_12-22-17

PR #1705, opened 2020-07-29
Add configurable api expanders

csenger:
This lets you configure api expanders in config.settings, e.g.:

```js
import {
  settings as defaultSettings,
} from '@plone/volto/config';
import {GET_CONTENT } from '@plone/volto/constants/ActionTypes';

export const settings = {
  ...defaultSettings,
  apiExpanders: {
    GET_CONTENT: ['mycustomexpander'],
  },
}
```

Doing a getContent on `http://localhost:3000/api/object` will result in a call to `http://localhost:3000/api/object?expand=mycustomexpander`.

This should have tests. We currently have no tests for the api middleware and no mocks for the parts to let it run.


tiberiuichim:
We've had the same need at one point, I like the idea of attaching them to the action types.

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 
 ### Feature
 
+- Configurable api expanders @csenger
```
tisto, line 9: "Add configurable api expanders" :)

src/middleware/api.js
```diff
@@ -16,6 +18,32 @@ import {
 
 let socket = null;
 
+/**
+ *
+ * Add configured expanders to an api call for an action
+ * @function addExpandersToPath
+ * @param {string} path The url/path including the querystring
+ * @param {*} type The action type
+ * @returns {string} The url/path with the configured epanders added to the query string
```
tisto, line 27: epanders -> expanders

tiberiuichim:
I like everything about this PR. Though I think it might be more future-proof to make the expander registration a callable instead of a simple string. This way I have a bit more control per site section on what expands to request.

CHANGELOG.md
alecghica: Fixed.

src/middleware/api.js
alecghica: Fixed.

tiberiuichim:
@nileshgulia1 Use matchPath to match the registered expanders per router path: https://github.com/plone/volto/blob/7fe7004af15d441950b31f367797e073fb9b3513/docs/source/recipes/appextras.md

sneridagh:
Yeah it's about time we have this available.

sneridagh:
@tiberiuichim @nileshgulia1 Umm now that I realise, expanders only applies to the GET_CONTENT action. There's not point to add them to all the requests in the API, or I'm missing something?

tiberiuichim:
The expanders that @csenger envisaged are registered per action type, so that's maximum flexibility.

The expanders in plone.restapi are inserted from the content serializer, which is used primarily by GET_CONTENT, indeed. We could decide now that we only care about configurable expanding on the getContent action, but this is not "fixed in stone" from plone.restapi. We, indeed, use the expanders to piggy-back additional info on top of a single "central" request and that is usually the getContent. That said, I like flexibility... we can't foresee all the scenarios where this sort of thing could be "abused".

Though we could elect to keep it simple, fine with me.

sneridagh:
Fine then, I would continue then with the documented approach that @csenger drafted in the first place, along with your suggestion to include the `matchPath`.

@nileshgulia1 do you plan to continue with it?

I can jump too if required.

sneridagh:
The config then would look like this, right?

```
  apiExpanders: [
    {
      match: '',
      GET_CONTENT: ['mycustomexpander'],
    },
    {
      match: '/de',
      GET_CONTENT: ['myothercustomexpander'],
    },
    ...
  ],
```

nileshgulia1:
@sneridagh I added `matchPatch` to apiExpanders config in the settings and pushed whatever I had, feel free to push in this PR.

nileshgulia1:
I think more like for navigation instance : 
`{
        match: '/events',
        GET_NAVIGATION: ['navigation'],
      },
      `

tiberiuichim:
@nileshgulia1 We want things to work as in comment https://github.com/plone/volto/pull/1705#issuecomment-815530171

That is, on GET_CONTENT action, we want to inject an expander.

sneridagh:
@nileshgulia1 I added a test for the `addExpandersToPath` function. It fails, because I think that we don't need to stringify the qs since all browsers support UTF-8 nowadays in URLs. @tiberiuichim what do you think?

src/middleware/api.js
```diff
@@ -15,6 +20,39 @@ import {
...
+      const match = matchPath(pathPart, reg.match);
+      return match ? reg[GET_CONTENT] : null;
+    })
+    .filter((reg) => reg);
+  let query = qs.parse(qs.extract(path));
+  let expand = join(compact([query.expand, ...Expanders]), ',');
+  if (expand) {
+    query.expand = expand;
+  }
+  const stringifiedQuery = qs.stringify(query);
```
sneridagh, line 49: I think we can live without it.

src/middleware/api.js
```diff
@@ -15,6 +20,39 @@ import {
...
+  const { settings } = config;
+  const { apiExpanders = [] } = settings;
+  if (!path) {
+    return path;
+  }
+  const pathPart = path.split('?')[0] || '';
+  const Expanders = apiExpanders
+    .map((reg) => {
+      const match = matchPath(pathPart, reg.match);
+      return match ? reg[GET_CONTENT] : null;
```
sneridagh, line 41: We need to filter this for all possible actions that could come in the config, not only `GET_CONTENT`, pass it as a parameter, then filter it too.

Also, I think this overall filter can be improved and simplified a lot.

src/middleware/api.js
```diff
@@ -15,6 +20,39 @@ import {
...
+ * @returns {string} The url/path with the configured expanders added to the query string
+ */
+export function addExpandersToPath(path, type) {
+  const { settings } = config;
+  const { apiExpanders = [] } = settings;
+  if (!path) {
+    return path;
+  }
+  const pathPart = path.split('?')[0] || '';
+  const Expanders = apiExpanders
```
sneridagh, line 38: `Expanders` not capitalised, it's not a Class.

sneridagh (changes requested):

tiberiuichim:
@nileshgulia1 Don't take my comment to literally, GET_CONTENT was mentioned as an example. @csenger original implementation was for all actions based on their type, we want to maintain that. What I meant was that if you're mentioning the GET_NAVIGATION action and you're showing that you're expanding "navigation" for it, I get confused.

nileshgulia1:
Sorry for the confusion, I just revert the things I added in this [commit](https://github.com/plone/volto/pull/1705/commits/9e50de6238a785cf530b68d7ef84d73eba81b94e) where I used action type as a param to filter expanders.

src/middleware/api.js
nileshgulia1: Done!

src/middleware/api.js
```diff
@@ -15,6 +19,39 @@ import {
...
+    matchPath(pathPart, reg.match) ? reg[type] : null,
+  );
+  let query = qs.parse(qs.extract(path));
+  let expand = compact([query.expand, ...expanders]);
+  if (expand) {
+    query.expand = expand;
+  }
+  const stringifiedQuery = qs.stringify(query, {
+    arrayFormat: 'comma',
+    encode: false,
```
nileshgulia1, line 47: @sneridagh Not sure, why `{encode: false}` wouldn't work.

sneridagh (approved):

tiberiuichim (approved):

nileshgulia1 (approved):

tiberiuichim:
@nileshgulia1 @sneridagh Thanks for pushing this forward!

Issue #1594, opened 2020-06-17
Bug report: Edit page through Contents list


terapyon:
I found a bug at the Edit page through the Content list.

Actions:

- View the top page `/`
- Click to Contents menu of left sidebar `/contents`
- Click to any folder in the contents list `/FOLDER_ID/contents`
- Select a dropdown menu of any item, then click the edit link, 
- Move to item URL `/FOLDER_ID/ITEM_ID/edit`
- Showing to edit for top page. This is a bug point.

I think this bug is related bottom of `/src/components/Edit/Edit.jsx`. (`asyncConnect`)

Anyone, did you have the same situation?

terapyon:
I tried to modify the code. But I could not make the test. Please help me.
https://github.com/plone/volto/commit/7ff21d0aa8d815ae67da0a8d2a0f819277571b87

pnicolli:
See also #1575

terapyon:
@pnicolli I think the same as me. Thank you for your notification.

pnicolli:
I guess the best solution for testing this kind of interactions is cypress. You could try to add a cypress test that follows the actions you described.
Meanwhile, right now we are trying your fix locally.

pnicolli:
FYI, the fix proposed by Manabu works in our site. I totally forgot about it and to come here and confirm it.

terapyon:
Thank you for merge!!

PR #1707, opened 2020-07-30
Bugfix/edit context

csenger:
See #1594


iFlameing:
@csenger I just re-run the task and now acceptance test is passing :)

PR #1709, opened 2020-07-30
Fix changelog for recurrence widget entry

nzambello:
The entry for the recurrence widget is gone under 6.5.0 release in the changelog, but that’s not correct


tisto:
@nzambello I usually fix and check those things before making a release. Thanks for making my life easier though. :)

Issue #1654, opened 2020-07-06
Uploading an image in the Image block replaces the global state.content store


tiberiuichim:
CREATE_CONTENT_SUCCESS reducer:

https://github.com/plone/volto/blob/master/src/reducers/content/content.js#L131

Upload of image:

https://github.com/plone/volto/blob/master/src/components/manage/Blocks/Image/Edit.jsx#L180

I propose that we implement my transformer-based approach to the Image upload.

tiberiuichim:
@nzambello Your ideas on this?

nzambello:
I don't know right now the best solution, but can't we manage the creation action using a subrequest or something like that?
That's how getContent works, you can call it in a subrequest and I think this is what we're missing here.

PR #1711, opened 2020-07-30
Form context clean breaking with hoc

tiberiuichim:
**What is this?**

At its base, this PR separates the internal state of Form.jsx into its own component and exposes it as a context to any of its children or grandchildren.

**What are the benefits?**

- Performance: the `properties` doesn't need to be passed to the EditBlocks anymore, so each edit block won't re-render when other blocks are changed.
- It enables intra-form communication (via the form global state) for edit widgets. Blocks are being passed `properties={formData}` but widgets don't have that
- It enables form state mutation from any of the form components

**What can we do with it?**

I have multiple uses and it enables some advanced scenarios. For example:

- Inter-widgets communication. I have a widget where I can set "ElasticSearch Hostname" and the next widget is a dropdown with a list with the indexes on that ES Server. That means that the second widget needs to be able to read the values from the first widget (so that it knows which server to query for index values). I've used [a similar idea](https://github.com/eea/volto-searchkit/blob/599deced62f69bd041c488cfa90c03d937b840de/src/components/Widgets/SelectField.jsx#L27) in the volto-searchkit prototype.
- "Batch" editing of blocks. Right now the `on*Block`  methods don't use promises, so it's not easily possible to ensure order of execution or state integrity, so I'm forced to wrap mutation calls in setTimeout, which seems to ensure a sort of "synchronicity" in multiple setState calls. See [this](https://github.com/eea/volto-slate/blob/387129e652740da45d424b590e0b64ca6b864cb1/src/utils/volto-blocks.js#L70) as a counter-example of the workarounds that are needed.

**Is this a breaking change?**

Not really. I haven't removed yet the [properties being passed to EditBlock](https://github.com/plone/volto/blob/0542851a68ff73f519a9e97cc06e2c41420d76a2/src/components/manage/Form/Form.jsx#L777), but I'd like to do that. Removing it would improve the performance of the edit form, as blocks wouldn't have to re-render when one of them is changed (unless they specifically watch the context formData). I've changed all the Volto blocks that needed any change, so that we are prepared to remove that passed `properties` parameter. I don't know if there's any other blocks "out there" that depend on the "properties" passed to them. 

All of this is quite cleanly implemented and the advanced behavior is "opt-in". Child components need to use `useFormStateContext()` or `static contextType = FormStateContext` to get access to the ancestor form's internal state/setState.

TODO to further enhance this PR:

- wrap the other form implementation (ModalForm, InlineForm) in the FormStateContext provider
- Explain all this in documentation

Some potentially breaking changes:

- onAddBlock returns a promise instead of the newly created block id. I've already adjusted Volto's blocks for this change. While it did work even before, I don't know about how correct it used to be. setState is an async operation, so [this code](https://github.com/plone/volto/blob/a68a175357f9d867d12252b5021d804cf3fac965/src/components/manage/Blocks/Title/Edit.jsx#L145) is effectively doing:

```
setState(blockdata)
setState(selectedBlock)
```

With this change, the above operation becomes:

```
setState(blockdata, () => setState(selectedBlock))
```
I'm not sure about the performance impact (React batches and optimizes multiple setState calls), but at least this way it guarantees that the block id really exists in the DOM when the promise is resolved.

- I've not removed the properties prop being passed to EditBlock in Form.jsx. I think we can provide a warning using propTypes about it.


tiberiuichim:
It seems that uploading an image breaks the form. I'm guessing it's because of https://github.com/plone/volto/issues/1654

tiberiuichim:
A sort of "technology demo" of what the form state context enables, a new plugin: https://www.youtube.com/watch?v=iTaPsWLGTSQ https://github.com/eea/volto-tabsblock

tiberiuichim:
The image upload problem will be fixed by https://github.com/plone/volto/pull/1730

nzambello:
I like this, but I will look after this to test it

src/components/manage/Form/FormContext.jsx
```diff
@@ -0,0 +1,75 @@
...
+          </FormStateContext.Consumer>
+        </FormStateProvider>
+      );
+    }
+  };
+
+  return React.forwardRef((props, ref) => (
+    <FormWithState {...props} formRef={ref} />
+  ));
+};
```
avoinea, line 75: Docs would be appreciated

avoinea (changes requested):

src/components/manage/Form/Form.jsx
```diff
@@ -278,45 +321,34 @@ class Form extends Component {
...
-            ...this.state.formData,
-            [blocksFieldname]: {
-              ...this.state.formData[blocksFieldname],
-              [id]: value || null,
-            },
-          },
-        });
-        return;
-      }
-    }
```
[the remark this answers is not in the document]
tiberiuichim: good catch. Bad git merge

src/components/manage/Blocks/LeadImage/Edit.jsx
```diff
@@ -95,7 +98,7 @@ class Edit extends Component {
           />
         )}
         <SidebarPortal selected={this.props.selected}>
-          <LeadImageSidebar {...this.props} />
+          <LeadImageSidebar {...{ ...this.props, properties }} />
```
nzambello, line 101:
```suggestion
          <LeadImageSidebar {...this.props} {...properties} />
```
This leads to the same result but it would be more readable IMHO

nzambello (changes requested):
I agree with @avoinea comments and I added another one

tiberiuichim:
I don't think it's a good idea to merge this PR right now.

The problem is that the "form state context" is not a generic concept. Now that I've started working on the idea of "subforms", I realize that we don't want to deal with multiple states when editing a page. If I'm editing a "subform" (represented as a column), I want to treat that subform as a stateless input component: it gets its content (value) as a prop, it calls a parent-provided callback to pass its changes. To have its own state would introduce concurency problems, I don't want to have those problems. 

I've stated to go at a different path: taking @nzambello idea of `onChangeFormData`, I'm not keeping form state in the parent. This leads to neat code, like in the ColumnsBlockEdit component here: https://github.com/eea/volto-columns-block/blob/b65e4f71c4071eea263b8a389d2caab887fac552/src/ColumnsBlock/ColumnsBlockEdit.jsx#L51

PR #1713, opened 2020-08-02
A new helper "applyConfig" that ensures that all addon config methods return config

tiberiuichim:
When writing volto-slate, which has an extensible architecture based on plugins and configuration insertion modeled after Volto's, I've noticed that it's very easy to do mistakes and create configuration methods that don't return the config object. So I've tweaked the addon loader code to use this safeWrapper and I've also added a helper called applyConfig that loads configuration methods. 

See the modifications in addons.md file in this PR for more details.


tiberiuichim:
This is ready to be merged

avoinea (approved):

iFlameing (approved):

PR #1714, opened 2020-08-06
Rename dummy-addons-loader to jest-addons-loader

tiberiuichim:
Also added some docs and tweaks on project configuration to load and test addons.


tisto:
LGTM. Though, we need to think about how to include this in the upgrade guide. Maybe we should wait for the next major release...

tisto:
@tiberiuichim I'd like to include this into the upcoming Volto 8 release. Would it be possible that you add a few lines to the Volto upgrade guide about this?

tisto:
@tiberiuichim @avoinea any updates here? I'd like to make the Volto 8 release today. Ping me. :)

tisto:
@tiberiuichim @avoinea crap. Just saw that Tiberiu already added the upgrade guide two days ago. Sorry folks! Will do a release now.

PR #1718, opened 2020-08-10
Adding sortby feature

### LINKED ISSUE #1716: Allow to sort search results
The Volto search should support sort_on parameter e.g.:

https://volto.kitconcept.com/search?SearchableText=Page&sort_on=Title

plone.restapi supports the basic sort params of the ZCatalog:

```
sort_on
The field name to sort the results on
sort_order
ascending or descending, with the default being ascending. Note that you can also use reverse as a synonym for descending
sort_limit
Since you’re likely to only want to use the first 20 or 50 or so items, we can give a hint to the ZCatalog not to bother to sort beyond this by passing along a sort_limit parameter, which is the number of records to sort.
```

https://zope.readthedocs.io/en/latest/zopebook/SearchingZCatalog.html#sorting
https://plonerestapi.readthedocs.io/en/latest/searching.html

iFlameing:
Fixes #1716


csenger:
@iFlameing 

* The handler ignores all parameters already set for the search. I suggest using the pagination change handler as a blueprint for the sort button handler. It uses the toSearchOptions() function to build the options for the contentSearch action. https://github.com/plone/volto/blob/63a2e1a4553c229c7baef4d39deea12f840b7eb8/src/components/theme/Search/Search.jsx#L139-L146
* You might also have to reset the pagination to page 1 if the sort oder changes. @tisto?
* Sorting needs 2 possible parameters: sort_on (the index to sort on) and sort_order ("ascending" or "descending"/"reverse"). If you search for newest date sort_on=effective and sort_order=reverse. You can see it in action here: [Example](https://volto.kitconcept.com/api/@@search?sort_on=Date&sort_order=reverse&SearchableText=Page&advanced_search=False&pt_toggle=%23&portal_type%3Alist=Event&portal_type%3Alist=Image&portal_type%3Alist=Link&portal_type%3Alist=Document&portal_type%3Alist=Collection&portal_type%3Alist=Folder&created.query%3Arecord%3Alist%3Adate=1970-01-02&created.range%3Arecord=min)
* All search parameters needed to be connected to URL-Parameters:
https://github.com/plone/volto/blob/63a2e1a4553c229c7baef4d39deea12f840b7eb8/src/components/theme/Search/Search.jsx#L296-L323

tisto:
Yes. Pagination should be reset to page 1 if the user chooses a different sort order.

iFlameing:
Fixed the problem related to the whole query string search and the current page issue.
.
```js
queryString= `sort_on=sortable_title&sort_order=reverse&SearchableText=Page*&advanced_search=False&pt_toggle=#&created.query:record:list:date=1970-01-02&created.range:record=min&portal_type:list:list=Event&portal_type:list:list=Image&portal_type:list:list=Link&portal_type:list:list=Document&portal_type:list:list=Collection&portal_type:list:list=Folder`
```

csenger:
* sorting by newest date needs sort_on=effective and sort_order=reverse.
* Using setState() immediately followed by searchContent is not guaranteed to work. setState is asyncronous and getContent might be called before the state is changed. This is the right form to handle that: https://github.com/plone/volto/blob/2048cf1c1e0d206b61e1e8027e5a19cd3612a015/src/components/theme/Search/Search.jsx#L141-L146
* The Search page behaves inconsistent now.
  * sort_on is not handled on page load. all other search related actions don't use the full this.props.location.search
  * sort_on is lost if you sort and then change to the second page of a bigger, paginated result set
  All search actions need to use all location.search parameters and the sort buttons need to change the url so it is picked up in all places (sort_on and sort_order)
* The button texts are not translated and inconsistent: "Alphabetically" and "date(newest first)". This should both be Camel or lower case and a space before the open bracket.
* The styling or UI needs to be reworked. You can't see which option is selected. The box with the 2 buttons does not look good enough. Maybe we should move that to the right side of the result count and make it smaller buttons or a drop down.

iFlameing:
@csenger Yeah!! I really don't know why we are using to search options. going to remove this and discuss about the UI with timo

iFlameing:
@tisto Travis is happy we can now test it locally :)

tisto:
@iFlameing LGTM. We have to do two more things:

1) Write Cypress tests to make sure the custom sort orders actually work
2) Fix the UX/UI of the sort order buttons (maybe you can come up with a proposal based on the book we purchased recently :))

tisto:
```
As anonymous user I can see the search results ordered by search rank

Given a document 'Colorless'
  and a document 'Color'
 When I search for 'Color'
 Then the search results will contain 2 results
  and the first search result is 'Color'

As anonymous user I can see the search results ordered alphabetically

Given a document 'B Color'
  and a document 'A Colorless'
 When I search for 'Color'
  and I sort the search results alphabetically
 Then the search results will contain 2 results
  and the first search result is 'A Colorless'

As anonymous user I can see the search results ordered by date (newest first)

TODO Alok
```

csenger:
Minor points:

* Is there a reason to have this.props.searchableText as it's own property
* the names of the handlers should be unified (`handleQueryPaginationChange`, `onsortHandle`). Looking at the code base it should be `onXxx`. And it should be properly camel cased to `onSortChange`
* The handler for the sort change should be a function you pass sort_on and sort_order to. The decision about the sort_order for a specific sort option shouldn't be done in an if block in the handler.

  Buttons should be 
  ```jsx
  <Button onClick={() => {this.onSortChange('effective', 'reverse')}}>
  ```
  and the handler function should take both arguments, sort_order maybe predefined to `ascending` or just undefined.

iFlameing:
@csenger I renamed the sort change function and also passing the argument.
Yes!! we need ` this.props.searchableText ` to show the searchable Text in UI. 
```js
              <h1 className="documentFirstHeading">
                {this.props.searchableText ? (
                  <FormattedMessage
                    id="Search results for {term}"
                    defaultMessage="Search results for {term}"
                    values={{
                      term: <q>{this.props.searchableText}</q>,
                    }}
                  />
                ) : (
                  <FormattedMessage
                    id="Search results"
                    defaultMessage="Search results"
                  />
                )}
              </h1>
```

cypress/integration/search.js
```diff
@@ -0,0 +1,126 @@
...
+        contentTitle: 'Color',
+      });
+      cy.waitForResourceToLoad('@navigation');
+      cy.waitForResourceToLoad('@breadcrumbs');
+      cy.waitForResourceToLoad('@actions');
+      cy.waitForResourceToLoad('@types');
+      cy.waitForResourceToLoad('?fullobjects');
+    });
+    it('As anonymous user I can see the search results ordered by search rank,', () => {
+      cy.visit('/colorless/edit');
```
tisto, line 23: @iFlameing please add our BDD scenario as comment inline, e.g. // Given a document 'colorless'

cypress/integration/search.js
```diff
@@ -0,0 +1,126 @@
+if (Cypress.env('API') !== 'guillotina') {
+  describe('Search', () => {
+    beforeEach(() => {
+      cy.autologin();
+      cy.visit('/');
+      cy.createContent({
```
tisto, line 6: @iFlameing I think this should go into the first test since it is only used there and bleeds into the other tests when added in beforeEach.

iFlameing: we are using this for the first and third test and I don't want to repeat the code. It only bleeding into the second test when we are adding A and B in front.

tisto: @iFlameing please do not try to avoid DRY in test. Tests should be explicit and readable This is far more important than DRY.

tisto:
@iFlameing pls rebase and implement the UX like here:

https://miro.medium.com/max/1400/0*miWi0WMP-TaU1yES.

left side: number of results
right side: sort by ...

Search (1)

Pls also show "No results" instead of "0" when there are not search results.

cypress/integration/search.js
iFlameing: ok, done

tisto:
@iFlameing seems that the search Cypress tests are freezing the tests on the CI.

iFlameing:
@tisto Ready for merge and release :) :)

tisto:
@iFlameing a few more amendments are necessary:

Search (1)

iFlameing:
@tisto the font-size of sortby is same i.e 1rem

tisto:
1rem is a relative font size that can result in different calculated font
sizes. What matters is the calculated and actual font size which seems to
differ.

tisto:
@iFlameing the font size of the sort-by options is: ".85714286rem" so the font-size set in CSS is not the same. Please make sure all three elements in the results/sort-by line have 14px calculated size (you can see this in dev tools under the "calculated" tab in CSS.

iFlameing:
@tisto you can now merge this when CI is green.

tisto (changes requested):
@iFlameing sorting by effective does not work. Create three documents and then sort by effective date. The document created last will not be the first entry in the search by effective date.

Please write a Cypress test for this scenario that proves that it is NOT working currently. Then we can move on to fix this problem.

Just to be clear: I want a red Cypress test on CI that only turns green once we fixed the issue.

tisto:
@plone/volto-team I could use a second review here please :)

tisto (approved):

Issue #1722, opened 2020-08-12
Listing Block crashes when switching to Image Gallery Template


jackahl:
Steps to reproduce: 

1. add a new listing block to page
2. Choose Image Gallery as template. (Do not add any criteria before!)
-> `TypeError: item[_config__WEBPACK_IMPORTED_MODULE_4__.settings.listingPreviewImageField] is undefined`

It does not happen when a criteria is set.

nileshgulia1:
fixed by #1723

PR #1723, opened 2020-08-12
Image gallery crash fix

jackahl:
Fix for  #1722 

Listingbody now detects whether it uses a the Gallery template and if so sends a request to the querystring endpoint to retrieve the fullobjects. This leads to the image gallery no longer crashing when no criteria are set (because than the block does not retrieve the fullobjects needed for that template)


src/components/manage/Blocks/Listing/ListingBody.jsx
```diff
@@ -68,9 +87,11 @@ const ListingBody = ({ data, properties, intl, path, isEditMode }) => {
     );
   }
 
+  console.log(listingItems);
```
tiberiuichim, line 90: console.log

jackahl:
@tisto can you check why cypress fails here? I dont really understand the trace.

tisto:
@jackahl did you try to run the Cypress tests locally? This is a regression on the listing block sort functionality.

src/components/manage/Blocks/Listing/ListingBody.jsx
tisto: @tiberiuichim I recall that I added an eslint rule to report console.log statements. No idea why this does not kick in...

PR #1727, opened 2020-08-15
Simplify README

tisto:


tiberiuichim:
@tisto: What's the idea with this change? I see Plone being recommended as the compatible backend, instead of "one of many".

tisto:
@tiberiuichim I'd like to simplify the introduction. We confuse the hell out of people if we tell them Volto is a frontend with three possible backends as the first elevator pitch. In real life there is only one possible backend which is Plone. Then there is the question if we pitch Volto as "React CMS" or as a new frontend for Plone. We won't be able to answer all those questions at once and we have to take into considerations discussions and decisions by the wider Plone community here. Though, I'd like to improve things bit by bit if possible. Therefore this draft PR is just a first step for an ongoing discussion.

tiberiuichim:
@tisto IMHO Volto goes beyond being a simple frontend for Plone. It's a natural evolution of Plone, it's closely related to Plone, it's a frontend for Plone, but it's also an extension of Plone. Stuff that you can't do with Plone (or can't do easily), become simple with Volto.

If it's to be a frontend for Plone, then it's not a complete frontend. We still have to implement quite a lot to reach feature-parity.

tiberiuichim:
It also changes some paradigms that are valid for Plone, how to structure a website, how to create its layout, what is customizable and what is not, so it's both a subset of Plone and a superset of Plone. I think it's valid to develop new functionality, while trying to reach feature parity with "classic" Plone.

PR #1730, opened 2020-08-17
Don't overwrite the global state content data when uploading images in blocks

tiberiuichim:
This solves the #1654 issue with the image block edit where it overwrites the redux state.content with the uploaded image data.


src/components/manage/Blocks/Image/Edit.jsx
```diff
@@ -95,8 +95,7 @@ class Edit extends Component {
   }
 
   /**
-   * Upload image handler (not used), but useful in case that we want a button
-   * not powered by react-dropzone
+   * Upload image handler
```
tiberiuichim, line 98: Don't delete the comment

tiberiuichim:
I think I'll rework the code a bit to avoid the { items: [] } list.

So, the idea is, in reducers the action.result can be:

- either an object (for example in `GET_CONTENT_SUCCESS`)
- or a list (in, for ex `GET_TYPES_SUCCESS`)

To make the code work and avoid the need for separate code in the `content` `GET_CONTENT_SUCCESS` + `CREATE_CONTENT_SUCCESS`, I've made the `action.result` fit the interface for `GET_CONTENT_SUCCESS`, which requires `action.result` an object and `action.result.items` as a list

But I think a better way and to avoid adding another special case, is to rewrite the GET_CONTENT_SUCCESS to be dealt as a separate code, different from GET_CONTENT_SUCCESS, and to deal with `action.result` as a list there, because GET_CONTENT_SUCCESS should never deal with `action.result` as a list. This would avoid the breaking change

src/middleware/api.js
```diff
@@ -68,15 +68,21 @@ export default (api) => ({ dispatch, getState }) => (next) => (action) => {
...
-              return api[item.op](item.path, {
+        ? request
+            .map((item) =>
+              api[item.op](item.path, {
                 data: item.data,
                 type: item.type,
                 headers: item.headers,
+              }),
+            )
+            .reduce((chain, apiCall) => {
```
tiberiuichim, line 79: Here it doesn't waterfall the requests anymore. To be fixed

iFlameing: fixed

iFlameing (approved):

tiberiuichim:
@iFlameing nice, thanks!

tisto:
@plone/volto-team @nzambello @giuliaghisini @avoinea @pnicolli @cekk I'd like to have at least one more review on this PR if possible.

giuliaghisini (approved):

avoinea (approved):
LGTM. 

If I understand well, this is not a breaking feature as the new components will write to `subrequests`, while the old ones (within custom projects) will continue to use the old way until they're updated.

tisto:
@tiberiuichim can you pls confirm that this is not a breaking change?

tiberiuichim:
@tisto It's not a breaking change, Alin had the same concern, we've looked together and concluded it's not. All "third-party" blocks that were based on the Image Block code will continue to be as broken as the image block used to be, but at least now the Image block is fixed.

The idea is: 

- even though it didn't look like, it was broken. I've noticed the problem, Nicola noticed the problem, I've added the ticket https://github.com/plone/volto/issues/1654 but it was based on discussions with him. 
- createContent now takes an optional "subrequest" parameter, just like getContent
- The image block now uploads images with createContent(..., this.props.block), so the content is "namespaced" for the block

PR #1732, opened 2020-08-17
fix bug that lead to data from one page being displayed in the edit-v…

jackahl:
…iew of another


jackahl:
This needs to be reverted ASAP as it completely wrecks the image upload!

Issue #1719, opened 2020-08-11
Users & Groups Control Panel: Add "show all users" button


tisto:
![Users-and-Groups](https://user-images.githubusercontent.com/632105/89869615-e58c6100-dbb4-11ea-83b2-d88179e35c02.png)

tisto:
![Nutzer-und-Gruppen](https://user-images.githubusercontent.com/632105/90372230-5d083780-e070-11ea-9b28-42c4bbeed9c4.png)

PR #1733, opened 2020-08-17
Adding show all user feature in user controlpanel

iFlameing:
Fixes https://github.com/plone/volto/issues/1719


avoinea:
@iFlameing I'm wondering how this feature works with thousands of users (e.g. LDAP)

tisto:
@avoinea that is something we have to take into account for sure. We are about to deploy Volto in an intranet with 10k users. My plan was to go ahead with this and see what happens. :)

(we are still in pre-production mode for another two months, therefore we can live with possible problems right now and we could take the bullet here)

avoinea:
@tisto I see. One small issue I've notice, component state is not reset/re-initialized on exit/enter:

users

tisto:
@iFlameing can you please look into this ^^^? And also re-run "yarn i18n"?

iFlameing:
@avoinea can you please review this once again.

avoinea:
@iFlameing can you please rebase.

iFlameing:
@avoinea done :) @tisto Why do we need this feature? Regarding showing large list we can use react-window :)

avoinea (approved):
LGTM.

A nice to have would be to also fix https://github.com/plone/volto/issues/1898 for the Users CPanel. But we can also do it within another PR if it's not straightforward.

iFlameing:
@tisto rebase is done

Issue #1734, opened 2020-08-18
Subtle bug in Form.jsx unsafe_componentWillReceiveProps


tiberiuichim:
I've merged Volto master in the "form state with context hoc" PR and the tests run into a recursive update caused by https://github.com/plone/volto/pull/1732/files

I think the code is wrong there, as shown by this simple example:
```
> {a:'1'} !== {a: '1'}
true
```

Of course, it could be argued that the form state context should not be so sensitive. In any case, using lodash.isEqual(nextProps.formData, this.state.formData) is better and more true to the original intention.

tisto:
@jackahl can you have a look into this?

jackahl:
should be fixed in here: https://github.com/plone/volto/pull/1736

PR #1736, opened 2020-08-18
fix issue in form that lead to broken image block

jackahl:


jackahl:
Fixes this: https://github.com/plone/volto/issues/1734

tiberiuichim:
This PR also fixes the image block issue: https://github.com/plone/volto/pull/1730

But I think both PRs are good to have in Volto

PR #1741, opened 2020-08-19
How to use an addon

ksuess:
- Clear separation of "Use of an addon" and "Creating an addon".
- Changed order to first "Use of an addon" then "Creating an addon".


tiberiuichim (approved):
Nice job!

PR #1744, opened 2020-08-19
Making minor changes regarding context aware of objectBrowser

### LINKED ISSUE #1740: Sidebar should use parent folder as starting point when editing an existing document.
We recently changed the sidebar to open in the context of the page. Though, we did not take into account that we have two possible states when editing an object:

1) object creation: the sidebar opens in the parent folder where the object is about to be created (that's fine)
2) object editing: when editing an existing object the sidebar shows the object itself, which is in most cases empty. This confuses the user. It would be better to keep this consistent with 1) and show the parent folder.

Edit-Page

iFlameing:
Fixes #1740


tiberiuichim (approved):

tisto:
@iFlameing this PR breaks the upload of an image when creating a new page. Steps to reproduce:

1) as admin
2) click create new page
3) add image block
4) upload image to image block
5) hit save button
-> image block goes away and js console is full of errors

tisto:
@iFlameing please write a Cypress test that creates a new document and adds an image block. I'd like to see this test in this PR and it should fail.

iFlameing:
@tisto This has nothing do with the broken image block you can see the jakob comment in this pr which broke the functionality https://github.com/plone/volto/pull/1732. I have merged the recent changes locally and it solves the above problem.

iFlameing:
If you still want I can write the cypress test.

cypress/integration/blocks-image.js
```diff
@@ -93,5 +91,23 @@ if (Cypress.env('API') !== 'guillotina') {
     //   cy.wait(5000);
     //   cy.url().should('eq', Cypress.config().baseUrl + '/my-page');
     // });
+    it('Add image via upload', () => {
```
tisto, line 94: @iFlameing this is a test for an existing document in edit view. The problem I ran into only occurred when I created a new document and added an image. That's a different test setup.

tisto: @iFlameing did you see my comment here? ^^^

iFlameing: I saw it that's why I have written another test in edit mode

tisto:
Works like a charm. I think this is what the user would expect. Though, we all have to test this in real life and see if that's the final thing we all want. cc @plone/volto-team

giuliaghisini (commented):
LGTM!

PR #1749, opened 2020-08-20
Now you can also paste video links from Volto CMS in video blocks

### LINKED ISSUE #1745: Assign a local video to video block is not working
Trying to create a video block from a local video on same volto server, doesn't work for me.

I've tried it in https://volto.kitconcept.com.

The process:

- Add a file (a MP4 file)
- Then add a page and add a video block inside.
- Inside the video url I have copied the path https://volto.kitconcept.com/videos/file_example_mp4_480_1_5mg.mp4 and it doesn't work.

But it worked if I add the url like https://volto.kitconcept.com/videos/file_example_mp4_480_1_5mg.mp4/@@download/file

I think the user don't need to know the @@download/file url to add a video.

The other thing is that video block doesn't have the browser object to search for videos in the current site (maybe the users should need to share videos from a local folder of the same site)

NOTE: In some cases, the /@@download/file doesn't work for me, I don't know if it depends on video type.

iFlameing:
Fixes #1745


nileshgulia1 (commented):
Tested and looks good.

avoinea (approved):
Tested. LGTM.

PR #1772, opened 2020-08-25
Refs #1771 - Handle custom Accept: headers in actions

### LINKED ISSUE #1771: Can't define custom action to get XML content from the backend
Trying to get content other format than `JSON` from the backend will end-up in a `null` response even if the request call is executed and the backend returns the right content:

``` jsx

import { GET_SVG } from './constants';

/**
 * Get SVG function.
 * @function getSVG
 * @param {url} url SVG URL.
 * @returns {Object} SVG Object.
 */
export function getSVG(url) {
  return {
    type: GET_SVG,
    request: {
      op: 'get',
      path: url,
      headers: {
        Accept: 'application/xml',
      },
    },
  };
}
```

getSVG

avoinea:
Fixes #1771


tiberiuichim (approved):
Maybe add some documentation about it?

tiberiuichim:
Another thing, is "files" a possible property in the response? See https://github.com/visionmedia/superagent/blob/v3.8.2/lib/node/response.js#L37

tisto (approved):

avoinea:
> e response? See [https://github.com/visionmedia/superagent/blob/v3.8.2/lib/node/respon](https://github.com/visionmedia/superagent/blob/v3.8.2/lib/node/response.js#L37)

I couldn't manage to get a response with `files` inside.

avoinea:
> Another thing, is "files" a possible property in the response? See https://github.com/visionmedia/superagent/blob/v3.8.2/lib/node/response.js#L37

I think this is already covered by https://docs.voltocms.com/developer-guidelines/redux/

tisto:
@avoinea just to be sure. This is not a breaking change, right?

avoinea:
@tisto Nope.

PR #1779, opened 2020-08-28
Fix translations by removing `<<<<<HEAD` artifacts from merge conflicts

steffenri:


tisto:
I wonder how we can prevent something like this via CI. Just grepping for ">>>>>" "<<<<<<" would do the trick I guess.

PR #1790, opened 2020-09-04
Vertical form tabs

giuliaghisini:
By default, forms are displayed with horizontal tab at the top.
If you have a content type with a lot of tab, the horizontal arrangement fails to contain them all.
You could define, in config.js to display form tab as a vertical list in all your project, changing the variabile `verticalFormTabs` to true. 
Schermata 2020-09-04 alle 10 31 45

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ### Feature
 
+- Render form with vertical tabs, setting the property `verticalFormTabs` in config.js @giuliaghisini
+
 ### Bugfix
 
 ### Internal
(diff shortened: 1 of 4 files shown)


tiberiuichim:
@giuliaghisini maybe set them automatically to vertical if more then 6 tabs are needed?

giuliaghisini:
> @giuliaghisini maybe set them automatically to vertical if more then 6 tabs are needed?

the idea was to make all volto form equals, to avoid confusing the user. But, we could discuss it!

tisto:
@tiberiuichim automatically switch would confuse the user. This should be consistent across the site IMHO. Therefore I think it makes sense to have this as a global setting for integrators.

tisto (approved):

tiberiuichim (approved):

PR #1793, opened 2020-09-04
enabled no-folderish CT to be translated

giuliaghisini:
This pr shows the 'add' button on the toolbar, if a content type is not folderish and multilingual is enabled, to enable ct translations.

### DIFF
--- a/src/components/manage/Toolbar/Types.jsx
+++ b/src/components/manage/Toolbar/Types.jsx
@@ -8,40 +8,45 @@ import { flattenToAppURL } from '@plone/volto/helpers';
 import { settings } from '~/config';
 
 const Types = ({ types, pathname, content, currentLanguage }) => {
-  return types.length > 0 ? (
+  return types.length > 0 ||
+    (settings.isMultilingual && content['@components'].translations) ? (
     <div className="menu-more pastanaga-menu">
-      <header>
-        <FormattedMessage id="Add Content" defaultMessage="Add Content…" />
-      </header>
-      <div className="pastanaga-menu-list">
-        <ul>
-          {map(filter(types), (item) => {
-            // Strip the type for the item we want to add
-            const contentTypeToAdd = item['@id'].split('@types/')[1];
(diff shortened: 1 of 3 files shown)


src/components/manage/Toolbar/Types.jsx
```diff
@@ -8,40 +8,45 @@ import { flattenToAppURL } from '@plone/volto/helpers';
...
+          <header>
+            <FormattedMessage id="Add Content" defaultMessage="Add Content…" />
+          </header>
+          <div className="pastanaga-menu-list">
+            <ul>
+              {map(filter(types), (item) => {
+                // Strip the type for the item we want to add
+                const contentTypeToAdd = item['@id'].split('@types/')[1];
+                // If we are in the root or in /contents, we need to strip the preceeding / and /contents
+                const currentPath = pathname
```
tiberiuichim, line 25: Use getBaseUrl, see https://github.com/plone/volto/blob/master/src/helpers/Url/Url.js#L15

tiberiuichim (commented):
I've never tested multilingual with Volto, but would it not make more sense to have a separate "Translate" button, if that is needed to provide translations?

giuliaghisini:
> I've never tested multilingual with Volto, but would it not make more sense to have a separate "Translate" button, if that is needed to provide translations?

Hi @tiberiuichim ,
maybe. I talked with Timo about this, and we thought it would be better to ask who designed the interface to think about it, because his goal was to minimize the buttons in the toolbar

tisto:
@tiberiuichim Albert recommended to put the "translate to..." button to the add menu because you are actually adding new content. Though, I guess we did not think about the use case when an object is non-folderish and can not contain any other content object. I reached out to Albert to ask about his opinion. This PR is about an intermediate solution to fix the use case explained above in short-term.

tiberiuichim (approved):

sneridagh (approved):

sneridagh:
almost ready, needs some final review

PR #1811, opened 2020-09-13
Hide block chooser button with React logic instead of CSS

tiberiuichim:
Instead of hiding the button from CSS, I've used the selected prop (which is required).

This makes it easier to have nested blocks  (because the CSS rule would apply too indiscriminate)


tiberiuichim:
image_2020_09_21T10_00_58_946Z

There's a very misleading bit of test. In effect the Cypress is doing a click on a hidden element, which the editor can't really do. My PR just uncovered this broken test, I'll push a fix

iFlameing:
LGTM!! and also tested locally.

iFlameing (approved):

sneridagh (approved):
LGTM

nzambello (approved):

Issue #1808, opened 2020-09-10
Volto forgets authentication when hot-reloading


tiberiuichim:
I've noticed (since forever) an annoyance that comes with developing with Volto. 

Steps to reproduce:

- start volto (`yarn start`)
- make some changes in code, wait for the hot reload to kick in (maybe make a "partial save", with incomplete code, to make React spill an error)
- now you're logged out of the website.

I've noticed a workaround: after the initial login I go and restart Volto, which makes the authentication "stick". My guess is that it's something related to this `setTimeout` [here](https://github.com/plone/volto/blob/master/src/helpers/AuthToken/AuthToken.js#L48) but I haven't really looked at the problem in a really long time.

tisto:
Yeah. I fully agree. Victor built a Chrome plugin to auto-login into Volto to work around this. Fixing this would be fantastic.

PR #1812, opened 2020-09-13
Improve hot-reload auth; use a one-hour refresh period for auth token

### LINKED ISSUE #1148: Unable to login if /acl_users/jwt_auth/manage_config Token Validity Timeout is 0
Trying to login with a Token Validity Timeout value set to 0 (configured inside Plone) immediately redirects to the logout page.

I think it has to do with this code in ``AuthToken.jsx``:

```
(jwtDecode(store.getState().userSession.token).exp * 1000 - new Date().getTime()) * 0.9);
```

In case the Token Validity Timeout is 0, the ``jwtDecode(store.getState().userSession.token).exp`` is undefined and the result of the expression is nan, which immediately triggers the setTimeout callback and redirects to the /logout page.

tiberiuichim:
… if token expiration is 0. This closes #1148

For the hot reload part, it only partially improves the story. This [might be relevant](https://medium.com/@denizism7/setup-and-configure-webpacks-hmr-and-react-hot-loader-in-a-react-redux-application-13a5bfdaacdb), although I think Volto already has some stuff for hot-reloading Redux store? Or am I wrong?


tiberiuichim:
I think the redux store is not reloaded, for example you can't trigger router location changes after hot reloading. In this case a full page reload is needed. At least it get rids (from my limited tests) of the forced logout.

tiberiuichim:
Also closes #1808

avoinea (approved):

tisto (approved):

tisto:
Note: we will test this in our projects and then merge it. Code looks good though.

PR #1737, opened 2020-08-18
Fixing bug related to closing the more menu in toolbar

iFlameing:
Fixes #1731

PR #1815, opened 2020-09-14
Profile toolbar fix

jackahl:


jackahl:
@iFlameing The bug was introduced with #1737 . My Pr fixes it bu reintroduces the problem with closing the  Add item menu. I'd argue that this is something we need to live with till fixed, because the issue with the profile menu is more severe.

iFlameing:
I think that we can solve it using CSS. I played with CSS in the browser and looks like we can solve it using margin-left.

jackahl:
@iFlameing than pls push whatever fix you found to this branch, so I can also test it.

PR #1816, opened 2020-09-14
Add Style component and AlignWidget

tiberiuichim:
The `<Style>` component is a wrapper that reads the `AlignWidget` information, which just wraps the AlignBlock helper.

I think we also should have a SizeWidget counterpart, not done yet.


src/components/manage/Widgets/AlignWidget.jsx
```diff
@@ -0,0 +1,131 @@
...
+          </button>
+          <button
+            aria-label={intl.formatMessage(messages.delete)}
+            className="item ui noborder button"
+            onClick={() => onDelete(id)}
+          >
+            <Icon name="close" size="large" color="red" />
+          </button>
+        </div>
+      )}
```
avoinea, line 120: Not needed anymore. Moved to `FormFieldWrapper`

avoinea (changes requested):

avoinea (approved):

sneridagh (commented):
@tiberiuichim may I be picky about naming? :P
Isn't `Style` too generic? Since we want (and we should) to have a SizeWidget, why don't name it accordingly? Like `SizeWrapper` or `SizerWrapper`. Just sayin'!

For the rest, LGTM.

tiberiuichim:
@sneridagh the Style component reads and applies both the size and align information.

Very high on my todo list for today is to come up with a generic styling framework for all blocks. I want to have a easily reusable component that can give me this: 

- when I edit a block, I want to chose a style for the block, from a palette of predefined styles. Think "quote block" type, or "green block" that would apply a specific background to the block, and so on. 
- I also want to be able to float blocks around, set their size, etc. (But this part is easy if the first part is done).

src/components/manage/Widgets/AlignWidget.jsx
```diff
@@ -0,0 +1,27 @@
...
+import { injectIntl } from 'react-intl';
+
+import { FormFieldWrapper } from '@plone/volto/components';
+import { AlignBlock } from '@plone/volto/helpers';
+
+const AlignWidget = (props) => {
+  const { id, onChange, value } = props;
+  return (
+    <FormFieldWrapper {...props} className="align-widget">
+      <AlignBlock
```
tiberiuichim, line 17: Needs a `<div className="align-tools">` around it.

sneridagh (approved):
LGTM
@tiberiuichim we should open an issue about all the things that we need to document like this one.

nzambello (approved):

nzambello:
WARNING: this will merge the CHANGELOG entries in a past release

PR #1820, opened 2020-09-15
Fixing break-line in view mode

iFlameing:
@giuliaghisini can you please review this because I am removing your changes. I am doing this because when we press shift+enter we want to keep the empty line. Before this Pr, the edit and view mode is not the same.


tisto:
@iFlameing is there any way we can write a Cypress test to make sure this regression does not occur again in the future? Since our DraftJS amendments are always very brittle we should try to make sure those features are properly tested so we can detect regressions.

giuliaghisini:
> @giuliaghisini can you please review this because I am removing your changes. I am doing this because when we press shift+enter we want to keep the empty line. Before this Pr, the edit and view mode is not the same.

@iFlameing i've never changed those lines! That code was introduced in this pr: https://github.com/plone/volto/commit/aa5258b7c033aa0adb6e4776f9b452fe2e77c38d#diff-0bd4999bb5f7e62daf4e3b68871bfa05.

This pr scares me a little, and I would like to try it on some of our sites.
Anyway, your intent was to show in view all new-lines at the end of the string? If so, isn't enough to add new empty text blocks after the block containing the text?

tiberiuichim:
IMHO we're beating a dead horse here, the whole shift+return is trying to come up with a workaround that's against the spirit of the Volto richtext editor: each paragraph is a separate block and this way Volto blocks can be mixed with the text.

giuliaghisini:
> IMHO we're beating a dead horse here, the whole shift+return is trying to come up with a workaround that's against the spirit of the Volto richtext editor: each paragraph is a separate block and this way Volto blocks can be mixed with the text.

yes, but shift+return will be usefull when you have contentypes without blocks, and you have a field that contains formatted text

tiberiuichim:
@giuliaghisini if it's a "metadata" field (so not inside the block) then the enter key works as well, it creates paragraphs. In any case, I think having the break option is fine with shift+return is fine

tisto:
@tiberiuichim SHIFT+RETURN is a pretty basic requirement for all kinds of blocks. Not in the standard Volto editor of course but in more complex blocks like blocks with image left and text right or our teaser grid blocks or rich-text fields that do not have the Volto editor.

We made the UX decision that a return creates a new block and that's fine. Though, there are lots of edge cases where SHIFT+RETURN is our escape hatch. And our editors are actually used to that pattern from Word and other editors (something I wasn't aware of TBH).

iFlameing:
@tisto It is ready for merge :)

PR #1823, opened 2020-09-16
Fixed image upload test

zotya:


tiberiuichim (approved):

tisto:
@iFlameing can you pls check this PR?

iFlameing:
@zotya can you please! explain, what problem this pr solves because we really don't need this method ```Cypress.Blob.base64StringToBlob```. we already using cypress.upload which takes care of everything regarding upload.

tiberiuichim:
@iFlameing there are random test failures on travis, which can't be explained by the code changes in the PR. This should fix those.

tiberiuichim:
@iFlameing for example this one: https://github.com/plone/volto/pull/1811/checks?check_run_id=1119773199#step:17:461

iFlameing:
@tiberiuichim thanks! for explaining :)

iFlameing (approved):

tisto:
Let's give it a shot and monitor if this improves the situation. I prefer to go with the standard Cypress solution but if that's not good enough, we have to find workarounds to make the tests run in a stable manner.

PR #1837, opened 2020-09-21
Adding placeholder image in Video block

iFlameing:


src/components/manage/Blocks/Video/VideoSidebar.jsx
```diff
@@ -68,21 +72,42 @@ const VideoSidebar = ({
...
+                    resetSubmitUrl();
+                    onChangeBlock(block, {
+                      ...data,
+                      url: '',
+                    });
+                  }}
+                  onChange={() => {}}
+                />
+                <TextWidget
+                  id="external"
```
tiberiuichim, line 92: same id, external, as above

iFlameing: Fixed

tiberiuichim (approved):

tisto:
@iFlameing todo

- add circle to -> symbol (maybe we have a more youtube-like icon for play)
- replace image url with content browser
- nice-to-have: show preview image in edit mode

iFlameing:
@tisto We can't pass the custom icon to the embed semantic UI components. We are storing video URL to `url` parameter, if  I implement the object browser then URL field will get overwritten. I can solve it by changing the `url` to video URL and storing the preview image at URL field.

tiberiuichim (commented):
There's a way to programatically create a thumbnail URL for youtube videos. (maybe also vimeo). See https://stackoverflow.com/a/12788320/1220919

tisto:
Maybe @sneridagh can have a look here and give his opinion if the large refactoring is necessary and worth the effort. This is still a good feature to have in Volto.

sneridagh:
> @tisto We can't pass the custom icon to the embed semantic UI components. We are storing video URL to `url` parameter, if I implement the object browser then URL field will get overwritten. I can solve it by changing the `url` to video URL and storing the preview image at URL field.

You can choose the field to be written by the ObjectBrowser. See: https://docs.voltocms.com/blocks/editcomponent/#openobjectbrowser-handler-api

sneridagh:
@iFlameing I think we can go if:
- You use the ObjectBrowser (using another property than url to store the URL) and allow to paste any URL too (TextLineWidget)
- The play button maybe we can place a (transparent) wrapper that has the icon in the middle centered using CSS?

iFlameing:
@sneridagh @tisto I implemented the selecting image from the object browser and fix everything regarding showing placeholder.

One thing I am not able to solve is Icon. The semantic UI icon is broken.

tisto:
@iFlameing pls give us a heads up on this.

iFlameing:
@tisto done. All the functionality is there like everything. Only the icon is missing.

sneridagh:
Now it's ready :)

PR #1845, opened 2020-09-25
Support allowedBlocks within Text block / BlockChooser

avoinea:
See integration: https://github.com/eea/volto-group-block


sneridagh (approved):
LGTM

@avoinea @tiberiuichim wondering if we should document this somehow, somewhere. (Also things like this one).

tiberiuichim:
@sneridagh I think we should start really building the styleguidist-based docs. Then it becomes obvious where things should be documented (or at least they're documented somewhere, instead of nowhere).

tiberiuichim:
@sneridagh At EEA it was decided to go for Styleguidist and showcase each addon, complete with its components, it becomes hard to keep track of everything that's available and makes communication across companies easier.

tisto:
Without having a strong opinion on the actual decision I think it would make things a lot easier if Volto core and the core contributing companies use the same technology to document stuff...just my 2c.

tiberiuichim:
@tisto It would be nice to have Storybook, if available and we'd use it at EEA as well (for sure), but there's also an urgent need to adopt something. If we can't use Storybook then we have to go for Styleguidist.

tiberiuichim (approved):

sneridagh:
@tiberiuichim crap, I merged from master into this one, I hope I didn't break anything in blocks_form branch

tiberiuichim:
@sneridagh no worries :) nothing is broken, it was bound to receive master at one point

PR #1858, opened 2020-09-28
Showing notification when user reorder the content

### LINKED ISSUE #1854: Users should see a notification after successful reorder in folder contents
1) as admin
2) go to folder_contents
3) re-order the items in a folder
-> current behavior: the user does not get any visual indication that the reorder was successful
-> should: the user sees a toast message "Reordering items successful"

iFlameing:
Fixes #1854


tiberiuichim:
I see you're using updateRequest, and a message about "sorting", instead of "ordering". Is it correct?

iFlameing:
@tiberiuichim I first thought that reordering happens when we sort the content. But we can also reorder content by moving up and down. I should also include in this pr. Thanks!! for pointing out :)

tiberiuichim:
@iFlameing I'd change the messages to "Item succesfully moved"

iFlameing:
@tiberiuichim Done!!

sneridagh (approved):

tisto:
@iFlameing the toast messages pile up too quickly. When you move the lowest item all the way up to the top (without dropping) you see a new message for every single item in the folder:

Contents (3)

It seems the toast message is not bound to the actual reorder request that we do after a user dropped an item. Can you please make sure we show the toast message only when the user drops an item and an actual request has returned a 204 for the PATCH request (see https://plonerestapi.readthedocs.io/en/latest/content.html?highlight=reorder#reordering-sub-resources)

tisto (changes requested):
Toast meesages should show up only after a successful REST API response with the HTTP status 204 after the PATCH request for the reorder has been sent.

tisto:
@iFlameing can you please check the CI failures tomorrow morning?

PR #1859, opened 2020-09-28
Changing exclude-nav widget from checkbox to select

### LINKED ISSUE #1856: Bulk action for "exclude from navigation" does not allow to set to no
There is currently no way to set "exclude from navigation" via the bulk action to no. Therefore I'd propose to add a select field instead of the checkbox:

Contents (2)

iFlameing:
Fixes #1856


tisto:
@iFlameing we also need a way to say "do not change anything", otherwise we would force the editors to override existing settings for all chosen pages.

iFlameing:
@tisto What should we do when the editor chooses `do not change anything` because as default if you create a file it will appear in the navigation. Keeping in mind that it is not a required property.

iFlameing:
@tisto I am thinking that I should set the value of `do not change anything` to `false`.

tisto:
@iFlameing "do not change anything" won't touch any exclude_from_nav setting whatsoever.

iFlameing:
@tisto It is ready for review. I am attaching a screenshot for a demo purpose
Screenshot from 2020-10-15 15-55-04

iFlameing:
It's a default value in selectWidget. But I will look into it.

iFlameing:
@tisto after talking to @sneridagh we conclude that we are unable to remove the `No Value` from the select widget. If we change the description then it will be shown by all the other select widgets. The current Schema is coming from the backend so we can't mark it as a special case. So, we decided to remove the `3rd` options and go with the `No value`. I tested out and `No value` work same as `3rd options`.

You will notice that I also updated the message in Content.jsx because during testing out I found that updateRequest called whenever we make some update to content that's why I changed the message from 'Item has been sorted` to `Item has been updated`.

sneridagh:
@tisto indeed, this is the best that we can do now. Since the widget is generic, we would need the widget to be able to customize the "no value" value, which I guess it's a lot of work for this tiny use case. I would stick with the default "no value" for now.

tisto:
@iFlameing what's the status of this? Ready for review? Can you pls rebase?

iFlameing:
@tisto done :)

Issue #1861, opened 2020-09-29
Querystring action uses fullpath instead of specified relative path


jackahl:
Steps to reproduce:

1. create page
2. add listing block
3. set location criteria and configure it as relative path "."
4. save page
5. check @querystring-search request in network tab:

the path specified for "plone.app.querystring.operation.string.relativePath" is the fullpath instead of a relative path.

This is caused by this bit in the querystringsearch action: https://github.com/plone/volto/blob/e39351e961d866b27e29141bc4d2aaf69a8398b6/src/actions/querystringsearch/querystringsearch.js#L13

PR #1862, opened 2020-09-29
Relative querystring

jackahl:
Fix for #1861

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Bugfix
 
+- querystingsearch action now uses correct relative path, if specified. Fixes #1861 @jackahl
+
 ### Internal
 
 ## 8.2.0 (2020-09-27)
--- a/src/actions/querystringsearch/querystringsearch.js
+++ b/src/actions/querystringsearch/querystringsearch.js
@@ -9,15 +9,6 @@ import { settings } from '~/config';
  */
 export function getQueryStringResults(path, data, subrequest, page) {
   // fixes https://github.com/plone/volto/issues/1059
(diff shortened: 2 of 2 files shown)


sneridagh:
@jackahl I think the code you removed is required in order for the relative query to work. If I recall correctly @pnicolli did that:

https://github.com/plone/volto/pull/747/commits/f0797dc90f9d63387d3abe6c5259340e1386e14a

Are you sure now we can remove it? Most probably we can, due to the relative querystring-search improvements, but we need to be sure.

sneridagh:
@pnicolli do you have time to take a look?

jackahl:
@sneridagh at least for the tests I did it worked. In one of our projects, we even had the issue of the current code not returning the correct items, but my changes actually doing so (admittedly I could not reproduce the failure in lain volto). Do you know who touched that part of the API endpoint last and could give some insight?

sneridagh:
@jackahl Piero did it, I asked him for a review. 

Anyway, it seems that was a temporary fix, waiting for the good solution (querystring context-aware) given this comment: https://github.com/plone/volto/pull/747#issuecomment-570940292

In case that Piero is busy, just go ahead.

tisto:
@iFlameing could you write a Cypress test for this please?

```
Given a page that contains two documents "news item one" and "news item two"
   and two documents "about us" and "contact" on the portal root
When I add a listing block with a relative path query criteria with the value "."
 Then the listing block shows two items
   and "news item one" shows up in the listing block
   and "news item two" shows up in the listing block
   and "about us" does not show up in the listing block
   and "contact" does not show up in the listing block
```

iFlameing:
@tisto You can see the cypress test in #1870

PR #1864, opened 2020-09-30
Storybook 6

sneridagh:


sneridagh:
@tiberiuichim I made it work on 6, the other storybook PR can be closed. Please take a look! As said, I'm not biassed towards any solution here. Hope you like it!

The idea would be write the .mdx story for each component (we can do it bit by bit). Use the unexported components for them, so we don't have to fight with mocked Redux stores, and we can setup the Storybook controls on the desired internal props for playground.

So we will have this features:

-     mdx support
-     auto playground setup
-     props autodoc
-     viewports playground

We can setup the other plugins, the a11y one is specially interesting. We can explore other addons too.

tiberiuichim:
@sneridagh Nice to see this thing rolling! As I have said, I'd rather have Storybook then styleguidist.

Will you continue on this PR? I'm interested to see how you'll load the Volto webpack config, that's what gave me problems. Basically [this](https://github.com/plone/volto/blob/8e5663f9d6610e56cc7cad43241f3a1be66074c1/styleguide.config.js#L80), but for Storybook

sneridagh:
@tiberiuichim yeah, I will try to add the remainings. I've already added the SVG icon support, we don't need to add every Webpack config there, but the ones that we require to render the components well. I will try to add the CSS too. Where did you get stuck?

tiberiuichim:
@sneridagh I was able to make it work for standalone components, but as soon as I've had components coming from `@plone/volto/components`, my mdx files just disappeared from Storybook's point of view, with no debugging message whatsoever.

sneridagh:
@tiberiuichim doh, it seems I've been overconfident on this one... I think that Storybook provokes a circular dependency when using `@plone/volto` paths... I've been trying for a while, and although I made it to compile it with no error, then I always get "undefined" in the import...

I might give another try but not soon I guess... that reminded me of another problem I have to discuss with you that is long due as well, and not easy to solve.

tiberiuichim:
@sneridagh Maybe you can push the state of your config now? At least as a start point for whatever the future holds.

sneridagh:
@tiberiuichim done!!

sneridagh:
Closing for now, to not add confusion and have it dangling around.

sneridagh:
closed in favor of: #2065

PR #1869, opened 2020-10-04
Fix broken current folder by default in content browser for image lin…

sneridagh:
…ks, solves #1860

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Bugfix
 
 - Fixing ObjectBrowser search input reload @iFlameing
+- Fix broken current folder by default in content browser for image links, solves #1860 @sneridagh
 
 ### Internal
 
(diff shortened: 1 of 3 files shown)


sneridagh:
:tada: First PR in four months 🎉

tisto:
w00t!

tiberiuichim:
I think the main issue at hand is that ImageSidebar should be using ObjectBrowserWidget. ObjectBrowser should be hidden away behind registered widgets. Which reminds me, we have this [doc](https://docs.voltocms.com/blocks/editcomponent/#objectbrowserwidget) which explains how one could register ObjectBrowserWidget for links and images. Why don't we do that directly? Why not just have them available by default and use them in Volto's codebase?

tiberiuichim:
> 
> 
> 🎉 First PR in four months 🎉

Hoping for many many more!

tiberiuichim (approved):

PR #1880, opened 2020-10-08
Fixing bug for editing the table block

### LINKED ISSUE #1879: Table block is not editable.
Once you created a table block on a page and save it. And once you visited the page for editing the table those changes will be not shown.

Step to reproduce:
1. Create a page and add a table block
2. save the page
3. Revisit the page
4.and try to edit the table block and save it.
5. You will find that those changes are not applied.

iFlameing:
Fixes #1879

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Bugfix
 
+- Fixing table block edit @iFlameing
+
 ### Internal
 
 ## 8.2.3 (2020-10-07)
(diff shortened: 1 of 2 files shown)


tisto (approved):
Works like a charm! A second review would be nice @tiberiuichim @sneridagh @plone/volto-team.

tiberiuichim:
Indeed, I've hit this problem as well, though I've consider it then a problem in my code. It happens because of code in getOnlyFormModifiedValues. It is a valid fix.

tiberiuichim (approved):

PR #1885, opened 2020-10-11
docs: upgrade guide to latest react-dropzone

nileshgulia1:


docs/source/upgrade-guide/index.md
```diff
@@ -1,17 +1,59 @@
 # Upgrade Guide
 
 This upgrade guide lists all breaking changes in Volto and explains the
- steps that are necessary to upgrade to the lastest version.
+steps that are necessary to upgrade to the lastest version.
 
 !!! note
-    There are times that updating the Volto boilerplate (the one generated by
```
tiberiuichim, line 7: Don't strip that leading whitespace, it's important. See https://squidfunk.github.io/mkdocs-material/reference/admonitions/#usage

tiberiuichim (changes requested):

docs/source/upgrade-guide/index.md
nileshgulia1: ah, my eslint does that. I'm seeing it in rest of my code. I have to disable eslint for this file.

nileshgulia1: @tiberiuichim Done!

tiberiuichim:
There's some white lies here. The idea is that no good Volto addon or project would break if it properly declared its dependencies. node_modules and the yarn workspaces system that we use ensures that multiple versions of the same library can coexist in a bundle. The problem is with addons or projects that use internally the react-dropzone and didn't declare it as a dependency. I don't know if we should treat every major Volto dependency as a breaking change because there's some project or addon "out there" that didn't properly declare its dependencies. And I say this as the author of such a mistake: volto-slate used react-dropzone but I didn't declare it as a dependency, instead relying on Volto to provide it. At least it should be specified as a peerDependency.

nileshgulia1:
IMO specifying `react-dropzone` as peerDependency mean  If this package doesn’t already exist in the `node_modules` directory or it is the wrong version, don’t add it. But, show a warning to the user that it wasn’t found. So we don’t want our library adding another version of a package to node-modules when that package could conflict with an existing version and cause problems.
We can Favor using Peer Dependencies when one of the following is true:
-  Having multiple copies of a package would cause conflicts.
- We want the developer to decide which version to install.
So a user can always decide if this particular package is needed won't cause any breaking changes. if it does, then can make code changes and migrate to newer API.

-cc @sneridagh @tiberiuichim

tiberiuichim:
@nileshgulia1 I'm not sure I understand what you mean. What are your proposals? I don't believe the problem in discussion, the react-dropzone upgrade, is worth discussing (except maybe as a blog article or whatever), but Volto was not at fault: there was never a breaking contract from Volto's side. Alin flagged the problem as breaking, but the deeper problem was with faulty dependency management of volto-slate.

nileshgulia1:
@tiberiuichim My main confusion was in general, Is it a good idea to include any package which has a breaking change on any of the transitive dependencies in Volto onto our project/addons dependency or peerDependency? Because I too faced this error in addons 'm writing probably similar to what you had faced in volto-slate.

tiberiuichim:
Addon packages could use peerDependency, with the understanding that that can break, then switch to real hard dependency when Volto upgrades to an incompatible version, though I imagine that the addon would immediately be upgraded to the version in Volto.

nileshgulia1:
This is obsolete.

PR #1886, opened 2020-10-11
Multi-block copy/paste

tiberiuichim:
Initial implementation, but works fine. 

Some things to discuss:

- I've created a new action/reducer pair called "blocksClipboard".
- I've introduced a new concept, the `settings.persistentReducers`. The information for these reducers is synced to localStorage, for which I've added a new dependency, "redux-localstorage-simple"
- which means that you can copy/paste blocks between two open windows without reload. Volto is a SPA, we shouldn't need to reload if an editor has two windows side by side.
- I've introduced a new type of interaction in the form. By holding "shift" key, you can click-select ranges of blocks. You can do the same with up/down arrow keys, but draft text blocks don't play nice. We need to special-case those.
- I've introduced (via portals) several buttons in the left toolbar: cut/copy blocks and paste blocks.
- I've added a new components, `manage/Form/BlocksClipboard`. Is it the right place for it?
- I've introduced a new concept, `block.blocksConfig.<blockName>.cloneData`. It should return null (in case the block doesn't want to be copied, like a Title block) or `[blockId, blockdata]`. This needs to be implemented for draftjs text blocks, because we don't want to end up with duplicate paragraph ids, as they become DOM element ids when rendered. Or is this done in redraft? I haven't looked at these in a while.
- We probably need some UX/css for this.

I'd really like to see some sort of clipboard preview, we can do it with css scaling, but I don't know how to trigger it... maybe as a "Clipboard" tab in the Block Chooser? Also, I don't know what to do about the @iFlameing's work. Copy/paste buttons in block sidebar.

This thing needs tests and documentation. Help on these would be highly appreciated!

And here's a quick demo video: https://www.youtube.com/watch?v=_VrR4GWlfd0


tiberiuichim:
Hmm... I've based my PR on @iFlameing's "copy" PR, there's a few changes that look a bit out of place.

tiberiuichim:
Maybe show the clipboard blocks preview as a tooltip for the paste button?

tiberiuichim:
Some issues with the current implementation:

- for some reason opening the Add form clears the clipboard
- maybe others related to sync of clipboard across tabs

Issue #1894, opened 2020-10-13
Adding an Inline Link always causes a line break 


steffenri:
steps to reproduce:
o) go to your local volto build
1) login
2) create a page
3) Create a text block with text e.g. "This is a {link}"

-> There's a line break in front of the link. This should not be the case as it is an inline link.

This issue may be caused by this PR: https://github.com/plone/volto/pull/1820/

I will open a PR to revert this change.

![Screenshot_2020-10-13 Edit Page](https://user-images.githubusercontent.com/54619639/95862808-b5ec0780-0d63-11eb-9113-dabf1b17fd3b.png)

![Screenshot_2020-10-13 Link Problem](https://user-images.githubusercontent.com/54619639/95862818-bab0bb80-0d63-11eb-9a1f-c482817812bc.png)

steffenri:
PR: https://github.com/plone/volto/pull/1895

PR #1895, opened 2020-10-13
revert PR number 1820 to fix linebreaks on inline links

steffenri:


steffenri:
Issue: https://github.com/plone/volto/issues/1894

PR #1899, opened 2020-10-13
added objectBrowser to UrlWidget, and attached UrlWidget to remoteU…

giuliaghisini:
…rl field of ContentType Link


tiberiuichim:
If this is what I think it is, then it's perfect! It's like the missing link. @sneridagh  

Though I wonder about the newly introduced dependencies. I don't see those any object proxy used in your PR, what are the changes to package.json and lock about?

giuliaghisini:
> If this is what I think it is, then it's perfect! It's like the missing link. @sneridagh
> 
> Though I wonder about the newly introduced dependencies. I don't see those any object proxy used in your PR, what are the changes to package.json and lock about?

i had to add identity-obj-proxy because when trying to re-create snapshot with jest it failed, because somewhere there's an import of a .css in a jsx in a file regarding Gallery.

avoinea (approved):

sneridagh (approved):
LGTM!

cekk:
For the backend part: https://github.com/plone/plone.restapi/pull/1005

I'm assuming that Volto send to the backend an "api url", so i can convert it into an internal reference like standard Plone Links.

IMHO the frontend user shouldn't see the api url when select an item (also in text editor), but the "frontend" one, or just only the relative path. It's ok to send the complete api url to the backend.

giuliaghisini:
> For the backend part: [plone/plone.restapi#1005](https://github.com/plone/plone.restapi/pull/1005)
> 
> I'm assuming that Volto send to the backend an "api url", so i can convert it into an internal reference like standard Plone Links.
> 
> IMHO the frontend user shouldn't see the api url when select an item (also in text editor), but the "frontend" one, or just only the relative path. It's ok to send the complete api url to the backend.

I've do adjustments to the UrlWidget component and to AddLinkForm to make sure that the url containing the API path is always sent to the backend, but the user sees only the part of the url after the api path

tiberiuichim (approved):

sneridagh:
@giuliaghisini please, could you take a look at the merging conflicts? Let's merge this one too

giuliaghisini:
> @giuliaghisini please, could you take a look at the merging conflicts? Let's merge this one too

ready!

sneridagh:
@giuliaghisini I am puzzled, but the PR is now empty! It only has the changelog. Could be that since I merged #1946 it was based on this one, then the changes are already in master?

giuliaghisini:
> @giuliaghisini I am puzzled, but the PR is now empty! It only has the changelog. Could be that since I merged #1946 it was based on this one, then the changes are already in master?

as i can see in commits log i think your'right.

PR #1902, opened 2020-10-14
Handeling null value in querystringwidget

iFlameing:


src/components/manage/Blocks/Listing/QuerystringWidget.jsx
```diff
@@ -307,7 +307,7 @@ class QuerystringWidget extends Component {
                           components={{ DropdownIndicator, Option }}
                           value={{
                             value: row.i,
-                            label: indexes[row.i].title,
+                            label: indexes[row.i]?.title,
```
tisto, line 310: @iFlameing did you try what happens when value label has when indexes[row.i] is Null (or something else)?

tisto (approved):

sneridagh:
@tisto @iFlameing Which is the use case? I mean, for all indexes, it should be a matching row.i, right? if not, that means that the value that comes from the widget is not valid/registered in Plone.

PR #1910, opened 2020-10-17
translation german

ksuess:
Inhaltsverzeichnis


locales/de/LC_MESSAGES/volto.po
```diff
@@ -1062,11 +1062,11 @@ msgstr "Mehr"
 
 #: components/manage/Contents/ContentsItem
 msgid "Move to bottom of folder"
-msgstr "Zum Anfang des Ordners wechseln"
+msgstr "Ans Ende des Ordners verschieben"
```
tisto, line 1065: @ksuess the container could be a page. Shall we just drop the "Ordner"? -> "Ans Ende verschieben"

tisto:
@ksuess Volto master is currently in alpha phase. Therefore I try to avoid merges. Did you merge this PR by accident? The changelog "Translation" is also something we should improve. :)

PR #1917, opened 2020-10-19
added UniversalLink component to handle internal/external/download links

nzambello:
As ConditionalLink wraps links under conditions, this helps with URL handling for external, internal and Plone download links.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 ### Internal
 
 - Fix sentry docs markdown format @avoinea
+- Add `UniversalLink` to handle internal/external/download links @nzambello
 
 ## 8.5.0-alpha.0 (2020-10-14)
 
--- a/src/components/index.js
+++ b/src/components/index.js
@@ -29,6 +29,7 @@ export Unauthorized from '@plone/volto/components/theme/Unauthorized/Unauthorize
 
 export Icon from '@plone/volto/components/theme/Icon/Icon';
 export ConditionalLink from '@plone/volto/components/manage/ConditionalLink/ConditionalLink';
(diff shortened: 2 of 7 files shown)


tisto:
@nzambello I guess you called it UniversalLink to make it distinctive from "Link". Though, maybe just "Link" might be a better name. Just asking here. I am undecided and naming things sucks.

nzambello:
I know 😂 
I thought that we could have:

- Link (that's a link, nothing more, unmanaged, the one from react-router)
- ConditionalLink (specific use-case wrapping a link)
- UniversalLink (because it manages its rendering basing it on the URL to adapt to any scenario)

But that's an idea.
I jotted here my feelings about that naming, but we can change it.

tiberiuichim:
That's a big commitment, a universal link :) 

There's missing support for email links and probably internal links. It would be cool if we could combine it with our volto-slate link representation, here: https://github.com/eea/volto-slate/blob/d9dd5e99460c8121d296ed8aa3b4f8c22989f513/src/editor/plugins/Link/render.jsx#L3

avoinea (approved):
I was about to comment that `<h1>` which is a block element is not allowed within an `<a>` inline element, but this seems to be allowed with HTML5. Thus:

LGTM.

sneridagh:
As p.a.m. co-author, I thought at the first glance it was about multilingual stuff (we named the "permalink" to a translation universal link).

BTW naming sucks and I'm fine with the universal naming. LGTM.

pnicolli (approved):
LGTM

tisto:
@sneridagh FYI: CI is red since this merge.

PR #1922, opened 2020-10-20
Fix runtimeConfig relative vs absolute import

avoinea:
This will allow runtime_config customization via volto-project / addons

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 ### Internal
 
 - Translation @ksuess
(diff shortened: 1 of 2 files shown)


tisto (approved):

tisto:
@avoinea released: https://www.npmjs.com/package/@plone/volto/v/8.5.0-alpha.2 let me know when we can do a final release. :)

avoinea:
@tisto Thank you! I think we can do the final release. Any other issue, if any, can be handled within the next minor releases.

PR #1926, opened 2020-10-21
Render nativeLang as language selector

mikejmets:
Show the nativeLang text instead of the 2 character language code


tisto:
@mikejmets thanks for your contribution!

Classic Plone indeed shows the full nativeLang string...I am undecided about which version I prefer. Opinions?

sneridagh (approved):
I would default to the long version, most probably the LanguageSelector would be later customized depending on the project look and feel anyways, so the developer have the chance to use the short one easily.

LGTM

sneridagh:
@tiberiuichim @pnicolli @tisto this might be one of those potential "UI breaking" changes... did we decide how to handle them? I mean, it's not a breaking change, but someone who relies on the initial implementation (visually) might find on update that it changed...

tiberiuichim:
@sneridagh We don't have (yet) multilingual sites, but that's a good question.

I'd still call this change a Patch rather then a major breaking change.

tisto:
@tiberiuichim @sneridagh since this is a UI facing issue I would not just release this via a patch release. When you do a minor upgrade of software you do not expect any breaking change or changes in the UI.

I would be pissed as a Volto integrator when I do a minor upgrade and this would change in my site without notice.

PR #1931, opened 2020-10-23
Return 404 on SSR, when appropriate

tiberiuichim:
Caveat: I haven't changed the 404 component and I've tried to reuse the DefaultView wrapper around it, to show the "site", but that does not work well because there's too much stuff missing. I think we should change the 404 component to include a link to the website root.

Screenshot 2020-10-23 060437

I think we might want to handle the other error pages like this, 404 is priority right now for me.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Bugfix
 
+- Properly return 404 in server response, when appropriate @tiberiuichim
+
 ### Internal
 
 ## 8.5.3 (2020-10-22)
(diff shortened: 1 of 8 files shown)


tiberiuichim:
Ugh, just to explain how this works, because it's a bit convoluted. I've tried initially to plug into superagent/api middleware, but I couldn't make it work. Then I was looking in the server.jsx code and I've noticed this line: https://github.com/plone/volto/blob/2a95dff232ed43d6c2a4af92b85ff988b7590112/src/server.jsx#L175

So, what we're doing is: in the NotFound component, we're writing in the "staticContext" prop (perhaps we should introduce that static context anyway, not just for the server?) that's passed down from the router. Then this is detected at the final top render step.

sneridagh:
LGTM for now, we can discuss the caveats and the points of improvement.

tiberiuichim:
Continued in #1948

PR #1933, opened 2020-10-23
fixed bug in render lists in draftjs, when there's some styles inside…

giuliaghisini:
… list-item (links or bold text ecc..)

Before this fix, if you create a text-block, create list inside text block, add some styles or link to an element, a error was thrown.
For example:

- [Link1](http://www.google.it)
- [Link2](http://www.google.it)
- Text **bold**

Fixed this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## 8.5.4 (unreleased)
 
+- fixed bug in render lists in draftjs, when there's some styles inside list-item (links or bold text ecc..) @giuliaghisini
+
 ### Breaking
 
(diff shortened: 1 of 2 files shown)


tisto:
@giuliaghisini can you share the actual error that is thrown?

iFlameing:
@giuliaghisini @tisto This is the same problem which I fixed in this pr.
https://github.com/plone/volto/pull/1929

giuliaghisini:
@tisto @iFlameing yes, sorry i didn't search for other pr. The merged pr from @iFlameing solves that problem. I close this pr

Issue #1936, opened 2020-10-23
A pathname like /policy/edit does not show the Unauthorized or Forbidden component when not logged in


silviubogan:
It instead shows an almost empty page. Any pathname that is in the form `/**/edit` and maybe others too are affected by this issue.

This JSX expression should be moved to a lower indentation level, next to what is currently its parent JSX expression.

https://github.com/plone/volto/blob/390fa3d1c1c079b692f2033ddbc2dd02c658560e/src/components/manage/Edit/Edit.jsx#L249-L257

The lines above should be moved outside of this conditional rendering:

https://github.com/plone/volto/blob/390fa3d1c1c079b692f2033ddbc2dd02c658560e/src/components/manage/Edit/Edit.jsx#L215

I tested it and it works.

I will make a PR too.

tiberiuichim:
@silviubogan Thanks!

PR #1937, opened 2020-10-23
A pathname like /policy/edit does not show the Unauthorized component when not logged in

silviubogan:
A pathname like /policy/edit does not show the Unauthorized or Forbidden component when not logged in, ref #1936.


sneridagh (approved):
LGTM

PR #1941, opened 2020-10-25
Update to use Node 12 or 14

sneridagh:
On Oct.27th, Node 14 is LTS, prepare accordingly.


tiberiuichim (approved):

package.json
```diff
@@ -210,7 +210,7 @@
     ]
   },
   "engines": {
-    "node": "^10 || ^12"
+    "node": "^12 || ^14"
```
tisto, line 213: @sneridagh @tiberiuichim this is a breaking change IMHO. So we either have to wait for the next major release or do a major release right away.

sneridagh: Oddly enough, 10, 12 and 14 will be "valid" LTS until 2021-04-30 when they will drop 10 as LTS.
https://nodejs.org/en/about/releases/

We could maintain the node restriction, but remove it from the builds tests. Then we minimize the impact, and no need for a major... Doing a major is ok for me too.

tisto: We have to come up with an official release policy here. I'd say we always support all LTS versions that are currently supported and drop them with a major release on the day they run out of support. We have to write this down somewhere in the docs. If we support a version we should test it.

It might also help to add docs how to override the pinned version in custom project. We had this problem in a project but I don't fully recall how I fixed it back then. :(

pnicolli: Changing this should definitely be a breaking change. Always supporting all LTS versions is a safe policy, it requires a second PR in a few months though.
If we want to save that time, we could go for the "add this PR to the next major version", whenever that happens (I assume it will happen before April 30th). It's not that much time to be saved, actually, so I'm fine either way.

sneridagh:
+1 to all comments, I updated the PR accordingly. 

We support all active LTS, added it to the docs and README.

package.json
tisto: We can always add a new LTS version. Which is something we should do as soon as it is out. With removing we should maybe wait for the next major version we do anyways.

sneridagh:
Meh, Netlify is already using 14 for the builds and the docs build fails.

Since in the form that it is now is not harmful in any case, are you all ok?

/cc @tisto @pnicolli @tiberiuichim

pnicolli (approved):

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@
 
 ### Internal
 
+- Enabling support for the new active LTS NodeJS version 14. NodeJS 10 eol will happen on 2021-04-30 and Volto will update accordingly. More information on https://nodejs.org/en/about/releases @sneridagh
```
tisto, line 13: @sneridagh I think I'd make this a feature release. We are actually adding a new feature which is that we commit ourselves to support a new node version (and test it properly). Technically we are not doing anything here. Though, it is important for outside communication IMHO and should not go into something subtle as "internal".

tisto (approved):
Apart from the comment regarding a feature release, I am good with this PR.

PR #1948, opened 2020-10-26
SSR errors: 401, 403, 404

tiberiuichim:
Screenshot_20201026_213952

See #1931 for some details on the initial principle of how it works. I'll write a bit of documentation later.

I think we should write a new chapter in the Volto docs, "Internal documentation"


tiberiuichim:
@sneridagh I think the non-hydrated version should work just as well and with a lot less hassle, I will implement it.

tiberiuichim:
@sneridagh Should be ready now. For some reason this configuration just works, no hydration setup needed.

tiberiuichim:
Hmm... now it hydrates. I'll change the Html component to eliminate scripts.

avoinea (approved):

src/server.jsx
```diff
@@ -174,6 +174,23 @@ server
...
+            });
+
+            res.status(context.error_code).send(
+              `<!doctype html>
+                ${renderToString(
+                  <Html
+                    extractor={extractor}
+                    markup={markup}
+                    store={store}
+                    extractScripts={false}
```
silviubogan, line 189: @tiberiuichim In development environment, the error pages, with this PR, do not load the styles that they should have. This change corrects this (I am not sure if this has or doesn't have other negative implications):

```suggestion
                    extractScripts={process.env.NODE_ENV !== 'production'}
```

HTH

sneridagh:
@tiberiuichim could be that now the Guillotina test fails because it expects for a 200 and now it returns a 401?

tiberiuichim:
@sneridagh I don't know... I guess somebody has to try it with guillotina. The bad side, though, is that the test runner doesn't provide a lot of information about the cause of failure.

sneridagh (approved):

tisto:
@plone/volto-team this is a bugfix that we will release with 9.x.x.

PR #1951, opened 2020-10-27
Multi-selected copy/paste blocks

tiberiuichim:
Continuation of #1886, everything written there still applies. Copied here:

Initial implementation, but works fine. 

Some things to discuss:

- I've created a new action/reducer pair called "blocksClipboard".
- I've introduced a new concept, the `settings.persistentReducers`. The information for these reducers is synced to localStorage, for which I've added a new dependency, "redux-localstorage-simple"
- which means that you can copy/paste blocks between two open windows without reload. Volto is a SPA, we shouldn't need to reload if an editor has two windows side by side.
- I've introduced a new type of interaction in the form. By holding "shift" key, you can click-select ranges of blocks. You can do the same with up/down arrow keys, but draft text blocks don't play nice. We need to special-case those.
- I've introduced (via portals) several buttons in the left toolbar: cut/copy blocks and paste blocks.
- I've added a new components, `manage/Form/BlocksClipboard`. Is it the right place for it?
- I've introduced a new concept, `block.blocksConfig.<blockName>.cloneData`. It should return null (in case the block doesn't want to be copied, like a Title block) or `[blockId, blockdata]`. This needs to be implemented for draftjs text blocks, because we don't want to end up with duplicate paragraph ids, as they become DOM element ids when rendered. Or is this done in redraft? I haven't looked at these in a while.
- We probably need some UX/css for this.

I'd really like to see some sort of clipboard preview, we can do it with css scaling, but I don't know how to trigger it... maybe as a "Clipboard" tab in the Block Chooser? Also, I don't know what to do about the @iFlameing's work. Copy/paste buttons in block sidebar.

This thing needs tests and documentation. Help on these would be highly appreciated!

And here's a quick demo video: https://www.youtube.com/watch?v=_VrR4GWlfd0


sneridagh (approved):
LGTM, I would add a couple of simple tests here and there.

BTW, question: Why the messages in that helper module? (Also why there are already messages there too)

sneridagh:
@tiberiuichim we didn't talked about this! What's the status?

tiberiuichim:
@sneridagh nothing changed yet, I haven't had time for it. I'll try to get back to it this week, maybe Thursday. 

So, if I remember correctly, the decision was to leave the current implementation as it is, the only wish was to be able to deselect specific blocks.
My wishes are that the clipboard doesn't clear on paste (or at least there's a way to paste but not clear the clipboard) and that there's also a delete button, in case I don't want to copy the blocks, but I want them removed.

tiberiuichim:
Screenshot 2020-11-29 082501

I've moved the buttons to the bottom. I've added the delete blocks button.

tiberiuichim:
Screenshot 2020-11-29 083039

And on mobile screen.

sneridagh:
@tiberiuichim Nice! I like it more than before. Just a reminder! the circled buttons in Pastanaga used to mean "primary action pointer" for the user, so the tools buttons shouldn't be circled.

tiberiuichim:
@sneridagh right, I'll remove the circled class. Hopefully tomorrow I have time and space to work on the cypress tests.

locales/de/LC_MESSAGES/volto.po
```diff
@@ -310,7 +310,7 @@ msgstr "Ziel auswählen"
 
 #: components/manage/Widgets/FileWidget
 msgid "Choose a file"
-msgstr "Datei wählen"
+msgstr "||||||| a5f08f67a"
```
sneridagh, line 313: ???

sneridagh (changes requested):

locales/de/LC_MESSAGES/volto.po
sneridagh: There are several of them

sneridagh (approved):
LGTM!!

PR #1953, opened 2020-10-29
Fix reducer errors

tiberiuichim:
I took the opportunity to review the code and clean up a bit razzle.config.js


zotya:
@tiberiuichim:
1. I've added back the SentryCliPlugin as it is needed for creating releases in sentry when we want to create the release on build time.
2. added an extra check in crashReporter, so we will send messages to sentry only if SENTRY_DSN, or RAZZLE_SENTRY_DSN is configured

zotya (approved):

sneridagh (approved):
LGTM

PR #1958, opened 2020-10-30
Allow customization of piped resources from the backend

tiberiuichim:
Introduce `settings.backendResourceMatch`. This enables additional "backend pipes" that are treated similar to ``@@images`` and ``@@downloads``


tiberiuichim:
Based on discussions with Victor, we should enable an optional virtual hosting path change (just like the dev proxy middleware) , to be integrated in getAPIResourceWithAuth.

tiberiuichim:
Somehow superseeded by eeacms/volto-vh-middleware.

PR #1961, opened 2020-11-02
Fixing overlap of labels with each other in select Widget

### LINKED ISSUE #1960: Select Widget labels overlap if they are bigger in length.
**Describe the bug**
The select widget label overlap if you have long text.

**Expected behavior**
The labels don't overlap with each other.
**Screenshots**
Textoverelapping

iFlameing:
Fixes #1960


tisto:
@sneridagh can we get this into the next release?

sneridagh (changes requested):
@tisto problem is that height is required, that will break other things. I'm pretty sure that it was important. Can't remember now where (I think it's because of the height in the forms). I need to look in the line history.

sneridagh:
@iFlameing Having a picture of what was the problem in the first place would be great to have to know what are we fixing, and the outcome of the fix as well.

tisto:
@sneridagh here you go:

Screenshot 2020-11-03 at 22 57 14

I am fine with postponing this to figure this out properly.

sneridagh:
@tisto let me take a look tomorrow. So the problem is that if the option is too long, then it jumps, right? We can do the release tomorrow morning.

sneridagh (approved):

PR #1967, opened 2020-11-03
Razzle3.3.7

sneridagh:
We need to make sure that the plugins use the new schema.

This is important alone because it prepares Volto for the upcoming Razzle 4 which will pack a lot of updated features.

But the most important thing is that it enables React "Fast Refresh": the new way of doing refresh while developing that will replace HMR, which in Volto has a glitch when the code reloads, the router got broken.


sneridagh:
Good thing is that it uses @babel/core@7.11.1 which is already the version we are using now, so we might avoid the hoisting problems.

sneridagh:
Also would be great to sneak in this upgrade the ability to load different jest.config.js files and Volto (using Razzle) loads them.

mtoepfl:
> But the most important thing is that it enables React "Fast Refresh": the new way of doing refresh while developing that will replace HMR, which in Volto has a glitch when the code reloads, the router got broken.

I'm not sure if we will benefit so much, because Volto makes heavy use of class components.

> Local state is not preserved for class components (only function components and Hooks preserve state).

I'm not sure if it's true for latest version of Fast Refresh.

tiberiuichim (approved):

tiberiuichim (approved):
Razzle deprecates modify(), so all projects should migrate their razzle.config.js, right?

sneridagh:
> Razzle deprecates modify(), so all projects should migrate their razzle.config.js, right?

@tiberiuichim if they fiddle with the razzle config object I guess so, it should comply with the new pattern:
https://razzlejs.org/docs/customization

If you haven't fiddled with it, then it works right away.

I guess now it's a bit more ordered, with the new pattern, and you have more flexibility to modify specific parts from the Razzle config more accurately.

sneridagh:
@tiberiuichim I do not hava any site at hand with addons that modify razzle config. Can you point me to one?

sneridagh:
@tiberiuichim btw I will need you for finding a sane way to make Razzle happy with the plugins.

tiberiuichim:
Here's how an addon can fiddle with razzle.config setup: https://github.com/eea/volto-bise/blob/master/razzle.extend.js

sneridagh:
It would be great to have it ready for Volto 9, however it's not a must.

sneridagh:
@tiberiuichim You'll love this one ^^^ :) turned out to be fairly easy.

I also made a PR here: https://github.com/jaredpalmer/razzle/pull/1464

I had to patch Razzle for make not released plugins (in the new object shaped plugins) to work.

razzle.config.js
```diff
@@ -186,12 +191,23 @@ const plugins = addonExtenders.reduce(
 
 module.exports = {
   plugins,
-  modify: (config, { target, dev }, webpack) => {
-    const defaultConfig = defaultModify(config, { target, dev }, webpack);
+  modifyWebpackConfig: ({
```
tiberiuichim, line 194: Maybe at some point, later, we could also support the additional plugin entry points for Razzle, coming from razzle.extend.js files. My proposal would be: keep the `modify()` as modifyWebpackConfig, but check if the extend file implements any of the other methods.

tiberiuichim (approved):

razzle.config.js
sneridagh: Unfortunately, if we do that, Razzle complains and throw a warning, and I guess then we won't be Razzle 4 compat. We can still support backwards compat by looking into the modify property, then treat it as the modifyWebpackConfig (same as Razzle 3 is doing now).

tisto:
@plone/volto-team decision: @sneridagh will add a few lines to the upgrade-guide.

PR #1970, opened 2020-11-04
Making inline Link toolbar, location aware in content browser & typing in the link widget detects if internal

iFlameing:


src/components/manage/Sidebar/ObjectBrowser.jsx
```diff
@@ -74,9 +74,11 @@ const withObjectBrowser = (WrappedComponent) =>
     closeObjectBrowser = () => this.setState({ isObjectBrowserOpen: false });
 
     render() {
-      let contextURL = this.props.pathname;
-      if (this.props.pathname?.endsWith('edit')) {
-        contextURL = getParentURL(this.props.pathname);
+      let contextURL = this.props.pathname
```
iFlameing, line 77: I think that I should remove `this.props.pathname` since we already extracting the contexturl from location. what you all think??

sneridagh: Could you be more specific? I don't see why we need to change this.

sneridagh:
@iFlameing I would do it differently, withRouter is no longer necessary, since you always can do a useLocation().

tiberiuichim (commented):
```
if (contextURL?.endsWith('edit')) {
```
There's getBaseURL in volto helpers. Would it be possible to use that instead? Not sure if it fits.

sneridagh:
@plone/volto-team we need that you weigh in with this one.
By adding the open the browser in the current context we introduced two behaviors:
- On creation (add): it opens in the parent (since the object is not created yet)
- On edit: users see inside the object and is rather common that they see an empty list is kind of confusing

Proposal is to open always on parent, what do you think? We could parameterize it though...

sneridagh:
> ```
> if (contextURL?.endsWith('edit')) {
> ```
> 
> There's getBaseURL in volto helpers. Would it be possible to use that instead? Not sure if it fits.

We need to detect if we are on edit view here.

tiberiuichim:
@sneridagh On edit, go to parent only if the current item doesn't have children.

As a quick fix is fine, but if I were a regular editing, doing the kind of tasks I've had to do recently, I'd get upset. IMHO it would be better if it opened in the "last picked item's location", once used in a page. Basically, if you're editing a page and maybe you store your images in a specific location, then you want to open there "always", so after you've inserted the first image, for the next, you don't want to navigate again to the same folder.

tiberiuichim (approved):

tisto:
@tiberiuichim let's merge this one as a quick fix and then discuss how we can improve things.

PR #1974, opened 2020-11-06
Use `flattenToAppURL` instead of `addAppURL` in link widget

sneridagh:
@giuliaghisini There was any reason for use `addAppURL` instead?


sneridagh:
I guess the rationale here is that we want always to save "flat" links into the field... or else then the resolveuids won't work?

/cc @cekk @tisto

PR #1975, opened 2020-11-06
Refactoring FileWidget

iFlameing:
@sneridagh I found out that in react-dropzone if we don't provide onChange to input it will use onDrop handler for request. So, I remove the onChange handler for the input. I also found that we are using semantic input which might don't work for some case. And also react-dropzone uses browser Input tag. I have just deleted the unnecessary code.

### DIFF
--- a/src/components/manage/Widgets/FileWidget.jsx
+++ b/src/components/manage/Widgets/FileWidget.jsx
@@ -5,7 +5,7 @@
 
 import React from 'react';
 import PropTypes from 'prop-types';
-import { Input, Button, Image, Dimmer } from 'semantic-ui-react';
+import { Button, Image, Dimmer } from 'semantic-ui-react';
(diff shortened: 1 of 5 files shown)


src/components/manage/Widgets/FileWidget.jsx
```diff
@@ -106,7 +105,11 @@ const FileWidget = (props) => {
                 className="image-preview"
                 id={`field-${id}-image`}
                 size="small"
-                src={value?.download ? flattenToAppURL(value.download) : null}
+                src={
```
iFlameing, line 108: I have added params so that browser doesn't cache the image. It is very misleading when you keep changing the image for the right fit and the browser doesn't update the preview image.

sneridagh:
image

@iFlameing I improve it a bit :)

sneridagh (approved):

PR #1977, opened 2020-11-08
Normalize relative imports for addons with webpack plugin

tiberiuichim:
This plugin normalizes imports in addons. An import such as:

```
import Progress from './Progress';
```

is "normalized" like:

```
import Progress from '@plone/datatable-tutorial/CellRenderer/Progress'
```

This means that I now can customize the `Progress` component without having to use package-level imports inside my addon (by creating a `src/customizations/@plone/datatable-tutorial/CellRenderer/Progress.jsx` file).

This also reduces the need for conventions and ceremonies in addons and ensures consistency in experience.

I need to do tests, I think it might be possible to test this with the stuff in `__tests__`. But it's tedious work :)

@sneridagh: Right now it filters all resolve requests by "is it an addon?". Do we want this enabled also for Volto?


tiberiuichim:
I've watched the compilation times, they don't seem to grow. If anything, they seem to shrink, but this needs more tests.

Without the new resolver plugin: Done in 59.84s

With the resolver plugin: Done in 46.17s.

sneridagh:
@tiberiuichim of course we want it in core Volto! 

However I'm wondering how this plays with the index files, we do require still to set the full path there I guess. I'm not a fan of the index files anyways, because of the circular deps errors and other loading issues, but truth is that they are handy

tiberiuichim:
@sneridagh It works fine with index files. I've actually only tested it with an index file.

(I think) this is what happens: in the index the import is seen with the absolute path, so it's not a problem when it is further exported, because it will be a reference (however many levels of index/exports) to the "final" import which will be properly aliased and shadowed.

tiberiuichim:
@sneridagh for my test I've shadowed this component: https://github.com/tiberiuichim/datatable-tutorial/blob/main/src/CellRenderer/Progress.jsx which is referenced here: https://github.com/tiberiuichim/datatable-tutorial/blob/9a9f510408c3020c35a1e6e1f391cc5298d47671/src/index.js#L38 this the import from here: https://github.com/tiberiuichim/datatable-tutorial/blob/main/src/CellRenderer/index.js

sneridagh (approved):
@tiberiuichim shall we add it for Volto 9 release?

tiberiuichim:
@sneridagh Yeah, otherwise I'll have to explain the intricacies of Volto imports/exports/customizations, etc at the training, and that's exactly what I was trying to avoid with this implementation.

I'll try to get the Volto support as soon as possible.

tiberiuichim:
@sneridagh I've added test. It looks good to me, I propose that we merge this, notify our user base about it, to be aware of this.

On that note, I'll drop a note in the upgrade guide.

sneridagh:
Awesome!! I plan to start merging Volto 9 prs this weekend

sneridagh:
@tiberiuichim is it ready?

tiberiuichim:
@sneridagh Yes, it should be. Fingers crossed that we don't find something horific, but it worked for my simple project + addon.

PR #1980, opened 2020-11-09
Fixing bug related to initiation of table block with previoius table …

### LINKED ISSUE #1979: Table block is initiated with the previous table block data
**Describe the bug**
The table block is initiated with the previous block data which is not good. Table block should always initiate with an empty value.

**To Reproduce**
1. Create a page
1. Add the table block with some data
1. Add another one you will see that the new table block is initiated with the above data.
1. If not try to add another table block. You will see this error.
**Expected behavior**

**Screenshots**
Edit-Page

iFlameing:
…block data.
Fixes #1979

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Bugfix
 
+- Fixing bug related to initiation of table block with previous table block data @iFlameing
+
 ### Internal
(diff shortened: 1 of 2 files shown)


sneridagh (approved):
LGTM!!

PR #1981, opened 2020-11-09
Refactor CT icons helper: add getContentIcons

nzambello:
Moved getIcon into unique helper getContentIcons, which is extendable in config.
Added default icons in config/ContentIcons.
Used getContentIcons helper for ContentsItem and ObjectBrowser.
Added this feature in docs.

In Volto projects, you can configure this for custom content types like:

```js
import * as config from '@plone/volto/config';
import courseSVG from './icons/course.svg';

export const settings = {
  ...config.settings,
  contentIcons: {
    ...config.settings.contentIcons,
    Course: courseSVG,
};
```


tiberiuichim (approved):

tisto (approved):
@nzambello awesome new feature! I am just wondering if this is a breaking change. getIcon still exists and can be used. Therefore this should be a backward compatible change even when people have customized views, right?

giuliaghisini (approved):

nzambello:
> @nzambello awesome new feature! I am just wondering if this is a breaking change. getIcon still exists and can be used. Therefore this should be a backward compatible change even when people have customized views, right?

Right, but the helper have been moved and this could be considered breaking even if was not used anywhere and in components were using their own methods.
So I guess this could be considered breaking but not much 😂
I think this should be released in a minor for semver.

sneridagh (approved):

tisto:
> > @nzambello awesome new feature! I am just wondering if this is a breaking change. getIcon still exists and can be used. Therefore this should be a backward compatible change even when people have customized views, right?
> 
> @tisto Right, but the helper have been moved and this could be considered breaking even if was not used anywhere and in components were using their own methods.
> So I guess this could be considered breaking but not much 😂
> I think this should be released in a minor for semver.

You mean this line here, right?

https://github.com/plone/volto/pull/1981/files#diff-f3a239449bbebaac8df3a75ce26114955a8ef7a2b7f768ce0a83bdae7e8972b4L21

That's technically a breaking change then. As long as we do not differentiate between a public and an internal API, all API/helpers changes are breaking changes.

tisto:
I'd say it is time for another major release. Can we add a short paragraph to the upgrade guide? @sneridagh what do you think?

sneridagh:
Indeed, in fact, we already have a good bunch of things in store for Volto 9. I pretend to make the proposal tomorrow at the meeting.

nzambello:
I'd have another feature to add to this I guess. I'm trying to add fontawesome icons for our project CTs but that's not simple right now.

nzambello:
> I'd have another feature to add to this I guess. I'm trying to add fontawesome icons for our project CTs but that's not simple right now.

Ok, I figured out that's ok, I needed to import svgs directly as in Volto

tisto:
@plone/volto-team we need a couple of lines in the upgrade guide. Then this one is good to go into 9.x.x.

nzambello:
> @plone/volto-team we need a couple of lines in the upgrade guide. Then this one is good to go into 9.x.x.

@tisto done

avoinea (approved):

sneridagh (approved):

PR #1984, opened 2020-11-10
Add Guillotina GH action

sneridagh:


.github/workflows/ci.yml
```diff
@@ -34,11 +34,6 @@ jobs:
       # node install
       - run: yarn install --frozen-lockfile
 
-      # node code analysis
```
tisto, line 37: @sneridagh why did you remove the code analysis section here?

sneridagh: yes, because it has its own GH action, right?

tisto: Not sure if this is the way to go. gh actions can only run jobs in parallel. therefore we'd rather remove the code-analysis job again instead of the code-analysis section from the main job.

sneridagh: I thought that was the intended thing to go, whatever you think it's the best way to go.

Travis is really starting to annoy me. I guess they are starting to let it go.

sneridagh:
@tisto Should we remove the static separate job then?

tisto:
@sneridagh yeah. I guess so. Maybe at some point gh actions provide sequential pipeline steps.

PR #1988, opened 2020-11-13
Fix leaking input CSS in the link widget in draftjs

sneridagh:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Bugfix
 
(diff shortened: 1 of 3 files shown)


sneridagh:
We didn't realised that the CSS for the inputs was leaking to it, that's why it was out of place and so fat! :)

tiberiuichim (approved):

tisto (approved):
Works like a charm! Let's do. a release right away. :)

PR #1995, opened 2020-11-16
Extend the internal proxy capabilities, now the target is overridable…

sneridagh:
… and SSL aware

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### Feature
 
 - Extend the internal proxy capabilities, now the target is overridable and SSL aware @sneridagh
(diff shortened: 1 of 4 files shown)


tiberiuichim (approved):

tiberiuichim (commented):
Should we read the proxyRewriteTarget from the RAZZLE_ environment variable?

I'm thinking of the scenario where you want to use this "for real" and you have an acceptance website. I can read the environment from my project and set those values, but that means an extra thing I have to lug around in my projects.

tiberiuichim:
I'm still musing if we should still keep the "dev" name in that "devProxyToApiPath"

tiberiuichim (approved):

PR #2001, opened 2020-11-19
Add extendible AppExtras

tiberiuichim:
# AppExtras component

The `AppExtras` component is a general use insertion point for things like
analytics, general purpose code spanning the whole application or third party
services code. AppExtras component inserts will be rendered by the `App`
component.

## How to use it

You can either use it by overriding it via Component Shadowing by placing
a custom `src/customizations/components/theme/AppExtras/AppExtras.jsx`...

Or you can use the new key of `config.settings`, the `appExtras`. This is
a list of registrations, each registration is an object with:

- `match`: The `match` key is for objects compatible with [react-router's
matchPath](https://reactrouter.com/web/api/matchPath), so it can be either
a simple string or a match object.
- `component`. Use the `component` to link the component to be used.
- `props`: Extra props to be inject to the actual component used.

For example:

``` jsx
import { settings as defaultSettings } from '@plone/volto/config'

export settings = {
  ...defaultSettings,
  appExtras: [
    ...defaultSettings.appExtras,
    {
      match: '',
      component: GoogleAnalyticsPing
      props: {
        'google-tag': '123456'
      }
    },
    {
      match: {
        path: '/blogs',
        exact: true,
      },
      component: WordClouds
    },
    {
      path: '/blog/**/edit',
      component: ExtraToolbarButton,
    }
  ]
}
```

You can use this to render, for example, `<Portal>` components to be inserted
somewhere in the website.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 ### Feature
 
+- Enhance `AppExtras` component to make it pluggable and lookup a new
+  `~/config` export, the `appExtras`. These are router-path filtered components
+  that are rendered inside the `AppExtras` component @tiberiuichim
+
 ### Bugfix
 
 ### Internal
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,9 @@
 
 ### Feature
 
-- Enhance `AppExtras` component to make it pluggable and lookup a new
-  `~/config` export, the `appExtras`. These are router-path filtered components
-  that are rendered inside the `AppExtras` component @tiberiuichim
+- Enhance `AppExtras` component to make it pluggable through the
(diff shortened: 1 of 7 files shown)


tiberiuichim:
The generators would need to be updated when this change is merged. I'll change it to use `config.settings` instead, as to not require this change.

sneridagh:
@tiberiuichim won't you add props to the Component in case you want to parameterise it?

tiberiuichim:
@sneridagh: I've added, although with the current architecture doing anything fancy is not possible. Only the simplest use case, as I have shown in the PR doc, like `'google-tag': '123456'`

Though in case someone wants something advanced, they can do:

```
{
    component: (props) => <GoogleAnalytics tag="1233" />
}
```

sneridagh (approved):

PR #2003, opened 2020-11-19
Lazy-load some js libraries

tiberiuichim:
WIP on lazy-loading external libraries.


tiberiuichim:
I'll give up on the current PR. The problems are:

- Although it's somewhat nice to be able to say `withLoadables(['toastify', 'Select'])` and such, I feel that there's a level of indirection, specially as there should be some level of inteligence in the component that uses the lazy component. I don't think we can provide a solution for all of the components, how they should behave when the lazy component hasn't been loaded yet.
- All tests then need to include a mock. We can automock it, but they still need to `beforeAll` force the lazy components to load.

It's been a great learning opportunity, but I'll pass for now. I'll try to use this knowledge to further enhance the HTML block, that one is a bit of a pain to lazy-load because of the Prism library.

nileshgulia1:
@tiberiuichim I see the following traceback when creating a page. 
Screenshot 2020-11-30 at 2 50 00 PM

src/components/manage/Widgets/SelectWidget.jsx
```diff
@@ -242,6 +243,9 @@ class SelectWidget extends Component {
    */
   render() {
     const { id, choices, value, onChange } = this.props;
+    // const { Select, AsyncPaginate } = this.props;
```
tiberiuichim, line 246: Cleanup

src/components/manage/Widgets/SelectWidget.jsx
```diff
@@ -318,6 +322,7 @@ class SelectWidget extends Component {
 
 export default compose(
   injectIntl,
+  // withLoadables(['Select', 'AsyncPaginate']),
```
tiberiuichim, line 325: Cleanup

src/components/manage/Widgets/SelectWidget.test.jsx
```diff
@@ -22,17 +28,21 @@ test('renders a select widget component', async () => {
...
+        <SelectWidget
+          id="my-field"
+          title="My field"
+          onChange={() => {}}
+          onBlur={() => {}}
+          onClick={() => {}}
+        />
+      </Provider>,
+    );
+  });
```
tiberiuichim, line 45: TODO: use the more explicit style to test, see Workflow.test.jsx

src/reducers/lazyLibraries/lazyLibraries.js
```diff
@@ -0,0 +1,16 @@
+import { LOAD_LAZY_LIBRARY } from '@plone/volto/constants/ActionTypes';
+
+const initialState = {};
+
+export default function lazyLibraries(state = initialState, action = {}) {
```
tiberiuichim, line 5: Add docstring

tiberiuichim:
TODO on this PR:

- fix the HTML block test (the snapshot was wrongly updated)
- provide "bundles", a way to preload a set of libraries. Instead of loading the extra libs "on demand" and be faced with empty components, why not preload them when we login, or when faced with the Edit screen?

tiberiuichim:
TODO:

- add ref forwarding
- rename the "withLoadables" to "injectLazyLib"
- provide a "preload" hoc/hook pair that can preload some libraries

tiberiuichim:
Rename to injectLazyLibs

PR #2006, opened 2020-11-20
Fix bundle analyzer plugin

tiberiuichim:
Fix `yarn analyze` command by packing our own version of
  webpack-bundle-analyzer integration. It has a few differences in
  configuration to the previous integration: it saves the bundle stats by
  default and doesn't open an HTTP server. Check output of `yarn analyze`
  command to identify the path to the `reports.html` file @tiberiuichim

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,12 @@
 
 ### Bugfix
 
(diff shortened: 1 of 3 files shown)


sneridagh (approved):

tisto:
@tiberiuichim good work! This is important!

PR #2020, opened 2020-11-25
Lazy-load prismjs libraries in HTML block

tiberiuichim:
Avoid PrismJS, as it's big. Also give up on storing value as state in the block.


tiberiuichim:
This saves 6kb from the main additional chunk that is loaded. A 1.25% decrease in size. 🥇

sneridagh (approved):

PR #2032, opened 2020-11-30
Search depth

giuliaghisini:
Added search **depth** in listing blocks. 
The depth field shows up only if there's at least one 'path' criteria in query.


nzambello (approved):

src/actions/querystringsearch/querystringsearch.js
```diff
@@ -10,23 +10,33 @@ import { settings } from '~/config';
...
-        ...data,
-        ...(!data.b_size && {
+        ...d,
+        ...(!d.b_size && {
           b_size: settings.defaultPageSize,
         }),
         ...(page && {
-          b_start: data.b_size
+          b_start: d.b_size
             ? data.b_size * (page - 1)
```
sneridagh, line 36: Missing data -> d ?
I'm not a fan of using such short var names, I see that the naming here might be redundant, but still... Can we stick  to meaningful name vars?

sneridagh (changes requested):

sneridagh (approved):

PR #2036, opened 2020-11-30
Generate language file of added missing German translations

ksuess:


tisto:
@ksuess did I forgot to run "yarn i18n"? Shouldn't that have been caught by our CI check?

ksuess:
Well, that would be super comfortable, if an additonal 'yarn i18n' would be run automagically.
I started manually.

tisto:
Ok. Seems we do not do this check any longer on gh actions:

https://github.com/plone/volto/blob/master/.github/workflows/ci.yml

Will look into re-adding it.

PR #2042, opened 2020-12-02
feat(restapi): add emailSend action

nzambello:
Add emailSend action for `/@email-send` restapi endpoint.

Ref: https://plonerestapi.readthedocs.io/en/latest/email-send.html


tiberiuichim (approved):

giuliaghisini (approved):

sneridagh (approved):

cekk:
Be aware that this endpoint only works if the "sender" has the "Use mailhost services" permission.

By default i think that only admin has this permission to avoid spam problems, so probably this action could be unusable in most of the cases.

We needed this for a simple form action and we ended up with a custom endpoint like email-notification (where you can't pass the recipient)

PR #2043, opened 2020-12-02
[Breaking] Listing default view

jackahl:
Change default Listing Template to include only Text and renamed the old default Template to Summary Template. I think this makes more sense as per default most content types dont even have the image behavior necessary to show something in the current default listing template.


tisto:
@sneridagh I guess it is time for Volto 11.x :)

sneridagh (approved):

PR #2056, opened 2020-12-11
Make sure prism is loaded before loading HTML block

tiberiuichim:


iFlameing (approved):
Tested it locally and now it works.

tisto:
I have no clue why the Guillotina tests fail. Seems to be related to the pg test setup. @bloodbare any clue what's the issue here?

tisto:
@sneridagh the guillotina tests are failing on master, so I guess we are good to merge. They started to fail after merging the PR from Manabu, which only contains changes to README.md.

PR #2058, opened 2020-12-11
PoC refactoring using React Context for passing down the config in co…

sneridagh:
…mponents

### DIFF
--- a/src/components/index.js
+++ b/src/components/index.js
@@ -5,7 +5,7 @@
 import loadable from '@loadable/component';
 
 //  Do not lazy load them, since it has not much sense (they will live in the main chunk)
-export App from '@plone/volto/components/theme/App/App';
+// export App from '@plone/volto/components/theme/App/App';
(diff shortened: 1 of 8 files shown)


sneridagh:
@tiberiuichim I don't know if I'm into something or not... we'll see.
What do you think?

tiberiuichim (commented):
For the function components we could also have a `useVoltoConfig` hook, like: https://github.com/eea/volto-slate/blob/master/src/hooks/useEditorContext.js

tiberiuichim:
How would we access the config in helper functions? They'd have to be refactored and passed a config registry, I believe?

tiberiuichim:
We also use settings in the server module.

sneridagh:
@tiberiuichim I have the faith that if we remove the requirement from the components alone, it would be enough to cut the chain. The others can still work as before with the direct imports.

Anyways, I still think that we should separate the registry from the core config.

sneridagh:
Closed, we now have the config registry.

PR #2063, opened 2020-12-12
Fix some key points to improve the circular imports problem

sneridagh:
Those are not certainly killers of the problem, but certainly helps.

Also, the tests seems to run faster. (55secs vs 85secs in a quad core machine)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,15 @@
 
 ### Internal
 
+- Fix some key points to improve the circular imports problem @sneridagh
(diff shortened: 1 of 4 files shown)


tiberiuichim (approved):
The hidden key to unlocking the Volto power.

sneridagh:
@tiberiuichim here we go

tiberiuichim:
@sneridagh Let's merge this ASAP, even just the speedup in tests makes this really worth it.

Issue #2009, opened 2020-11-22
Users and groups settings does not store role and group membership on creation or change


fredvd:
After installing a Volto site with the docker plone backend I tried to create a new user having the Manager role to stop using the default Plone/Zope admin/admin.  

**To Reproduce**
In a fresh installed Volto site, go to the users & groups settings, create a new user, assign the user to the Manager role or add the User to the Administrators Group.

**Expected behavior**
When I log out of Volto and try to log in with my new created user, I should be able to create and edit pages. I can't. 

Checking in the Plone backend control panel I see that the created user does not have the Manager role and/or is not added to the Administrators group.
<img width="834" alt="users and groups settings" src="https://user-images.githubusercontent.com/394784/99903125-ae087700-2cc2-11eb-92a8-0364b879088a.png">

Also, If I check the 'Manager' role in the User/role matrix in the Users and Groups settings, markinet the checkbox is also not persisted.   Setting the role through the Plone backend does show the role checked on reload in the Volto controlpanel

Volto 9.1.0
Plone 5.2.2 
Zope 4.5.1
Python 3.8.6

nileshgulia1:
I too observed this. When I try to change the roles and try to refresh, the role doesn't seems to apply. I wrote the user and groups controlpanel way long ago, somehow didn't notice the problem. Will have a look maybe this week!

tisto:
I'd recommend writing a failing Cypress test first, then push it on a PR to prove the problem. After this write a fix and prove with a green CI that the problem has been fixed.

fredvd:
Fixed in other issues/PR's

PR #2064, opened 2020-12-12
Reimplement the architecture of user/groups controlpanel

nileshgulia1:
https://github.com/plone/volto/issues/2009


nileshgulia1:
@fredvd You can have a look at it. I also added cypress tests to detect regression in future.

sneridagh (approved):

avoinea:
I think we should have the save button here as saving on checkbox select is bugous. 

For example:

1. You select a role, click back and then Users & Roles:

role

You need to hard refresh to see the changes.

avoinea:
This one seems even worse:

groups

nileshgulia1:
Updating a role doesn't return the roles updated for a specific user/group. Instead we have to re-call the /@users endpoint to fetch the updated user.
In standard plone we do have a save button. Here are [designs](https://app.zeplin.io/project/5c76b54e0e8c876239917ea3/screen/5c76b59686d8b86292eecca7) in mobile view.
For web views, adapting to these can be a little tricky or maybe @albertcasado can have better options...

sneridagh:
In @pbauer training, his first reaction was to look for the save button and when not finding one, feeling like unsure that by just clicking a checkbox, the role was set. In such a delicate action (setting user roles), I think we need a proper save button here.

@nileshgulia1 I know we might do an extra call, but in such operation I guess it's worth it.

@avoinea @tiberiuichim thoughts?

avoinea:
Indeed, I also think that all the back-end persistent actions should be done via a `Save` button for UX consistency.

tisto:
We could either show a toast message when the setting has been stored in the backend or add a dedicated save button. Though, the current situation is not very user friendly.

nileshgulia1:
@sneridagh I gave this a try and now kinda works, see the demo. Also added the save button to indicate the roles have been saved in the backend. 
In Plone we also have "inherited" icon to show the user role belonging in a group. I can tackle that and some other cosmetics in a seperate PR.
Since we fetch all users and groups on mount. I removed the need for "show-all users" button. Anyway the right place for that should not be inside the `<Table/>` component as it blocks the view on initial mount.

https://user-images.githubusercontent.com/22280901/103573306-094d9c00-4ef4-11eb-9cad-2cf7e0731bc7.mp4

sneridagh:
@nileshgulia1 @avoinea how about this one? is it ready?

nileshgulia1:
@avoinea I made some UI changes you suggested. 
 - switch the order of label and icon, so first text then icon
 - made both icon and label clickable.
Screenshot 2021-01-21 at 1 46 51 PM

tisto:
I would recommend to either use the standard Pastanaga save button or use a toast message. I have a slight preference for a toast message here because we do not have those inline save buttons anywhere else in Volto if I am not mistaken. Would be nice to hear @albertcasado's opinion here as well...

avoinea:
@tisto , @nileshgulia1, @sneridagh 

IMHO the best approach would be to split this panel in two: 
- Users
- Groups

This way we can move the buttons within Toolbar where they belong.

I don't see any good reason to have them together :wink:

tisto:
+1. Since this control panel does not rely on registry entries we shouldn't have a problem here.

nileshgulia1:
@tisto @sneridagh This is how the controlpanel looks right now.
Screenshot 2021-01-23 at 12 18 44 AM

-  I have separated the "users" and "groups" to different components in the Controlpanel site.
- moved the action buttons in the toolbar.
- The toast is displayed on adding user and saving roles.

-cc @avoinea

avoinea (changes requested):
@nileshgulia1 Niice!

There are still some issues to be solved:
1. When you click cancel it should go back to /controlpanel (both users and groups)
2. When adding a new group, the modal form seems to be adding a new User
3. Please add cypress also for the above scenarios

avoinea:
> When adding a new group, the modal form seems to be adding a new User

Actually the required `email` property for group confused me. The form should respect the same `required` fields as the Plone one http://localhost:8080/Plone/@@usergroup-groupdetails

I get this error when I add a new group:

group-error

avoinea:
This note also seems not to work:
```
Note that roles set here apply directly to a user. The symbolindicates a role inherited from membership in a group.
```

**Plone:**
plone

vs

**Volto:**
volto

avoinea:
These panels should display the unauthorized error on backend errors:

unauthorized

See example:

https://github.com/plone/volto/blob/master/src/components/manage/Controlpanels/ContentTypes.jsx#L349-L352

nileshgulia1:
@avoinea I've made some changes according to comments.
https://github.com/plone/volto/pull/2064#issuecomment-765903020
I can't see this error when adding a new group. Can you please specify, how to reproduce it?
105579687-e48b6c80-5dad-11eb-9b8d-9bc5d3040616




Also the modalForm, shows the "Add group" for groups. Let me know if this isn't happens at your end.
Screenshot 2021-01-23 at 7 00 10 PM

We should keeps the roles, here in the ModalForm atleast I guess?

cypress/integration/user-control-panel.js
```diff
@@ -6,27 +6,61 @@ if (Cypress.env('API') !== 'guillotina') {
...
-    it('Should add User to controlPanel', () => {
       // when I added a user from controlPanel
-      cy.get('.addSVG').click();
+      cy.get('Button[id="toolbar-add"]').click();
       cy.get('input[id="field-username"]').clear().type('iFlameing');
       cy.get('input[id="field-fullname"]').clear().type('Alok Kumar');
       cy.get('input[id ="field-email"]').clear().type('info@example.com');
       cy.get('input[id="field-password"]').clear().type('test@test');
-      cy.get('button[title="Save"]').focus().click();
+      cy.get('button[title="Save"]').click(-50, -50, { force: true });
```
tiberiuichim, line 22: I think I'd rather have focus() then the coordinates.

nileshgulia1:
> This note also seems not to work:
> 
> ```
> Note that roles set here apply directly to a user. The symbolindicates a role inherited from membership in a group.
> ```
> 
> **Plone:**
> plone
> 
> vs
> 
> **Volto:**
> volto

@avoinea I found an issue here. If we change `authentictedUsers` role , then somehow the roles are not changing in the backend. Infact that particular role is 'inherited' and applied to the specific group as happens in the standard plone. 

Due to this we cannot able to remove the inherited roles.

avoinea:
If the user exists we should display the error within the modal or within a toast:

existing-user

avoinea:
I get the following error when I try to add a group that already exists:

existing-group

avoinea (approved):

nileshgulia1:
@sneridagh You wanna take a final review to this! You can merge after that :)

CHANGELOG.md
```diff
@@ -370,6 +370,8 @@
 
 ### Bugfix
 
+- Make sure that prism is loaded before rendering HTML block @tiberiuichim
```
sneridagh, line 373: This is outdated for sure, why is this being added?

sneridagh: A merge leftover probably

cypress/tests/core/user-control-panel.js
```diff
@@ -1,30 +1,66 @@
...
-    cy.visit('/');
-    cy.autologin();
-    cy.visit('/controlpanel/users');
-    cy.waitForResourceToLoad('@navigation');
-    cy.waitForResourceToLoad('@breadcrumbs');
-    cy.waitForResourceToLoad('@actions');
-    cy.waitForResourceToLoad('@types');
-  });
+if (Cypress.env('API') !== 'guillotina') {
+  describe('User Control Panel Test', () => {
```
sneridagh, line 2: We removed all the conditionals if Guillotina ... This can be removed as well as it's in cypress/tests/core

src/actions/authRole/authRole.js
```diff
@@ -0,0 +1,13 @@
+/**
+ * AuthRole actions.
+ * @module actions/authRole/authRole
+ */
+
+import { AUTH_ROLE } from '@plone/volto/constants/ActionTypes';
+
+export function authenticatedRole(role) {
```
sneridagh, line 8: Why do we need it? And what does it do? Why is it persistent?

nileshgulia1: I created a new action-reducer pair named `AUTH_ROLE` for keeping the authenticated role available (persistant) throughout the userSession. As the other roles are dependent on this, it needs to be available on initial mount of the controlpanel.
Screenshot 2021-04-08 at 2 13 36 PM

At the first place I haven't used it as "persistant", causing abnormal behaviour when updating authenticated roles. As this also responsible for behaving as "inherited" roles for users.

tisto:
@avoinea @nileshgulia1 @sneridagh I just tried to create a user and set permissions in a current Volto and the users and groups controlpanel is in a very sorry state right now. We should really try to finish this PR and improve the situation. With the seamless mode, we do not have the fallback to Plone Classic any longer and then the most basic control panels need to work (I ran into the problem that I can not check any checkboxes any longer, not sure if this is fixed by this PR as well).

nileshgulia1:
@tisto The current state of "Users and groups" controlpanel is improved a lot in this PR. I've tried to cover most of the edge cases especially working with "inherited roles" (Decided to store inherited`AuthRole` as a persistant reducer, For more info see: https://github.com/plone/volto/pull/2064#discussion_r609472618 ). For me its good to go, however it'd be great also @sneridagh to have a final look.

sneridagh (commented):
@nileshgulia1 I think the persistent reducers has absolutely no sense to me... I do not see why do we need them, also it holds a single role? It should be a list of roles, right?

Also, given the huge amount of changes is almost impossible to follow the changes flow... So, my 5c is: could we start over and go step by step fixing one thing at a time? So we go back two or three steps in order to put the right things in the right place and face one challenge and give the right solution one at a time.

sneridagh:
Also, we need an army of Cypress tests for it.

nileshgulia1:
@sneridagh I'll try to create subsequent PR's by splitting this one.

pnicolli:
Finally had a chance to test this. The Users controlpanel works fine 👍 
I found a few minor things I'm not a fan of:

1. IMHO, "No value" should never be present in select inputs
image

2. I see this Tick appearing when hovering the delete button, which is totally non-blocking for the merge, imho, but I think it isn't needed 
image

3. In the Groups controlpanel, I would change the "Add new group" icon to one that represents more than one person, just to make it clear that it's different from the one in the Users controlpanel
image

Despite these comments and the facts that these controlpanel still lack the possibility to add a user to a group and to reset a user's password, I would totally suggest to merge this and work on the other things separately, since it's a major improvement over the current one in master.

If cypress tests are coming, it's even better, of course :)

Thanks @nileshgulia1

nileshgulia1:
> In the Groups controlpanel, I would change the "Add new group" icon to one that represents more than one person, just to make it clear that it's different from the one in the Users controlpanel.

@pnicolli Currently we don't have `add-group.svg` or something similar which will better indicate adding of groups. Maybe can just use the "+" icon?

As far as I remember, I looked into going deeper in "adding user to a group", might have to do some changes in restapi idk.  I agree to make separate PR's on small fixes further. Let's not overfill this PR.

Apart from the icon issue, I think we can go ahead merging  this.

-cc @sneridagh

sneridagh:
@nileshgulia1 I'm moving forward also some fixes to the selectWidget, it might get some time to make it through (see my last draft PR). I would like the users and groups use its own select, with its special use case on it, so the default one is more agnostic and simple. How much would that cost?

nileshgulia1:
@sneridagh I think there isn't much special case attached to the "Select" component that we use in this PR. We only need "selects" when creating a user/group in a `ModalForm` which is schema based. Its only with the "no-value" field that we needed to get rid of, at the same time I don't think its a blocker for the PR.

nileshgulia1:
@sneridagh I've used the type: 'array' to indicate the use of ArrayWidget. I would wait for your select refactor PR to go in first and see if we can make it work. Separating/creating a new Select component( not a new widget ) for this specific use case, doesn't work for me as a quick feature and I think it will cost a lot more to make it work. 

Also for this we would have to make a refactor to `<ModalForm/>` , which in many places we are already using it like this.

@sneridagh From the yesterday's meeting, We should only use SelectWidget with "plone.restapi" based json data, right?

Here, the use case we have is for listing roles coming from `/@roles` backend endpoint, so no hardcoded values.

pnicolli:
Hi everyone, what is the state on this PR and related things that were mentioned in the last comments? The current volto user controlpanel is really not working and it really needs some sort of fix. As I mentioned the last time we talked about it in a volto meeting, I would totally accept merging a PR and reworking some of the stuff later. About a month has passed and I think something should be done about this, because it would be otherwise unacceptable to skip fixing a vital part of a cms like I believe this one is.

nileshgulia1:
@pnicolli I'm okay to merge the PR. @sneridagh will merge once he gets back from holidays!

nileshgulia1:
Sorry, accidentally merged wrong branches. Fixed! Have a nice weekend!

nileshgulia1:
@sneridagh We can merge this once you give green flag. I accidentally rebased on wrong branches while testing and cleaned it up thereafter. However I request you to "squash and merge" this (If you have option).

src/actions/index.js
```diff
@@ -89,7 +89,6 @@ export {
   resetPassword,
   updatePassword,
   updateUser,
-  showAllUsers,
```
sneridagh, line 92: @nileshgulia1 this was introduced by
https://github.com/plone/volto/pull/1733

I think we are removing it with this changes... Can you take a look?

src/actions/users/users.js
```diff
@@ -12,7 +12,6 @@ import {
   UPDATE_PASSWORD,
   INITIAL_PASSWORD,
   RESET_PASSWORD,
-  Show_All_USERS,
```
sneridagh, line 15: Same as above, missing https://github.com/plone/volto/pull/1733 changeset.

src/actions/users/users.js
```diff
@@ -79,17 +78,6 @@ export function listUsers(query) {
...
- * Show All users function
- * @function showAllUsers
- * @returns boolean
- */
-export function showAllUsers() {
-  return {
-    type: Show_All_USERS,
-  };
-}
-
```
sneridagh, line 92: https://github.com/plone/volto/pull/1733

src/constants/ActionTypes.js
```diff
@@ -85,7 +86,6 @@ export const SYSTEM_INFORMATION = 'SYSTEM_INFORMATION';
 export const DATABASE_INFORMATION = 'DATABASE_INFORMATION';
 export const SET_BLOCKS_CLIPBOARD = 'SET_BLOCKS_CLIPBOARD';
 export const RESET_BLOCKS_CLIPBOARD = 'RESET_BLOCKS_CLIPBOARD';
-export const Show_All_USERS = 'Show_All_USERS';
```
sneridagh, line 88: https://github.com/plone/volto/pull/1733

src/constants/ControlpanelIcons.js
```diff
@@ -0,0 +1,23 @@
+/**
```
sneridagh, line 1: Not used any longer, deleted in #2383 Please checkout and delete.

src/reducers/users/users.js
```diff
@@ -14,13 +14,11 @@ import {
   UPDATE_USER,
   INITIAL_PASSWORD,
   RESET_PASSWORD,
-  Show_All_USERS,
```
sneridagh, line 17: https://github.com/plone/volto/pull/1733

src/reducers/users/users.test.js
```diff
@@ -8,15 +8,13 @@ import {
   UPDATE_USER,
   INITIAL_PASSWORD,
   RESET_PASSWORD,
-  Show_All_USERS,
```
sneridagh, line 11: https://github.com/plone/volto/pull/1733

sneridagh (changes requested):
@nileshgulia1 I finally had time to go through it properly. I've found some things, could you please check them and fix them?

The main one is that seems that a feature introduced in https://github.com/plone/volto/pull/1733 is missing (probably due to a merging issue) or did you already took care of it in your code?

sneridagh:
@nileshgulia1 let's merge this asap after you take a look at the last issues!

nileshgulia1:
> @nileshgulia1 I finally had time to go through it properly. I've found some things, could you please check them and fix them?
> 
> The main one is that seems that a feature introduced in #1733 is missing (probably due to a merging issue) or did you already took care of it in your code?

Yes @sneridagh, I removed it intentionally, We "showAll" `users/groups` on intial load of the corresponding controlpanel now. Anyway the correct place for "ShowAll" button should not be inside the table, just my 2c.

I removed the obsolete icon file. We can merge now. 

Thanks

sneridagh:
We do show all users on initial load? How about if we have 100k users, eg. coming from ldap?

Could you quickly point me Where do we have that call?

nileshgulia1:
@sneridagh I've tried to replicate what currently Plone has, here's the GET call: https://github.com/plone/volto/blob/2230371d2c8c8c07d14debde764cb0b104ec0370/src/components/manage/Controlpanels/Users/UsersControlpanel.jsx#L112

Screenshot 2021-09-01 at 10 34 03 PM

In the current scenario( volto.kitconcept.com), we don't show any user onload and have to click on "showAll" everytime,
which intern gonna list "all users" without any batching. See below:
https://github.com/plone/volto/blob/master/src/components/manage/Controlpanels/UsersControlpanel.jsx#L482

I don't remember the exact reason but the correct place for that button should be somewhere in the segment, not inside the table as it interferes with the content.

nileshgulia1:
Maybe I'm missing something here, @sneridagh Let me know your thoughts, I'm okay to re-add the changes lost.

nileshgulia1:
> Also, per documentation, all users listings are not allowed if you are not manager:

Okay, this feature also lacks in current volto, as its there in Plone with LDAP accounts. I can take care of it too after this PR gets merged.

sneridagh (approved):

PR #2080, opened 2020-12-20
Integrate the generator in the Volto repo

sneridagh:


sneridagh:
@plone/volto-team 
I had the feeling that the generator should live inside the Volto project itself. For a variety of reasons, but the main one, being the fact that we need to test the current Volto changes in a project environment too. We've been bitten already a couple of times with changes in Volto worked well, but it a project not. So this move tries to address it. Leaving the generator in it's own repo and test it here wasn't going to solve the problem, since we could encounter use cases where changes in both would be needed in order to make both test pass at the same time (in different PRs) leading to a chicken-egg problem.

Also, my intention is not to use a classic monorepo project way of doing things. Although I used `packages` naming, it is merely aesthetic. So Volto remains in the root, accessory packages in the `packages` directory and they are totally detached. At some point we can seek a more tight integration, but personally I'm not a fan of the monorepo story.

I also moved the test-backend-server to docker too, eventually we could get rid of the python build...

So I guess it's the best move for improving Volto's resilience. Let me know your thoughts!

tiberiuichim (approved):
I like it, it makes sense to have it in Volto.

tisto:
The generator belongs into the main Volto package, no question about that. We always (should) release and test those components together. If two components belong to each other and, at least, partly depend on each other they should be tested and released together. Separating those two components gives us no additional value whatsoever. That's something that we often got wrong in the Plone community in the past and we should overcome this false notion of loose coupling.

nzambello:
Completely agree

avoinea (approved):

sneridagh:
@plone/volto-team FYI I will merge this during the weekend, I assume everybody agreed on this, if not, shout out!

tiberiuichim:
Go for it, as far as I'm concerned.

Issue #2025, opened 2020-11-26
Optimize Volto performance score


tiberiuichim:
Just dumping some ideas.

- critical css embedded in html
- async css loading 

```
<link rel="preload" onload="this.onload=null;this.rel='stylesheet'" href="https://www.eea.europa.eu/portal_css/EEADesign2006/ploneStyles1606245946918895960.css" as="style" />
```

For Firefox this is needed: https://www.eea.europa.eu/loadcss.js

- generate multiple bundles with different browser polyfills level. These bundles would not polyfill features that are supported by evergreen browsers. 
```
    const extractor = new ChunkExtractor({
      statsFile: path.resolve('build/loadable-stats.json'),
      entrypoints: ['client'],
    });
```
- use service workers to load non-critical JS

PR #2088, opened 2020-12-22
WIP to improve DevTools Lighthouse scores

silviubogan:
Maybe useful one day: https://gist.github.com/silviubogan/5d01886907d3b0ec981b25dc06d1eeda.


src/helpers/Html/Html.jsx
```diff
@@ -82,7 +84,7 @@ class Html extends Component {
           <meta name="viewport" content="width=device-width, initial-scale=1" />
           <meta name="apple-mobile-web-app-capable" content="yes" />
           {/* Add the crossorigin while in development */}
-          {extractor.getLinkElements().map((elem) =>
+          {extractor.getLinkElements({ rel: 'preload' }).map((elem) =>
```
tiberiuichim, line 87: I don't think this is needed, make sure to test in "production" mode by running "yarn build; yarn start:prod"

tiberiuichim (changes requested):

tisto:
@silviubogan @tiberiuichim nice! At kitconcept we integrate lighthouse (and some other perf tests) to continuously monitor our performance over time. We could do the same for Volto core (this has been on my todo list for quite some time actually). The only problem we have to overcome is to automatically deploy the latest version of Volto (master) somewhere as part of our CI job so that a performance test can run against this instance.

src/helpers/Html/Html.jsx
silviubogan: I think it helps: https://developer.mozilla.org/en-US/docs/Web/HTML/Preloading_content. I tested with the given commands. The score goes from 17 to 16 on my laptop when compared to Volto master, but it is WIP.

cekk:
@tisto have you automated it in your projects or do you run perf tests manually when needed?

Btw we started looking into Volto perf problems too and this pr could be a good starting point ;)

tiberiuichim:
See also https://github.com/plone/volto/issues/2025

tisto:
@cekk we auto deploy to a dev server after all tests have been passed. After the deployment we run a set of perf. tests against the dev instance (that contains example content that always stays the same). We then include the lighthouse HTML reports in Jenkins. I am happy to help with setting this up. Though, we have to solve the problem of the auto-deployment first.

tisto:
@sneridagh what if we set up a small hetzner VM where we auto-deploy the latest volto master automatically on every commit? With this we could at least track the performance over time on master (not on PRs and branches though).

I am open to any other thoughts. Ideally we would auto-deploy every single PR and run lighthouse against it and add the lighthouse score in the GH pull request. I could take care of the CI part but I am no expert on the auto-deployment part.

cc @cekk

mtoepfl:
Perhaps Vercel can be used. Should not matter that Volto doesn't depend on NextJS. RestAPI can be hosted separate (CORS or proxy), or not? They offer a plan for Open Source projects:

https://vercel.com/pricing

tiberiuichim:
Closed, there's critical css support now in Volto 10.10

PR #1985, opened 2020-11-10
Work on #1932 to lazy-load react-toastify in Volto

silviubogan:

PR #2090, opened 2020-12-23
Use react-toastify through loadable everywhere

silviubogan:
and update the App component's snapshot.

Related:
1. https://github.com/plone/volto/pull/1985
2. https://github.com/plone/volto/pull/2003


tiberiuichim:
For some reason this PR only as 3 checks, no cypress tests. @tisto, @sneridagh any idea why?

sneridagh:
None of the gh actions tests have run. Maybe because the branch is not local to the plone org?

nzambello:
I see here all the tests are passed.
I didn't run a live application with this branch nor tested with lighthouse or similar, but I think this is a good step because this was one of the slowest parts being loaded and have significant impacts on rendering efficiency (time and resources).

sneridagh:
Not sure if the indirection from loadable justifies it, from an approachability point of view. The resultant code is super ugly...
@silviubogan is react-toast that big? It's really worth it?

sneridagh:
@silviubogan I would rather use a more friendly approach like the one @tiberiuichim was drafting here #2003

silviubogan:
@sneridagh I like the approach in #2003 and I think the react-toast is loaded in App.jsx in any use case, so it is not very well suited for a change like the one in this PR. I think @tiberiuichim knows better what was the plan behind.

tiberiuichim:
This PR offloads 23 KB from the main bundle, from 1.49 MB to 1.47 MB.

silviubogan:
I close this PR because it was replaced with work in https://github.com/plone/volto/pull/2003 and https://github.com/plone/volto/pull/2170.

tiberiuichim:
Indeed, it's old work, though we still have to do the actual work, now that the injectLazyLibs is in Volto core.

pnicolli:
> Indeed, it's old work, though we still have to do the actual work, now that the injectLazyLibs is in Volto core.

I agree. Even if toastify is used a lot around the codebase and it is not that big, I would be happy to load it lazily since it is not used right away when the page loads, but only after some input or some other things happen.

davisagli:
Closing since it sounds like a new PR implementing this using `injectLazyLibs` would be more appropriate.

PR #2091, opened 2020-12-23
Add navportlet implementation

tiberiuichim:


sneridagh (approved):

sneridagh:
@tiberiuichim is it ready?

tiberiuichim:
@sneridagh No, I need to rename the endpoint in plone.restapi. It also depends on release in plone.restapi

CHANGELOG.md
```diff
@@ -10,6 +10,17 @@
 
 ### Feature
 
+- Add `NavPortlet` component, it can fetch the `@navportlet` plone.restapi
```
sneridagh, line 13: renaming

sneridagh: Not sure if we want to be that verbose in the changelog.

src/components/theme/Navigation/ContextNavigation.stories.js
```diff
@@ -0,0 +1,216 @@
...
+  // subcomponents: { ArgsTable },
+};
+
+export const Renderer = () => (
+  <ContextNavigationComponent navigation={navigation} />
+);
+
+export const Connected = () => <ContextNavigation />;
+
+const style = `
```
sneridagh, line 131: Why the style is needed?

src/components/theme/Navigation/ContextNavigation.stories.js
```diff
@@ -0,0 +1,216 @@
+import ContextNavigationDefault, {
```
sneridagh, line 1: We are all newbies writing Stories, I guess we will get through and learn as we write them. Having them in MDX is not that important, if we never export them or write prose around enough to be a doc on its own.

src/components/theme/Navigation/ContextNavigation.stories.js
tiberiuichim: There's no style in Volto right now for the NavPortlet.

src/components/theme/Navigation/ContextNavigation.stories.js
tiberiuichim: I need good editor integration with MDX, otherwise it's pretty painful. Maybe I could live with it, let's say, but doing something basic like integrating that style string proved too difficult.

Issue #2098, opened 2021-01-04
Problem with large files download


cekk:
**Describe the bug**
When you try to download a large File from Plone, the ssr can't handle it and the call is never resolved.

**To Reproduce**
1. Add a new File object with a ~50mb file attachment
2. Click on newly created object in Volto
3. Click on file name to download it (the @@download/file link)

**Expected behavior**
Download file browser popup appears

The problem seems that superagent does not wait the end of the api call to serve the response, and this leads to an error on SSR:

<img width="612" alt="Schermata 2021-01-04 alle 11 22 39" src="https://user-images.githubusercontent.com/1402078/103525233-2d22ca80-4e7f-11eb-8184-9c1e4f00e4f3.png">

Probably we need to pipe the response instead of listening to the end event, but i'm not a super expert in streams.

cekk:
Ok, found the problem. See #2099  and #2100 

I can close this issue.

PR #2099, opened 2021-01-04
Handle errors on file and image download

cekk:
This i related to #2098 because i had an error from superagent but it wasn't managed correctly by the SSR.

Now if the superagent call returns an error, we reject the promise and print the error.


sneridagh (approved):
LGTM

tiberiuichim (approved):

PR #2100, opened 2021-01-04
Increase maxResponseSize for superagent calls

cekk:
By default superagent drops responses larger than 200mb.
This leads to an error (aka infinite page loading because errors are not well managed..se #2099 ) if we want to download files > 200 mb from Plone.

I don't know if you limit file sizes on Plone usually, but we have several files > 200 mb.

With this change i changed the default limit to 500 mb but i don't know if it's a good idea setting it higher or not.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@
 - Fix redirection for Link objects. @cekk
 - Fix import order in server.jsx. @cekk @tiberiuichim
 - Make sentry config more resilient to edge cases (SPA, storybook) @sneridagh
-
+- Increase maxResponseSize for superagent calls. Now is 500mb (#2098) @cekk
 ### Internal
 
 - Translations german: Unauthorized, Login/Register @ksuess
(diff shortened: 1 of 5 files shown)


tiberiuichim:
@cekk I'm ok with the change, what about making this a configurable setting? `maxResponseSize`

sneridagh:
@tiberiuichim I'm also ok but maybe we can completely unset it? If you need to put some limit, you can implement it in Nginx/web server side anyway... just wondering.

tiberiuichim:
@sneridagh On superagent's side, it looks like the default was added to avoid a [DOS](https://github.com/visionmedia/superagent/issues/1259) 

I'd rather we have "secure by default", document this setting and let people know how to solve it in case they have special needs. 200 MB is more then enough for everybody and then @cekk can configure their instance, from the project, with a higher setting.

cekk:
i've update the pr with this value into settings and add some infos in the readme

sneridagh (approved):

tiberiuichim (approved):

PR #2102, opened 2021-01-05
[Breaking] Removing id from form field wrapper

iFlameing:
I think that we don't need this or we have to change the "field" to something like "wrapper" so that if someone doesn't pass the fieldset we are assigning two ids with the same value. The Form.Field and label  you can see by visiting these two links
https://github.com/plone/volto/blob/master/src/components/manage/Widgets/FormFieldWrapper.jsx#L106
https://github.com/plone/volto/blob/master/src/components/manage/Widgets/FormFieldWrapper.jsx#L113


tisto:
Either the wrapper should have a unique ID or the field, not both.

tisto (changes requested):
@iFlameing seems you need to amend a Cypress test. I am actually wondering why we only have to amend one Cypress test. Seems we do not have a good forms test coverage.

tisto:
Note: we remove the id from the wrapper. We have to make sure the fieldset is in the ID of the input field itself to avoid collisions when using the same form fields in different fieldsets.

tiberiuichim (approved):

sneridagh:
@plone/volto-team Timo and I think that the PR might be potentially breaking since someone might rely on the ids for their acceptance tests, so I marked it as breaking and schedule it for the next major.

sneridagh:
@iFlameing please can you write docs about this breaking change? docs/source/upgrade-guide/index/md

iFlameing:
@sneridagh yes!

docs/source/upgrade-guide/index.md
```diff
@@ -12,6 +12,12 @@ This upgrade guide lists all breaking changes in Volto and explains the
 
 ## Upgrading to Volto 11.x.x
 
+### id is removed from FormFieldWrapper
+
+We have removed the id from the FormFieldWrapper because it is coincide with the label id if we don't provide the fieldset. 
+If you have cypress test which depends on this id then just remove the id from the test and if test fails then just 
+add `.react-select-container` instead of your id.
```
tisto, line 19: @iFlameing maybe we can show an example diff here or at least point to this PR, so people can check what actually changed.

tisto (approved):
I'd add a link to this PR in the upgrade guide. Apart from this the PR looks good!

PR #2097, opened 2021-01-03
image loading lazy

mamico:
add the loading=lazy attribute to the img tag for the Image and HeroImage blocks

* currently 73% browser support https://caniuse.com/loading-lazy-attr but no side effect for others
* no need to add 3rd-party javascript code https://web.dev/browser-level-image-lazy-loading
* improve performance and lighthouse score #2088

PR #2103, opened 2021-01-05
Add Image for srcset and lazy loading

nzambello:
Features:
- Add ImageFromUrl component which selects Plone miniatures using srcset attr and using lazy loading as in #2097 
- Add getSrcSet helper for srcset and sizes basing on selected miniature

Added code and tests, refactored components with images to use ImageFromUrl

This should improve performances avoiding loading heavy images if not needed.


giuliaghisini (approved):
wow

tiberiuichim (commented):
Maybe rename ImageFromUrl to ScalableImage or something along these lines?

sneridagh (commented):
Awesome PR! Thanks!
Only two things:
- naming: Since it's the embodiment of a how a Plone image should render, why not just call it `Image`?
- scales: @tisto made some research based on our latest projects, and tried to push for a (new) standard scales for Volto:
https://github.com/kitconcept/kitconcept.volto/pull/5

still pending, but we want to push it for the upcoming Volto integration package as default for Volto.

datakurre:
My experiment, which passed the validator, is here https://github.com/datakurre/my-volto-app/commit/89c86ba93db2ad404f45a7dfc5fc9dde686390e6

I don't remember my use of `sizes` any more :see_no_evil:

But I recall that we had to hack Plone image sizes with unlimited heights, e.g.

```
huge 1600:65536
great 1200:65536
large 800:65536
preview 400:65536
mini 200:65536
```

to avoid browser getting scale, which required blocky upscaling :thinking:

sneridagh:
@datakurre thanks a lot!! I searched for it, but I couldn't find it!

nzambello:
Updated the PR:

- Refactor of the new component and the helper method
- Renamed both, now it's Image
- Loaded low resolution placeholder loading afterwards the <source> with other resolutions using the width descriptor making the browser to decide which image is the best match to load based on the actual size of the image on the page

nzambello:
Preview of the loading with the placeholder.
https://user-images.githubusercontent.com/21101435/104717825-37b84c00-572a-11eb-847f-d1dc75237bde.mov

Heavily inspired by Gatsby image

CHANGELOG.md
```diff
@@ -61,6 +61,7 @@
 ### Feature
 
 - Lazy load image in blocks Image and HeroImage @mamico
+- Add ImageFromUrl with srcset and lazy loading using Plone miniatures @nzambello
```
nzambello, line 64:
```suggestion
- Add Image with srcset and lazy loading using Plone scales @nzambello
```

avoinea:
Sizes we use for the EEA website:

```
print 2000:2000
panoramic 1920:1080
landscape 1370:771
portrait 771:1370
xlarge 950:950
wide 325:183
large 768:768
preview 400:400
mini 200:200
thumb 128:128
tile 64:64
icon 32:32
listing 16:16
```

tiberiuichim:
It doesn't seem like the srcset properly responds to the images. I can see the /listing being loaded, but it's always loading the big image instead of one of the smaller sizes. Maybe the `sizes` attribute needs to be specified?

datakurre:
I'm sorry to have not been able to try this pull yet, but reading the Smasing Magazine article and trying out the linter really helped me:

* https://www.smashingmagazine.com/2014/05/responsive-images-done-right-guide-picture-srcset/
* https://github.com/datakurre/my-volto-app/commit/89c86ba93db2ad404f45a7dfc5fc9dde686390e6
* https://ausi.github.io/respimagelint/

So `srcset` could describe all the available versions and their widths, it may not be enough for browser to properly choose the size, because browser might load the image before it knows the eventual size of the rendered image area. That's why `sizes` is needed to tell the browser the maximum size of the area up to each responsive breakpoint.

Also, because sizing is based on widths, Plone image scaling must provide those widths or images may be upscaled on browser. That's how I ended with image sizes like `large 800:65536`.

tiberiuichim:
Thanks @datakurre!

I ran the validator on the PR, on a page with a listing block and a hero block. Here's the things it has to say:

Screenshot 2021-01-23 124120

tiberiuichim:
Isn't the @@images/image/listing size too small? 16px, as is the default, is no bigger then a small icon. Yes, it loads fast, but it's also ugly. I suggest maybe use tile (64px) or thumb (128px) as default?

datakurre:
For me the validator results look like I expected: Plone image scaling is too smart for srcset and that's why widths do not match. My indefinitive heights in image scale forces Plone always to scale by width, what should make it match.That should also prevent cropping mentioned by validator. Also sizes-attribute should be set (and that's the hardest one, because its values probably depend on themes.

tiberiuichim:
@datakurre No worries, I haven't yet plugged the new image scales into Plone, this was with the default image scales. I am trying to get a feeling of what this whole problem space is about. The above comment, with the screenshot, wasn't aimed at you. Just documenting things as I encounter them

tiberiuichim:
@nzambello My hope for this PR is that, in the end, it provides a comprehensive solution for responsive images. On issue I have though, and maybe I have read the code wrong, but it seems to only take Plone image content types into account (so basically, only the "image" field). I think this should be handled better and allow also handling content types with either non-standard image field or multiple image fields.

datakurre:
No problem. For me the key for understanding this was to acknowledge that there is no magic involved, but browser really must be able to decide the optional scale from HTML only.

tiberiuichim:
Another idea: when definining image scales, I think it could be interesting to also define resizing parameters. Looking at [these parameters](https://sharp.pixelplumbing.com/api-resize#resize), for example (most important, fit type and positions.

tisto:
Roadmap:

* [ ] Agree on the current settings
* [ ] Implement current settings in kitconcept.volto
* [ ] Document the image scales Volto expects from the backend
* [ ] Gain experience with the new image sizes in production
* [ ] Write PLIP to update Plone 6 core

src/components/theme/Image/Image.jsx
```diff
@@ -0,0 +1,73 @@
...
+      {!critical && (
+        <noscript
+          dangerouslySetInnerHTML={{
+            __html: `
+              <img
+                src="${src}"
+                ${srcSet?.length && `srcset="${srcSet.join(', ')}"`}
+                alt="${alt}"
+                className="${className}"
+                role="${role}"
```
silviubogan, line 55: In this HTML string I think that attribute values strings should be escaped based on the rules in [this SO answer](https://stackoverflow.com/a/9189067/258462).

src/components/theme/Image/Image.jsx
```diff
@@ -0,0 +1,80 @@
...
+      </picture>
+      {!critical && (
+        <noscript
+          dangerouslySetInnerHTML={{
+            __html: `
+              <img
+                src="${src}"
+                ${srcSet?.length && `srcset="${srcSet.join(', ')}"`}
+                alt="${alt}"
+                className="${className}"
```
nzambello, line 58:
```suggestion
                class="${className}"
```

nileshgulia1 (commented):
@nzambello I did a quick smoke test, didn't got time to review full code yet. The page crashes when adding a new image from an image block. 

Screenshot 2021-11-15 at 4 37 23 PM
Also some of the images don't load due to this, until you do a full page reload. looks like the `source` elements are not being loaded or Idk maybe I'm missing something here....? 
Thank you!

giuliaghisini:
@nzambello it should be updated with the new way of showing images that is now used in the image block

cypress/tests/core/blocks-listing-templates.js
```diff
@@ -48,7 +48,7 @@ describe('Folder Contents Tests', () => {
     cy.url().should('eq', Cypress.config().baseUrl + '/my-folder/my-document');
     cy.get('.listing-item img')
       .should('have.attr', 'src')
-      .and('contain', '/my-folder/my-document/my-image/@@images/image/preview');
+      .and('contain', '/my-folder/my-document/my-image/@@images/image/icon');
```
avoinea, line 51: @nzambello Maybe we should use thumb or something else as there are discussions to remove low resolution scales. @tisto ^

nileshgulia1 (approved):
LGTM 🎉

nzambello:
@plone/volto-team this is ready IMO.
Fixed remaining things, tested in a production environment and updated the documentation.

docs/source/developer-guidelines/images.md
```diff
@@ -0,0 +1,65 @@
...
+| imageField         | string                                |                    | `image` | Image field (see https://github.com/plone/volto/pull/2731)                     |
+| alt                | string                                |                    | `''`    | Alternative text for the image                                                 |
+| className          | string                                |                    |         | CSS class for the image                                                        |
+| containerClassName | string                                |                    |         | Additional CSS classes for the container picture element                       |
+| floated            | `left` or `right`                     |                    |         | Float the image to left or right                                               |
+| size               | `thumb`, `small`, `medium` or `large` |                    |         | Image size (CSS-level, not related to img resolution)                          |
```
sneridagh, line 55: Not related to plone scales? Could be kind of confusing.

src/components/manage/Blocks/HeroImageLeft/View.jsx
```diff
@@ -5,7 +5,7 @@
 
 import React from 'react';
 import PropTypes from 'prop-types';
-import { flattenToAppURL } from '@plone/volto/helpers';
+import Image from '@plone/volto/components/theme/Image/Image';
```
sneridagh, line 8: Not from components?

`import { Image } from '@plone/volto/components`

src/components/manage/Blocks/Image/View.test.jsx
```diff
@@ -4,11 +4,23 @@ import { render } from '@testing-library/react';
...
+  icon: 32,
+  listing: 16,
+};
+
 describe('Image View Component', () => {
   test('renders a view image component with a local image', () => {
     const { getByRole } = render(<View data={{ url: '/image.jpg' }} />);
     const img = getByRole('img');
-    expect(img).toHaveAttribute('src', '/image.jpg/@@images/image');
+    expect(img).toHaveAttribute('src', '/image.jpg/@@images/image/listing');
```
sneridagh, line 23: `listing` scale by default? why is that? that scale is not in the plone.volto ones.

src/components/manage/Blocks/Listing/SummaryTemplate.jsx
```diff
@@ -26,13 +27,20 @@ const SummaryTemplate = ({ items, linkTitle, linkHref, isEditMode }) => {
         {items.map((item) => (
           <div className="listing-item" key={item['@id']}>
             <ConditionalLink item={item} condition={!isEditMode}>
-              {!item.image_field && <img src={DefaultImageSVG} alt="" />}
+              {!item.image_field && (
```
sneridagh, line 30: Wondering if the Image component could take care of the default image as well, passing a prop with the default image to use and use it if the image is not present.

src/helpers/Image/Image.js
```diff
@@ -0,0 +1,88 @@
...
+    imageType = 'internalUrl';
+
+  switch (imageType) {
+    case 'svg':
+      attrs.src = flattenToAppURL(image.download);
+      break;
+
+    // Scales object from Plone restapi
+    // ideal use of Plone images
+    case 'imageObject':
```
sneridagh, line 46: I'm missing the use case more common in blocks. The one that you know the image field (coming from the catalog), and you want to render it, eg. an image in a teaser.

Granted, that is difficult to get right with the srcset, but might be you want it anyways (no srcset info).

Maybe, the possibility to fetch the info from the image object anyways?

src/components/manage/Blocks/Listing/SummaryTemplate.jsx
tiberiuichim: Maybe do it as part of PreviewImage from here? https://github.com/plone/volto/pull/2888

Or maybe push that functionality in this PR?

src/components/theme/Image/Image.jsx
```diff
@@ -0,0 +1,146 @@
...
+    <>
+      <picture className={pictureClassName}>
+        {srcset?.length > 0 && <source srcSet={srcset} />}
+        <img
+          src={src}
+          alt={alt}
+          className={className}
+          role={role}
+          loading="lazy"
+          style={{ width: '100%', objectFit: 'cover' }}
```
pnicolli, line 99: Why the default styles? These look quite opinionated, even if they can be overriden with the `imageProps` prop. Is there a reason for these two styles?

nzambello: If I remember it correctly, I set those inline to avoid layout shifts while loading CSS.

docs/source/developer-guidelines/images.md
```diff
@@ -0,0 +1,65 @@
+# Images
+
+Images are rendered with the Volto component `Image`.  
+It renders images using Plone scales from [plone.app.imaging](https://github.com/plone/plone.app.imaging) and handles:
```
mauritsvanrees, line 4: Minor documentation comment: `plone.app.imaging` was only used for Archetypes, so not relevant here. Scales for dexterity are done in `plone.namedfile`, so the link should point there.

MrTango:
I'm not sure if understand the code right, but you did not seem to define sizes right?
I think it would be better, so that the browser knows the image size before it is loaded, even more so when using lazy loading.
I'm currently trying to solve this for classic UI.
On one hand enabling the browser to automatically select the right image scale is awesome.
But i would like to also allow art direction.
In the form of providing different cropped versions for smaller screens, like i could create via plone.app.imagecropping.
Art direction is only possible via source tags in combination with media queries.  
This is more explicit and you can force the browser to choose e special scale which contains your cropped image.
With srcset and sizes the browser will choose what ever fit's in the moment, even a bigger scale, when the bigger scale is already in the cache. This way you might see the bigger scale on small screens which is fine from the quality point of view, but might not what you want. The bigger scale does not have the crop and looks bad on mobile.
Here is a good example of that problem and the solution: https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images

The challenge now is to find the right amount of flexibility without making the configuration to complex.
I would like to exchange ideas on that topic.
I'll join the team in Bucharest for the Beethoven Sprint, i hope we can make some progress then.

giuliaghisini:
> I'm not sure if understand the code right, but you did not seem to define sizes right? I think it would be better, so that the browser knows the image size before it is loaded, even more so when using lazy loading. I'm currently trying to solve this for classic UI. On one hand enabling the browser to automatically select the right image scale is awesome. But i would like to also allow art direction. In the form of providing different cropped versions for smaller screens, like i could create via plone.app.imagecropping. Art direction is only possible via source tags in combination with media queries. This is more explicit and you can force the browser to choose e special scale which contains your cropped image. With srcset and sizes the browser will choose what ever fit's in the moment, even a bigger scale, when the bigger scale is already in the cache. This way you might see the bigger scale on small screens which is fine from the quality point of view, but might not what you want. The bigger scale does not have the crop and looks bad on mobile. Here is a good example of that problem and the solution: https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images
> 
> The challenge now is to find the right amount of flexibility without making the configuration to complex. I would like to exchange ideas on that topic. I'll join the team in Bucharest for the Beethoven Sprint, i hope we can make some progress then.

@MrTango 
image sizes are defined here https://github.com/plone/volto/blob/3660875628a9c721994a7de4a10309ad5735f1d9/src/config/ImageScales.jsx and inject in config here https://github.com/plone/volto/blob/3660875628a9c721994a7de4a10309ad5735f1d9/src/config/index.js#L117.

Browser try to determine the size of image here https://github.com/plone/volto/blob/3660875628a9c721994a7de4a10309ad5735f1d9/src/components/theme/Image/Image.jsx#L65 and loads the src-set maximum image size based on available width and pixel ratio.

In Volto we don't have yet the concept of 'cropped image' only the scaled image. 
I think crop will be a next feature and is not a theme to take care now on Volto.

MrTango:
Do you set the sizes attribute somewhere?
This is important for the browser to know how big the image will be on the screen.
This will be used to calculate the needed image size by looking at the screen resolution (2x,3x aso) and the [Intrinsic_Size](https://developer.mozilla.org/en-US/docs/Glossary/Intrinsic_Size) of the image.

For example:
```
<picture class="image-size-large">
  <source
    srcset="https://picsum.photos/id/1011/800 800w,
            https://picsum.photos/id/1011/1000 1000w,
            https://picsum.photos/id/1011/1200 1200w,
            https://picsum.photos/id/1011/1600 1600w"
    sizes="800px"
    >
  <img src="https://picsum.photos/id/1011/800" sizes="800px">
</picture>
```
The image intrinsic size of the image is 800px. On a older 1x desktop screen that means the browser might prefer the image scale: https://picsum.photos/id/1011/800, but on a high dpi screen let's say 2x, the Browser would require the scale https://picsum.photos/id/1011/1600 1600w

If you don't provide the sizes attribute, the intrinsic size has to be calculated from CSS definitions, that takes more time.

PS. I'll be at the sprint in Bucharest the following days, let me know if we could sit together to discuss this and also my ideas for the srcset config in the backend. I made the default config much smaler now, but it is still valid to have it and gives us to the option to improve the frontends in time more.

datakurre:
@giuliaghisini Does those w-sizes really work?

Referring my original POC two years ago: https://github.com/datakurre/my-volto-app/commit/89c86ba93db2ad404f45a7dfc5fc9dde686390e6

`src-set` should really just list all the available images and their real pixel width

then

`sizes` attributes is there for telling the view port size; it supports responsive breakpoints and CSS math

Just with these in place, browser will pick whatever size it sees best fit. I didn't even see `2x` or `3x` src-set items to be really required, because browser was able know its pixel density and choose the right image, as long as one was available.

My only problem was that Plone image scaling choose the optimal side of the image for scaling. That's why I practically disabled scaling by height with settings these weird sizes

```
huge 1600:65536
great 1200:65536
large 800:65536
preview 400:65536
mini 200:65536
```

With these sizes in place, I know that image scaling always scaled by width,  src-set was giving correct information and browser was able choose the right size.

`sizes` attribute, of course, is more tricky, because its optimal value really depends on your theme.

datakurre:
Oops... deleted my comment regarding w-sizes, because it was me, who was originally proposing those. I'll re-read my original PoC and comment back why did they work :)

datakurre:
@giuliaghisini I'm sorry for my previous comment (I now delete) that made no sense.

Regarding `sizes`, could `sizes="100vw"` be better than nothing? At least then browser would never choose a size greater than the window itself.

giuliaghisini:
> @giuliaghisini I'm sorry for my previous comment (I now deleted) that made no sense.
> 
> Regarding `sizes`, could `sizes="100vw"` be better than nothing? At least then browser would never choose a size greater than the window itself. (Or does browser act like that by default?)
> 
> Or does it already work properly? Just loading listing size at first and the correct scale when CSS has been calculated?
> 
> I recall that my old PoC [datakurre/my-volto-app@89c86ba#diff-3f5fdd396d4f51d380bccc043229925fc08f0b5c5bffa414d8e4e1cf9325e55bR19](https://github.com/datakurre/my-volto-app/commit/89c86ba93db2ad404f45a7dfc5fc9dde686390e6#diff-3f5fdd396d4f51d380bccc043229925fc08f0b5c5bffa414d8e4e1cf9325e55bR19) only loaded a single scale, but I did not try that with `loading="lazy"` 🤔

yes, it could be better than nothing.. it's the sizes of the 'placeholder' slots into which the browser can insert images from the srcset.. 
Maybe 100vw will be too much, it will cause some 'flickers' area in page when real image is loaded.. i think.. we have to try...

Maybe sizes could be setted in applySrcSet fn where we know the real image size.. but i don't know it could be correct.. 
what do you think?

giuliaghisini:
@datakurre @MrTango i've added sizes="100vw" as default and make some improvements. 
Now i think it's working much better than before..
Could you try it? 
for sizes.. i think we cannot calculate them.. it depends on page layout.. i've added sizes as props to pass to the Image component and setted default to '100vw'

datakurre:
@giuliaghisini Did you try https://ausi.github.io/respimagelint/ already?

datakurre:
I think sizes as a prop is good start. Once someone is really building responsive theme with optimal sec-sets, that someone probably gets an idea, how this should be enhanced.

MrTango:
>>> giuliaghisini

> we didn't set the real size of image in attributes because we don't know the real size of image here, we only know the miniature size.. this is the big problem of this implementation and need to think about it..

while defining the srcset we give all the needed info's by appending the size like `800w` of that scale.
The sizes attribute, is not only useful together with media queries, if you leave them of as i did in my example, you go just with the fallback size of 800px. This size is the size the image would take on the screen, no matter of the screen resolution.

On a 1x resolution screen it will be 800px as well as on a 2x resolution screen.
Think of it as the physical size of the image.

> Another problem that we have here, is that srcset is injected in page only when image meets intersection observer.. it was a way to do the lazy loading of image... but i don't think it's the best way to do it, because this causes two image loading for one image:
> 
>     the image set in src attribute of (/huge.jpg/@@images/image/listing)
>     and the needed image size from srcset.

i don't why you think you need this, if you add the loading="lazy" attribute, the browser should be able to load lazy. Right now chrome will still load to fast all images, firefox and safari will save more data and only load images when you are getting close to them.
So with that in mind, if you want the optimum control and have it working on chrome you have to do it with intersection observer.

https://www.ctrl.blog/entry/lazy-loading-viewports.html

For Plone 6 classic UI we will probably stick with the browser defaults and just use loading="lazy".
Chrome will eventually fix the behavior.

I guess when you add the srcset and the src attributes via the observer it should work, the browser can't load anything without either a src or a srcset.




>>> datakurre:
> src-set should really just list all the available images and their real pixel width
> 
> then
> 
> sizes attributes is there for telling the view port size; it supports responsive breakpoints and CSS math
> 
> Just with these in place, browser will pick whatever size it sees best fit. I didn't even see 2x or 3x src-set items to be really required, because browser was able know its pixel density and choose the right image, as long as one was available.

not quite, the sizes attribute with out a media query defines the Intrinsic Size of the image, the same you can archive by setting the width attribute on the img tag. But it is true that the Browser is able to do the math for us, by calculating Intrinsic Size * screen resolution >> needed image size.



> 
> With these sizes in place, I know that image scaling always scaled by width, src-set was giving correct information and browser was able choose the right size.
> 
> sizes attribute, of course, is more tricky, because its optimal value really depends on your theme.

it depends if you want to change the Intrinsic Size for different viewports, most of the time that not needed.
With my example above, the image is 800px on a Desktop. For mobile that resolution works too.
As we need a bit less than 400px usually * screen resolution (somewhere between 2x and 4x), leaves us with needed sizes of (800px up to 1600). Together with the general definition of `img {max-with: 100%;}` it only make sense when the image is smaller on Desktop than on a smaller screen. 

For example if our image is 250px and we want it to be almost full screen on a mobile device.  
Than something like this make sense:
`sizes="(max-width: 768px) 90vw, 250px"`.

In most cases it's not worth optimizing to much.

https://css-tricks.com/a-guide-to-the-responsive-images-syntax-in-html/

MrTango:
also one thing i noted while playing around with [my example page](https://mrtango.github.io/responsive-images-example/) is, that we need to set the height, when we are using lazy loading or everithing jumps arround :dancers: 
I hope we can get together tomorrow on discord for a chat.

giuliaghisini:
> also one thing i noted while playing around with [my example page](https://mrtango.github.io/responsive-images-example/) is that we need to set the height when we are using lazy loading or everything jumps arround 👯 I hope we can get together tomorrow on discord for a chat.

sure! we could talk about this on discord. 
i'm not seeing some 'jumping' since yesterday afternoon improovments... but we could take a look together !

giuliaghisini:
@MrTango i've investigated on 'jumping' images... and... i found that maybe it's to the miniature fault....
I'll explain..
Until real image to display is loaded, the image with the lowest scale among those configured in the config in _config.settings.imageScales_ is loaded as placeholder.
In our case the smallest scale is 'listing' which has a width of 16px.
For example, my image have a listing miniature that is 16x10px
and the actual image will be shown by the srcset is 833x555px.
The proportions of the two images are not the same, in fact:
16x10px proportion is 1,6:1
and 833x555px proportion is 1,5:1

this is the reason for the 'jumpings' i think..
and these images are coming from plone that maybe rounded x or y dimensions when scaling...

MrTango:
The jump will always happen when the browser does not have a hight to start with, before it lazy loads the image.
So i think we should set width and height attributes on the img tag.

giuliaghisini:
> The jump will always happen when the browser does not have a hight to start with, before it lazy loads the image. So i think we should set width and height attributes on the img tag.

it could be a solution.. but we don't know the real image height..

pnicolli:
Image scaling discussion notes (Beethoven sprint 2022):

- we need to render images with explicit width and height attributes on the html elements
- we need to have scales info in volto somehow in order to do that, because we don't have that info from restapi in listings (querystring-search), for example
- current idea is to add an expander to get scales information (urls and sizes) for the images of the items returned
- how do we handle preview_image against image_field? We could actually move the logic to the new restapi expandable service and do that by default, and simplify the frontend
- keep in mind that images sometimes have copyright info, so maybe add those info to the backend service, or add a different expander. This can be discussed later on.
- remove loading="lazy" from the placeholder image https://github.com/plone/volto/blob/add_srcset_imgs/src/components/theme/Image/Image.jsx#L116
- keep using IntersectionObserver instead of loading="lazy" because Chrome behaves badly (loads images way too soon)
- remove the style attribute when we add the explicit width and height
- there is a PR in plone.base (https://github.com/plone/plone.base/pull/5) to add a control panel to handle image scales properly in plone 6, discussion is happening there for that point. When that is finished, a dedicated controlpanel can be added to Volto

tisto:
@MrTango @giuliaghisini modern browsers allow you to set the width and height params without "px", so you basically just set the aspect ratio. This allows the browser to fully render the image properly (avoid jumps) and at the same time provides full flexibility of the srcset attribute:

https://www.smashingmagazine.com/2021/04/humble-img-element-core-web-vitals/#the-basics

MrTango:
Today's testing showed that it does not matter much, what size we are putting into width/height attributes.
As long as they have in the same aspect ratio.
So width="1" and height="1" will do the same as width="300" and height="300", together which some CSS to have the image width at 300px. Both "px" and "vw" work well in CSS, but "%" does not, image are jumping when using "%" for width.

So as long as we set width and height with the correct aspect ration, we can change it via CSS width as we want to.
It event works well with max-width="100%", the Browser will recalculate continuously and load images from the provides list of options in the srcset attribute.

pnicolli:
Superseded by #3337

PR #2118, opened 2021-01-13
'All' button in contents and batch size

nzambello:
As it's now, we can't fetch all the contents from a query with the All button, if this is what it's meant for.
Adding b_size=100000000 to the querystring, restapi can return all the contents of that query (fake-ish but working).


giuliaghisini (approved):

tiberiuichim:
This should be batched, though, right?

Batching also implies a sorting order and ES, for example, gives you a unique id for a search result, to be able to continue the batch. That may be overkill for Volto, but that's correct, otherwise things are not safe.

sneridagh:
@tiberiuichim you mean batched, the 1000000 records shown?

tiberiuichim:
@sneridagh Nevermind, I was wrong, I was confusing things.

Issue #2120, opened 2021-01-13
More action is broken with Plone 4 backend


avoinea:
Clicking the `More` action button within Toolbar will raise the following error,

![2021-01-13-194033_1052x287_scrot](https://user-images.githubusercontent.com/323385/104488609-4b06d280-55d7-11eb-8f44-b0e9756fa141.png)

PR #2121, opened 2021-01-13
Fix broken More menu when no history action

avoinea:
Refs #2120

### DIFF
--- a/cypress/integration/history.js
+++ b/cypress/integration/history.js
@@ -1,0 +1,39 @@
+describe('History Tests', () => {
+  beforeEach(() => {
+    // give a logged in editor and the site root
+    cy.autologin();
+    cy.visit('/');
+    cy.waitForResourceToLoad('@navigation');
+    cy.waitForResourceToLoad('@breadcrumbs');
+    cy.waitForResourceToLoad('@actions');
+    cy.waitForResourceToLoad('@types');
+    cy.waitForResourceToLoad('');
+  });
+  it('As editor I can add a page and access history', function () {
+    // I add a page
+    cy.get('#toolbar-add').click();
+    cy.get('#toolbar-add-document').click();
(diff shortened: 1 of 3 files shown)


cypress/integration/history.js
```diff
@@ -0,0 +1,39 @@
+describe('History Tests', () => {
+  beforeEach(() => {
+    // give a logged in editor and the site root
+    cy.autologin();
+    cy.visit('/');
+    cy.waitForResourceToLoad('@navigation');
+    cy.waitForResourceToLoad('@breadcrumbs');
+    cy.waitForResourceToLoad('@actions');
+    cy.waitForResourceToLoad('@types');
+    cy.waitForResourceToLoad('');
```
tiberiuichim, line 10: I wonder if this does anything (I think I've introduced this style when I got rid of ?fullobjects).

sneridagh: I hope it does!

src/components/manage/Toolbar/More.jsx
```diff
@@ -126,30 +126,18 @@ class More extends Component {
               {editAction && <Display pathname={path} />}
             </li>
             <li>
-              {historyAction ? (
```
sneridagh, line 129: I think we need the history conditional for not showing it in the root, isn't it?

sneridagh (changes requested):
I'll leave the conditional for the root use case, where it seems weird to show it, at least until we have DX-based roots...

src/components/manage/Toolbar/More.jsx
avoinea: @sneridagh The problem is that in Plone 4 `historyAction` is undefined everywhere, but `/history` is a valid path.

sneridagh:
@avoinea I see, I just checked in the current version and we are showing it at root (I thought we didn't).
Anyways, LGTM!

sneridagh (approved):

PR #2124, opened 2021-01-14
Improve CSS in the inner toolbar for the image block to support narro…

sneridagh:
…wer width (like for using it inside grid blocks)

### DIFF
--- a/theme/themes/pastanaga/extras/blocks.less
+++ b/theme/themes/pastanaga/extras/blocks.less
@@ -304,7 +304,7 @@ body.has-toolbar.has-sidebar-collapsed .ui.wrapper > .ui.inner.block.full {
 
   .toolbar-inner {
     display: flex;
-    width: min(100%, 450px);
+    width: ~'min(100%, 450px)';
     height: 40px;
     padding: 4px;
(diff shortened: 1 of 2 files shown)


sneridagh:
@avoinea @tiberiuichim you'll like this one :)

avoinea:
@sneridagh Cool. Still I get the following error within a volto project, not sure if related or I'm doing something wrong:

2021-01-14-114003_1022x201_scrot

theme/themes/pastanaga/extras/blocks.less
```diff
@@ -304,7 +304,7 @@ body.has-toolbar.has-sidebar-collapsed .ui.wrapper > .ui.inner.block.full {
 
   .toolbar-inner {
     display: flex;
-    width: 450px;
+    width: min(100%, 450px);
```
avoinea, line 307: @sneridagh The issue reproduce also on Volto core. This may help https://stackoverflow.com/questions/64255505/less-js-error-evaluating-function-min-incompatible-types

sneridagh:
@avoinea yes, less is picky with this...

PR #2126, opened 2021-01-14
Fix regression in toolbar inner CSS for image blocks

sneridagh:


sneridagh:
I went too quick without taking an actual look at the thing in wide screens :')

avoinea:
Love it! 
2021-01-14-145038_1224x878_scrot

avoinea (approved):

PR #2128, opened 2021-01-15
Use public/critical.css as critical CSS in production

tiberiuichim:


src/server.jsx
```diff
@@ -224,6 +226,9 @@ server
                     markup={markup}
                     store={store}
                     extractScripts={process.env.NODE_ENV !== 'production'}
+                    criticalCss={readFileSync('public/critical.css', {
```
tiberiuichim, line 229: Test for css file existance.

tiberiuichim (commented):
@silviubogan Please make the required changes according to comments.

public/critical.css
```diff
@@ -0,0 +1,6 @@
+/* TODO: generate this file automatically with
```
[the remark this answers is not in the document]
silviubogan: Is the placement of `critical.css` file good, in dir `public`? Should I also add it to .gitignore?

silviubogan:
@tiberiuichim Should we also allow for an `uncritical.css` file if a `critical.css` file exists? I think that this change would practically make it possible to replace all the CSS in Volto.

nileshgulia1:
I think for the uncritical part we should allow as we are handling now, `rel = 'preload'` preloading them when needed. I was wondering how will prevent duplication of CSS when `above the fold content` has already been downloaded. or its worth to do an extra call to server. see: https://css-tricks.com/authoring-critical-fold-css/#an-automated-approach

src/critical-css.js
```diff
@@ -0,0 +1,20 @@
+import { existsSync, lstatSync, readFileSync } from 'fs';
```
tiberiuichim, line 1: Right now the module does too little to stand on its own (I'd move the code to server.jsx). Maybe we can cache the css content (if found) for 5 minutes? Then it will be ok to be standalone.

silviubogan: I added the cache feature.

tiberiuichim:
Closed in favour of #2136

PR #2132, opened 2021-01-15
added og tags for social sharing

giuliaghisini:
Added og tags for social sharing


src/components/theme/View/View.jsx
```diff
@@ -229,6 +230,27 @@ class View extends Component {
...
+                this.props.content.image?.scales?.large?.download,
+              )}
+            />
+          )}
+          {this.props.content.description && (
+            <meta
+              property="og:description"
+              content={this.props.content.description}
+            />
+          )}
```
nzambello, line 253:
```suggestion
          <meta property="og:title" content={this.props.content.title} />
           {this.props.content.description && (
            <meta
              property="og:description"
              content={this.props.content.description}
            />
          )}
          <meta
            property="og:url"
            content={toPublicURL(this.props.content['@id'])}
          />
          {this.props.content.image?.scales?.large?.download && (
            <meta
              property="og:image"
              content={toPublicURL(
                this.props.content.image.scales.large.download,
              )}
            />
          )}
          <meta name="twitter:card" content="summary_large_image">
```

nzambello (changes requested):
Could you add some tests for https://github.com/plone/volto/blob/0569cea9ee59b52fb5d38c4a20762ad5ce73c644/src/helpers/Url/Url.js?

src/components/theme/View/View.jsx
```diff
@@ -229,6 +230,27 @@ class View extends Component {
           )}
           <title>{this.props.content.title}</title>
           <meta name="description" content={this.props.content.description} />
+
+          <meta property="og:title" content={this.props.content.title} />
+          {/* <meta property="og:type" content="" /> */}
```
nzambello, line 235:
```suggestion
```

src/components/theme/View/View.jsx
nzambello, line 253:
```suggestion
          )}
          <meta name="twitter:card" content="summary_large_image">
```

nzambello (changes requested):

src/components/theme/View/View.jsx
nzambello: https://css-tricks.com/essential-meta-tags-social-media/

tisto:
@giuliaghisini @nzambello we have to take into account that there might be add-ons (like kitconcept.seo) that override the og tags (and allow to set a different title) and that might collide with those tags in Volto core.

src/components/theme/View/View.jsx
```diff
@@ -229,6 +230,27 @@ class View extends Component {
           )}
           <title>{this.props.content.title}</title>
           <meta name="description" content={this.props.content.description} />
+
+          <meta property="og:title" content={this.props.content.title} />
```
tisto, line 234: Isn't that something that the editor would like to override? I would guess that if you embed a page without this tag, that there is a fallback for the site title anyways (haven't checked).

src/components/theme/View/View.jsx
```diff
@@ -229,6 +230,27 @@ class View extends Component {
...
+          {/* <meta property="og:type" content="" /> */}
+          <meta
+            property="og:url"
+            content={toPublicURL(this.props.content['@id'])}
+          />
+          {this.props.content.image?.scales?.large?.download && (
```
tisto, line 240: Where does the image come from? Isn't that just empty for most content types?

src/components/theme/View/View.jsx
nzambello: AFAIK we currently don't have access to the site title

src/components/theme/View/View.jsx
nzambello: If present (it's the lead image) we have the image here
We could consider a fallback but I don't know what we could use

nzambello:
> @giuliaghisini @nzambello we have to take into account that there might be add-ons (like kitconcept.seo) that override the og tags (and allow to set a different title) and that might collide with those tags in Volto core.
> 
> Just wondering. Did you consider using kitconcept.seo for your use case? It offers all the social tags and allows to the editor to actually override the settings (with fallbacks when you don't add anything).

@tisto I saw kitconcept.seo just recently. I think we should ship something on Volto by default from standard fields and let developers override that with additional fields from that addon, and I don't see blockers for this.

tisto:
@nzambello not sure I fully get your point. Though, if we add meta tags in the header and then an add-on needs to override those fields this is a lot harder to accomplish than coming up with a plugin architecture that allows add-ons to add additional meta tags to the header.

We should be careful not to overengineer stuff here. You either need SEO capabilities, then you should use an add-on that allows editor to override settings, or you don't (for an intranet use case for instance). I am ok discussing bringing SEO capabilities into Volto core. Though, then we should go all the way (and add backend fields as well). Mixing things up and providing all the options that one could imagine introduces an overhead that has no value for us IMHO.

What we could do is to add those meta fields to the Volto templates and gracefully degrade them. If you have a SEO add-on with additional fields, Volto takes those values, if you don't it falls back to the page title/description/ etc (exactly what this PR does). Though, this is actually what kitconcept.seo already does. See:

https://github.com/kitconcept/kitconcept.seo/blob/master/README.rst#add-to-volto-site

tisto:
@jackahl can you share our template from the DVV project here? I have the feeling we are re-inventing the wheel here...

src/components/theme/View/View.jsx
```diff
@@ -229,6 +230,27 @@ class View extends Component {
...
+            <meta
+              property="og:image"
+              content={toPublicURL(
+                this.props.content.image?.scales?.large?.download,
+              )}
+            />
+          )}
```
nzambello, line 246:
```suggestion
          {this.props.content.image?.scales?.large?.download && (
            <>
              <meta
                property="og:image"
                content={toPublicURL(
                  this.props.content.image?.scales?.large?.download,
                )}
              />
              <meta
                property="og:image:width"
                content={this.props.content.image?.scales?.large?.width}
              />
              <meta
                property="og:image:height"
                content={this.props.content.image?.scales?.large?.height}
              />
            </>
          )}
```

src/components/theme/View/View.jsx
tiberiuichim: Can we make a configurable utility?
```
export default applyConfig(config) {
    const {extractOpenGraph} = config.settings;

    config.settings.extractOpenGraph = (data) => {
        if (data['@type'] === 'Contact']) {
            return {'og:image': data['portrait']........}
        }
       return extractOpenGraph(data);
    }
}
```

PR #2136, opened 2021-01-19
Critical css

tiberiuichim:


tiberiuichim:
@silviubogan In the jest tests the __SERVER__ variable is not defined, so we need to guard against that.

tiberiuichim:
### After applying critical.css
Screenshot 2021-01-19 213108

### No critical.css applied
before-critical

tiberiuichim:
That was for the Desktop device in Lighthouse. For the mobile device, Volto master get score 18, with the critical.css I get 55 as score. I've used volto.kitconcept.com to extract the critical.css and this tool: https://github.com/eea/critical-css-cli

tiberiuichim:
I think the fonts need to have a fallback, maybe it's the optimizer that's stripping it, I need to investigate.

tiberiuichim:
### Lighthouse score without critical.css, volto master

non-critical-lighthouse-scores

### Performance graphs without critical.css, volto master

non-critical-performance

### Lighthouse score with critical.css

critical-lighthouse-scores

### Performance graphs

critical-performance

This was tested on my local machine. With deployment environments (gzip enabled) the score will go up for both cases.

PR #2148, opened 2021-01-22
Adding classname in TextWidget and objectBrowserBody so that we can t…

iFlameing:
…arget those element in tests


sneridagh:
@iFlameing I think we should give that classname something more meaningful, since it's the button, we cannot refer it as "field-X", since it's not the field itself, it's the button of the assigned action, right?

sneridagh (changes requested):
See comment

iFlameing:
@sneridagh done :)

PR #2152, opened 2021-01-22
"seamless" mode for API

sneridagh:
The idea is to host both Volto and API under the same URL. See below: https://github.com/plone/volto/pull/2152#issuecomment-766110431


sneridagh:
I made it work even with a plain Nginx:

```
   location ~ / {
            expires +1y;
            proxy_pass http://volto;
        }
            expires +1y;
            proxy_pass http://volto;
        }

        if ($http_accept = 'application/json') {
           rewrite ^(.*) /VirtualHostBase/https/myservername:443/Plone/VirtualHostRoot$1 break;
           proxy_pass http://api;
           break;
        }

        proxy_set_header        Host $host;
        proxy_set_header        X-Real-IP $remote_addr;
        proxy_set_header        X-Forwarded-For $proxy_add_x_forwarded_for;
        proxy_set_header        X-Forwarded-Proto $scheme;
        proxy_redirect http:// https://;
        proxy_pass http://volto;
    }

```

sneridagh:
How come that we never reached to this conclusion? Now that's on the table, it seems to me so "natural"... is there any drawback we cannot think of?

Having in production this kind of setup, where both Volto and the API lives under the same sun. The content negotiation stays, so any API call is made with the `Accept` header, so it's routed to the backend.

Pros:
- All internal links are app ones, so a link to a page /my-page will be returned by the API as is. So bye bye to `flattenToAppURL`.
- As the components does not need to flatten anything, nor know which the apiPath is, then lots of imports to ~/config are spared, so less intricacy that will lead to less circular dependencies problems...
- ...

Cons:
- ...

/cc @plone/volto-team

tiberiuichim:
If the setup is like:

```
nginx -> varnish ---> volto
                            \-> Plone
```
Varnish will need to be configured to differentiate between requests by inspecting accept headers. Probably doable. @avoinea ?

sneridagh:
Fixed the problems with the file resources middleware. Merged with the PR with the publicURL introduction, make them both work together.

sneridagh:
closing in favor of: https://github.com/plone/volto/pull/2320

PR #2154, opened 2021-01-24
WIP: Image proxy

tiberiuichim:
This PR replaces the default `@@images` Express "proxy" with one that does the image processing in the Volto server. Here's the overall architecture:

- it's extending the `@@images` middleware and adds an `ImageProxy` representation for images
- it uses 2 caches: one for "originals" and one for "processed" (thumbnails)
- the first thing it does, always, is to make a HEAD call to the "context url" and based on the received `etag` (or `last-modified`, if etag doesn't exist), it computes cache keys
- if a request comes from a thumbnail (`@@images/imagefield/somesize`), it tries to find it in the processed cache. If it can't find it, it tries to find the "original" (the `@@images/imagefield` image data) in the "originals" cache. If it's not there, it fetches it from the backend and stores that in the originals cache, resizes a thumbnail, stores it in the "thumbnails" cache.
- in case of non-thumbnail, there's still an image processing done, to optimize the output files size

I've based the PR on @nzambello's srcset PR, as that one has image size definitions in Volto, but other than that it can fit into the current Volto.

There are several improvements that I'd like to make:

- introduce a flag in the Image component, to signal that non-lossy format should be kept. Right now this PR transforms images only in their original format, so if the backend provides a PNG file, the thumbnail image will still be a PNG file. If an original image is non-lossy, then no image processing will be done on them, as there's no optimization possible.
- make the Image component smarter and introduce webp formats in the srcsets, have the middleware detect webp parameter from the request
- maybe allow specifying some more image resize parameters for the image sizes (like quality, maybe), so you could, for example, have a higher or lower quality on some of the sizes, or specify a different cropping algorithm. Making this configurable in Volto means that frontend developers are free to come up with extra image sizes, fit for their custom components and not care too much about side-effects (blobs in the Plone backend will be only the originals). Too much flexibility in reading parameters from the URL is not good, though, as it opens the proxy to DOS attacks.
- I've mentioned the two caches that are used by default. They can be configured from the projects and we're using the Keyv library as an adapter, so it's really flexible and easy to come up with new implementations. Already existing cache storage adapters include redis, for example. The default caches use QuickLRU, which  uses memory to store the files. I'd like to write a new adapter on top of QuickLRU which stores data in the filesystem and use that for the "originals" cache.


src/express-middleware/image-proxy/file-cache.js
```diff
@@ -91,7 +91,8 @@ class FileCache {
     try {
       const dir = path.join(__dirname, `${key}`);
       const metadataPath = `${dir.replace(path.extname(dir), '.metadata')}`;
-      fs.outputFileSync(dir, data.data);
+      const imageData = data.data.replace(/^data:image\/\w+;base64,/, '');
+      fs.outputFileSync(dir, imageData, { encoding: 'base64' });
```
tiberiuichim, line 95: Please don't write files encoded in base64. Please write them in the original binary format. No processing needed. If outputFileSync doesn't allow you to write binary files, find something else.

sneridagh:
Closing this PR and marking as "Abandoned". 
This feature became uninteresting for the community, since we pushed for a more tight and performant integration with Plone backend image handling. However, I do think it's interesting to have in mind this use case.
The enhanced `Image` component was merged in Volto 17 already.
If someone is interested in rescuing it, you can take it as a reference and open a new PR.

PR #2157, opened 2021-01-25
added Linkify library to parse emails and links in text into clickabl…

giuliaghisini:
…e links

If you insert a mail or link in text, without adding manually a link, link with mailto is auto generated for email addresses and simple link is generated for link.


tiberiuichim (changes requested):
I think it's not optimal from several points of view:

- this is code that needs to run every time on all text, so it might represent a performance hit
- the task it does (link detection) can be implemented in the richtext editor (and lazy loaded), so that it runs only once to detect and create draft links
- I'm concerned about raising the bundle size by adding libraries that absolutely have to be present in the main bundle.

giuliaghisini:
I have completely changed the approach to create automatically links when writing text. 

I've added a plugin to draftjs that detects links and mail-addresses and automatically creates LINK entity in draftjs, that will be converted into links by UniversalLink in ToHTML.jsx. 

I've also improved the UniversalLink component, to allways add target _blank to external links if openLinkInNewTab property is unsetted.

@tiberiuichim @sneridagh @nzambello

tiberiuichim (approved):

nzambello (approved):

giuliaghisini:
@tiberiuichim @sneridagh moved to draft. 
There's a conflict with existing links.. trying to understand..

giuliaghisini:
> @tiberiuichim @sneridagh moved to draft.
> There's a conflict with existing links.. trying to understand..

understood and fixed. Now it's working

sneridagh (approved):

Issue #1914, opened 2020-10-18
Configurable redux reducers dumped to global window.__data


tiberiuichim:
The size of the HTML generated is increased in some cases because the redux store is dumped to the window. I think it should be configurable, in my case, I don't see an optimization, my components behave on the client side like there hasn't been any data provided to them serverside. Probably I have to async-connect them or who knows what else, but in the end what I end up with is huge HTMLs (around 1MB) with no visible benefits to this.

tiberiuichim:
I'm now dumping only `userSession` and `router` to `window.__data`, seems ok so far.

PR #2158, opened 2021-01-26
supply configurable initial reducers in SSR

nileshgulia1:
so that we can control what preloaded state we want in our initial client app. #1914


tiberiuichim:
This is the only place where they should be filtered: https://github.com/plone/volto/blob/147b3cd1d6b3b326cd7225b132f214c9bb3670a3/src/helpers/Html/Html.jsx#L151

The big concern is about the data that's transfered from the server to the browser. See this place for a starting point: 
https://github.com/eea/volto-bise/blob/master/src/customizations/volto/helpers/Html/Html.jsx#L14

Though we want to have these reducers configurable, so that part is ok (though those reducer names are not part of serverConfig). Right now that optimization is disabled in BISE, see this commit:
https://github.com/eea/volto-bise/commit/001728a2e4a39114017a7c3d570d050ef882da8a#diff-eab43a4915429976633277419de252a8790fe08898718804e0c3c62b95ebcf37

sneridagh (approved):

PR #2179, opened 2021-02-01
Show toolbar menu label with general font @fontName

ksuess:
Willkommen_bei_IGIB_GRIF

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 ### Internal
 
 - Translation de: aria-label "Contents". "Inhalte" not "Inhaltsverzeichnis" @ksuess
+- show toolbar menu label with general font @fontName @ksuess
 
 ## 11.0.0 (2021-01-29)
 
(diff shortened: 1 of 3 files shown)


tiberiuichim (approved):

sneridagh (commented):
I have to differ on this one... the font in the CMSUI should be (at least) different from the main theme one. One could make it the same, but then the config must reflect it.

We had projects that the toolbar (and the sidebar) looked particularly ugly because the theme font was ugly... same applies for the sizing, etc 

The CMSUI should be able to stand out from the theme font settings. Quanta will use this approach.

What I would do now: Create a new variable in the pastanaga theme `@fontName-CMSUI` then make the toolbar (and sidebar) use it. If possible I would also do the same for the size.

ksuess:
@sneridagh Yes, the font and most of all the font-size of the CMSUI should be different from the main font and font-size. I will close this PR. 
My current project has font-size 19px. Imagine how the admin views like dropdown menu manager look like!
It was just a quick fix, cause the situation now is not uniform for CMSUI **labels that are obvious** like the 'more' menu. 

Maybe a sprint topic to clean up all ...

PR #2183, opened 2021-02-02
robots.txt from plone as default 

giuliaghisini:
robots.txt in /public folder wins.
If /public/robots.txt doesn't exists:
 - uses VOLTO_ROBOTSTXT env variable if is defined
 - else get robots.txt from plone (${settings.apiPath}/robots.txt)


src/helpers/Url/Url.js
```diff
@@ -93,6 +93,20 @@ export function flattenToAppURL(url) {
...
 }
 
+/**
+ * Given a URL if it starts with the API server URL
+ * this method removes the /api or the /Plone part.
+ * @method toPublicURL
+ * @param {string} url URL of the object
+ * @returns {string} public URL
+ */
+export function toPublicURL(url) {
```
tiberiuichim, line 103: Could be a separate setting. It's already added as a configurable in Plone via kitconcept.volto, right?

tiberiuichim (approved):
I like it, it makes sense.

src/helpers/Robots/Robots.js
```diff
@@ -0,0 +1,37 @@
...
+/**
+ * Generate robots. Get robots from plone
+ * @function generateRobots
+ * @param {Object} req Request object
+ * @return {string} Generated robots
+ */
+export const generateRobots = (req) =>
+  new Promise((resolve) => {
+    //const url = `${req.protocol}://${req.get('Host')}`;
+    const request = superagent.get(`${settings.apiPath}/robots.txt`);
```
tiberiuichim, line 20: should also use settings.internalApiPath

giuliaghisini: in production we use only RAZZLE_API_PATH. Isn't it correct?

We could change to 
``${settings.internalApiPath ?? settings.apiPath}/robots.txt``

what do you think?

src/helpers/Url/Url.js
giuliaghisini: sorry, right. We don't use kitconcept.volto and we dont'know it. We are adapting our backend to return the correct url for sitemap in robots.txt

tisto:
Nice addition. Nitpicking: I would call the changelog entry and title of the PR something with "fallback" since it is more a fallback than a "default" mechanism.

src/helpers/Robots/Robots.js
tiberiuichim: Yes, correct. superagent calls should use the internalApiPath, if set, as much as possible.

giuliaghisini: correct!

> Yes, correct. superagent calls should use the internalApiPath, if set, as much as possible.

fixed

sneridagh (approved):

ichim-david:
@giuliaghisini @sneridagh @tiberiuichim  this work is already committed however I am bringing to your consideration about
the idea of perhaps having VOLTO_ROBOTSTXT env variable be the first option followed by the public/robots.txt file.

Think about  a docker deployment where the image has something in robots.txt and you quickly need to set something else.
By starting the instance with the  VOLTO_ROBOTSTXT env set you can quickly change that before creating a new image with
the modified file.

tiberiuichim:
@ichim-david the resolution order should be: env var -> plone robotstxt -> static file in public. Could you create a PR for this?

giuliaghisini:
> @ichim-david the resolution order should be: env var -> plone robotstxt -> static file in public. Could you create a PR for this?

i think the best order should be:
env var -> static file in public -> plone robots.txt

what do you think?

tisto:
Do we plan to make the robots.txt field editable in Volto? If this is the case we have to think about how to make this clear to the user. We have something similar in c.solr where you can set the connection via zcml and via the solr controlpanel. This always confuses the hell out of me and this here has the same potential.

ichim-david:
The safest/least breaking  change would be to do as @giuliaghisini  mentioned as I've also suggested:
env var -> static file in public -> plone robots.txt

This way if no variable is defined then it behaves the same as volto 11.1.0 where public folder is first
then robots.txt from plone.

@tisto it might be is a bit confusing to have editable public folder robots.txt given the fact that Plone's robots.txt is already
editable in volto through controlpanel/site controlpanel.
Probably it could be housed within Volto Settings control panel alongside a good enough description that it takes precedence
over the setting from site controlpanel however with the env var taking precedence I find it less of a need for such a feature.

giuliaghisini:
@tisto I think the best solution, in addition to giving precedence to any env variabile and robots.txt file in static, is to allow the user to edit the plone robots.txt from Volto, because now he has to access plone in order to fix it.

PR #2184, opened 2021-02-02
UniversalLink and ConditionalLink accepts also an item to link to.

giuliaghisini:
if item is of @type Link (has the remoteUrl property), a direct link to remoteUrl is generated if user is not logged.

This avoids the passage from the Link detail view page even when users are not authenticated. Especially when Link-type objects appears in listing blocks.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 ### Feature
 
 - Internationalization story for add-ons @sneridagh
+- UniversalLink and ConditionalLink accepts also an item to link to. If item is of @type Link, a direct link to remote url is generated if user is not logged. @giuliaghisini
 
 ### Bugfix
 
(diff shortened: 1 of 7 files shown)


sneridagh (approved):

tiberiuichim (approved):
Can we have some documentation for the UniversalLink? It starts to get big and handle all sort of cases.

giuliaghisini:
> Can we have some documentation for the UniversalLink? It starts to get big and handle all sort of cases.

where we could put this components'documentation?

nzambello (approved):

tiberiuichim:
One option would be a story for Storybook, something like this:

https://github.com/plone/volto/blob/7216d1db01ee391332014152af37ca1ef7cb883b/src/components/manage/Toolbar/Dropdown.stories.js

Or a new page under https://docs.voltocms.com/recipes/

giuliaghisini:
> One option would be a story for Storybook, something like this:
> 
> https://github.com/plone/volto/blob/7216d1db01ee391332014152af37ca1ef7cb883b/src/components/manage/Toolbar/Dropdown.stories.js
> 
> Or a new page under https://docs.voltocms.com/recipes/

i've added the storybook for UniversalLink

tiberiuichim (approved):

Issue #2181, opened 2021-02-02
Toast "email: required input is missing" is raised when hit Cancel (X) on the password reset form


tiberiuichim:
**Describe the bug**
Toast "email: required input is missing" is raised when hit Cancel (X) on the password reset form

**To Reproduce**
1. Go to http://localhost:3000/password-reset
2. Click on 'X' to cancel
4. There's a toast popup message

**Expected behavior**
No toast

tiberiuichim:
This also happens on the contact form.

nileshgulia1:
Also in Register Form when you click 'save' with empty values.

nileshgulia1:
Cannot reproduce anymore.

PR #2185, opened 2021-02-03
fix secondary button behaving as type submit

nileshgulia1:
related to #2181 
In `Form.jsx` we do validation checks and show toast in case there're errors. At the same time we let semantic-ui Form handle  these errors and render them inside the Form itself. Can we just rely on one of the two? or move the validation errors on bottom and remove the toast completely. what do you think @sneridagh @giuliaghisini ?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 ### Bugfix
 
 - German translation: aria-label of '/contents' button : "Inhalte" not "Inhaltsverzeichnis" @ksuess
+- Fix secondary button behaving as type submit @nileshgulia1
 
 ### Internal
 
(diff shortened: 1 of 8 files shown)


giuliaghisini:
i think toast is super usefull if you have many tabs in your form on submit, and fields with errors are  not in current tab.. i'm not fully agree to remove the toast notification

sneridagh:
totally agree with @giuliaghisini ... the toasts display other kind of errors too... I do not think it's redundant.

nileshgulia1:
Thanks @giuliaghisini for the comments. I identified the real issue and modified PR accordingly.

giuliaghisini (approved):

avoinea (approved):

nileshgulia1:
Cannot reproduce anymore.

PR #2187, opened 2021-02-03
Slots using the first iteration of the registry singleton

sneridagh:


tiberiuichim:
Work continues in #2212

PR #2188, opened 2021-02-03
PoC - Registry with Circular detector plugin

sneridagh:
Based on slotswithregistry, added a few more entry points to the `Registry`, modified the usual suspects in helpers, reducers, actions and middleware,  from 192 it went to 152 :)
A lot more are due... a proof that it improves the situation, and still maintain backwards compat.


tiberiuichim (commented):
How about:

```
const { settings } = config;
```

PR #2194, opened 2021-02-05
Documentation: Set up project with Volto at a specific branch

ksuess:
For testing how to use and override the new pluggable toolbar configuration I was urged to set up a project with Volto from repo branch. I am definitly not familiar with yarn workspaces. But hey, it works, hurray! Feedback welcome.

### DIFF
--- a/docs/source/contributing/developvoltoinproject.md
+++ b/docs/source/contributing/developvoltoinproject.md
@@ -1,0 +1,97 @@
+Contributing to Volto can be done by setting up Volto with
+
+```js
+clone https://github.com/plone/volto.git volto_coredev
+cd volto_coredev
+yarn start
+````
+
+Sometimes it's necessary to develop in a project. For example the configuration is extended and you need to check how it can be used and overridden in a project.
+
+This is a setup for a vanilla project that uses Volto at branch 'somenewfeature'.
+
+## Set up project with Volto at a specific branch
+
+Create vanilla Volto project
+
+```bash
+yo @plone/volto  --skip-addons --skip-workspaces
+```
+
+Add mrs.developer
+
+```bash
+yarn add mrs-developer -WD
+```
+
+Create 'mrs.developer.json' and add the following content with the branch you want to work on.
+
+```json
+{
+  "volto": {
+    "package": "@plone/volto",
+    "url": "git@github.com:plone/volto.git",
+    "branch": "somenewfeature",
+    "path": "src"
+  }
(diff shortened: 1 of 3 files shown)


tiberiuichim (approved):

tiberiuichim:
Looks good! I'm glad you're documenting this.

The method I've used, but I think yours is much more preferable, is with yalc, something I've [documented some time ago](https://gist.github.com/tiberiuichim/165bdd167d320ab389904ab9148ec512).

ichim-david:
@ksuess I am not in favor of documenting this approach this way because right now with mrs.developer you can only have one output path and we already have documentation about the using addons folder.
The default for the yo generator is to add the output develop to addons folder.

This means that if someone has a project with addons and tries to follow  your guide things will bring some extra artifacts since mrs.developer will checkout the addons also in voltodevelop folder.

Although it isn't an addon, it would be simpler to add volto as an addon, then you no longer need a yarn develop:volto command, yarn develop being enough.

One issue that I've found is the fact that mssdev add /src to volto everytime you run develop:volto and then yarn start crashes
that it can't find razzle.config and you need to manually edit the volto path to remove /src.

If mrs.developer would be smarter about reading the json file and having the ability per package where to checkout I think it
would be easier to have indeed volto in src/voltodevelop/volto and addons to continue to be in src/addons however right not
it is not.

The info is valuable however I feel that some things need to be changed before  this info is presented in the Volto documentation.

docs/source/contributing/developvoltoinproject.md
```diff
@@ -0,0 +1,100 @@
...
+Sometimes it's necessary to develop in a project. For example the configuration is extended and you need to check how it can be used and overridden in a project.
+
+This is a setup for a vanilla project that uses Volto at branch 'somenewfeature'.
+
+## Set up project with Volto at a specific branch
+
+Create vanilla Volto project
+
+```bash
+yo @plone/volto  --skip-addons --skip-workspaces
```
avoinea, line 18: I think here you can just say:
```
yo @plone/volto my-volto-project --no-interactive --skip-install
```

sneridagh:
@ksuess I was reviewing old PRs when I stumbled upon this one. I don't know if you have seen it already (from this week):

https://sneridagh.dev/blog/a-new-build-setup

https://github.com/plone/cookiecutter-volto

I think we can close this now, after all these years, I think we finally got "the thing" right.

PR #2196, opened 2021-02-07
New configuration registry

sneridagh:
So, this one will be hard, in the unit tests part... seems that all is broken when you put all the sensible changes in place (`yarn start` works well).

The tests freak out with the lack of the central omnipresent config in place 😩😩


sneridagh:
Not sure that this https://github.com/plone/volto/blob/c831a77ffc6f49c525d352a5dc2f34892bb9abfa/test-setup-config.js is a good idea... although it works at first glance, it seems that breaks a lot of things.

Still not sure if this "setupFiles" are the things we think they are... :( we should find another way to "populate" the registry with the default config.

tiberiuichim:
```
Test Suites: 2 failed, 223 passed, 225 total
Tests:       4 failed, 639 passed, 643 total
Snapshots:   215 passed, 215 total
Time:        23.676s
Ran all test suites.
```

Only 4 tests fail on my machine... one in the html block test, which is a weird special case, and some other related to cookies, which I think we can fix.

sneridagh:
@tiberiuichim thing is that without the setupFiles populating the registry, they work... the authToken ones are super weird... I guess that it breaks something internally, no clue why.

Trying another (less elegant approach) now...

tiberiuichim:
@sneridagh I think it's weird that the registry setup is still done as an import sideeffect. I think that's a problem. I think there should be an explicit config bootstrap that can be triggered from the client and server code, and this bootstrap could also be triggered from test-setup-config.js.

sneridagh:
@plone/volto-team So, introducing the "new" Volto's Configuration Registry™

## The problem
It's been a while since we are experiencing undesired side-effects from "circular import dependency" problems in Volto. Although they aren't much noticeable, they are there, waiting for bitting us. In fact, these kind of problems are common in NodeJS world, and the very nature of how it works make them "workable" thanks to the NodeJS own import resolution algorithm. So the "build" always works, although we have them.

The most preoccupying ones are the ones introduced by the config "imports" since the level of intricacy is quite high and affecting modules living in `components`, `helpers`, `actions`, `middleware`... since all of them "require" access to the config. We've been using the `config` module using it as container of Volto's config taking advantage of the ES6 module system. So we "import" the config, then the exported config data in that module "magically" is there.

However, lately we found it biting us under several use cases that we managed to dodge with workarounds. I had the feeling that we need to address that somehow. That's why (@tiberiuichim and I) we've been trying to find a nice and elegant solution to it during the last months. After trying and discussing several approaches, we reached to an agreement on the solution using a centralised singleton (the Configuration Registry) that is populated from the `config` module.

## The solution
We were searching a solution that is:
- Non-breaking, backwards compatible (if possible)
- Still works for add-ons (if possible)
- Cut the circular dependency

After considering several other approaches, we've found that the singleton approach was the best, and did a PoC that demonstrates that it works, and reduced the circular dependencies (using a Webpack plugin that detects them). This singleton is instantiated once, and returned afterwards, and has an API to store the same keys we had before (`settings`, `views`, `widgets`, `blocks`). It's inmutable, and you cannot modify its signature or extend it (prevents from unintentional tampering).

https://github.com/plone/volto/blob/40aa975f40742d4028d0650b0cb34281d2f93adf/src/registry.js

The old code:
```js
import { settings } from '~/config'
...

console.log(settings.apiPath)

...
```

new code:
```js
import config from '~/registry'
...

console.log(config.settings.apiPath)
config.settings.isMultilingual = true
...
```

### Non-breaking, backwards compatible
The registry is populated on bootstrap time, after the `~/config` module is loaded, so the old way and the new way coexists without interference. Of course, we should use only the new way in core (this PR addresses that). People in their projects can slowly migrate to use the registry (although not required).

### Add-ons
The add-ons still can use the old (config passing though) of the config and modify it in bootstrap time, or use the new registry, so backwards compatibility in add-ons is assured. It is uncertain at this moment if we will deprecate the old way, although there is no real need to do so right now.

### Cut the circular dependency
The most preoccupying circular import problems get solved, by cutting the loop at the code module level, since all the modules now uses the registry to access the config.

### Caveats
The tests have to populate a "mocked" version of the config, and it was a bit difficult to make it work, since although the circular imports seems to "play well" on the code itself, the testing framework is a bit more non-tolerant to have them. So load the config in the setup of the test led to circular imports in the test framework itself breaking a lot of tests. So I had to come up with this mocked version of the testing setup that populates the config with well known (and from now on consistent and maintained) set of configurations that will apply to only the tests, making them more consistent (in the snapshotting) too.

Unfortunately this led that the refactor has been a bit hellish (in the tests part). The good thing is that now I have a grasp of the current testing state for Volto, and I have to say, we need a full sprint on it 😂

## Actions
We will need to make a quick decision on this one, since the refactor has been very big and the merging with the upcoming changes could be hellish too.

So, what do you think? Feedback is very appreciated!

Thanks!

src/components/manage/Display/__snapshots__/Display.test.jsx.snap
```diff
@@ -1,69 +1,48 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
 exports[`Display renders an actions component 1`] = `
-Array [
+<div>
   <label
-    htmlFor="display-select"
```
tisto, line 6: @sneridagh can you explain why the snapshots contain the HTML version (for, class) instead of the React version (htmlFor, className)?

src/components/manage/Preferences/__snapshots__/ChangePassword.test.jsx.snap
```diff
@@ -20,178 +20,28 @@ exports[`ChangePassword renders a change password component 1`] = `
           className="ui segment"
         >
           <div
-            className="inline required field help"
+            className="password-widget"
```
tisto, line 23: @sneridagh if this reflects a real change in the HTML/Class structure we should expect lots of breaking styles. Or is this just a follow up on the test setup changes?

src/components/manage/Widgets/__snapshots__/FileWidget.test.jsx.snap
```diff
@@ -1,81 +1,39 @@
...
               My field
             </label>
           </div>
         </div>
         <div
-          className="eight wide column"
+          class="eight wide column"
         >
           <div
-            className="file-widget-dropzone"
```
tisto, line 31: @sneridagh is the fileWidget component completely gone in this test snapshot?

tisto (commented):
w00t! That's a great addition and a large improvement over the traditional way! We should push this forward and get this into core asap. I added my questions inline.

One other question: is there a way we could expose the reporting of circular dependencies in our CI? It would be good if we would be able to reach zero circular dependencies and then make gh actions report any violations.

tisto:
Changelog entry is missing :)

src/components/manage/Display/__snapshots__/Display.test.jsx.snap
sneridagh: I used @testing-library on them instead of the react-testing-renderer, and it outputs proper html.

I think we should move all of them to use it, sooner or later, and force all new developments to use it.

src/components/manage/Preferences/__snapshots__/ChangePassword.test.jsx.snap
sneridagh: It's because of the new testing config mocks. In this case, the widgets are all mocked, to a more readable output where at a glance, you can see that the intention in the test was to just render a type of widget, and not the actual widget structure, which does not bring anything but a bunch of html. Also, when you change something from a widget, it only reflects on its test, not in a random number of tests where that widget is actually used. I guess it's more accurate and isolated than the previous state.

src/components/manage/Widgets/__snapshots__/FileWidget.test.jsx.snap
sneridagh: Something weird is going on in this test, we have to figure it out. Previously, the test output was an array (!?!?) which is weird. I changed to use @testing-library and it behaved better, but I overlooked that the widget was gone. It's a matter of timing issues in testing rendering... the widget is there for sure, if not, cypress won't pass...

sneridagh:
> One other question: is there a way we could expose the reporting of circular dependencies in our CI? It would be good if we would be able to reach zero circular dependencies and then make gh actions report any violations.

There are the "good" acceptable circular ones like:

src/components/manage/Widgets/RecurrenceWidget.jsx -> src/components/index.js -> src/components/manage/Widgets/RecurrenceWidget.jsx

which are provoked by the index.js pattern that are accepted and even considered not harmful by the NodeJS community.

Then there are the "harmful" ones that involves several jumps through module groupings (`components`, `helpers`, `actions`, ...). Those are the ones that we managed to avoid because they are a time bomb waiting to explode.

I don't know if we could implement a detection system in CI that detects them... maybe we can using the circular dependency plugin for it and add a step in the CI. Worth exploring!

tiberiuichim (approved):

docs/source/upgrade-guide/index.md
```diff
@@ -8,7 +8,148 @@ This upgrade guide lists all breaking changes in Volto and explains the
...
+  navDepth: 3,
+};
+```
+
+then you'll add the `applyConfig()` function as default export and copy that settings key in it:
+
+```js
+export default function applyConfig(config) {
+  config.settings = {
+    ...config,
```
tiberiuichim, line 110: ...config.settings

sneridagh: T_T thanks!

src/components/manage/Display/__snapshots__/Display.test.jsx.snap
tisto: Ok. That makes a lot of sense. In the future, it might be better to create a separate PR for this move to make it obvious where this comes from. Not worth the effort for this PR now I guess though.

src/components/manage/Preferences/__snapshots__/ChangePassword.test.jsx.snap
tisto: So we do not expect any breaking changes whatsoever when it comes to HTML/CSS, correct?

tisto (approved):
My open issues/questions have been addressed and I am +1 to merge this and do a major release.

src/components/manage/Preferences/__snapshots__/ChangePassword.test.jsx.snap
sneridagh: Nope :)

sneridagh:
@tisto Going through your comments made me remind me about the FileWidget output was weird. I will look into that.

sneridagh:
@plone/volto-team btw, I will also leave some numbers that I realised the other day:

## master Volto 11.1.0
Test Suites: 225 passed, 225 total
Tests:       643 passed, 643 total
Snapshots:   216 passed, 216 total
Time:        72.113s
Ran all test suites.

## Volto 12.0.0
Test Suites: 225 passed, 225 total
Tests:       643 passed, 643 total
Snapshots:   216 passed, 216 total
Time:        57.595s
Ran all test suites.

20% of improvement, 15 secs seems not much but still!

avoinea (approved):

src/config/index.js
```diff
@@ -34,6 +34,8 @@ import { contentIcons } from './ContentIcons';
 
 import applyAddonConfiguration from 'load-volto-addons';
```
sneridagh, line 35: @tiberiuichim since this is no longer true, shouldn't we rename it to `applyConfig` or so?

create-addons-loader.js
```diff
@@ -67,14 +68,18 @@ const safeWrapper = (func) => (config) => {
   return res;
 }
 
+const projectConfig = (config) => {
+  return typeof projectConfigLoader.default === "function" ? projectConfigLoader.default(config) : config;
```
sneridagh, line 72: @tiberiuichim wondering if we could squeeze this point even a bit more and lookup for a `volto.config.js` too, and if it exists, then load its configuration too. It could be useful for (cypress) testing...

pnicolli (commented):
+1 for the Singleton approach, I like it.
I left just a few minor comments here and there.

src/registry.js
```diff
@@ -0,0 +1,87 @@
...
+    return this._data.slots;
+  }
+
+  set slots(slots) {
+    this._data.slots = slots;
+  }
+}
+
+const instance = new Config();
+Object.freeze(instance);
```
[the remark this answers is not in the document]
sneridagh: By definition a singleton is not extensible, and complies with the internal intended API, so it's a safeguard to not allow people to fiddle with it. I don't have a strong opinion with this, we might want to allow it... but it felt sensible and reasonable.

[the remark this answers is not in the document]
tiberiuichim: Maybe add a new `config.addon` key? Or `config.custom` ? To have a place for addons to place any config that doesn't fit in one of the existing keys?

PR #2200, opened 2021-02-10
Let toolbar buttons and menues be restrictable to route

ksuess:
```
const applyConfig = (config) => {
  config.toolbar.activities.view.top.push(ButtonA)
  config.toolbar.activities.view.bottom.push({
    match: '/documentation',
    component: ButtonB,
  })
  return config;
};
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Breaking
 
-- Add a new config registry, the `toolbar`. You should migrate your project's `config.js` to export the `toolbar` registry. @tiberiuichim @ksuess
+- Add a new config registry, the `toolbar`. You should migrate your projects `config.js` to export the `toolbar` registry. @tiberiuichim @ksuess
 
 ### Feature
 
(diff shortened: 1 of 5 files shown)


tiberiuichim:
@ksuess I think it might be better to always require the `{ match, component }` data structure, so that we don't have to special-case it. It would mean that the registration of buttons would always have to include that information. Another option would be to have another, router-aware helper component that can be used to wrap the button. Something like:

```
const applyConfig = (config) => {
  config.toolbar.activities.view.top.push(ButtonA)
  config.toolbar.activities.view.bottom.push(withPathRestriction(ButtonB, (pathname) => pathname.includes('/documentation))
  return config;
};
```

And the withPathRestriction would be:
```
import { matchPath } from 'react-router';
const withPathRestriction = (WrappedComponent, matcher) {
    const wrapped = (props) => {
        const location = useLocation();
        const { pathname } = location;
        return matcher(pathname) ? (<WrappedComponent {...props} >) : (<></>);
    }
    return wrapped;
}

```

In any case, I've noticed that the matching function that you've used is a simple equality check, we should use `matchPath`, similar to https://github.com/plone/volto/blob/master/src/components/theme/AppExtras/AppExtras.jsx#L11

@plone/volto-team what do you think?

tiberiuichim (approved):

PR #2204, opened 2021-02-12
Configurable image proxy

tiberiuichim:


silviubogan:
@tiberiuichim I think that file-cache.js should be renamed to FileCache.js (similarly, the corresponding test file). Do you agree?

src/express-middleware/image-proxy/file-cache.js
```diff
@@ -14,18 +15,19 @@ function isNumber(val) {
 
 class FileCache {
   constructor(opts = defaultOpts) {
-    this.basePath = opts.basePath;
+    this.basePath = process.env.VOLTO_FILE_CACHE_BASE_PATH || opts.basePath;
```
nileshgulia1, line 18: @silviubogan  we can remove `process.env.VOLTO_FILE_CACHE_BASE_PATH` from here? below we're resolving path already, I think.

silviubogan: @nileshgulia1 I think no because below we are resolving the path relative to the current working directory which is the root directory of Volto, and `VOLTO_FILE_CACHE_BASE_PATH` preferably contains a path relative to that root directory although it can contain absolute paths too.

sneridagh:
Leaving this open, since I really think we could get back to it at some point in the future, and the code is really amazing.
However, there are also third party solutions that do the job well (Thumbor, other CDN SASS services), and we don't have to maintain them.

@tiberiuichim Should we leave this open, or let's just archive it? What do you think?

Message for future maintainer: If one year passes after this message and no further developments, close it.

davisagli:
Closing as abandoned. (Also in the long run I think volto should not be in the business of serving images.)

PR #2213, opened 2021-02-17
New breadcrumbs `INavigationRoot` aware for the *Home* icon

sneridagh:


tiberiuichim:
https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/services/breadcrumbs/get.py#L44

Nice, we're getting closer to our Plone roots :)

tiberiuichim (approved):

PR #2215, opened 2021-02-18
feat: add extendable asyncConnect entries

nzambello:
As referred in the [docs](https://github.com/plone/volto/blob/master/docs/source/developer-guidelines/redux.md) we cannot add actions to SSR handling if not in routes' components.
I added a new entry in config with an extendable list of these actions to be used in `asyncConnect` from App.jsx so addons or site projects can use it to improve performances and be more SEO compliant.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ### Feature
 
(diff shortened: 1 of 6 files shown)


giuliaghisini (approved):

sneridagh:
Superseded by: #2223

PR #2223, opened 2021-02-20
Router-enabled configurable asyncConnect extenders

tiberiuichim:
This is an enhancement on Nicola's PR, #2215. It basically solves the same problem, with a few differences:

- it allows per-route customizing the asyncConnects. Not only adding new connects, but potentially also removing them. Imagine that I want to create a website without the breadcrumb component, so I don't want the `@breadcrumb` service to be called. I can either customize App.jsx or the getBreadcrumb action, but I end up with customized files. With this you could write an extender like:

```
config.settings.asyncPropsExtenders = [
    {
      path: '/',
      extend: (dispatchActions) => dispatchActions.filter(asyncAction=> asyncAction.key !== 'breadcrumb')
     }
]
```

- we internalize some of the code from redux-connect. Why is this a good thing? While redux-connect is a stable package that does what we need, it's also not something that we can extend, for example to support proper SSR of "async-powered blocks" such as the listing block. There's not a lot of internal documentation, for us Volto people it is a scary monster that we sometimes have to battle. Its code is sometimes hard to read not because the code is complex, but because that code is poorly written in functional style while it operates in an imperative way.

The `@plone/volto/helpers/asynConnect` can work as a drop-in replacement for `redux-connect/asyncConnect`, so to finish this PR I'd like to move the rest of redux-connect code to Volto (right now we're missing only https://github.com/makeomatic/redux-connect/blob/master/modules/components/AsyncConnect.js), remove the package from dependencies and provide a webpack alias to fake it.


sneridagh (approved):
LGTM!
We need to add the example in the documentation, but I already see it mergeable!
Do you want to finish the migration of the entire package into Volto?

tiberiuichim:
@sneridagh Yeah, I want to finish the migration of that package, optimize some code and do some more testing for it. I hope tomorrow all this will be ready.

sneridagh (approved):

PR #2233, opened 2021-02-23
Improved comments by @rexalex

avoinea:
- Reply comment
- Better icons for users with no icon set
- Tree view of comments

comments


src/components/manage/Form/Form.jsx
```diff
@@ -518,6 +524,24 @@ class Form extends Component {
     return id;
   }
 
+  /**
+   * will prevent event from triggering submit
```
tiberiuichim, line 528: Comment should follow the style of the rest of the file.

tiberiuichim (approved):
Awesome! I've left a comment. Also, the Form API changes are worth mentioning in the changelog

tisto:
Nice. Without having time to look at the code right now: how do we decide if we show threaded comments or flat ones? Do we take the p.a.registry setting from Plone classic into account here?

I am +1 on adding avatar images to the comments (which is the case in Plone classic if I am not mistaken). Though, we might want to do this in a separate PR.

avoinea:
> Nice. Without having time to look at the code right now: how do we decide if we show threaded comments or flat ones? Do we take the p.a.registry setting from Plone classic into account here?

@tisto We do, but I don't remeber seeing settings about flat vs threaded comments.

> I am +1 on adding avatar images to the comments (which is the case in Plone classic if I am not mistaken). Though, we might want to do this in a separate PR.

Indeed, I think we should consider this checkbox from Site Setup > Discussion:

discussion

@rexalex ^

sneridagh (approved):

PR #2239, opened 2021-02-25
Addon loading and discovery through dependencies

tiberiuichim:
With this change, it's now possible to list addons in another addon as a dependency. For example volto-slate can now list `@eeacms/volto-object-widget` as a dependency, in volto-slate/package.json "addons" key, just like the project does. It's also possible to use extra loaders and internally we use a dependency graph to resolve the order of loading the addon loaders.


nzambello (approved):
Haven't tested yet, but LGTM.
This is an interesting feature for addon developers

tiberiuichim:
There's a few edge cases that I need to solve before being ready to be merged.

tiberiuichim:
Problems fixed, good to go. Give it a try, though there's a ton of cases that can be encountered "in the wild":

- main addon with node_modules installed dependency addons
- various of combinations of installed (node_modules) vs development addons
- various levels of inter-addon dependencies (addon A dependens on addon B, addon B depends on several extra loaders from addon B)
- development addon can be symlinks to parent folder

sneridagh (approved):

PR #2257, opened 2021-03-05
Updated "Volto in Production" list

alecghica:


tiberiuichim (approved):

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 
 ### Feature
 
+- Updated "Volto in Production" list @alecghica
```
tisto, line 9: @alecghica since this has nothing to do with Volto software I wouldn't add a changelog entry here.

tisto (changes requested):
Awesome! We should remove the changelog entry and then merge this right away.

CHANGELOG.md
alecghica: @tisto check my initial commit. I added the changelog because "Changelog verifier" failed. Shall I just remove the changelog entry and ignore that?

tisto: Since when do we have a changelog verifier on the Volto repo? If we need a changelog entry this should be under "internal" then.

alecghica: Done.

sneridagh:
@alecghica merging with the changelog error check is fine too :)

PR #2258, opened 2021-03-05
Improved object browser with manual input field

sneridagh:
WIP

- [x] allow manual input for URLs, tokeniser detector for internal ones, only for non-"multiple" widgets (eg. relation fields)
- [x] new helpers for making that possible - isInternalURL
- [x] new config setting for indicating the URL the app is deployed on (`publicURL`) so we can determine even in SSR isInternalURLs
- [x] new prop `allowedExternal` for the widget, if we do not want the widget to allow external links
- [x] tests for the new helpers required
- [x] documentation for the new `allowedExternal` prop
- [ ] cypress tests for it
- [ ] add icon for external and internal in the token?
- [x] fix icon in the popup

I wonder if we would like to reuse this in other scenarios, like the link chooser in the text block editor... we could refactor it to make it happen...

(sound on) 😄 

https://user-images.githubusercontent.com/486927/110161342-82cb1300-7ded-11eb-9b35-ee47d45ccf07.mov

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -3,6 +3,8 @@ import { StaticRouter } from 'react-router-dom';
 import { IntlProvider } from 'react-intl';
 import enMessages from '~/../locales/en.json';
 
+import '../test-setup-config';
+
 import '~/theme';
 
 export const parameters = {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Bugfix
 
+- Fix storybook initial config registry setup @sneridagh
+
 ### Internal
 
 ## 12.2.0 (2021-03-03)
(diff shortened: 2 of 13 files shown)


tiberiuichim (approved):

tiberiuichim (approved):

nzambello (approved):

sneridagh:
Ok, literally there's no way to write a Cypress test it in Volto for this, since there is none instantiated in the core anywhere... How do we test these use cases? Should we create a "test page" under a test route with all these components that are available but not instantiated? 

Say `++test++playground`?

Also, I think I'll leave the icon (internal/external) out for now, waiting for Quanta.

/cc @tiberiuichim

tiberiuichim:
My personal wish is to have an addon integrated in the generator tests. That addon can be a dump for all sort of helpers and cases that we don't encounter in Volto core, and bonus is that we also test the addon infrastructure.

tiberiuichim (approved):

avoinea (approved):

PR #2267, opened 2021-03-12
Add testing add-on for enable special testing use cases and configuration options

sneridagh:


create-addons-loader.js
```diff
@@ -68,6 +74,14 @@ const safeWrapper = (func) => (config) => {
   return res;
 }
 
+const voltoConfFileConfig = (config) => {
```
sneridagh, line 77: @tiberiuichim Also implemented the `volto.conf.js` thingy. What do you think?

sneridagh:
meh tests fails, but works locally (?) :(

sneridagh:
does not work either... weird.

create-addons-loader.js
tiberiuichim: whohooo!

tiberiuichim:
@sneridagh What's the intention with the voltoconfig? 

For me the idea was to be able to take the stuff that we put right now in package.json (and we only do addons, right now) and be able to put it in a separate json or js file, using the rc type of libraries https://github.com/davidtheclark/cosmiconfig or https://www.npmjs.com/package/rc-config-loader

sneridagh:
So, this PR enables a way to add addons (mainly for testing) then run the tests with these custom builds. This might be handy on adding features that plain Volto does not provide (eg widgets, multilingual or Guillotina specifics). 

Not happy with the outcome, since the method seems a bit cumbersome (using a script to modify package.json and add a custom `jsconfig.json`), I realised that might be that we could also do it in a more simple way using a higher config override, using a `volto.config.js` file. 

I like to have the addon approach (specially for G), this will allow us to reorganize the tests in a more sane way). But having a better way to override thing in the build might be also useful. Not sure if having both around is more confusing than useful.

I would like also to extract things from `package.json` but for now we can only extract the 'addons' key... the rest is required for workspaces to work. Other option would be to read them from `volto.config.js` in the addonsloader.

We have to talk about this PR... not sure of the outcome, and if we really need them both or only one of them.

tiberiuichim:
@sneridagh I see, so it opens the posibility of loading addons directly in volto, instead of going via Volto project. For us, developers, it would be nice, just cloning Volto and quickly enabling an addon. Though it could be confusing for newbies. 

Even so, I think it would be better to just enable rc-style configuration of that "addons" package.json key. Then we can drop a .voltorc.js or .voltorc.json (Similar to https://eslint.org/docs/user-guide/configuring/configuration-files#configuration-file-formats) with extra addons (and this file can be simply copied in CI envs). And the code paths would be the same, discover voltorc file in "project root", which will be volto itself when running standalone.

sneridagh:
So, `RAZZLE_TESTING_ADDONS=test-addon yarn start` I thought it might be better for exactly this use case.

tiberiuichim:
@sneridagh Looks good!

sneridagh:
@tiberiuichim cleanup and docs

docs/source/developer-guidelines/testing.md
```diff
@@ -22,6 +22,11 @@ environment variable.
 
 This is specially useful in CI while developing add-ons, so you can pass an specific configuration that deals with the addon config properly.
 
+## Add addons for testing purposes
+
+Quite often, you need different configurations and enabling components for testing purposes. You can use the `RAZZLE_TESTING_ADDONS` environment variable to dafine them. This addons does not have the capacity to install dependencies (as opposed to define them properly as a true addon) but all the rest of the addon features works as expected.
```
tiberiuichim, line 27: type, "define them". "This addons" => "These addons" or "This addon".

PR #2284, opened 2021-03-18
Added SidebarPopup component for extra sidebar handling

avoinea:
- Custom Generic Sidebar outside Blocks scope.
- Extra configuration for widgets within SidebarPortal.

### DIFF
--- a/src/components/index.js
+++ b/src/components/index.js
@@ -93,6 +93,7 @@ export Messages from '@plone/volto/components/manage/Messages/Messages';
 export BlockChooser from '@plone/volto/components/manage/BlockChooser/BlockChooser';
 export Toolbar from '@plone/volto/components/manage/Toolbar/Toolbar';
 export Sidebar from '@plone/volto/components/manage/Sidebar/Sidebar';
+export SidebarPopup from '@plone/volto/components/manage/Sidebar/SidebarPopup';
 export SidebarPortal from '@plone/volto/components/manage/Sidebar/SidebarPortal';
 export PersonalTools from '@plone/volto/components/manage/Toolbar/PersonalTools';
 export More from '@plone/volto/components/manage/Toolbar/More';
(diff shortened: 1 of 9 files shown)


tiberiuichim (approved):
We might need a special doc chapter on things we include that are not directly used in Volto?

nileshgulia1 (approved):

src/components/index.js
```diff
@@ -93,6 +93,7 @@ export Messages from '@plone/volto/components/manage/Messages/Messages';
 export BlockChooser from '@plone/volto/components/manage/BlockChooser/BlockChooser';
 export Toolbar from '@plone/volto/components/manage/Toolbar/Toolbar';
 export Sidebar from '@plone/volto/components/manage/Sidebar/Sidebar';
+export SidebarPopup from '@plone/volto/components/manage/Sidebar/SidebarPopup';
```
tiberiuichim, line 96: I'm not sure we want this line. Would webpack drop this component from the bundle if we don't do this? In theory only addons that use this component would activate the inclusion of the module in the bundle.

tiberiuichim (changes requested):

tiberiuichim (approved):

src/components/manage/Sidebar/SidebarPopup.jsx
```diff
@@ -0,0 +1,46 @@
...
+        <aside
+          role="presentation"
+          onClick={(e) => {
+            e.stopPropagation();
+          }}
+          onKeyDown={(e) => {
+            e.stopPropagation();
+          }}
+          ref={ref}
+          key="sidebarpopup"
```
sneridagh, line 27: Now I see it here :)

sneridagh (approved):

sneridagh (approved):

PR #2288, opened 2021-03-22
as in Plone, hide controlpanel for users that are no 'Manager' or 'Si…

giuliaghisini:
…te Administrator'


src/components/manage/Toolbar/PersonalTools.jsx
```diff
@@ -143,12 +145,17 @@ class PersonalTools extends Component {
                 <Icon name={rightArrowSVG} size="24px" />
               </button>
             </li>
-            <li>
-              <Link to="/controlpanel">
-                <FormattedMessage id="Site Setup" defaultMessage="Site Setup" />
-                <Icon name={rightArrowSVG} size="24px" />
-              </Link>
-            </li>
+            {isControlpanelEnabled(this.props.user) && (
```
tiberiuichim, line 148: Maybe `isAdminUser` ?

giuliaghisini: Yes! It's better. I've updated the code

nzambello (approved):

cekk:
i think that we should start thinking about how to check user permissions and not roles with restapi, because we are missing a lot of Plone's power like this.

src/helpers/User/User.js
```diff
@@ -0,0 +1,16 @@
...
+ * Is admin user method.
+ * @method isAdminUser
+ * @returns {undefined}
+ */
+export function isAdminUser(user) {
+  return user?.roles
+    ? user.roles.includes('Site Administrator') ||
+        user.roles.includes('Manager')
+    : false;
+}
```
pnicolli, line 16:
```suggestion
import { intersection } from 'lodash/intersection';

/**
 * Checks if the user has one or more given roles.
 * @method userHasRoles
 * @returns {boolean}
 */
export function userHasRoles(user, roles = []) {
  return intersection(user?.roles ?? [], roles).length > 0;
}
```

How about a slightly different and more generic approach, like this for example? I can't decide whether I would prefer having a direct ``isAdminUser`` method or a more generic one, or maybe both.

PR #2290, opened 2021-03-23
Pluggable components

tiberiuichim:
WIP


sneridagh:
Love it...

sneridagh (approved):

sneridagh:
Some narrative docs would be great!

tiberiuichim:
@sneridagh I'll add narative docs.

Should we merge the slots with the "pluggables" implementation? It would simplify the story we tell.

The difference between slots and pluggables right now is:

- slots are more declarative, the "inserts" are declared as configuration (or they come from the content)
- the "pluggables" are more dynamic and short-lived in their nature, similar to the Portal components that we use. The pluggables might also be used as an extension mechanism that's not possible with slots. Each pluggable slot can process the "inserts" in any way it wants (not just rendering them), so they can be used as an adhoc mini-plugin framework.

sneridagh (approved):

package.json
```diff
@@ -9,7 +9,7 @@
     }
   ],
   "license": "MIT",
-  "version": "12.10.1",
+  "version": "12.8.0",
```
sneridagh, line 12: Watch out!

PR #2294, opened 2021-03-24
Add ObjectWidgets and improvements to the InlineForm

sneridagh:
## ObjectListWidget
Screenshot 2021-04-01 at 19 38 34

It's a widget that takes a schema itself to render a collection of objects that are saved in an array as value.

The shown widget has this schema:

https://github.com/plone/volto/blob/179a45f895d99e161ca37ea90432ef3c42351ba0/packages/core-sandbox/src/components/Blocks/TestBlock/schema.js#L38-L102

## BlockDataForm

A convenience component that wraps a schema for a block. 

https://github.com/plone/volto/blob/179a45f895d99e161ca37ea90432ef3c42351ba0/packages/core-sandbox/src/components/Blocks/TestBlock/Data.jsx#L1-L25


tiberiuichim (approved):

src/components/manage/Form/InlineForm.jsx
```diff
@@ -86,32 +83,31 @@ const InlineForm = ({
...
-                }}
-                key={field}
-                error={errors[field]}
-                block={block}
-              />
-            ))}
-          </Segment>
-        </div>
-      )}
+      <div id={`blockform-fieldset-${defaultFieldset.id}`}>
```
tiberiuichim, line 86: shouldn't use here the blockform prefix for the id.

src/components/manage/Form/InlineForm.jsx
```diff
@@ -137,8 +133,12 @@ const InlineForm = ({
                   {map(fieldset.fields, (field) => (
                     <Field
                       {...schema.properties[field]}
-                      id={fieldIndex ? `${field}-${fieldIndex}` : field}
-                      value={schema.properties[field].value || formData[field]}
+                      id={field}
```
tiberiuichim, line 136: should be made unique. The problem is with fields such as "title" that are very common, will have duplicate id

tiberiuichim (approved):

PR #2226, opened 2021-02-22
Adapt to BlocksForm in Blocks Engine

nileshgulia1:
Related to #2207 .

PR #2295, opened 2021-03-25
Fix multiselected includes

avoinea:
Follow-up #2226


src/components/manage/Form/InlineForm.jsx
```diff
@@ -38,7 +38,7 @@ const InlineForm = ({
   const defaultFieldset = schema.fieldsets.find((o) => o.id === 'default');
   const other = schema.fieldsets.filter((o) => o.id !== 'default');
   return (
-    <Segment.Group raised className="form">
+    <div className="ui form">
```
sneridagh, line 41: already changing this in #2294 is there a strong reason?

PR #2299, opened 2021-03-25
Fixed InlineForm boolean false value

razvanMiu:
**Issue:** If you have a boolean type Field (CheckboxWidget) with no default value and `props.value = false` then this condition `formData[field] || schema.properties[field].default` from InlineForm will evaluate as `false || undefined` which is equal to `undefined` hence the CheckboxWidget value will be `undefined` and will create the issue from here https://github.com/plone/volto/pull/2296.
**Fix:** Use the `default` property only if `default` is a defined property otherwise use the value of the Field.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Bugfix
 
+- Fixed InlineForm boolean false value @razvanMiu
+
 ### Internal
 
 ## 12.4.0 (2021-03-25)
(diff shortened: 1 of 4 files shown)


tiberiuichim (approved):

avoinea (approved):

tiberiuichim:
it should be a rule, we don't want state in any of the widgets (of course, there may be exceptional cases)

sneridagh (approved):

PR #2306, opened 2021-03-27
This fixes addons in Volto itself

sneridagh:
by using the root config using the nowadays accepted naming for the babel config:

https://babeljs.io/docs/en/config-files#project-wide-configuration

Projects are already using it.

Also, updated the adding the nodeExternals for testing addons

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,3 +1,0 @@
-{
-  "presets": ["./babel"]
-}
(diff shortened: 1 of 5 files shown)


tiberiuichim (approved):

sneridagh:
I messed things up, force pushed a better commit set.

PR #2307, opened 2021-03-27
fix warning on ref forwarding sidebarPopup

### LINKED ISSUE #2305: Warning message in Volto bootstrap: "Warning: forwardRef render... "
Warning: forwardRef render functions do not support propTypes or defaultProps. Did you accidentally pass a React component?

image

nileshgulia1:
fixes #2305
Sorry for missing this.....

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Bugfix
 
 - Fixed InlineForm boolean false value @razvanMiu
+- fix warning related to ref forwarding during ssr @nileshgulia1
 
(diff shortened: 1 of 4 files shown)


sneridagh:
@nileshgulia1 thanks for the quick fix, however I've been thinking about it, and I think it's better to move the responsibility to open/close detection to the aside itself, instead to force the inner component to deal with it. So the inner component can be totally independent of it.

So if you don't mind, I did it here: #2309

nileshgulia1:
see: #2309

PR #2309, opened 2021-03-28
Fix warning message in console, move open/close detection to the asid…

sneridagh:
…e itself

### DIFF
--- a/src/components/manage/Sidebar/SidebarPopup.jsx
+++ b/src/components/manage/Sidebar/SidebarPopup.jsx
@@ -7,13 +7,13 @@ import { doesNodeContainClick } from 'semantic-ui-react/dist/commonjs/lib';
 const DEFAULT_TIMEOUT = 500;
 
 const SidebarPopup = (props) => {
-  const { children, open, overlay } = props;
+  const { children, open, close, overlay } = props;
 
   const asideElement = React.createRef();
 
   const handleClickOutside = (e) => {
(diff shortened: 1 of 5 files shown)


sneridagh:
I like this solution more than the other, so we move the responsibility to open/close detection to the aside itself, instead to force the inner component to deal with it. So the inner component can be totally independent of it.

This superseedes: https://github.com/plone/volto/pull/2307

src/components/manage/Sidebar/SidebarPopup.jsx
```diff
@@ -2,11 +2,27 @@ import React from 'react';
...
 
-const SidebarPopup = (props, ref) => {
+const SidebarPopup = (props) => {
   const { children, open, overlay } = props;
+
+  const asideElement = React.createRef();
+
+  const handleClickOutside = (e) => {
+    if (asideElement && doesNodeContainClick(asideElement.current, e)) return;
+    props.closeObjectBrowser();
```
nileshgulia1, line 16: @sneridagh I think its okay with handling open/close inside sidebarPopup. We should also change the name of closer prop (`props.closeObjectBrowser()`)as the component can be used elsewhere apart from ObjectBrowser.

sneridagh: Indeed, that is true. Will look into it.

avoinea (approved):

PR #2319, opened 2021-03-29
fix: sitemap url generation

nzambello:
Now sitemap has correct URLs for crawlers with correct public URL.


tiberiuichim (approved):

tiberiuichim:
./src/helpers/Sitemap/Sitemap.js
Attempted import error: 'toPublicURL' is not exported from '@plone/volto/helpers'.

sneridagh:
@nzambello maybe an uncommitted change?

src/helpers/Url/Url.js
```diff
@@ -101,6 +101,21 @@ export function flattenToAppURL(url) {
...
 }
 
+/**
+ * Given a URL if it starts with the API server URL
+ * this method removes the /api or the /Plone part.
+ * @method toPublicURL
+ * @param {string} url URL of the object
+ * @returns {string} public URL
+ */
+export function toPublicURL(url) {
```
sneridagh, line 111: Are you aware of the new `publicURL` setting? It has good a zero config default already:

https://github.com/plone/volto/blob/c0a983cc2bf63e89709215e420b3d52a7880814e/src/config/index.js#L59-L64

and it's an env var if required something else outside of the "defaults". It would also cover other use cases out of "/Plone" and "/api".

sneridagh: FlattenToAppURL is already aware of it, but we could fancy a helper that removes anything else and sets the publicURL. Maybe flatten first, then add publicURL?

sneridagh (approved):

PR #2320, opened 2021-03-30
[Breaking] Introducing Seamless Mode

sneridagh:
Presenting the "seamless" mode. The goal? Achieve Zero configuration (avoiding at the same time hardcoded API_PATH or other vars in code) with good sensible defaults when setting up deployments.

Having in production this kind of setup, where both Volto and the API lives under the same sun. The content negotiation stays, so any API call is made with the `Accept` header, so it's routed to the backend.

The idea was still use the same API_PATH meaning, and do not introduce any breaking change, so old deployments in place continue working as they are.

## Development
- Seamless mode by default
- Calls to API is always routed through internal proxy, http://localhost:3000 is both Volto and Plone by default.
- API_PATH sensible default is http://localhost:8080/Plone (no change of behaviour)

## Production
- Nothing changes for old deployments, all work the same way if API_PATH is set
- If you want seamless in production as well you need to setup your reverse proxy (see attached Nginx reference config) properly, so the reverse proxy detects the `accept` header and route the requests properly, and remove API_PATH env var from your build
- In seamless mode no API_PATH is required (and should not be present at build time), but the `Host` header must be set in the reverse proxy (see attached Nginx reference config)

## Pros

- The zero config and the lack of having to rely on hardcoded variable names in the build
- Theoretically, you could deploy several sites using the same Volto SSR server without recompiling (just using the `Host` header in the reverse proxy)
- Opens the door for http://servername/sitename deployments as well, so several sites hosted the same Volto SSR server, we could use headers (same as Pyramid) to accomplish that as well.
- All internal links are app ones, so a link to a page /my-page will be returned by the API as is. So flattenToAppURL will still be required (for old deployments) but if seamless is adopted, it won't be required anymore in mid-term.
- The Plone classic UI is not public (which in some points clients might find it ugly and problematic from the SEO point of view), with the bonus that the indexers cannot reach them.
... (add more pros)

## Cons

- The reverse proxy gets a bit more of config, but quite simple overall and repeatable.
... (add your own cons)

```
# This defines in which IP and port Plone is running.
# The default is 127.0.0.1:8080
upstream backend {
    server host.docker.internal:8080;
}
upstream frontend {
    server host.docker.internal:3000;
}

server {
  listen 80;
  server_name local.kitconcept.io;

  client_max_body_size 1G;

  access_log /dev/stdout;
  error_log /dev/stdout;

  location ~(.*)$ {
        expires +1y;
        proxy_pass http://frontend;
    }
        expires +1y;
        proxy_pass http://frontend;
    }

    if ($http_accept = 'application/json') {
        rewrite ^(.*) /VirtualHostBase/http/local.kitconcept.io/Plone/VirtualHostRoot$1 break;
        proxy_pass http://backend;
        break;
    }

    location ~ /@@images/ {
        rewrite ^(.*) /VirtualHostBase/http/local.kitconcept.io/Plone/VirtualHostRoot$1 break;
        proxy_pass http://backend;
        break;
    }

    location ~ /@@download/ {
        rewrite ^(.*) /VirtualHostBase/http/local.kitconcept.io/Plone/VirtualHostRoot$1 break;
        proxy_pass http://backend;
        break;
    }

    proxy_set_header        Host $host;
    proxy_set_header        X-Real-IP $remote_addr;
    proxy_set_header        X-Forwarded-For $proxy_add_x_forwarded_for;
    proxy_set_header        X-Forwarded-Proto $scheme;
    proxy_pass              http://frontend;
  }
}
```


nzambello:
Maybe `@@download` should be managed like `@@images` too

sneridagh:
@plone/volto-team I would schedule this for Volto 13, technically not a breaking change, but the change I think it's worth to make it noticeable. What do you think?

tiberiuichim (approved):

avoinea:
I wonder how easy would be to configure haproxy instead of nginx :thinking:

avoinea (approved):

PR #2324, opened 2021-03-30
Add showTags setting to be able to configure tags visibility

avoinea:
Add possibility to disable tags on default View via config.settings.showTags

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ### Feature
 
+- New setting, `config.settings.showTags` to be able to configure tags visibility on default View @avoinea
+
 ### Bugfix
 
 ### Internal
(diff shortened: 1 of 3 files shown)


tiberiuichim:
meh... but ok.

In plone the tags is a viewlet, so it is configurable, in a sense, as part of the overall framework. Here (in Volto) the tags are hardcoded to a certain position that may not match what what you want. There's always the possibility to customize the View.jsx, but I guess that's not desired.

Ideally we would create slots and move the tags to slots, making them configurable and thus replicate the Plone's viewlets.

sneridagh:
+1 with @tiberiuichim but the workaround is ok for me now.

sneridagh (approved):

PR #2326, opened 2021-04-01
Proposal: missdev fetch on https on ci

nzambello:
May we have this as a standard?
In projects, I usually have to fix this in order to make it work on CI for docker.

Is it a common way or is it only my issue/way to do things?


tiberiuichim:
I'm not sure what the implications are of --fetch-https (I need to use git+ssh to push), but worst case scenario, maybe we can have develop:ci for that?

nzambello:
> I'm not sure what the implications are of --fetch-https (I need to use git+ssh to push), but worst case scenario, maybe we can have develop:ci for that?

Yeah maybe, but then we should have something to understand if we are in dev env or ci env for `yarn preinstall`

tiberiuichim:
> > I'm not sure what the implications are of --fetch-https (I need to use git+ssh to push), but worst case scenario, maybe we can have develop:ci for that?
> 
> Yeah maybe, but then we should have something to understand if we are in dev env or ci env for `yarn preinstall`

I think there's some standard env variable, like CI=true ?

nzambello:
I think it could work, but I wouldn't add a `preinstall` to the template, but rather I'll document it somewhere it will be clear and easy to find.

avoinea (approved):
Great. I already did this into my projects due the same issue https://github.com/eea/ims-frontend/blob/develop/package.json#L45

If you have both `https` and `url` within `mrs.developer.json` it will configure git the right way:
```
 6 [remote "origin"]
  7     url = https://github.com/eea/volto-slate.git
  8     fetch = +refs/heads/*:refs/remotes/origin/*
  9     pushurl = git@github.com:eea/volto-slate.git
```

pnicolli:
LGTM, the configuration created by ``--fetch-https`` already sets https as the fetch method and leaves git+ssh as the push methos, as stated by @avoinea 
Imho, update changelog and this is good to go

pnicolli (approved):

tiberiuichim (approved):

sneridagh (approved):

Issue #2330, opened 2021-04-02
Stop changing block id onMutateBlock


avoinea:
This approach generates a lot of issues when used together with DX Blocks Layout Editor.

**How to reproduce:**

* Go to DX Blocks Layout Editor on you content-type:
* Add a new text block and make it fixed, required, etc.
* Now click on the plus sign and add a new image block
* All the settings on the text field are lost

Proposed solution:
* Stop using the text uid for the new block on mutateBlock and generate new one:

https://github.com/plone/volto/blob/5daa6a128a226a3e54da05e53e33c72a0a8fe60b/src/helpers/Blocks/Blocks.js#L154

tiberiuichim:
I don't think it will be a problem to fix it as you propose. Push a PR

PR #2334, opened 2021-04-05
fix: loosing block settings when adding new block

nileshgulia1:
addresses: #2330

### DIFF
--- a/src/helpers/Blocks/Blocks.js
+++ b/src/helpers/Blocks/Blocks.js
@@ -123,7 +123,7 @@ export function addBlock(formData, type, index) {
   const blocksLayoutFieldname = getBlocksLayoutFieldname(formData);
   const totalItems = formData[blocksLayoutFieldname].items.length;
   const insert = index === -1 ? totalItems : index;
-
+  const trailId = formData[blocksLayoutFieldname].items[index];
   return [
     id,
     {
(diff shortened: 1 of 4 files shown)


nileshgulia1:
@sneridagh @tiberiuichim In the above scenario, we're "adding" another block right after the current one instead of mutation.  I intend to move the 'block adding' logic inside the addBlock helper. 
Add an extra text block after adding a block (let's say image) from the last one.

I also propose another solution by keeping the block mutation intact by just copying the existing block settings into the new block.

avoinea:
@nileshgulia1 We should try not to change the current behavior and insert new blocks before the Text one.

nileshgulia1:
Demo: 


https://user-images.githubusercontent.com/22280901/113610901-5ebc5500-966b-11eb-8c37-5d0d7b338f1c.mov

avoinea:
Duplicate #2337

src/helpers/Blocks/Blocks.js
```diff
@@ -160,31 +160,39 @@ export function mutateBlock(formData, id, value) {
...
+    return {
+      ...formData,
+      [blocksFieldname]: {
+        ...formData[blocksFieldname],
+        [idnewBlock]: value || null,
+      },
+      [blocksLayoutFieldname]: {
+        items: [
+          ...formData[blocksLayoutFieldname].items.slice(0, index),
+          idnewBlock,
```
nileshgulia1: Replacing the current block with the newer block created. I like this :)

nileshgulia1:
Continued in #2337

PR #2336, opened 2021-04-05
Remove InlineForm default focus on first input

avoinea:
As InlineForm is mostly used within SidebarPortal focusing on the first input by default would lead to a lot of issue within blocks main area.
The most annoying one is:
* Scroll-to-top effect when adding new block with title input within SidebarPortal


tiberiuichim (changes requested):
Let's make focus-stealing as an optional param to InlineForm?

avoinea:
@tiberiuichim done

PR #2337, opened 2021-04-05
Use `onInsertBlock` callback when adding new blocks if available

avoinea:
See #2330

### DIFF
--- a/src/helpers/Blocks/Blocks.js
+++ b/src/helpers/Blocks/Blocks.js
@@ -151,45 +151,34 @@ export function addBlock(formData, type, index) {
   ];
 }
 
-export function mutateBlock(formData, id, value) {
-  const { settings } = config;
+export function addBlockBefore(formData, id, value) {
   const blocksFieldname = getBlocksFieldname(formData);
   const blocksLayoutFieldname = getBlocksLayoutFieldname(formData);
-  const index = formData[blocksLayoutFieldname].items.indexOf(id) + 1;
+  const index = formData[blocksLayoutFieldname].items.indexOf(id);
 
-  // Test if block at index is already a placeholder (trailing) block
-  const trailId = formData[blocksLayoutFieldname].items[index];
-  if (trailId) {
-    const block = formData[blocksFieldname][trailId];
-    if (!blockHasValue(block)) {
-      return {
-        ...formData,
-        [blocksFieldname]: {
-          ...formData[blocksFieldname],
-          [id]: value || null,
-        },
-      };
-    }
-  }
-
-  const idTrailingBlock = uuid();
-  return {
-    ...formData,
-    [blocksFieldname]: {
-      ...formData[blocksFieldname],
-      [id]: value || null,
-      [idTrailingBlock]: {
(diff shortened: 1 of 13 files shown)


src/helpers/Blocks/Blocks.js
```diff
@@ -152,44 +152,32 @@ export function addBlock(formData, type, index) {
...
+    {
+      ...formData,
+      [blocksFieldname]: {
+        ...formData[blocksFieldname],
+        [newBlockId]: value || null,
+      },
+      [blocksLayoutFieldname]: {
+        items: [
+          ...formData[blocksLayoutFieldname].items.slice(0, index),
+          newBlockId,
```
nileshgulia1, line 175: Replacing the current block with newer block created. I like this :)

nileshgulia1, line 175: Replacing the current block with newer block created. I like this :)

src/helpers/Blocks/Blocks.js
```diff
@@ -152,44 +152,32 @@ export function addBlock(formData, type, index) {
 }
 
 export function mutateBlock(formData, id, value) {
```
nileshgulia1, line 154: Where are we calling this method?

src/helpers/Blocks/Blocks.js
avoinea: :wink:

src/helpers/Blocks/Blocks.js
avoinea: `mutateBlock` in not called in the core, but maybe some add-on rely on it. If we remove it, I think we need to have a breaking release.

@sneridagh ^

sneridagh:
I just experience the issue but I think here they are a lot of things in play, not only the mutate issue.

Let's talk about it tomorrow, but here my two cents:

The "mutate" story comes from the initial blocks implementation design: Rob, Timo and I designed it back in the day with the mutation feature in mind so given a text block as canonical block is able to mutate into any other block as a feature. It makes a lot of sense because in theory potentially you could transform any block into another preserving its content (and reuse it) in the mutated block. What happened is that we never really exploited this. IMO, this is still relevant and we might want to use it in the Quanta block toolbar,  (btw, Gutenberg has it).

Inside the DX block editor we do not want to mutate the block, but add a new one when clicking into the + button, probably do not add a new text block at the bottom. That is a complete different story IMO and should be addressed differently.

I guess we are using the `BlocksForm` for render the edit form, can we somehow overload the + button in that context?

tiberiuichim:
hah... I just wrote to Alin saying the same thing... that mutate block is not just adding a new block, it's block transformations.

avoinea:
While I'm in for the block transformation idea, I don't think the add button is the right one for this purpose. Also the current implementation of the `mutateBlock` helper method is far away from this purpose.

Thus, IMHO, right now it's doing more harm than good.

avoinea:
But indeed, maybe we shouldn't touch `onMutateBlock` for this PR purpose and add new handler.

sneridagh (approved):

PR #2341, opened 2021-04-06
fix universal link when no item content obj passed

nileshgulia1:


tiberiuichim:
@nileshgulia1 A more accurate changelog entry is needed. A testcase might be nice to have as well

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 
 ### Bugfix
 
+- fix universal link @nileshgulia1s
```
pnicolli, line 13:
```suggestion
- fix universal link when no item content obj passed @nileshgulia1
```
I would add a bit more explanation here, just to make it clearer. The PR title is fine imho

pnicolli (commented):
I'll add this as a comment review since there's just really a non-blocking minor suggestion I made in the comments.
Apart from my comment on the changelog, the PR is good to go.
Another way to improve this component, imho, would be to add ``.isRequired`` to the ``'@id': PropTypes.string,`` field of the item prop, because we want that in case the ``item`` prop is passed.

CHANGELOG.md
nileshgulia1: I fixed the typo in changelog and made it more accurate. Also added `isRequired` propType to `item[@id]`

pnicolli (approved):

nileshgulia1:
@sneridagh We can also merge this?

tiberiuichim (approved):

sneridagh (approved):

PR #2343, opened 2021-04-07
fix default value bug if default = true

AlexBueckig:
Using a default value of true for the checkbox widget using the InlineForm makes the widget unuseable since the logical operator always returns true for the comparison in the InlineForm.jsx line 97. Using nullish coalescing operator `??` instead of logical or `||` fixes the issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
 ### Feature
 
 ### Bugfix
-
+- Fix default value for checkbox widget @alexbueckig
 ### Internal
 
 ## 12.6.1 (2021-04-06)
--- a/src/components/manage/Form/InlineForm.jsx
+++ b/src/components/manage/Form/InlineForm.jsx
@@ -94,7 +94,7 @@ const InlineForm = ({
               focus={index === focusIndex}
               value={
                 'default' in schema.properties[field]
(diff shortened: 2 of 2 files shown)


AlexBueckig:
Btw. I don't know why the plone contributor aggreement check fails. I signed the agreement back in December 2019 on the volto sprint in Bonn. As you can see on my github profile I'm also verfified member of the Plone Foundation. Maybe you could look into it? :)

tisto:
@polyester AlexBueckig signed the plone contributor agreement at the Beethoven Sprint 2019 and I sent it to the foundation back then. Could you please double-check if the contributor agreement reached the foundation properly? If that's the case I can take care of adding him to the list on gh.

tisto:
@AlexBueckig thanks for your PR! We will look into the contributor agreement issue.

tiberiuichim (approved):

nileshgulia1 (approved):
LGTM. I was also facing the issue. Thanks for the fix!

avoinea (approved):

polyester:
@tisto yes, his agreement is in place

tisto:
@AlexBueckig your contributor agreement is valid and you are member in the plone org on gh. The check is based on your email address, so maybe that doesn't match. Let's see if it works for your next PR. :)

PR #2345, opened 2021-04-07
Fix for forms in content types, the fieldset was not being passed ove…

sneridagh:
…r to the field. This affected form generation ids and labels.


sneridagh:
@polyester we recently added to Volto core the `objectListWidget` (a sort of improved DataGridField, Volto way). I did my best on a11y side, but it would be great if someone else could take a look. I hope the improvement in this PR is in the right direction, in such a complex widgets, I'm kind of lost... Any suggestion or pointer to good resources (shouldn't be necessarily React ones) are welcome.

You can find the current shape of it in the StoryBook: https://docs.voltocms.com/storybook/?path=/story/widgets-object-list-widget--object-list-widget

The other that might need love at some point is the `ObjectBrowser`:
https://docs.voltocms.com/storybook/?path=/story/object-browser--connected

Sprint worth material for sure but if we can already pave the way it would be great.

tiberiuichim (approved):

PR #2351, opened 2021-04-10
Razzle 4 - Webpack 5

sneridagh:
Made it work enough to see the main build pass. I had to disable several plugins, but the result is not very worthy from my point of view for now. Builds are slower, and output files are great in size... so I don't see any immediate value to pursue see this done.

We put a lot of work and effort in the build system and adapt the add-on infrastructure to match it, make them work again in Webpack 5 is going to be hard. Also found that most of the changes introduced (eg. resolver plugins API) the documentation sucks, and no examples out there to look upon.

I leave it here if someone want to continue the work.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,6 +10,7 @@ junit.xml
 eslint.xml
 yarn-error.log
 build
+cache
 
 # Other
 .DS_Store
--- a/package.json
+++ b/package.json
@@ -228,10 +228,10 @@
     "node": "^10 || ^12 || ^14"
   },
   "dependencies": {
-    "@babel/plugin-proposal-export-default-from": "7.10.4",
-    "@babel/plugin-proposal-export-namespace-from": "7.10.4",
-    "@babel/plugin-proposal-json-strings": "7.10.4",
-    "@babel/plugin-proposal-throw-expressions": "7.10.4",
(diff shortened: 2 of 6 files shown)


pnicolli:
@sneridagh I was looking into this since we use razzle-plugin-scss and the version for razzle 3 still uses node-sass :/
Maybe we should consider using webpack 4 with razzle 4? It's supported and looks like its documented as of today. I think I will try to find some free time to experiment a little with this.

davisagli:
> Maybe we should consider using webpack 4 with razzle 4? It's supported and looks like its documented as of today. I think I will try to find some free time to experiment a little with this.

@sneridagh @pnicolli I started some work in this direction in #3536

tiberiuichim:
`src/babel.js` is used by current projects. Just to keep this in mind. https://github.com/eea/freshwater-frontend/blob/master/babel.config.js#L1

davisagli:
@tiberiuichim I guess that comment was meant for #3536. I brought back babel.js there

sneridagh:
Superseded by: #3536

ichim-david:
@sneridagh  I've deleted the branch since the pull request was closed and we now have razzle 4 and webpack 5 since 17.
There are way too many open branches left that should be deleted so I will do some cleaning up when time permits.

PR #2356, opened 2021-04-13
Add node support section to README.md

tisto:


tisto:
@sneridagh Node 10 release cycle is about to end at 2021-04-30. We have to come up with an official policy to drop support for old node versions. My proposal would be that we support all Node versions that are currently supported and drop them as soon as they run out of official support.

sneridagh (approved):

sneridagh:
Also FYI https://docs.voltocms.com/releases/#releases and https://docs.voltocms.com/releases/#nodejs are in place.

PR #2358, opened 2021-04-14
Improve error handling in SSR when an error occurs in the code

sneridagh:


src/server.jsx
```diff
@@ -132,6 +132,9 @@ function setupServer(req, res, next) {
     // Displays error in console
     console.error(error);
 
+    // If error happens in Volto code itself
+    if (!error.status) res.status(500);
+
     res
       .status(error.status)
```
ksuess, line 139:
```suggestion
    // If error happens in Volto code itself: 500
    res
      .status(error.status || 500)
```

ksuess (changes requested):
Maybe I'm wrong but I think otherwise the change has no effect.

src/server.jsx
sneridagh: Ok, yeah, I will correct it, I also think it's better.

PR #1901, opened 2020-10-14
Babel view

giuliaghisini:
Implemented babel-view when translating an object and when editing an object that is already translated, to compare them.

PR #2366, opened 2021-04-20
Adapted babel view to volto12

giuliaghisini:
this pr replaces this one https://github.com/plone/volto/pull/1901

Quick demo:

https://user-images.githubusercontent.com/486927/115866712-58cfbd80-a43a-11eb-9ea8-12364c40fbfb.mov


src/components/manage/Blocks/Block/EditBlockWrapper.jsx
```diff
@@ -19,11 +19,11 @@ const messages = defineMessages({
 
 const EditBlockWrapper = (props) => {
   const hideHandler = (data) => {
-    return !!data.fixed || !blockHasValue(data);
+    return !!data.fixed || !(blockHasValue(data) && props.blockProps.editable);
```
tiberiuichim, line 22: Not sure about the order of operations here. I'd add some parens.

tiberiuichim:
```
return (!!data.fixed || !(blockHasValue(data)) && props.blockProps.editable);
```

src/components/manage/Blocks/Block/BlocksForm.jsx
```diff
@@ -32,6 +32,7 @@ const BlocksForm = (props) => {
     manage,
     children,
     blocksConfig = config.blocks.blocksConfig,
+    editable,
```
tiberiuichim, line 35: can we do `editable=true` here as default? Otherwise we'll have a ton of code to migrate, we use BlocksForm in other places as well.

src/components/manage/Blocks/Description/Edit.jsx
```diff
@@ -50,6 +50,16 @@ class Edit extends Component {
     onAddBlock: PropTypes.func.isRequired,
     onFocusPreviousBlock: PropTypes.func.isRequired,
     onFocusNextBlock: PropTypes.func.isRequired,
+    editable: PropTypes.bool,
```
tiberiuichim, line 53: This means that we have to migrate all our existing blocks, to add this prop?

There's already support for readOnly blocks, if you pass data.readOnly. Can we hitchike on that?

@avoinea Is it correct what I'm saying?

src/components/manage/Form/Form.jsx
```diff
@@ -82,6 +82,9 @@ class Form extends Component {
     description: PropTypes.string,
     visual: PropTypes.bool,
     blocks: PropTypes.arrayOf(PropTypes.object),
+    isFormSelected: PropTypes.bool,
+    onSelectForm: PropTypes.func,
```
tiberiuichim, line 86: This has a smell. What's the idea?

tiberiuichim (changes requested):

tiberiuichim:
@giuliaghisini What about using the BlockView components instead of introducing the editable?

I'm conflicted about it. There's already readOnly support in some of the blocks (maybe all?) if you mixin data.readOnly for the blocks (@avoinea will know more), but I'm more concerned about introducing a heavy burden on the rest of the ecosystem. All existing blocks will have to be migrated, to include the editable option and implement a certain behavior.

src/components/manage/Blocks/Block/BlocksForm.jsx
giuliaghisini: yes sure

src/components/manage/Blocks/Description/Edit.jsx
giuliaghisini: no, you don't have to migrate your blocks. For volto blocks, default value for editable is 'true'. 
For readonly blocks implemented by @avoinea i can only see the readOnlyBehavior field in ContentTypeLayout, but i can't understand how it works.. maybe @avoinea can you show us?

src/components/manage/Form/Form.jsx
giuliaghisini: isFormSelected is a boolean that tells if form is selected. This is needed when we are in babel view and we focus on a field of source translation object to view, for example, the configuration of a block in sidebar. 

onSelectForm is a function to set the state variable 'formSelected' to know in Edit or Add mode, which form is selected, again for the reason of showing the correct data in the sidebar.

You could better understand if you look at Edit.jsx and Add.jsx components

giuliaghisini:
> BlockView



> @giuliaghisini What about using the BlockView components instead of introducing the editable?
> 
> I'm conflicted about it. There's already readOnly support in some of the blocks (maybe all?) if you mixin data.readOnly for the blocks (@avoinea will know more), but I'm more concerned about introducing a heavy burden on the rest of the ecosystem. All existing blocks will have to be migrated, to include the editable option and implement a certain behavior.

I don't know wich are the BlockView components, maybe @avoinea could explain us. Anyway, existing volto blocks doesn't need to be migrated, because i've introduced the editable variabile wich default is 'true'.

tiberiuichim:
@giuliaghisini Sorry, somehow I've edited your comment. 

> What about using the BlockView components instead of introducing the editable?

Sorry for the "short hand". Any block view component, such as:

https://github.com/plone/volto/blob/master/src/components/manage/Blocks/Text/View.jsx
https://github.com/plone/volto/blob/master/src/components/manage/Blocks/Table/View.jsx
and so on

giuliaghisini:
> @giuliaghisini Sorry, somehow I've edited your comment.
> 
> > What about using the BlockView components instead of introducing the editable?
> 
> Sorry for the "short hand". Any block view component, such as:
> 
> https://github.com/plone/volto/blob/master/src/components/manage/Blocks/Text/View.jsx
> https://github.com/plone/volto/blob/master/src/components/manage/Blocks/Table/View.jsx
> and so on

oh, i understand. No, because i think in edit mode is useful to the editor to view block configuration available in sidebar..

src/components/manage/Blocks/Description/Edit.jsx
sneridagh: Maybe then the wording is not appropriate? I had the same confusion as @tiberiuichim did. @giuliaghisini what is the ultimate purpose of this prop?

sneridagh:
@giuliaghisini OMG, I merged the engine improvement, and now it's another conflict 😞 sorry about that.

On the other hand, since the changes modify how the blocks are selected and they relate to the sidebar, I think it's relevant to add them to the babel view.

src/components/manage/Blocks/Description/Edit.jsx
giuliaghisini: This props enable user to select the block and view it's configuration in the sidebar, to compare or replicate it in the translation content that it's editing in that moment, but without editing the block.

src/components/manage/Form/Form.jsx
```diff
@@ -110,6 +113,9 @@ class Form extends Component {
     blocks: [],
     pathname: '',
     schema: {},
+    isFormSelected: true,
+    onSelectForm: null,
+    editable: true,
```
sneridagh, line 118: I see it now 👍

sneridagh (commented):
Overall LGTM, code wise. 

A bit concerned about the complexity that we are adding to the overall Add/Edit/Form/Blocks Engine pack, but we have to devote an entire sprint to simplify it and make it more modular yet...

sneridagh:
Now, regarding UI/UX 😊

I like the toolbar, after a bit of hesitation about the fact that if it should be enclosed into a ... (more) button for consistency and after a check to Albert, I'm fine with it. In edit mode, we are not going to have much more of that "Actions on edit" (I hope/guess) We can always group them at a later point.

However some issues I would like to improve: 

Screenshot 2021-04-21 at 12 23 06

the vertical space is not even with the other buttons (maybe because the other has the circle?)

image

Then the menu opened is also displaced, because it's too short I guess... could we improve this somehow?

Regarding naming of the languages. It's something that I think we can improve too, in this menu and in the add menu too. So instead of the abreviature, use the full language name (like in the language selector).

Talking about the language selector, we don't have any indicator of the current language. Could we set the current language to bold (like in Classic).

Thanks for the effort!!

sneridagh:
One more thing, in case you are not aware of, we can now implement multilingual Cypress tests :) /cc @tisto 

I enabled not long ago the ability to load an optional add-on for test purposes from the command line (see the Sandbox core tests ones). I can take care of the boilerplate If you help me a bit with the Cypress tests :) 

What do you think?

sneridagh:
BTW, priority is merge this PR. Then Volto 13. @plone/volto-team what do you think?

giuliaghisini:
> Now, regarding UI/UX 😊
> 
> I like the toolbar, after a bit of hesitation about the fact that if it should be enclosed into a ... (more) button for consistency and after a check to Albert, I'm fine with it. In edit mode, we are not going to have much more of that "Actions on edit" (I hope/guess) We can always group them at a later point.
> 
> However some issues I would like to improve:
> 
> Screenshot 2021-04-21 at 12 23 06
> 
> the vertical space is not even with the other buttons (maybe because the other has the circle?)
> 
> image
> 
> Then the menu opened is also displaced, because it's too short I guess... could we improve this somehow?
> 
> Regarding naming of the languages. It's something that I think we can improve too, in this menu and in the add menu too. So instead of the abreviature, use the full language name (like in the language selector).
> 
> Talking about the language selector, we don't have any indicator of the current language. Could we set the current language to bold (like in Classic).
> 
> Thanks for the effort!!

I've analyzed the space between buttons, but.. it's the same. Maybe the the visual effect is different because save and cancel buttons are rounded... and compare translantions button have only icons.. I'll use the Button component from semantic-ui but the result will be the same..

for the others things.. now i'll check what is possibile to do.

sneridagh:
> I've analyzed the space between buttons, but.. it's the same. Maybe the the visual effect is different because save and cancel buttons are rounded... and compare translantions button have only icons.. I'll use the Button component from semantic-ui but the result will be the same..

I see, I suspected that. Maybe add the circle, but with transparent color, so it "fakes" the same vertical space and the perceived vertical space will be the same? Just wondering, haven't tried it.

nzambello:
@sneridagh another thing we would like is to take all the actions and menus under a single toolbar menu for translations like we have a single translations menu on Plone classic.
We cannot refactor the whole thing now because we have no more time left for this, but it will surely be a nice (near?) future improvement

giuliaghisini:
> Now, regarding UI/UX 😊
> 
> I like the toolbar, after a bit of hesitation about the fact that if it should be enclosed into a ... (more) button for consistency and after a check to Albert, I'm fine with it. In edit mode, we are not going to have much more of that "Actions on edit" (I hope/guess) We can always group them at a later point.
> 
> However some issues I would like to improve:
> 
> Screenshot 2021-04-21 at 12 23 06
> 
> the vertical space is not even with the other buttons (maybe because the other has the circle?)
> 
> image
> 
> Then the menu opened is also displaced, because it's too short I guess... could we improve this somehow?
> 
> Regarding naming of the languages. It's something that I think we can improve too, in this menu and in the add menu too. So instead of the abreviature, use the full language name (like in the language selector).
> 
> Talking about the language selector, we don't have any indicator of the current language. Could we set the current language to bold (like in Classic).
> 
> Thanks for the effort!!

For the other things, i tryed to position the ComparingLanguage menu beside the toolbar button, as it was possibile, displayed the full language name, and added style for selected language in LanguageSelector

sneridagh:
@giuliaghisini 

`RAZZLE_TESTING_ADDONS=core-sandbox:multilingualFixture yarn start`

sneridagh:
@plone/volto-team I need someone review my changes too. Additional eyes are always welcomed :)

sneridagh (approved):
And from my side, we are good to go.

sneridagh:
Added a demo in the description for future references.

tiberiuichim (approved):

tiberiuichim:
For the future refactoring, we need the toolbar dropdown menus to work in "levels", similar to User menu > Personal Preferences.

In the Add menu, instead of having all the possible translations listed, we'd have "Translate...", click on it, then we get to the second page where we see the list of available languages.

src/components/manage/Blocks/Description/Edit.jsx
sneridagh: However, question: If a block does not implement a mean for being "readonly" it could be modified, right? and if so, we will not only to change all the existing blocks, but also transfer that burden to the block code itself. 

Can't we make it depend on the Engine?

sneridagh:
@avoinea could you weigh on this one?
https://github.com/plone/volto/pull/2366#discussion_r619667562

is there a way in the engine for making this easier?

src/components/manage/Blocks/Description/Edit.jsx
giuliaghisini: yes.. i don't think it's easy to make it dependent from engine, because the editable param passed to block has to be passed to the Editor text component to make it disabled..

avoinea: For consistency I would use `props.readOnly` also. I don't think we can use `data.readOnly` for this purpose as this is persistent to ZODB.

PR #2367, opened 2021-04-20
Add Node 16 support

tisto:
Volto starts with Node 16 and all unit tests pass. I guess we should wait for gh actions to provide Node 16. Then do a release. @sneridagh can you point me to our Node policy in docs again? I guess it would make sense to add a paragraph there that we start to support a new node version as soon as it is out, what do you think? The sooner we start with a new version the longer an old Volto version will support it (in the future).

### DIFF
--- a/.github/workflows/core.yml
+++ b/.github/workflows/core.yml
@@ -32,6 +32,7 @@ jobs:
       - run: yarn install --frozen-lockfile
 
       # node test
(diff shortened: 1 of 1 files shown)


pnicolli (changes requested):

sneridagh:
On 26/10/2021 it becomes LTS, so scheduled to release by then!

PR #2369, opened 2021-04-20
Fix some BlockDataForm bugs

tiberiuichim:
- don't add variations field multiple times
- always clone the schema before applying enhancers


CHANGELOG.md
```diff
@@ -4,19 +4,22 @@
 
 ### Breaking
 
+- (minor) Refactor the block schema and variations mechanism
```
sneridagh, line 7: I won't call it a breaking, since the core is not using it yet, really. I would say it's a "part 2" instead, if not, we would have to leave it for 13...

docs/source/blocks/settings.md
```diff
@@ -42,13 +42,14 @@ const customBlocks = {
...
     // A block can define variations (it should include the stock, default one)
-    variations: {
-      default: { label: 'Default' },
-      custom: {
-        label: 'Custom',
-        // The variation level schema Enhancer function
-        schemaEnhancer: CustomSchemaEnhancer
-      }
+    extensions: {
+      variation: [
```
sneridagh, line 46: `extensions` but `variation` ? I would call them all in plural.

sneridagh (commented):
I'm a bit torn also in putting components in helpers, even if they are HOCs, we already regret it a couple of times... What do you think?

docs/source/blocks/settings.md
tiberiuichim: it's because it will be saved into the block data. So we'll have, for example:

```
{
    "variation": "summaryListing",
    "text": ...
}
```

The Listing block right now saves its "variation" in the "template" key: https://github.com/plone/volto/blob/f07c4e74699d7fbd202348cc1197783861eb386b/src/components/manage/Blocks/Listing/TemplateWidget.jsx#L68, though it names its "variations" as "templates" in the config.

But from my point of view, is like this: "extensions" are the available extensions. "variation" is one kind of extension, as a generic concept.

CHANGELOG.md
tiberiuichim: ok, not breaking then

tiberiuichim:
> I'm a bit torn also in putting components in helpers, even if they are HOCs, we already regret it a couple of times... What do you think?
 
We have hooks and HOCs in helpers: https://github.com/plone/volto/blob/master/src/helpers/Loadable/Loadable.js and https://github.com/plone/volto/blob/master/src/helpers/AsyncConnect/index.js#L93 to name a few. 

From my perspective they don't belong in the components folder, but to a more neutral place, though I can understand the thinking behind keeping helpers "js-only".

tiberiuichim:
- "variation" is fine as a name
- isDefault into the "variation" choice
- rename "variation choices" to "items"
- move the title into the extension definition
- refactor the hoc to get the extension id
- remove / adjust the old doc into extensions.md

sneridagh (approved):

PR #2370, opened 2021-04-21
Makefile in addon generator

tiberiuichim:
I've added a Makefile in the addon generator, it provides the ability to spin up a Volto development instance from inside an addon.


packages/generator-volto/generators/addon/templates/Makefile.tpl
```diff
@@ -0,0 +1,54 @@
...
+# From: https://coderwall.com/p/izxssa/colored-makefile-for-golang-projects
+RED=`tput setaf 1`
+GREEN=`tput setaf 2`
+RESET=`tput sgr0`
+YELLOW=`tput setaf 3`
+
+project:
+	npm install -g yo
+	npm install -g @plone/generator-volto
+	npm install -g mrs-developer
```
sneridagh, line 16: Globally?

sneridagh (approved):
LGTM, how about testing ;)

packages/generator-volto/generators/addon/templates/Makefile.tpl
tiberiuichim: It's based on Volto docs :) https://docs.voltocms.com/getting-started/install/#install-volto

tiberiuichim:
Actually I've noticed that there's no "api" part in the generator. The whole backend setup needs to be removed from the Makefile.

packages/generator-volto/DEVELOP.md
```diff
@@ -0,0 +1,18 @@
...
+    cd volto/packages/generator-volto
+    yarn
+    npm link
+
+### Generate new project
+
+    yo @plone/volto my-volto-project --skip-install
+    cd my-volto-project
+
+### Generate new add-on based on remote template
```
avoinea, line 15: I think the right title here is `Generate new add-on`

packages/generator-volto/generators/addon/templates/Makefile.tpl
avoinea: I always install these globally 😄

avoinea (approved):

PR #1916, opened 2020-10-18
handle isMulti & null value in selectWidget

nileshgulia1:
Fixes #1915 and probably #1878

PR #2371, opened 2021-04-21
Correct the selected values rendering at isMulti SelectWidget

ionlizarazu:
move isObject condition below isArray condition for a correct multi-select selected values rendering.

See: https://github.com/plone/volto/pull/1916#issuecomment-823415191

### DIFF
--- a/src/components/manage/Widgets/SelectWidget.jsx
+++ b/src/components/manage/Widgets/SelectWidget.jsx
@@ -96,17 +96,17 @@ function getDefaultValues(choices, value) {
       value: 'no-value',
     };
   }
-  if (isObject(value)) {
-    return {
-      label: value.title !== 'None' && value.title ? value.title : value.token,
-      value: value.token,
-    };
-  }
+
   if (isArray(value) && choices.length > 0) {
     return value.map((v) => ({
       label: find(choices, (o) => o[0] === v)?.[1] || v,
       value: v,
     }));
+  } else if (isObject(value)) {
+    return {
+      label: value.title !== 'None' && value.title ? value.title : value.token,
+      value: value.token,
+    };
   } else if (value && choices.length > 0) {
     return { label: find(choices, (o) => o[0] === value)?.[1] || value, value };
   } else {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@
 
 ### Breaking
 
+### Bugfix
+
+- Correct the selected values rendering at isMulti SelectWidget @ionlizarazu
+
 ### Feature
 
 - Include a pluggable architecture for pluggable render-time insertions (similar to <Portal>) @tiberiuichim


sneridagh (approved):
Hey @ionlizarazu , I vaguely remember that this had a reason (see https://github.com/plone/volto/commit/176059f82190c034b88927ba59afb536789f7cfb), at some point in some edge case, Plone returns an object (instead of an array). Might be because of fixed (choice) arrays?

My fault for not providing a proper test to catch this up... Anyways, the code seems fine since we test for all the use cases.

Maybe you have another edge case in your hands right now. If you have time, could you add a quick unit test for it? Thanks!!

ionlizarazu:
I'm trying to add a multi-selectable field in one of my custom Blocks.
I add this code  in schema.properties rendering it as multi-select field:

```
export const CardContainerSchema = (types) => ({
[...]
properties: {
    contentTypes: {
        title: 'Content Types',
        choices: types,
        isMulti: true,
    }
```

The `types` parameter is the object composed by the contentType values in the state which are modified for a correct rendering of the choices in my Block Edit.jsx file:
```
const state_types = useSelector(state => state.types.types); 
const types = state_types.map((type) => ([type.title, type.title]))
```
This is the behaviour of my block before the modification :
selectMultiple not working

After the pull request:
selectMultiple after modifications

tiberiuichim:
@ionlizarazu May I suggest using the Listing block? We need to modernize the Listing block, perhaps it's better if you rewrite some of the listing block innards to use BlockDataForm and then add a new listing template to that.

src/components/manage/Widgets/SelectWidget.jsx
```diff
@@ -96,17 +96,17 @@ function getDefaultValues(choices, value) {
...
+
   if (isArray(value) && choices.length > 0) {
     return value.map((v) => ({
       label: find(choices, (o) => o[0] === v)?.[1] || v,
       value: v,
     }));
+  } else if (isObject(value)) {
+    return {
+      label: value.title !== 'None' && value.title ? value.title : value.token,
+      value: value.token,
```
nileshgulia1, line 108: Notice that choices behave differently when comes as an object (`value.token`), I think that's what Victor was talking about!
Nonetheless, I'm okay with the schema you mentioned as that gonna work with Array of objects passed.

nileshgulia1 (approved):
LGTM when it comes to handle multiple values with the given schema!

ionlizarazu:
> @ionlizarazu May I suggest using the Listing block? We need to modernize the Listing block, perhaps it's better if you rewrite some of the listing block innards to use BlockDataForm and then add a new listing template to that.

Nice suggestion, we will look into it, even so, with this pull request we want to get the expected functionality of a multi-select field working, using the schema configuration.

tiberiuichim (approved):

PR #2374, opened 2021-04-22
[Breaking] a11y: change login form

nzambello:
- Refactor login form: it had keyboard navigation issues with tabIndex
- Segment secondary text had not minimum contrast ratio, so I fixed it to textColor
- Login form is simpler and has better UX
- Add self registration config option

Screenshot_2021-04-22 Login


src/components/theme/Login/Login.jsx
```diff
@@ -222,25 +229,7 @@ class Login extends Component {
...
-                                    id="registration form"
-                                    defaultMessage="registration form"
-                                  />
-                                </Link>
-                              ),
-                            }}
-                          />
-                        </p>
-                      </Grid.Column>
+                      <Grid.Column stretched width="12"></Grid.Column>
```
tiberiuichim, line 232: I know this should be conditional on self-registration enabled. Do we have an endpoint for Plone settings?

sneridagh: Not yet... it's part of that inglorious endpoint for "bootstrap" some things here and there, we could make it a static config for now.

sneridagh (commented):
LGTM, but I would create that config for the register.

For the record, this is Quanta login form proposal, we are very close with this one :)

Screenshot 2021-04-25 at 10 56 20

Screenshot 2021-04-25 at 10 56 13

sneridagh:
As we talked in the meeting, can we quickly implement a config setting that controls the "Register" toggle? eg. `config.settings.showSelfRegistration` ? Then tie it to the AnonTools link and the link in the login.

Thanks!

nzambello:
@sneridagh added config option for self registration

sneridagh:
Ok, let's merge this for Volto 13, ok? I think it's worth it.

tiberiuichim (approved):

PR #2378, opened 2021-04-26
Remove console.error from server

nzambello:
Logging NotFound errors are Plone' stuff and responsibility, because it's not Volto's responsibility.
This could be useful to debug SSR but it's generating a lot of noise in prod env


sneridagh:
@nzambello well, not only NotFound errors are shown. I think it's valuable to log errors in there, if not, you are blind on what happens in the SSR world. I agree that NotFound errors might be useful to be filtered from here.

Could we filter them somehow?

tiberiuichim:
`@errorlog` endpoint :) so that we can read the ignored exceptions list.

nzambello:
@sneridagh I agree with that. IMO it is valuable to have that.
@tiberiuichim that could be super cool, do we have access to that in some way?

I now used an array with Plone's default for that.

sneridagh (approved):

PR #2380, opened 2021-04-26
Making objectBrowser Widget context aware

iFlameing:


src/components/manage/Widgets/ObjectBrowserWidget.jsx
```diff
@@ -264,10 +265,10 @@ export class ObjectBrowserWidgetComponent extends Component {
     ev.preventDefault();
     this.props.openObjectBrowser({
       mode: this.props.mode,
+      currentPath: this.props.location.pathname,
       onSelectItem: (url, item) => {
         this.onChange(item);
       },
-      propDataName: 'value',
```
giuliaghisini, line 270: i'm not sure about this.. why did you remove this prop?

iFlameing: because if we pass this prop then we are not passing the contexturl to ObjectBrowserBody. Also, I don't know the useCase you are having. We can make it optional like if the propDataName is passed by props we will set it otherwise null. 

`propDataName: props.propDataName ?? null ; `

sneridagh: @iFlameing it seems to me we need it for this feature: https://docs.voltocms.com/blocks/editcomponent/#propdataname-vs-dataname

That's why we need (even) better exhaustive Cypress tests, to detect this.

@iFlameing please, provide tests for this use case, and if possible for the others as well. Now we can use the package/codesandbox add-on for that.

sneridagh (changes requested):

src/components/manage/Widgets/ObjectBrowserWidget.jsx
iFlameing: ok, I will write the test tomorrow

iFlameing: I reverted this change and now it is working without breaking anything.

sneridagh (approved):

PR #2387, opened 2021-04-29
Improve Github Actions names

sneridagh:
@tisto Since all are running in parallel and once one fail, the other are cancelled, what do you think also into split the Core Main Tests into Code Analysis, Unit Tests and Acceptance Tests?

It would help to fail quick and spot problems in the pushed code and do not waste computation time.


tisto (approved):
I started to do the same for restapi. My main reason was that developers couldn't figure out that for instance flake8 or black failed. To make this more obvious I split up things. Though, that's a slippery slope and we should over-do this because a CI pipeline (running expensive steps after less expensive steps) still makes sense to me in general.

PR #2389, opened 2021-04-30
Better handling of the click outside events

sneridagh:


avoinea:
volto-slate toolbar basic text manipulation:

slate-bold

avoinea:
[slate-metadata-mentions](https://github.com/eea/volto-slate-metadata-mentions)

slate-metadata-mentions

avoinea:
[accordion-block](https://github.com/eea/volto-accordion-block)

accordion-block

avoinea:
[columns-block](https://github.com/eea/volto-columns-block)

columns-block

sneridagh:
Still not working, since floating elements attached outside (like the objectBrowser and the slate toolbar which are created and attached in the Body) falls outside. In the browser we have control, but in the toolbar we don't. That also prevents any other custom implementation that does the thing right, since from a11y perspective attaching things in the body is the correct way to go.

avoinea (approved):

PR #2394, opened 2021-05-03
feat: upgrade husky to v6 and support hooks in projects/addons dev

nileshgulia1:
We have "husky hooks" in volto-core, the idea is to add support in `volto-projects` as well as in `addons` to maintain consistency across volto projects. The problem with this is we do not have "git initialised" project generated with our scaffolding tool. As husky required `.git/` to be present at the root, we have to do it manually everytime using `git init`.

This PR copies husky configs into generator, perhaps I have to re-think about "git initialization" mechanism, whether to make it optional?( CRA make that default, wheareas create-razzle-app does not )

For addons, the husky config might come/used from projects like we have for `eslint`.


tiberiuichim:
@nileshgulia1 You could wrap the preinstall/postinstall hooks in package.json in a "exists .git dir" bash test. Something like `ls .git && husky ...` (Note, I don't know bash, just suggesting something that would avoid triggering husky if not in a git clone).

nileshgulia1:
superseded by #3842

Issue #2397, opened 2021-05-04
Sort order should not be boolean with latest restapi 


nzambello:
**Describe the bug**
With latest versions of plone.restapi, the querystring-search sort order index has changed.

So, if we were using `sort_order: true` or `sort_order: true`, now we should query for `sort_order: descending` or `sort_order: ascending`.

**To Reproduce**
Just create a listing block and set sorting order.
Easy to see using sortable title as `sort_on`

**Expected behavior**
Using a listing block ordered by title, in Edit mode when selecting "Reverse order" you should see contents with titles starting with Z or W, while finding the ones starting with A or B deselecting "Reverse order".

**Software (please complete the following information):**
 - Volto 12.14.0
 - Plone 5.2.3
 - Plone REST API 7.3.5

nzambello:
Implementing a solution (WIP)

tisto:
@nzambello maybe plone.restapi should issue a warning or so when we pass parameters that do not work as expected. We could create an issue on plone.restapi and write a test to prove the behavior, then think about the best way plone.restapi could handle this. Just an idea...

nzambello:
@tisto I think the restapi team should notify the volto team if something that can affect volto changes.
Otherwise, we should keep track of restapi changes and verify every release to work with volto having a "tested version" pinned for volto.

We are insiders and we fixed it fastly (and it's still annoying because we discovered this in production) but it is a big issue for a plone developer who's not so in the core community.

tisto:
@nzambello absolutely. To be honest I wasn't aware that we changed anything Volto-related in plone.restapi. Maybe we need to do some research to figure out how that happened and then document that properly in Volto.

PR #2375, opened 2021-04-24
Improvequerystringwidget

sneridagh:

PR #2398, opened 2021-05-04
Fix sort order: using 'ascending' and 'descending'

nzambello:
Fixes: #2397

Now listing blocks can order items with "reverse" using latest plone.restapi's rules.

Using a listing block ordered by title, in Edit mode when selecting "Reverse order" you can see contents with titles starting with Z or W, while finding the ones starting with A or B deselecting "Reverse order".

This PR uses two approaches:
- fix in the action which prevents us to migrate blocks to fix it, working with existing listing blocks
- ListingData saves correctly new blocks

### DIFF
--- a/src/actions/querystringsearch/querystringsearch.js
+++ b/src/actions/querystringsearch/querystringsearch.js
@@ -24,7 +24,7 @@ export function getQueryStringResults(path, data, subrequest, page) {
 
   // fixes https://github.com/plone/volto/issues/2397
   if (requestData?.sort_order !== null) {
-    if (typeof requestData.sort_order === "boolean") {
+    if (typeof requestData.sort_order === 'boolean') {
       requestData.sort_order = requestData.sort_order
         ? 'descending'
         : 'ascending';
--- a/src/actions/querystringsearch/querystringsearch.js
+++ b/src/actions/querystringsearch/querystringsearch.js
@@ -24,7 +24,7 @@ export function getQueryStringResults(path, data, subrequest, page) {
 
   // fixes https://github.com/plone/volto/issues/2397
   if (requestData?.sort_order !== null) {
-    if (requestData.sort_order === true || requestData.sort_order === false) {
+    if (typeof requestData.sort_order === "boolean") {
       requestData.sort_order = requestData.sort_order
         ? 'descending'
         : 'ascending';
--- a/src/actions/querystringsearch/querystringsearch.js
+++ b/src/actions/querystringsearch/querystringsearch.js
@@ -9,9 +9,10 @@ import config from '@plone/volto/registry';
  */
 export function getQueryStringResults(path, data, subrequest, page) {
   const { settings } = config;
-  // fixes https://github.com/plone/volto/issues/1059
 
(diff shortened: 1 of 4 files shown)


giuliaghisini (approved):

src/actions/querystringsearch/querystringsearch.js
```diff
@@ -21,6 +22,15 @@ export function getQueryStringResults(path, data, subrequest, page) {
     });
   }
 
+  // fixes https://github.com/plone/volto/issues/2397
+  if (requestData?.sort_order !== null) {
+    if (requestData.sort_order === true || requestData.sort_order === false) {
```
pnicolli, line 27:
```suggestion
    if (typeof requestData.sort_order === "boolean") {
```

How about this as a simpler check?

pnicolli (approved):
LGTM

tiberiuichim (approved):

pnicolli:
I like that this is backwards compatible.
I am wondering if we should (in a separate PR is fine) add a deprecation warning in the action, if the old behavior is detected. Then we could plan removing that fix in the action and only keep the updated ``ListingData`` with the updated ``onChange`` in che ``CheckboxWidget``.
Thoughts?

sneridagh (approved):
Ok, we are in the middle of a major release AND a refactor of the whole listing block (see #2375, which I'm almost done) :)

Given the special situation, I would merge it not in master but in a 12 branch, make just another 12 release. What do you think?

pnicolli:
@sneridagh Seems fine to me.

sneridagh:
@pnicolli So it's a fix of the current version, and like this is fine to me if it covers the previous values, although it's not elegant.

So, as I planned to include the listing refactor in 13, using schemas, let's see if we can deal with it (and also the previous value), if not, include a breaking change.

sneridagh:
@nzambello prettier! :)

nzambello:
@sneridagh so will you replicate or check this with the new refactored QuerystringWidget and we'll have a 12.x branch with this?
I think should keep trace of this in the changelog of master anyway, or we'll have a release on npm without changelog.

sneridagh (approved):

sneridagh:
@pnicolli btw, I think the solution in this PR is enough elegant :) Do you think we should provide a proper dropdown choices "ascending/descending"? or continue with the boolean? I think it's a boolean because in classic it's a boolean.

/cc @plone/volto-team

pnicolli:
@sneridagh I'm fine with the boolean, it's always been clear to users afaik. I'm also fine with the PR, it's a good fix for the situation.

My proposal was just to plan a moment in the future where the change to the querystringsearch action is removed, since it will be useless code at some point, maybe 3-4 majors from now. The deprecation warning, if we want to do this, should be logged when that code is run, and an upgrade guide should be written about this, of course. This is why I suggested to do that in a separate PR :)
Two years from now, I wouldn't want to see a bunch of BBB code like this stuck in Volto core, it's a big enough project to maintain 😬 
Of course, all of this does not take your QuerystringWidget rework into consideration, I haven't looked into it yet, so maybe all of what I'm saying is useless 😛

PR #2407, opened 2021-05-06
Make listing block ssr

tiberiuichim:


tiberiuichim:
So, my feeling after implementing this is that we should use asyncConnect() on top of the blocks and not use the blocksConfig.listing.getAsyncProps 

Wrapping the blocks view component is an edge case, so we can document the things to be aware of. 

If we keep the asyncConnect then we still keep the semantic of "async prop injection", otherwise we're just dispatching some actions. By keeping the asyncConnect as API we also have a chance to cleanup and overhaul this story.

If asyncConnect wouldn't already exist in Volto, then probably the config setting would be no-brainer.

sneridagh:
wooohoooo! Looks great already!!

ksuess:
This helps a lot. 
BTW this does affect first level blocks, not listing blocks in grids or columns blocks, doesn't it? That would be super cool, if even these were SSR…

tiberiuichim:
@ksuess it works block-in-block story (columns block, tabs, etc). I wouldn't have any any other way: https://github.com/plone/volto/blob/77a4363f6784aca8d138dfc973174eca033f692d/src/helpers/Blocks/Blocks.js#L344

tiberiuichim:
@ksuess I think using the asyncPropExtenders it would be possible to replace the `content` asyncConnect promise and use the implementation here. So if it's something that you need right now, that's a way to go. 

Overall, we're still thinking about API, I'd like to explore the asyncConnect version first, as expressed in my comment https://github.com/plone/volto/pull/2407#issuecomment-833827725

tiberiuichim:
@sneridagh Maybe we can get this in time for Volto 14? I'm ok with the current shape... maybe rename `getAsyncProps` to `getAsyncData`, as we don't have the props injection aspect.

giuliaghisini (approved):

sneridagh:
@tiberiuichim let's do the renaming and go for it then? Not sure also about the "visitor" naming :)

tiberiuichim:
@sneridagh I've asked a colleague to write a test for the visitor function. I'm using the [visitor pattern](https://en.wikipedia.org/wiki/Visitor_pattern) there, so I thought I'd make it obvious just to reference something familiar. That `const visitor` could be renamed... but it's more or less internal name in the `fetchContent` function.

PR #2421, opened 2021-05-10
Generic link plugin

tiberiuichim:


sneridagh (approved):
LGTM, might need some additional cleanup (comments leftovers) and eventually move it to a more semantic/sensitive naming schema? AnchorPlugin was the name of the DraftJS plugin that we forked.

Issue #2434, opened 2021-05-12
UniversalLink: respect target for internal links


ksuess:
**Describe the bug**
link with target "same window" should open in same window, not new window.

**To Reproduce**
1. Slate: Create external link with href "/dokumentation/suche?q=&f=kompasscomponent_agg.kompasscomponent_token%3ABEB2k&l=list&order=asc&p=1&s=10&sort=bestmatch" and target "same window"
2. save
3. click on link and see a new tab opening with the target, instead of opening in the same window.

**Expected behavior**
link opens in same window

**Software (please complete the following information):**
 - Volto Version 12.4.2

ksuess:
I close this issue, as with **no selection of a target** the link opens in the same window as expected.
![image](https://user-images.githubusercontent.com/1144737/118230971-b16d0600-b48e-11eb-89e0-b10a5431293b.png)

PR #2436, opened 2021-05-12
Universallink: keep target of internal link

ksuess:
if target='_self' this should not be overridden by '_blank'

fixes https://github.com/plone/volto/issues/2434

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Bugfix
 
+- UniversalLink: Respect target of internal link. @ksuess
+
 ### Internal
 
 ## 12.14.2 (2021-05-10)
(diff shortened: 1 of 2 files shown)


sneridagh:
Could be that this fix only applies to slate? In drafts, the openLinkInNewTab is a boolean. Maybe I'm mistaken... do we have tests for this?

ksuess:
Yes that reduces the issue to the question if there is a need for a choice of the target or if it's enough to let the editor decide if the href is internal or external like draftjs does.

sneridagh:
I think we need it! Sorry when I meant draftjs, I meant in the default blocks too. But that's a boolean now, so the condition should be different I guess. I assume the value in Slate is "_blank"? Ok, since tomorrow we have to start working in Volto's Slate adoption and it's surroundings, it's a good way to start :) 

Also, related, @tiberiuichim already started to work on a PR to make the actual link plugin to be "universal" and not only drafts related. We can talk about it tomorrow too.

PR #2443, opened 2021-05-13
Fix edit warning

tiberiuichim:
There's a forward ref that's passed to Toolbar, but the toolbar component doesn't forward refs. I think it was left out by mistake.

This fixes a browser console warning.


nzambello (approved):

sneridagh:
@giuliaghisini is this correct? 
https://github.com/plone/volto/blob/40fe11e23e249155575584bab1f9169b4ceaa42e/src/components/manage/Edit/Edit.jsx#L227

(it's not called)

Then it's passed to:
https://github.com/plone/volto/blob/40fe11e23e249155575584bab1f9169b4ceaa42e/src/components/manage/Edit/Edit.jsx#L388

sneridagh:
@tiberiuichim I think the purpose of this was not to be used in that component but as the reference to make the calculations in https://github.com/plone/volto/blob/ba582b98e5731b2dd45c6109223368f8df549160/src/components/manage/Multilingual/CompareLanguages.jsx#L45

What I don't know is why it was complaining, it seems legit, right? So I would be wary about this...

tiberiuichim:
https://github.com/plone/volto/blob/40fe11e23e249155575584bab1f9169b4ceaa42e/src/components/manage/Edit/Edit.jsx#L227

That function is not called, right? :)

tiberiuichim:
Yeah, so the proper fix is probably to call the function.

giuliaghisini:
I don't remember what it was done for in particular.. you need to test what happens when you have your site in multilingual mode and you have some translations of the page/content you're editing..

tiberiuichim:
@giuliaghisini Sorry, I don't have access to a multilingual website. Can we fix this warning? It's annoying :)

giuliaghisini (approved):

PR #2445, opened 2021-05-13
Adding hover effect on ObjectBrowserNav icon

iFlameing:
screenshot

Screenshot 2021-05-13 at 7 48 20 PM


src/components/manage/Sidebar/ObjectBrowserNav.jsx
```diff
@@ -90,7 +90,13 @@ const ObjectBrowserNav = ({
               {item.title}
             </span>
             {item.is_folderish && mode === 'image' && (
-              <Icon name={rightArrowSVG} size="24px" />
+              <span>
```
tisto, line 93: @iFlameing was this span element added by accident? If we change the HTML structure we might have to consider this as a breaking change. I don't see the use of this span element at first, could you elaborate?

sneridagh: +1 seems unrelated or a left over?

iFlameing: sorry!  I added this to increase the hover effect's space, but it doesn't look good as per UI and I forget to remove it. I will fix it today.

sneridagh:
Once selected, the hover on the arrow (which works in master) is gone. Can we have both?

iFlameing:
@sneridagh tried to fix it please review it.

sneridagh (approved):
LGTM!
image

PR #2446, opened 2021-05-13
Remove unused pathname parameter from Blocks/Listing/View.jsx

ionlizarazu:
The Listing block with criteria is not rendering correctly on a non-multilingual homepage.
the Listing View.jsx component supposedly receives a `pathname` [line 9](https://github.com/plone/volto/blob/bd1201a4f918c3e2d47c0360b3b3734a3c8aec0f/src/components/manage/Blocks/Listing/View.jsx#L9) parameter, but it never comes and everytime is _undefined_. See [DefaultView](https://github.com/plone/volto/blob/bd1201a4f918c3e2d47c0360b3b3734a3c8aec0f/src/components/theme/View/DefaultView.jsx#L46) and [RenderBlocks](https://github.com/plone/volto/blob/bd1201a4f918c3e2d47c0360b3b3734a3c8aec0f/src/components/theme/View/RenderBlocks.jsx#L32) parameters.
Being pathname _undefined_, if the path parameter value is an empty string (it is at root/homepage) ListingBody component is called with `path = undefined` at [line 13](https://github.com/plone/volto/blob/bd1201a4f918c3e2d47c0360b3b3734a3c8aec0f/src/components/manage/Blocks/Listing/View.jsx#L13).
If any criteria is defined, the getQueryStringResults action is called with a undefined path, that becomes into a error.

Here the visual example:
listing block not rendering

In this commit I removed pathname parameter from the View.jsx, because it has never any data.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Bugfix
 
+- Fix the Listing block with criteria to render correctly on a non-multilingual homepage. @ionlizarazu
+
 ### Internal
 
 - Only log changes to po (`poToJson`) if running as a script @sneridagh
--- a/src/components/manage/Blocks/Listing/View.jsx
+++ b/src/components/manage/Blocks/Listing/View.jsx
@@ -6,11 +6,11 @@ import { withBlockExtensions } from '@plone/volto/helpers';
 import { ListingBlockBody as ListingBody } from '@plone/volto/components';
 
 const View = (props) => {
-  const { data, path, pathname } = props;
+  const { data, path } = props;
 
   return (
     <div className={cx('block listing', data.variation)}>
-      <ListingBody {...props} path={path || pathname} />
+      <ListingBody {...props} path={path} />
     </div>
   );
 };


ionlizarazu:
I forgot to add the gif after modifications:
listing block rendering correctly

sneridagh:
@ionlizarazu I did this because it fixes the listings in babel view... in a normal listing, the path is always present.

sneridagh:
Also there ar multiple cypress tests testing the listing block, why didn't they break? Given the gif you supplied might be something only happenning in the root? 

We need: 

* a cypress test for this use case
* a cypress test for a listing in the babel view

I can help with them if needed

sneridagh:
Talking from memory, what should happen is that in root, path is "", thus, false then the other is undefined. What about a ?? operator? (Whatever is called XD)

ionlizarazu:
I've add a test in the branch [block-listing-root-cypress-test](https://github.com/codesyntax/volto/tree/block-listing-root-cypress-test). This test fails in root page, which I fix with this pull request.
As you say, if this pathname prop is needed we can modify the [line 13](https://github.com/plone/volto/blob/bd1201a4f918c3e2d47c0360b3b3734a3c8aec0f/src/components/manage/Blocks/Listing/View.jsx#L13) with something like: `      <ListingBody {...props} path={path || pathname || '/'} />
`

sneridagh:
`<ListingBody {...props} path={path ?? pathname} />` could also do, whatever fits best!

sneridagh:
mmm strange, this PR is not launching the Actions might be because it's coming from an external repo?
/cc @tisto

ionlizarazu:
`<ListingBody {...props} path={path ?? pathname} />` this look nice, I didn't know that operator, thanks for teaching me ;-) .
I closed this PR to open one directly at plone/volto repository #2451

PR #2452, opened 2021-05-14
add new cypress test for listing block

ionlizarazu:
Cypress test for listing block at root page.
Now gives an error, trying to fix it at #2451

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 ### Internal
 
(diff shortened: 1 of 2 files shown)


sneridagh:
@ionlizarazu it passes :/

sneridagh:
Ok, that was weird... I had to reload the page, it seems Cypress was stuck with the previous data??? Anyways...

Issue #2469, opened 2021-05-21
Reset password: info to user on not registered email address


ksuess:
**Describe the bug**
User tries to reset his password, but the email is not registered (typo or just some wrong email address).
Click on "send" button does not do anything. An error message could tell the user, that this is not a valid email address.

**Expected behavior**
Error message "This is not a registered email addres."

**Software (please complete the following information):**
 - Volto 13.0.1

Issue #1921, opened 2020-10-20
Sharing: no visual feedback when adding a new local role


tisto:
# Steps to Reproduce

- as admin
- create a new page
- go to the sharing view (e.g. https://volto.kitconcept.com/new-page/sharing)
- assign a new local role to any user or group
- click the "->" button
-> nothing happens (from the user perspective)
-> expected: the "->" button shows a loading indicator (like in the left toolbar)
-> expected: when the operation was successful, show a toast message "Permissions have been updated"

<img width="1237" alt="Screenshot 2020-10-20 at 12 32 26" src="https://user-images.githubusercontent.com/632105/96575428-04a02100-12d1-11eb-8c80-bc85695cc2b9.png">

tisto:
@iFlameing can you please look into this tomorrow?

PR #2470, opened 2021-05-21
password reset: error message user if identifier not registered

ksuess:
Fix https://github.com/plone/volto/issues/2469


ksuess:
What's the policy for translations in reducers? I think it's a no go policy. 
So, where can I place a translation?

tiberiuichim:
@sneridagh @giuliaghisini

src/reducers/users/users.js
```diff
@@ -174,6 +173,17 @@ export default function users(state = initialState, action = {}) {
...
         },
       };
+    case `${RESET_PASSWORD}_FAIL`:
+      return {
+        ...state,
+        [getRequestKey(action.type)]: {
+          loading: false,
+          loaded: false,
+          error: {
+            message: 'No user registration found for this email address.',
```
giuliaghisini, line 183: ok, but could you change with an intl message to be translatable?

ksuess:
I thougt about this, if the error message should be inline or maybe better be a toast.

this is what this PR does:
image

What's the policy for error and info messages? @timo, I've seen you issued a related here: https://github.com/plone/volto/issues/1921

sneridagh:
@ksuess ummm the login uses a toast, at least currently, because of the outcome should be present when the route changes, to the next step/state (logged in), is that the case? ... in order to make them consistent, for now I would make it a toast as well. Although I see that it might also be sensible to have it inline if there's no next state...

@plone/volto-team insights?

ksuess:
I missed the feedback to the user, that the email is not registered, which he requests to reset the password for. Now I needed to tweak some more as replacing labels by field placeholders, theming and replacing the inline field info with a consistent use of toast messages.

So I ended with two new packages. One for a general form with placeholders and red bottom border for required fields and no labels anymore: https://github.com/rohberg/volto-tweaks-rohberg
And another package that provides the forms for Login, Password Reset and also the contact page with the above sleek form.: https://github.com/rohberg/volto-rohberg-login-tweak

So I am out here. If the above would be of general interest, just ping me.


Form

tiberiuichim:
@ksuess Can I pursuade you to add your packages to the volto addons list here? https://github.com/collective/awesome-volto

PR #2471, opened 2021-05-21
Apply the `schemaEnhancer` from the main block even if no variations …

sneridagh:
…are found


sneridagh:
@tiberiuichim I have no clear opinion if this should also be backported to the `withBlockSchemaEnhancer` helper.

tiberiuichim (approved):

PR #2477, opened 2021-05-25
Volta support

nzambello:
Support for  https://volta.sh/
Non intrusive change, enables correct  versioning for volta toolchain if used


nzambello:
BTW I know. Kinda confusing namings here. 🤣

sneridagh (approved):

PR #2479, opened 2021-05-26
Changed 'batch_size' attribute in 'b_size' in querystring widget.

giuliaghisini:
The problem was that changing the 'Batch size' field value in query string widget in sidebar it has no effect, because the field that plone.restapi likes is b_size and not batch_size

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 # Change Log
 
-## 13.0.3 (unreleased)
+## 13.1.0 (2021-05-24)
 
 ### Feature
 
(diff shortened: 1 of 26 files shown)


sneridagh (approved):

nzambello (approved):

tiberiuichim:
ah... sorry, I just saw this PR. I've pushed a new one, solving partially the same issue, https://github.com/plone/volto/pull/2495

The approach in this PR would require a migration on the existing listing blocks.

sneridagh (approved):

giuliaghisini:
@sneridagh @tiberiuichim what about this? could we merge it in master?

tiberiuichim:
@giuliaghisini The Cypress test I've added is flaky, it has no default sorting so the results vary and it fails. I'll try to look at it tomorrow. I think I might have fixed it in the superceeding PR, but then you have another PR that conflicts with that one :)

Issue #2492, opened 2021-06-04
A11y: Skiplinks not contained in Landmark


ThomasKindermann:
**Describe the bug**
Skiplinks are not contained in a landmark

**To Reproduce**
1. Go to frontpage
2. Open Dev Tools
3. Run Axe dev tool test
4. Axe shows "all elements must be contained in a landmark"

**Expected behavior**
Axe test should pass, Skiplinks must be contained in a landmark.
Skiplinks can be contained in a "aside" / "Complementary" landmark with aria-label 

**Screenshots**
![Screenshot 2021-06-04 at 11 17 53](https://user-images.githubusercontent.com/50659809/120778345-85412400-c526-11eb-9133-5735ee9c08bc.png)

**Software (please complete the following information):**
 - OS: osx
 - Browser chrome
 - Volto Version 13.2.1

ThomasKindermann:
PR: https://github.com/plone/volto/pull/2493

@tisto @sneridagh

ichim-david:
Fixed by the merging of the pull request referenced in https://github.com/plone/volto/issues/2492#issuecomment-854542186

PR #2493, opened 2021-06-04
#2492 added landmark role to skiplink-container

ThomasKindermann:
added a landmark role to the skiplink container for a11y
the role is "complementary", the side can have multiple complementary landmarks if they have labels

issue: #2492


sneridagh (approved):

nzambello (approved):
LGTM, but won't it be too redundant or too noisy maybe? I've never used that role on a screen reader

PR #2495, opened 2021-06-04
Fix listing pagination

tiberiuichim:
From the querystringwidget / listing block refactoring, the batch size and limit was no longer respected.

### DIFF
--- a/src/components/manage/Blocks/Listing/ListingBody.jsx
+++ b/src/components/manage/Blocks/Listing/ListingBody.jsx
@@ -24,32 +24,27 @@ const ListingBody = React.memo(
     );
     const dispatch = useDispatch();
 
+    const adaptedQuery = {
+      ...(data.batch_size ? { b_size: data.batch_size } : {}),
+      ...(data.limit ? { limit: data.limit } : {}),
+      ...(data.query ? { query: data.query } : {}),
+      fullobjects: 1,
+    };
+
     React.useEffect(() => {
       if (data?.query?.length > 0) {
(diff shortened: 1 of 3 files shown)


src/components/manage/Blocks/Listing/ListingBody.jsx
```diff
@@ -14,29 +14,37 @@ import paginationRightSVG from '@plone/volto/icons/right-key.svg';
...
     const content = properties;
+    const { settings } = config;
+    const { batch_size = settings.defaultPageSize } = data;
+
+    const adaptedQuery = {
+      ...(data.limit ? { limit: data.limit } : {}),
+      ...(data.query ? { query: data.query } : {}),
+      b_size: batch_size,
+      fullobjects: 1,
+    };
```
giuliaghisini, line 26: I'm not a fan of rewriting parameters for the query .. because some parameters may be forgotten ..
for example, **sort_on** and **sort_order** are missing here .. and maybe some other params that i'm not thinking in this moment also i fixed the b_size param here in this pr: https://github.com/plone/volto/pull/2479, maybe you colud improve that pr adding your default settings.defaultPageSize

giuliaghisini (changes requested):

tiberiuichim:
@giuliaghisini I'll restructure my PR as a push to yours. 

I think we should be explicit about what parameters we include in the querystring request. We made the blocks extendible, they can get a lot more data then we expect, we shouldn't push that into the request.

PR #2508, opened 2021-06-09
Sharing: no visual feedback when adding a new local role 

iFlameing:
Fixes https://github.com/plone/volto/issues/1921

Screenshot

Screenshot 2021-06-09 at 7 04 25 PM


nileshgulia1 (approved):
LGTM

tisto (approved):
@iFlameing @sneridagh LGTM. I am ok to merge and release. 

I am just wondering if it makes much sense to show the same message string twice, as headline and body text. How does Plone classic do it? The headline says "Sharing", the toast messages say "permissions". Technically it is a "local role" that is updated. 

We won't be able to solve Plone's naming mess here. Though, maybe we can think about a step in the right direction though. Let's discuss this after our stand-up today. :)

tisto:
Just checked Plone the classic message. "Changes saved", even more unspecific than what we have in this PR...

Eine-Nachricht-—-Deutsch

src/components/manage/Sharing/Sharing.jsx
```diff
@@ -159,6 +164,13 @@ class SharingComponent extends Component {
   UNSAFE_componentWillReceiveProps(nextProps) {
     if (this.props.updateRequest.loading && nextProps.updateRequest.loaded) {
```
pnicolli, line 165: I don't remember that perfectly at this time, but is there a possibility that the nextProps are ``loaded`` but with an error on the request response? That should be checked if it's the case.

pnicolli:
I would agree with changing the popup title, probably something as simple as "Success!" would be good

src/components/manage/Sharing/Sharing.jsx
iFlameing: If the request failed in that case loading is set to false not the loaded true. You can see the code here
https://github.com/plone/volto/blob/master/src/reducers/sharing/sharing.js#L82

pnicolli: Awesome, thanks for checking this for me, I was in a bit of a rush but didn't want to forget this :)

tisto:
@iFlameing @pnicolli what about "Permissions updated" as title and "Permissions have been updated successfully." as text?

pnicolli:
Works for me

tisto:
@iFlameing pls update the labels and fix the CI job.

iFlameing:
@tisto I will create a new pr.

Issue #2483, opened 2021-05-31
Object Browser closes clicking on it on new context


nzambello:
**Describe the bug**
When you edit new content (a Page), so in a new Plone context, if you try to use the Object Browser for a link, clicking two times on the Sidebar (everywhere) will close the ObjectBrowser.
My issue started for the inability to search for a content: first click on the search (lens) icon button, then focusing on the search input was causing the sidebar to close. Then I saw that it was triggered by the clicks, anywhere in the sidebar. 

It doesn't happen on an existing content.

**To Reproduce**
1. Add a new page
2. Create a text block with some text
3. Select some text to add a link
4. Open the object browser to retrieve a content
5. Click on the search button
6. Focus on the search input
7. The sidebar will close

**Expected behavior**
It shouldn't close

**Screenshots**

https://user-images.githubusercontent.com/21101435/120179092-7434a200-c20a-11eb-961d-40cdeb6d266c.mov

**Software (please complete the following information):**
 - OS: any
 - Browser: Chrome, Firefox
 - Volto Version: 12.14.0 -> 13 (master) for sure
 - Plone Version: 5.2.4
 - Plone REST API Version: 7.3.1

nzambello:
@sneridagh can this be related to https://github.com/plone/volto/pull/2389#issuecomment-830027413 ?

avoinea:
I couldn't reproduce this in Chrome, but indeed it reproduces on Firefox.

giuliaghisini:
I cannot reproduce this in Chrome neither in Firefox. 
I think this is fixed and there's nothing more to do.

@sneridagh @tiberiuichim @avoinea

davisagli:
fwiw, I saw behavior similar to this in a case where I was accidentally using withObjectBrowser to wrap 2 different components. So there were two different SidebarPopup components listening for mouse clicks, and the click could be inside one but outside the other.

PR #2509, opened 2021-06-09
fix: sidebar popup click outside

nzambello:
Fixes: #2483

- Fixes outside click handling
- Closes SidebarPopup on pressing ESC key


src/components/manage/Sidebar/SidebarPopup.jsx
```diff
@@ -2,26 +2,43 @@ import React from 'react';
...
 
 const SidebarPopup = (props) => {
   const { children, open, onClose, overlay } = props;
 
-  const asideElement = React.createRef();
-
   const handleClickOutside = (e) => {
-    if (asideElement && doesNodeContainClick(asideElement.current, e)) return;
-    onClose();
+    let sidebarContainer = e?.target?.closest('#sidebar-popup-container');
```
pnicolli, line 12: I don't remember browser compatibility that we are granting with Volto, but please note that ``closest`` is not IE compatible. There's a [tiny polyfill at MDN](https://developer.mozilla.org/en-US/docs/Web/API/Element/closest#polyfill), in case it's needed.

src/components/manage/Sidebar/SidebarPopup.jsx
```diff
@@ -2,26 +2,43 @@ import React from 'react';
...
-  });
+    if (open) {
+      document.addEventListener('click', handleClickOutside, false);
+    } else {
+      document.removeEventListener('click', handleClickOutside, false);
+    }
+    return () =>
+      document.removeEventListener('click', handleClickOutside, false);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [open]);
```
pnicolli, line 31: I would consider refactoring the ``handleClickOutside`` function out of the component like ``function handleClickOutside(e, open);`` and fix the eslint comment

src/components/manage/Sidebar/SidebarPopup.jsx
```diff
@@ -2,26 +2,43 @@ import React from 'react';
...
+
+  React.useEffect(() => {
+    if (open) {
+      document.addEventListener('keydown', handleESCPress, false);
+    } else {
+      document.removeEventListener('keydown', handleESCPress, false);
+    }
+    return () => document.removeEventListener('keydown', handleESCPress, false);
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [open]);
```
pnicolli, line 41: I would consider refactoring the ``handleEscPress`` function out of the component like ``function handleEscPress(e, onEscPress);`` or, as an alternative, ``handleEscPress(e) => boolean;`` and fix the eslint comment

giuliaghisini (approved):

src/components/manage/Sidebar/SidebarPopup.jsx
nzambello: IE is not supported: https://github.com/plone/volto/blob/2888ad2d927601cc1b1472086dd1cdd5d713994f/src/config/index.js#L107

src/components/manage/Sidebar/SidebarPopup.jsx
nzambello: I also considered that, but even defining handleClickOutside as `function` eslint was expecting to have it as dep 🤷

pnicolli: Yes but IIRC that would be fine to add since it won't be redefined on every component render.

nzambello: uh, yeah right

nzambello: I need a function to pass to addEventListener and it would have not the firm of a listener function, so I should have a function which returns the listener and it would be more complicated since the ilstener is recognized by reference.
I'm not sure the performances will benefit more than readability would be worse.
IDK if I'm explaining this.

pnicolli: Yeah my point wasn't about performance, it was more about removing eslint fixes, but it's not that much important :)
Anyway yes, you are right, it would be a little more complicated than I imagined.

tiberiuichim: What about useCallback? Than you could pass the result from useCallback as dependency to the useEffect.

pnicolli: Yes that's actually the exact solution but for some reason I didn't think about it when I wrote the comment

nzambello: Done in 5f56ae3

pnicolli (approved):
LGTM

avoinea (approved):

sneridagh:
@nzambello I just tried out, and the ESC event bubbles up and deselects the block as well... any chance we can cut it somewhere with a `stopPropagation`?

nzambello:
@sneridagh I tried to fix it, but the event is handled by something in a higher level (it's attached to body) and with `stopPropagation` and the others workaround I cannot stop it.
It's happening in master, too: if I remove the event handler here, the field still loses focus, but at least I get the sidebar closing.

Here's what I tried:
```diff
diff --git a/src/components/manage/Sidebar/SidebarPopup.jsx b/src/components/manage/Sidebar/SidebarPopup.jsx
index 348a1f223..b6c058e3f 100644
--- a/src/components/manage/Sidebar/SidebarPopup.jsx
+++ b/src/components/manage/Sidebar/SidebarPopup.jsx
@@ -20,7 +20,13 @@ const SidebarPopup = (props) => {
 
   const handleESCPress = React.useCallback(
     (e) => {
-      if (e.code === 'Escape') onClose();
+      if (e.code === 'Escape') {
+        e.preventDefault();
+        e.stopImmediatePropagation();
+        e.stopPropagation();
+        onClose();
+
+        return false;
+      }
     },
     [onClose],
   );
@@ -37,11 +43,11 @@ const SidebarPopup = (props) => {
 
   React.useEffect(() => {
     if (open) {
-      document.addEventListener('keydown', handleESCPress, false);
+      document.addEventListener('keydown', handleESCPress, true);
     } else {
-      document.removeEventListener('keydown', handleESCPress, false);
+      document.removeEventListener('keydown', handleESCPress, true);
     }
-    return () => document.removeEventListener('keydown', handleESCPress, false);
+    return () => document.removeEventListener('keydown', handleESCPress, true);
   }, [handleESCPress, open]);
 
   return (
```

Do you have any other ideas? Otherwise I think it could become a separated topic.

sneridagh:
@giuliaghisini I think this one is not fully resolved, right? For what I can recall, I gave it a try (se my comment). I don't know if that behavior is desirable (ESC on it, deselecting the block).

thoughts?

nileshgulia1:
> @giuliaghisini I think this one is not fully resolved, right? For what I can recall, I gave it a try (se my comment). I don't know if that behavior is desirable (ESC on it, deselecting the block).
> 
> thoughts?

Also happens on main branch right now as Nicola mentioned. I'm not sure about the first point in the PR, the outside click handling works well. The ESC key needs to be taken care.

giuliaghisini:
> > @giuliaghisini I think this one is not fully resolved, right? For what I can recall, I gave it a try (se my comment). I don't know if that behavior is desirable (ESC on it, deselecting the block).
> > thoughts?
> 
> Also happens on main branch right now as Nicola mentioned. I'm not sure about the first point in the PR, the outside click handling works well. The ESC key needs to be taken care.

this is a very old pr and maybe in that period the outside click handling was not workinkg. 
Esc key need to be taken care, i agree with you. 
Could you work on this?

nileshgulia1:
> > > @giuliaghisini I think this one is not fully resolved, right? For what I can recall, I gave it a try (se my comment). I don't know if that behavior is desirable (ESC on it, deselecting the block).
> > > thoughts?
> > 
> > 
> > Also happens on main branch right now as Nicola mentioned. I'm not sure about the first point in the PR, the outside click handling works well. The ESC key needs to be taken care.
> 
> this is a very old pr and maybe in that period the outside click handling was not workinkg. Esc key need to be taken care, i agree with you. Could you work on this @nileshgulia1 ?

Thanks for heads up. I plan to work on it during the sprint.

sneridagh:
@giuliaghisini If this is not reproduceable anymore, let's close it. Could you please take a look?

nileshgulia1:
@sneridagh We can close this as its not reproducible anymore. The second part of this PR i.e handling ESC key is done in https://github.com/plone/volto/pull/6315

PR #2141, opened 2021-01-20
implement interface for plone seo extensions to add values to metadat…

jackahl:
…a fields

Issue #252, opened 2018-08-22
Full SEO support


sneridagh:
and add ability to configure meta tags in the <head>

Maybe include a configuration item that includes them in Html helper or alternativelly using Helmet in the App component.

robgietema:
We could add this to the config.js

vinay72:
Configure Meta title and Meta description

Perform the following:

    Open the static module whose content items you want to configure with specific meta title and description and in the right pane, click Custom Fields for <content type>.
    For dynamic modules, navigate to Administration » Module builder, click the module and then, the content type.
    Create the following custom fields:
        Type: Search engine optimization, Meta field: Meta title.
        Type: Search engine optimization, Meta field: Meta description.

        NOTE: When you create the fields, you can make them localizable. This way, you can display different content items for different language versions.
        For more information about creating custom fields, see Create custom fields.
    Create a new content item, or edit an existing one.
    You will see the new fields.
    Fill out the Title for search engines and the Description fields.
    If you leave Title for search engines field blank, the system will automatically fill it out with the Title of the item.

    IMPORTANT: If you leave the Description field empty, Sitefinity CMS  automatically fills it out with the content of the first field after the content of the content item, regardless of whether it is a long or short text field. 
    For example, if after the news item, you have a Summary field, the Description field will be populated with the content of the Summary field. For dynamic modules, you can use Modify the backend screens of a dynamic module to change the order of the fields.

RESULT: When you open the content item in detailed view, the runtime value of these fields is used to set the page meta title and meta description tags in the <head> tag of the page. If you do not set them, the meta title is taken from the title field of the static content item or the identifier field of the dynamic content item.

tisto:
FYI: there is https://github.com/kitconcept/kitconcept.seo which provides all the basic SEO fields (and more). I guess we would have to write and release a frontend part for this (that overrides the helmet component) or add some kind of hooks that kitconcept.seo (and possibly other extensions) can hook in.

tiberiuichim:
I think BodyClass can be used as an example on how  to lift stuff from the child route components to the top level, together with asyncConnect they might do the job

nzambello:
This thread covers wider aspects, but a part of this has been worked in https://github.com/plone/volto/pull/2132

PR #2510, opened 2021-06-10
Add og tags and seo support

nzambello:
Fixes: #252 

Closes: #2141 
Closes: #2132 

Merged both the PRs into a single one.


src/components/theme/View/View.jsx
```diff
@@ -220,8 +222,65 @@ class View extends Component {
...
+          <meta
+            property="og:url"
+            content={
+              this.props.content.seo_canonical_url ||
+              toPublicURL(this.props.content['@id'])
+            }
+          />
+          {this.props.content.image?.scales?.large?.download ||
+            (this.props.content.opengraph_image?.scales?.large?.download && (
+              <>
```
giuliaghisini, line 253: this throws errors.. each <meta> it must not be enclosed in a fragment

giuliaghisini (changes requested):

src/components/theme/View/View.jsx
nzambello: 👍

giuliaghisini (approved):

src/components/theme/View/View.test.jsx
```diff
@@ -20,6 +20,7 @@ beforeAll(() => {
       ECONNREFUSED: () => <div className="ECONNREFUSED" />,
     },
   });
+  config.settings.publicURL = 'https://plone.org';
```
tiberiuichim, line 23: Can we have a safer default? Maybe take RAZZLE_API_PATH, or a variant, as the default?

nzambello: This is only the test, the setting itself is in the config

sneridagh: For the test is fine I guess. The config in the tests resets on every test.

sneridagh (commented):
LGTM, check the single comment.

tisto:
@jackahl can you please look into this PR as well and confirm that this is in sync with the settings that we have/use in our projects?

jackahl:
> @jackahl can you please look into this PR as well and confirm that this is in sync with the settings that we have/use in our projects?

Checked. Is in sync with how we have it.

sneridagh:
@nzambello @giuliaghisini there was an error in the code, could you take a look and confirm that a lead `image` wins over an `opengraph_image`?

sneridagh (approved):

tiberiuichim:
Can we separate all this SEO and content image functionality in a new component? Ideally even one that's plugged, so that it can be overriden. I feel that View.jsx should act just as a "switch" and decide which view component to actually use. If we need something that should be rendered along with everything else, fine, but let's split that into another component, as it's a "separate concern".

tiberiuichim (changes requested):
Can we separate all this SEO and content image functionality in a new component? Ideally even one that's plugged, so that it can be overriden. I feel that View.jsx should act just as a "switch" and decide which view component to actually use. If we need something that should be rendered along with everything else, fine, but let's split that into another component, as it's a "separate concern".

pnicolli:
> Can we separate all this SEO and content image functionality in a new component? Ideally even one that's plugged, so that it can be overriden. I feel that View.jsx should act just as a "switch" and decide which view component to actually use. If we need something that should be rendered along with everything else, fine, but let's split that into another component, as it's a "separate concern".

@tiberiuichim 
If I understand correctly, you mean two things:

- refactor ``getContentImageInfo`` in a separate helper function
- create a separate component for all the content meta tags

Is this right? Any suggestions about where you would put the two new items in the source tree?

tiberiuichim:
I'd put all of this into a ContentMetadataTags component. @sneridagh what's your take on this?

```
          {this.props.content.language && (
            <html lang={this.props.content.language.token} />
          )}
          <title>
            {this.props.content.seo_title || this.props.content.title}
          </title>
          <meta
            name="description"
            content={
              this.props.content.seo_description ||
              this.props.content.description
            }
          />
          <meta
            property="og:title"
            content={
              this.props.content.opengraph_title ||
              this.props.content.seo_title ||
              this.props.content.title
            }
          />
          <meta
            property="og:url"
            content={
              this.props.content.seo_canonical_url ||
              toPublicURL(this.props.content['@id'])
            }
          />
          {contentImageInfo.contentHasImage && (
            <meta
              property="og:image"
              content={toPublicURL(contentImageInfo.url)}
            />
          )}
          {contentImageInfo.contentHasImage && (
            <meta property="og:image:width" content={contentImageInfo.width} />
          )}
          {contentImageInfo.contentHasImage && (
            <meta
              property="og:image:height"
              content={contentImageInfo.height}
            />
          )}
          {(this.props.content.opengraph_description ||
            this.props.content.seo_description ||
            this.props.content.description) && (
            <meta
              property="og:description"
              content={
                this.props.content.opengraph_description ||
                this.props.content.seo_description ||
                this.props.content.description
              }
            />
          )}
          <meta name="twitter:card" content="summary_large_image" />
```

sneridagh:
I agree, it's more sane and flexible. Once done also, as you suggested, it could also be wrappped into a pluggable.

sneridagh:
I'll take care

sneridagh:
@nzambello @giuliaghisini @pnicolli @tiberiuichim I double checked the add-ons and I think 
that it was the other way around, so the `opengraph_image` wins over the set `image`. Please could you confirm this?

Also, I added a new setting for specifying the image field to use, just in case.

@tiberiuichim moved to its own component, put a Pluggable in place `content-metadata-tags`. Also, time to open a Default Pluggables present in Volto documentation section?

docs/source/configuration/settings-reference.md
```diff
@@ -246,3 +246,9 @@ in the `config.settings.serverConfig` object.
     For the moment it admits only one property: `errorPages` whose value is a Boolean.
 
     If `extractScripts.errorPages` is `true`, the JS will be inserted into the error page.
+
+### contentMetadataTagsImageField
+
+!!! block ""
+
+    The SEO default image field name for using it in the metadata HEAD tag. Defaults to `image`.
```
tiberiuichim, line 254: > The OpenGraph image that will represent this content item, will be used in the metadata HEAD tag as `og:image` for SEO purposes. Defaults to `image`. See the [OpenGraph Protocol](https://ogp.me/) for more details.

tiberiuichim (approved):
LGTM, I've added a doc suggestion. @sneridagh I think you'll also need to add yourself to the changelog

nzambello:
@sneridagh Thank you for taking this up, I'd be late on this.
I cannot approve my own PR, but I approve the chages 👍

PR #2513, opened 2021-06-10
Set default value selected for variation in listing block

giuliaghisini:
Before this fix, when you creating a listing block, variation field was empty in sidebar, even though a default option was set. 

Now, when you create a new listing block, in sidebar default option is selected for field 'variation'


tiberiuichim (approved):
Yes yes, another widget to support defaultValue :)

tisto (approved):
LGTM. Just out of curiosity, where do we set the default value for that particular block? In config.js?

giuliaghisini:
> LGTM. Just out of curiosity, where do we set the default value for that particular block? In config.js?

in config.js when you define the variations list, you could add the prop **isDefault** to a variation!

PR #2517, opened 2021-06-11
In listing blocks, scroll to start of listing block instead page start

giuliaghisini:
Let's imagine we have 3 listing block in page, and each one has pagination. 
If we change page of the 3rd block, window page was scrolled to top and i had to manually scroll down to see results of second page. 
With this change, on listing page change, window is scrolled to the start of listing block.


pnicolli (approved):
LGTM

tiberiuichim (approved):

sneridagh (approved):

PR #2527, opened 2021-06-17
Fix German translations for leadimage and listing block

tisto:


locales/de/LC_MESSAGES/volto.po
```diff
@@ -1016,7 +1016,7 @@ msgstr "Layout"
 
 #: components/manage/Blocks/LeadImage/LeadImageSidebar
 msgid "Lead Image"
-msgstr "Leadbild"
+msgstr "Lead-Bild"
```
tisto, line 1019: @ksuess I thought about this as well. Both versions are correct. It is just important that we are consistent here. Could you imagine writing down a few sentences about our current best practices when it comes to German translations? We could create a section in the docs for this and get @erral on board as well.

PR #2530, opened 2021-06-18
Complete es translation

erral:
As a first step for the i18n audit, we need to have a fully translated language.


tisto:
@erral you need to run "yarn run i18n" to make the build pass. If I am not mistaken German is fully translated.

tisto:
@erral seems I forgot to push my work yesterday: https://github.com/plone/volto/pull/2531

tisto:
@erral how do you plan to check a translation for completeness? Do we have an existing check in Plone classic? I guess a gh action would be a good idea. Maybe we can find people that are willing to make sure one particular language is always fully translated. If that's the case we could even have language-specific gh actions (e.g. fail if ES/DE is not fully translated).

erral:
In Plone Classic there is a Jenkins job that runs regularly (https://jenkins.plone.org/view/Translations/) I was planning to ask one of those for Volto too.

erral:
> @erral seems I forgot to push my work yesterday: #2531

Ah Ok, that's why the .json files weren't updated I will run it right now and push again.

tisto:
@erral I think @sneridagh refactored this so that the compile step is not necessary. Though, for some reason this does not work yet.

tisto:
@erral we are about to make a release this morning. Do you want us to wait for your PR and include it?

sneridagh:
json are compiled at build time and included in released versions and excluded from the repo (as in Plone). The do not need to be in there to work. Some old PR should have put them there again. They should be removed. The check is in there still for the released versions, it does not harm.

@erral The checker should trigger yarn i18n before doing its work I guess.

erral:
> @erral we are about to make a release this morning. Do you want us to wait for your PR and include it?

Yes please



> json are compiled at build time and included in released versions and excluded from the repo (as in Plone). The do not need to be in there to work. Some old PR should have put them there again. They should be removed. The check is in there still for the released versions, it does not harm.
> 
> @erral The checker should trigger yarn i18n before doing its work I guess.

Ok, noted.

tisto:
@sneridagh let's get a new release out of the door and then we can remove those files from the repo again. What do you think?

sneridagh:
@tisto ok! Just resalised a use case that the checker should be removed indeed... my bad.

tisto (approved):

PR #2532, opened 2021-06-18
Add DictWidget (new name: VocabularyTermsWidget)

ksuess:
Replace DictWidget by VocabularyTermsWidget

and **map new widget to field with widget attribute set to 'vocabularyterms'**

See use case vocabulary for Choice fields in VocabularyTermsWidget.jsx

Requires updates on other packages:
plone.schema = git git@github.com:plone/plone.schema.git branch=field-attribute-widget
plone.restapi = git git@github.com:plone/plone.restapi.git branch=add-dict-field-widget

See storybook.

https://user-images.githubusercontent.com/1144737/124275499-e9b1cc00-db42-11eb-9f98-ef0c75d75ca1.mov


tiberiuichim:
Whooop! A new widget! And excelent code quality!

tiberiuichim (approved):

sneridagh (changes requested):
Awesome work! LGTM!

However, a small UI thing, I would keep the UI consistent as much as possible between the different widgets. Is it much of a trouble to copy the button and the styling from a Pastanaga button like in the one in ObjectListWidget? 

image

Thanks a lot!!

ksuess:
> Awesome work! LGTM!
> 
> However, a small UI thing, I would keep the UI consistent as much as possible between the different widgets. Is it much of a trouble to copy the button and the styling from a Pastanaga button like in the one in ObjectListWidget?
> 
> image
> 
> Thanks a lot!!

oh, yes, sure. 
done: add new entry button pastanaga style like in ObjectListWidget

ksuess:
I should mention, that this does only cover most use cases. If you want to edit also the keys of the dictionary this widget is not suitable. But, for vocabulary it's essential, that the key (token or correct: value of the vocabulary term) does NOT change ever, as this is stored on the content type instance. And the reason to use a dict field and not a tuple ore list is exactly this: the title of the vocabulary entry can change or even be translated leaving the 'connection' to consumers intact.

tiberiuichim:
Sorry to come in so late, maybe it would be better to rename this as "TermWidget" ? If I can't change the keys, as well, then maybe it shouldn't be DictWidget?

ksuess:
@tiberiuichim TermWidget: good idea, will rename it.

It would be nice to have an attribute 'widget' on Plone Dict field.

sneridagh (approved):
LGTM!!

sneridagh:
+1 for the renaming, also +1000 on making all the fields possible "widget" keyword aware.

ksuess:
I am about to remove the widget mapping on field type 'dict' for DictWidget as the required widget for a dict field is not unambiguous. But,

At the moment there is no default widget for the dict field.
IMHO most cases need this widget to edit the dictionary values, only some cases need a widget to edit both, keys and values.

? So, do we want to leave it open, or do we map field type to DictWidget and leave it to the developer to overrde with an explicit widget configuration via (future) dict field widget attribute or via field id? Opinions?

tiberiuichim:
I don't think it's a disaster if we register the `widgets.widget.dict` to DictWidget, though maybe `widgets.widget.terms` would be more appropriate. For 

Just for a little diversity, I realized that we have a "mapping widget", that associates a set of fixed keys to values: https://github.com/eea/volto-object-widget#mapping-widget

sneridagh:
@ksuess 
Having a default mapping is fine, although one could consider that a mapping of a dict field is more suitable for a single ObjectWidget...

Ping me when it's ready!

sneridagh:
@ksuess is it ready? I just sync'ed

nzambello:
Super nice! Happy to get rid of ArrayWidget for dict entries, it's good but for other use cases.

Did you think on the translations of these entries? This is a big limit of our configurations right now.
I have  https://github.com/collective/volto-multilingual-widget but we wouldn't have the same key, then.

Not a blocker comment here, only a thing to think about 😄

ksuess:
@nzambello yes, I will definitly tune this widget to make term titles translatable.
For the moment I prefer to separate this PR from a following one for translatable term titles.

ksuess:
So this is a makeover. I removed the default mapping of field type 'dict' to this new widget. Instead I added a mapping of widget name 'vocabularyterms' to this widget.

Translations are updated to vocabulary, terms, etc. to clearly distinguish between a general Python dictionary and this vocabulary tool.

I understand that this could look, at a first glance, like an edge case and should not go into Volto core. But IMHO it is essential to feed Plone Choice fields with a vocabulary that can change over time without breaking the data base. To achieve this I think one straight way is to provide a vocabulary UI via a controlpanel Dict field.

@sneridagh, @tiberiuichim and who is interested, would you re-review?  Many thanks and a sunny weekend!

tiberiuichim (approved):

tiberiuichim:
@ksuess It's easier to deal with a centralized "bag of goodies" right now, so no worries if it's not dirrectly used by Volto, I think that's @sneridagh take on this as well. Have a great weekend!

sneridagh:
I think it's fits perfectly, and covers a quite important set of use cases. Thanks for it and have a good weekend too!

Issue #2548, opened 2021-06-26
Hyphenation block chooser


ksuess:
Hyphen block chooser labels

expected:
![Seite_bearbeiten](https://user-images.githubusercontent.com/1144737/123507337-178da100-d669-11eb-8205-4109ac532ed4.png)

current state:
![Cursor_and_Seite_bearbeiten](https://user-images.githubusercontent.com/1144737/123507364-3d1aaa80-d669-11eb-9099-7f4edfacf7b5.png)

**Software 
- Volto 13.2.2

PR #2549, opened 2021-06-26
Hyphenation block chooser labels

ksuess:
Fix https://github.com/plone/volto/issues/2548

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 
 - Fixed docs for config.settings.externalRoutes @giuliaghisini
 - Fix `Login` component navigation for `INavigationRoot` structures @sneridagh
+- Hyphenation block chooser labels @ksuess
 
 ### Internal
 
--- a/src/components/manage/BlockChooser/BlockChooser.jsx
+++ b/src/components/manage/BlockChooser/BlockChooser.jsx
@@ -125,10 +125,12 @@ const BlockChooser = ({
                       }}
                     >
                       <Icon name={block.icon} size="36px" />
-                      {intl.formatMessage({
-                        id: block.id,
-                        defaultMessage: block.title,
-                      })}
+                      <span>
+                        {intl.formatMessage({
+                          id: block.id,
+                          defaultMessage: block.title,
+                        })}
+                      </span>
                     </Button>
                   </Button.Group>
                 ))}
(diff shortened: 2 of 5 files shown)


tiberiuichim (approved):

sneridagh (approved):

tisto:
@sneridagh that's definitely a helpful fix, especially for German labels. Thanks, @ksuess!

I am playing devil's advocate here, isn't that a breaking change since we change the HTML structure? I guess it is highly unlikely that someone customized the blocks chooser. Though, I guess we should make a rule that every change in the HTML structure is at least fully documented, what do you think?

tiberiuichim:
I don't think it's a breaking change in this case, as the "external api" of BlockChooser hasn't changed.

nzambello:
I agree with @tiberiuichim

tisto:
@tiberiuichim @nzambello didn't we add a new span element here and therefore change our public API (the HTML structure of the component)? 

https://github.com/plone/volto/pull/2549/files#diff-301c0d5d89f802bfc2c42e8dab539e6a5a7a5706aa8f8f64e01b3341084e7da5R128

ksuess:
Let me see if there is a way to avoid a change of the html.

tisto:
Just for the record: the only reason I am so pedantic here is that I would like us to come up with a set of clear rules about breaking changes that we all can rely on in the future. This discussion is important. This has nothing to do with the PR itself, which is fine.

ksuess:
It's great that you have an I on such subjects, @tisto!

tiberiuichim:
IMHO the HTML structure of components is allowed to change without declaring it a breaking change. Here's my reasons of thinking this:

- if we regard Volto as a self-contained entity and the HTML change has side-effects in other parts (for example, css styling rules), then Volto also has the "fix" for that change, in the form of an updated CSS stylesheet.
- I don't consider breaking component shadowing to be a sufficient form of breaking change. The shadowing mechanism is at least on the level of monkey-patching in terms of "things that are possible, but shouldn't be encouraged". We've worked hard and continue to work on creating an official API so that shadowing is not required. Even more, I think right now the recommended approach should be "if you find that you need shadowing, do it now, be prepared that it would break in the future, but also try to fix your need in Volto, by coming up with a proper extension point".

nzambello:
I guess that we have two different concerns everybody is feeling, surely is what I think myself:
- As a maintainer of the product Volto, I want to consider a breaking change something is literally breaking projects or changing internal APIs without retrocompatibility (semver)
- As Volto integrator, having some components customized or a custom theme, I'd like to have an highlight of what is changing under the hood in the HTML because I'll need to take action for that

I wouldn't consider this case as breaking because we added a span, but I understand that similar cases could affect a project theme or something else, so we should highlight a change in HTML.

So, could we consider this type of cases as minor or patch but show a change in html to integrators and developers through the changelog? Maybe using a label or a custom section.

PR #2557, opened 2021-06-30
Add some missing dependencies

tiberiuichim:
There were hoisted by yarn, but they should be listed as actual dependencies


package.json
```diff
@@ -295,13 +298,16 @@
...
+    "filesize": "^6.4.0",
     "glob": "7.1.6",
     "hamburgers": "1.1.3",
     "handlebars": "4.7.7",
     "history": "4.10.1",
+    "hoist-non-react-statics": "^3.3.2",
     "http-proxy-middleware": "1.0.4",
     "husky": "4.2.5",
     "identity-obj-proxy": "3.0.0",
+    "immutable": "^4.0.0-rc.12",
```
sneridagh, line 310: This is a hard dependency for draftjs I pin it via "resolutions" but it has its own problems, like the resolutions are not passed to projects. However, I don't know if draftjs works with bleeding edge versions of immutable.

package.json
```diff
@@ -316,13 +322,16 @@
...
+    "lodash.sortby": "^4.7.0",
     "mem": "4.0.0",
     "mini-css-extract-plugin": "0.9.0",
     "moment": "2.24.0",
+    "object-assign": "^4.1.1",
     "pofile": "1.0.10",
     "postcss-flexbugs-fixes": "4.2.1",
     "postcss-loader": "3.0.0",
     "postcss-overrides": "3.1.4",
+    "prepend-http": "^3.0.1",
```
sneridagh, line 334: It seems that this is the cause of the failing test. I would use the same that is right now in the lock.

sneridagh (changes requested):

sneridagh:
How about adding "levenary" to mitigate the problem with the previous update?

package.json
tiberiuichim:
```
 WARN  deprecated babel-eslint@10.1.0: babel-eslint is now @babel/eslint-parser. This package will no longer receive updates.
 WARN  deprecated babel-plugin-react-intl@5.1.17: this package has been renamed to babel-plugin-formatjs
 WARN  deprecated draft-js-buttons@2.0.2: use @draft-js-plugins/buttons >=v4 instead
 WARN  deprecated draft-js-plugins-editor@2.1.1: use @draft-js-plugins/editor >=v4 instead
 WARN  deprecated draft-js-inline-toolbar-plugin@2.0.3: use @draft-js-plugins/inline-toolbar >=v4 instead
```

tiberiuichim:
@sneridagh Somehow it feels strange that we treat Volto as a standalone application but also as a library.

When it's a standalone app, then it would make sense that its development dependencies would be marked as devDependencies. But we also use it as a library and all our projects depend on having volto bring what would be the project's devDependencies. 

Any insights on this? I guess it's something we have to live with... for sure having Volto's development centralized as it is right now it's easier for the times we find ourselves in.

package.json
tiberiuichim: I've pinned it to 3.8.2. But I get this message when running yarn:

```
warning Resolution field "immutable@3.8.2" is incompatible with requested version "immutable@~3.7.4"
```

tiberiuichim:
@sneridagh With this it's possible to use pnpm instead of yarn.

tiberiuichim:
I've added the pnpm-lock.yaml file. I'm not sure what's the best workflow for the future... We could just ignore the pnpm lock and I'll make PRs to update it.

The workflow to update the pnpm lock file is:

```
rm -rf node_modules
pnpm i
```

PR #2559, opened 2021-07-01
Inlineform default values

rexalex:
Change for nullish condition needed here for default values.
This is needed to get the desired "false" value and not default to the defaultValues.


tiberiuichim (approved):

tiberiuichim:
@rexalex please add a changelog entry under "bugfixes"

sneridagh (approved):

PR #2560, opened 2021-07-01
Make a separate component of the BlockChooser wrapper logic

tiberiuichim:
This logic is repeated and reimplemented all over in such as https://github.com/kitconcept/volto-blocks-grid/blob/main/src/components/Grid/NewBlockAddButton.jsx or https://github.com/eea/volto-columns-block/blob/master/src/ColumnsBlock/EditBlockWrapper.jsx#L141 and so on. This PR encapsulates this logic into a component button so that it can be reused. Note that the component allows passing a custom "buttonComponent" prop, which can customize how the + button actually looks like.


src/components/manage/Blocks/Block/MutateBlockButton.jsx
```diff
@@ -0,0 +1,108 @@
...
+import { defineMessages, useIntl } from 'react-intl';
+
+const messages = defineMessages({
+  addBlock: {
+    id: 'Add block',
+    defaultMessage: 'Add block',
+  },
+});
+
+export const ButtonViewComponent = (props) => {
```
giuliaghisini, line 16: isn't it better to call it 'AddButtonComponent'?

src/components/manage/Blocks/Text/Edit.jsx
```diff
@@ -317,32 +287,18 @@ class Edit extends Component {
...
+            data={this.props.data}
+            block={this.props.block}
             onInsertBlock={(id, value) => {
-              this.setState((state) => ({
-                addNewBlockOpened: !state.addNewBlockOpened,
-              }));
               this.props.onSelectBlock(this.props.onInsertBlock(id, value));
             }}
-            currentBlock={this.props.block}
+            onMutateBlock={() => {}}
```
giuliaghisini, line 297: why mutating text block is not allowed?

tiberiuichim: There's something fishy in Volto right now:

According to this, the onMutateBlock [parameter is required](https://github.com/plone/volto/blob/master/src/components/manage/BlockChooser/BlockChooser.jsx#L146)

But [it's not passed down](https://github.com/plone/volto/blob/master/src/components/manage/Blocks/Text/Edit.jsx#L335) from the `Text/Edit.jsx` component

I think the wrapping of the component somehow swallows the proptypes warning. I made the params optional, for good measure.

src/components/manage/BlockChooser/BlockChooser.jsx
```diff
@@ -143,10 +144,15 @@ const BlockChooser = ({
...
   allowedBlocks: PropTypes.arrayOf(PropTypes.string),
   blocksConfig: PropTypes.objectOf(PropTypes.any),
 };
 
-export default injectIntl(BlockChooser);
+export default injectIntl(
+  React.forwardRef((props, ref) => (
+    <BlockChooser {...props} blockChooserRef={ref} />
+  )),
+  { forwardRef: true },
```
sneridagh, line 157: I would use `useIntl` because of this is a bit shady and not sure that it's supported in latest react-intl versions.

sneridagh (changes requested):
LGTM, but, naming (what else!)

Given the duality insert/mutate of the chooser, didn't we should reflect this in the name?

How about simply `BlockChooserButton`?

src/components/manage/BlockChooser/BlockChooser.jsx
tiberiuichim: Much better indeed. I've extracted this code from another, older PR that predates onInsertBlock... I should have given it a fresh look.

sneridagh (approved):

PR #2562, opened 2021-07-02
Pluggables on SSR

sneridagh:
Let the viewlets in Volto era begin!


sneridagh:
Well, it seems this is going to be complex, nevertheless, we are trying to achieve something that does not exist in the ecosystem :) . Status: It works, but the rehydration process complains, since it finds a DOM that it's not supposed to be there. Brainstorming:

- We can remove the node before rendering
- We can build a dynamic module that gets loaded and in the same tree renders the Plugs one after another, so the plugs are already are available inside their pluggables at the first client shot, given the information from the first pass.

/cc @tiberiuichim just come up with the second this morning... what do you think? Would it work?

sneridagh:
Closing, as it's a super long shot, and now we have slots

Issue #2521, opened 2021-06-15
URL validator doesn't accept path URLs


tiberiuichim:
When picking an internal document in the Event edit page, the URL validator for the Event URL field doesn't allow it.

**To Reproduce**
![Screenshot 2021-06-15 065221](https://user-images.githubusercontent.com/141568/121990458-7f5d0580-cda6-11eb-9988-54345d7e2ff3.png)

PR #2566, opened 2021-07-08
Cypress link test

tisto:


tisto:
@sneridagh just realized that we do not have a single test on the link redirect, is that possible?

tiberiuichim:
@tisto I wonder if the Link content type uses the URL widget? There's this issue with the Event ct: https://github.com/plone/volto/issues/2521

tisto:
@tiberiuichim it does use the URL widget.

PR #2580, opened 2021-07-13
add image gallery fallback if listingPreviewImageField is not available

jackahl:


src/components/manage/Blocks/Listing/ImageGallery.jsx
```diff
@@ -85,9 +85,19 @@ const ImageGalleryTemplate = ({ items }) => {
...
-    } else {
+    } else if (item.url) {
       imageSRCOriginal = `${item.url}/@@images/${settings.listingPreviewImageField}/large`;
       imageSRCThumb = `${item.url}/@@images/${settings.listingPreviewImageField}/thumb`;
+    } else if (item.url) {
+      imageSRCOriginal = flattenToAppURL(
+        `${item['@id']}/@@images/${settings.listingPreviewImageField}/large`,
+      );
+      flattenToAppURL(
+        (imageSRCThumb = `${item['@id']}/@@images/${settings.listingPreviewImageField}/thumb`),
```
sneridagh, line 96: This is not correct, an assignment in a function?

sneridagh (changes requested):

src/components/manage/Blocks/Listing/ImageGallery.jsx
jackahl: yea, obviously... No idea why that did not trigger an error.

jackahl: Fixed. No idea why I added the flattenToAppURL() there in the first place

sneridagh (approved):

PR #2583, opened 2021-07-15
fix: objectbrowsernav key + aria label fields

nzambello:
Fixes the `key` so if elements in different folders have the same name, we won't have rendering issues (I had), @id is unique.
Fixes some aria-labels because the id is not much human-readable, while the title does so it's more pleasant to hear.


pnicolli (approved):
LGTM

tisto (approved):
LGTM

sneridagh:
A number of Cypress tests already relies on them, should be fixed.

Could break a lot of project specific tests...

giuliaghisini (approved):

nzambello:
> A number of Cypress tests already relies on them, should be fixed.
> 
> Could break a lot of project specific tests...

Fixed that 👍

PR #2592, opened 2021-07-27
Route change scroll behavior

razvanMiu:
* Added anchor link functionality to `UniversalLink` component.
* New logic for `ScrollToTop` to handle anchor links and scroll restoration with configurable scroll behavior (`auto` or `smooth`)

Scroll behavior

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ### Feature
 
+- Added anchor link functionality to `UniversalLink` component. New logic for `ScrollToTop` to handle anchor links and scroll restoration with configurable scroll behavior (`auto` or `smooth`) @razvanMiu
+
 ### Bugfix
 
 ### Internal
--- a/src/components/manage/UniversalLink/UniversalLink.jsx
+++ b/src/components/manage/UniversalLink/UniversalLink.jsx
@@ -5,7 +5,7 @@
 
 import React from 'react';
 import PropTypes from 'prop-types';
-import { Link } from 'react-router-dom';
+import { Link, useHistory } from 'react-router-dom';
 import { useSelector } from 'react-redux';
 import { flattenToAppURL, isInternalURL } from '@plone/volto/helpers/Url/Url';
 import URLUtils from '@plone/volto/components/manage/AnchorPlugin/utils/URLUtils';
@@ -21,8 +21,13 @@ const UniversalLink = ({
   children,
   className = null,
   title = null,
+  stripHash = false,
+  delay = 0,
+  offset = null,
+  onClick = () => {},
   ...props
 }) => {
+  const history = useHistory();
   const token = useSelector((state) => state.userSession?.token);
 
   let url = href;
@@ -50,6 +55,10 @@ const UniversalLink = ({
     )?.length > 0;
   const isExternal = !isInternalURL(url) || isBlacklisted;
   const isDownload = (!isExternal && url.includes('@@download')) || download;
+  const appUrl = flattenToAppURL(url);
+
+  const path = stripHash ? appUrl.split('#')[0] : appUrl;
+  const hash = appUrl.split('#')[1];
 
   return isExternal ? (
     <a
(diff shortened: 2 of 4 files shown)


tiberiuichim (approved):

nzambello:
Just to say: I'm not a huge fan of big JavaScript manipulations, if I click on a link I'm not expecting much JS to run, I'd only expect a navigation.
This said, I think this can be very pleasant for users navigating the website.

Maybe is there a lighter way to do this? I know there are libraries like https://www.npmjs.com/package/react-scroll-restoration to accomplish a part of this, thus we can lighten UniversalLink.
UniversalLink is my worry and I'd like to keep it as simple as possible.

tisto:
This is indeed a big UX problem we currently face in Volto. From the user's perspective scroll sometimes happens randomly and ends up. That being said I fully agree with @nzambello here. We tried to tinker around with scrolling in our projects and failed every single time. This is a very complex use case and the right response to this is to simplify things, not add more complexity. We should discuss this in our next @plone/volto-team meeting.

cc @iFlameing

tiberiuichim:
This really needs to be fixed: https://github.com/plone/volto/blob/master/src/helpers/ScrollToTop/ScrollToTop.jsx#L32

It interferes with the SearchBlock facets.

iFlameing:
@tisto @nzambello @razvanMiu  I think that this is a complex problem. I already spend several days on this. Every time I tried to fix this, it is not 100%. I also tried the use the library like  https://www.npmjs.com/package/react-scroll-restoration but had hard luck. When I throttle my CPU and network I always see the browser scrolling to the top. Same with the browser back button. I really would like to check how the browser back button works with this implementation because changing scroll behavior to manual has some consequences.

iFlameing:
@tiberiuichim @tisto I did a fair amount of testing on this branch as well as master. Here is some finding.

1. The current implementation changes how the route changes. Previously, we first got to the top of the page and then changed the route, but this implementation changes the route first and then scroll to the top of the page. The smooth transition doesn't make much sense in the current implementation.
2. We still have the scrolling effect which we don't want.
3.Then only major difference this implementation makes is the first point. Other than that I don't think it adds anything because the current implementation is able to handle all the cases. Yes! there will be some time when the browser messes the scroll position using the back and front buttons. The current implementation handles it more effectively.

sneridagh:
@iFlameing when you say the current implementation, what are you referring to? This PR's? or the one in master?

iFlameing:
@sneridagh this pr.

razvanMiu:
@nzambello @tisto @sneridagh  Hi, sorry for not answering, I've been in holidays so:
1. For anchor links the current implementation scrolls to top for any changes of the route and doesn't handle anchors so to keep the scroll to top behavior and also handle the anchor links I overwrite the click event on UniversalLink (https://github.com/plone/volto/pull/2592/files#diff-c0021190896d8196577b4d304ac09c44abff5437fcf56791f7832c522f43ea01R86-R107), if the url has hash string into it, only to change the route (`history.push`) with a state (`volto_scroll_hash` -> needed so that we can strip the hash from url and keep the corresponding element id, `volto_scroll_offset` -> needed so that we can add y-offset) and also to add some optional delay before the link get triggered (a use case for this would be if you have an animated dropdown menu with anchor links and if you press a link you want to scroll to the corresponding element after the closing animation of the dropdown menu has finished - I can get into more details about this). @iFlameing You said `Previously, we first got to the top of the page and then changed the route, but this implementation changes the route first and then scroll to the top of the page` but in both implementation, master and this, the scroll happens after the change of the route so I'm not sure if I understand you correctly (https://github.com/plone/volto/blob/master/src/helpers/ScrollToTop/ScrollToTop.jsx#L33). Maybe your problem is that the scrolls happens after the content is loaded (https://github.com/plone/volto/blob/route_change_scroll_behavior/src/helpers/ScrollToTop/ScrollToTop.jsx#L80 -> line 83) but this is needed,  in my opinion, because the route is changed before the content of the new page is loaded. 
2. For the scroll restoration part my implementation is based on https://www.npmjs.com/package/react-scroll-restoration and is needed only to add the smooth scroll behavior and make use of the scrollTo function that waits for content load before scrolling. So we can keep the basic javascript scroll restoration but we can't make it smooth and we can't make sure that it will scroll to the right position (https://github.com/plone/volto/pull/2592/files#diff-1712f4ab0dc421cacb610886108bb1e5cd7978212604901774cf341a5fcf9ed1R9).
@tiberiuichim I'll take a look to see what is the incompatibility with SearchBlock facets.

tiberiuichim:
> @tiberiuichim I'll take a look to see what is the incompatibility with SearchBlock facets.

No need, it was a side-issue, not relevant in this PR. Just something to keep into consideration.

PR #2595, opened 2021-07-30
Disable change workflow for Image and Files in FolderContents view

giuliaghisini:
This is because in the manual selection of multiple items, or all items, if there is an image or a file (which do not have a workflow by their definition) it failed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Bugfix
 
 - Prevent form submit when clicking on BlockChooserButton @giuliaghisini
+- Disable change workflow for Image and Files in FolderContents view @giuliaghisini
 
 ### Internal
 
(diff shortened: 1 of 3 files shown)


tiberiuichim:
I think this problem should be handled in plone.restapi. (Probably) an invalid workflow action for a batch of items shouldn't trigger an error. @tisto thoughts on this?

giuliaghisini:
> I think this problem should be handled in plone.restapi. (Probably) an invalid workflow action for a batch of items shouldn't trigger an error. @tisto thoughts on this?

i totally agree. Maybe this pr will be only a quick fix for us..

tiberiuichim (approved):

giuliaghisini:
closed because handled here in plone.rest.api:  https://github.com/plone/plone.restapi/pull/1190

giuliaghisini:
https://github.com/plone/plone.restapi/pull/1190 closed because handled here in plone.rest.api

PR #2599, opened 2021-08-02
Fix SearchWidget search by path

giuliaghisini:


nzambello (approved):
LGTM

src/components/theme/SearchWidget/SearchWidget.jsx
```diff
@@ -69,30 +67,16 @@ class SearchWidget extends Component {
...
    * @param {event} event Event object.
    * @returns {undefined}
    */
   onSubmit(event) {
-    const section = this.state.section ? `&path=${this.props.pathname}` : '';
-    this.props.history.push(
-      `/search?SearchableText=${this.state.text}${section}`,
-    );
+    const path =
+      this.props.pathname?.length > 0 ? `&path=${this.props.pathname}` : '';
```
pnicolli, line 78:
```suggestion
      this.props.pathname?.length > 0 ? `&path=${encodeURIComponent(this.props.pathname)}` : '';
```

src/components/theme/SearchWidget/SearchWidget.jsx
```diff
@@ -69,30 +67,16 @@ class SearchWidget extends Component {
...
+    this.props.history.push(`/search?SearchableText=${this.state.text}${path}`);
```
pnicolli, line 79:
```suggestion
    this.props.history.push(`/search?SearchableText=${encodeURIComponent(this.state.text)}${path}`);
```

pnicolli (commented):
What do you think about using [encodeURIComponent](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent?retiredLocale=it) to safely build the search url?

nzambello:
> What do you think about using [encodeURIComponent](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent?retiredLocale=it) to safely build the search url?

@pnicolli nice one! We should only check if the parsing part would handle this correctly, but I think it's a nice to have!

src/components/theme/SearchWidget/SearchWidget.jsx
tiberiuichim: I like this API: https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams

pnicolli: Yes, that one's very nice. In a situation as simple as this one, though, I wouldn't go with URLSearchParams because of the worse browser compatibility it offers.

tiberiuichim: Works fine for our stack: https://github.com/plone/volto/blob/master/package.json#L188

tiberiuichim (approved):

sneridagh (approved):

PR #2596, opened 2021-08-02
implement item selection search filter in folder contents

nileshgulia1:
The search filter when selecting content items via dropdown doesn't seems to be implemented. 
Just to be sure, I'm talking about this:
Screenshot 2021-08-03 at 12 08 28 PM

PR #1697, opened 2020-07-24
fix: folder contents overflow

nileshgulia1:
fixes #1689

PR #2600, opened 2021-08-03
Fix overflow contents view table

giuliaghisini:
Added horizontal scrollbar if table contents exceed content width.

https://user-images.githubusercontent.com/51911425/128003389-1be228ba-d8da-4ec5-9ce2-8ed55c4622f2.mov


nileshgulia1:
@giuliaghisini same PR here: https://github.com/plone/volto/pull/1697

giuliaghisini:
> @giuliaghisini same PR here: #1697

oh, see now. 
But.. i think this PR is more complete, because scrollbar will appear only if needed. 
If you add only the 'overflow-x: auto' without handle the table styles, scrollbar will appear even if not needed because of .ui.attached.table styles. 
This PR takes care of this and of table styles.

nileshgulia1 (approved):
LGTM

tiberiuichim:
:) how do we get that last column pinned?

nzambello (approved):

nzambello:
> :) how do we get that last column pinned?

I guess we don't have any better option if not something _really_ tricky and hacky, and I wouldn't go for that right now

tiberiuichim (approved):
(not related to the PR, just a place to vent) That Contents.jsx is one of those big component files that need to be refactored, to allow us to "grow" with new functionality

ksuess:
I do not aggree with this PR. The folder contents view is for information and is being used to select an item for editing. The action 'edit' is already hidden in an '...' menu. So this menu should at least be visible, my opinion.
See this suggestion from last year: https://github.com/plone/volto/pull/1568#issuecomment-639467899 and imagine a beautiful designed version of it.

nileshgulia1:
> (not related to the PR, just a place to vent) That Contents.jsx is one of those big component files that need to be refactored, to allow us to "grow" with new functionality

@tiberiuichim one such example https://github.com/plone/volto/pull/2596

sneridagh:
I kind of agree with @ksuess, although is the kind of thing we were all searching for, it has the drawback that the ... menu is potentially not visible at first sight.

For reference, this is the Quanta mockup:

Screenshot 2021-08-18 at 10 02 37

Albert designed it even without it being always visible, I guess to remove visual clutter. I kind of like Google Drive's approach (it's behind a right click on the element), but agreed that it can be a bit too hidden for the average user.

Also, there's the ongoing agreement that the contents folder should have an app-ish look and feel, not to be constrained by the website "frame" structure. So I lean more towards Katja solution + Quanta additions + app-ish look and feel.

sneridagh:
I wonder if using react-table is sensible here:
https://codesandbox.io/s/react-table-fixed-columns-official-kowzlm5jp7

Using fixed columns seems easy. Then we could have all the filtering options.

tiberiuichim:
+1 on adding react-table as a dependency to Volto

ksuess:
- +1 for fixing column 'title' and column 'actions' ('react-table' seems to be an elegant implementation for fixing columns)
- +1 for showing an actions icon like it is done now (not a right click menu like in Google Drive. IMHO it is not obvious enough to be understood by rarely working users)

tiberiuichim:
@plone/volto-team Let's merge this, if possible.

sneridagh:
Agreed to merge unless we have a better approach based on the comments above in:

https://github.com/plone/volto/issues/2991

PR #2619, opened 2021-08-16
Fix singular and plural form of Placeholder in objectBrowserWidget

### LINKED ISSUE #2253: In mode='link', the label for the object_browser widget should not use plural
It says "No items selected", while only one item can be selected.

iRohitSingh:
Fix #2253


davisagli:
The PR is outdated, and I don't think it's incorrect as is, so I'll close the PR.

PR #2621, opened 2021-08-17
Search block

tiberiuichim:


src/start-client.jsx
```diff
@@ -24,7 +24,7 @@ initSentry(Sentry);
 function reactIntlErrorHandler(error) {
   if (process.env.NODE_ENV !== 'production') {
     /* eslint no-console: 0 */
-    console.info(error);
```
tiberiuichim, line 27: TODO: remove this comment before push

tiberiuichim:
A short demo, curtesy of Krisztina: https://www.youtube.com/watch?v=QbMeIlUtEr4

sneridagh:
What a terrific piece of code it is... Congrats and kudos for this!

I think having this is key and a differentiating factor for Volto (pretty much as the listing block). Also, as an example of the Volto technologies and patterns involved in its development, it's a small jewel that we should push forward.

What's left to do? Can I jump and help with something?

It would be great to have it ready for Volto 14.

tiberiuichim:
@kreafox Please look at this when you have the time:

- The tests are failing
- I don't think the reading URL parameters like ?SearchableText=something works anymore. The intention was to be able to replace the default Volto search page with this block, so it should be able to act like a drop-in replacement for that
- We don't have any support for sorting, maybe we should support that in some form.

sneridagh:
@kreafox the sort feature is awesome!! Btw, I noticed that the top variation is missing it?
Also a suggestion, we can improve it afterwards but how about to have an enable/disable toggle?

Thanks a lot!

kreafox:
@sneridagh Thanks! I've added for the top variation as well, I will check why is missing it.
And yes, it's good idea to have an enable/disable toggle! I will implement that too.
Thanks for the feedback!

tisto:
CI is failing.

sneridagh (approved):
LGTM

tisto (changes requested):
Awesome work @tiberiuichim and @kreafox!!!

This block is a feature monster and it will take some time to review and test every detail that we have. I will do an incremental review here:

Edit-Page (3)

Edit-Page (4)

Edit-Page (5)

Edit-Page (6)

Edit-Page (7)

CHANGELOG.md
```diff
@@ -6,8 +6,15 @@
 
 ### Feature
 
+- Add enable/disable toggle for sorting section in the search block, and fix missing sorting for the top variation @kreafox
```
tisto, line 9: @tiberiuichim @kreafox can we have a single changelog entry here please that says "Add search block"?

sneridagh:
Fixed:
- [x] Results literal
- [x] Use Headline instead of Title
- [x] Overall layout inconsistencies and polishments, group elements where necessary

Top:
image

Right:
image


## Comments

### Results empty instead of returning all content

Tricky one, since the default of an empty query string search is empty. We would have to modify the default behaviour or add a parameter to indicate that we want to trigger a wild full search at the API level.

### Number of search results matches with the limit number

This is another limitation of the querystring API, since the "items_total" is the limited number. Don't know if this is a query string along the Plone catalog limitation as well.

### Full facets working on edit mode

Is that possible? @tiberiuichim you told me about a "read only" mode for some widgets?

### Removal (or better defaults) for controls blocks settings

After careful consideration I think that having them is ok. Maybe agree on the defaults? I would leave the search box enabled by default, since seems key to me. Also, I can see the value of the button: Having it means "no live search", which is sensible to have. What I would remove is the labels, and the ability to customize them. Overall it has a problem: i18n! , and I think it's too much for the average user. Either we hide them (in a way that could be enabled later in the config) or remove them completely. Also, having them seems to me like redundant.

Some other bugs I've found on the way:
- [ ] Editing the query string not updating on save (the old one is triggered)
- [ ] `searchedText` never changes (in search results)
- [ ] Some facets cannot be set? Only the ones that have "lists"?
- [ ] Clear filters does not work, it gets bubbled by the unfold filters one

sneridagh:
### Sync URL-state

For now, I can see that there's one main problem present (already known by @tiberiuichim). It's about the state sync between URL-Block's State. It has some edge cases that still need to be fixed. Also, I think that in edit mode, the URL should not be playing along, and we should remove it. It has no sense. Also the "cleaning" of the state:

- When you save the block then edit again, the URL settings take precedence over the "config ones"
- When you navigate to other route, the URL updates a last time (console error shown about an update made from an unmounted component)
- When you edit the content, remove the block, add it again, it gets the "old config" (I guess it gets it from the URL, which in turn we remove it from edit, it will fix it)

### Facets that does not show (because they are not "lists")

There are a good bunch of indexes that have no value, then they are not rendered. Ideally we will have to write a proper FacetWidget for them and apply them depending on the index widget type (oddly, that's inside the operations). However, could we have a default one (like in the listing) using an Input component? and make the search work, of course... I pushed some code (I had problems with the value for it) in https://github.com/plone/volto/commit/0aca13fa1da35d4765479448bb00d8c01484c76a

I will push some more aesthetics today, but that's mostly all.

tiberiuichim:
Another bug: if there's no facet defined, the columns are still spaced for it

tiberiuichim:
Something else that should be consolidated: there's "editable", "mode", "isEditMode" just passed wildly around. It should be consolidated. This is also a problem in Volto, but here is like amplified.

tiberiuichim:
I propose that we show an activity indicator (spinner) in the search input.

tiberiuichim:
The sorting also needs to be redone. My plan for the sorting was to allow the editor to choose which sorting criterias to show in the dropdown, (maybe using the objectlistwidget or something along those lines)

tisto:
> The sorting also needs to be redone. My plan for the sorting was to allow the editor to choose which sorting criteria to show in the dropdown, (maybe using the objectlistwidget or something along those lines)

My 2c: the complexity of the search block is very high already and I guess we all would love to see the block in Volto 14 sooner than later. The more features we bring in, the more features we have to discuss, test, refactor, etc.

I would recommend that we focus on the existing features first. We can always add new features at a later point. Refactoring existing features might require another breaking change release. If the sorting is not ready and needs to be refactored, let's leave it out in the first version of the block and add it later. The search block with just the facets is more than impressive enough IMHO.

tiberiuichim:
> Full facets working on edit mode
> Is that possible? @tiberiuichim you told me about a "read only" mode for some widgets?

It would imply additional work. The main problem is converting the current facet value (or even resetting the whole search state) when the facet type is changed.

tiberiuichim:
@sneridagh We're now showing the Select and Checkbox facets as read only. A band-aid, but it's explicit and no more suprises for the editor. But I don't think it would be that difficult to reset the facet state when the search block data changes.

tisto:
> > Full facets working on edit mode
> > Is that possible? @tiberiuichim you told me about a "read only" mode for some widgets?
> 
> It would imply additional work. The main problem is converting the current facet value (or even resetting the whole search state) when the facet type is changed.

As mentioned before. I am personally totally fine if we polish and stabilize the features that we currently have (or even remove some). We can always add stuff later and we don't have to conquer the world in one piece. :)

tiberiuichim:
> In some estrange circumstances, the filters applied component appears, because I guess it gets confused between the URL and the current block settings.

@sneridagh Referencing the current code state as checkpoint, it happens when you copy/paste a URL with a search block and applied URL params.

sneridagh (approved):
I tested it and all problems with the URL-state sync are solved. As @tiberiuichim and @tisto pinpointed there might still be room for improvements, and new features, but the shape that it has now looks good to me for enter in the next alpha.

tisto:
@tiberiuichim @sneridagh FYI:

Search-Page

sneridagh:
@tisto @tiberiuichim I can reproduce it, it's an edge case when no query string is set up in the config, only the search text is to be search, since it evaluates to empty in the querystring query.

tiberiuichim:
@sneridagh @tisto I've traced the empty search text bug, it's like this:

- The ListingBody component from the listing block will not trigger a backend call when its passed query is empty. We should probably do that, and that's the main cause for this bug
- The querystring state in redux doesn't contain the original query parameters, so there's no way to determine if the content of the `querystring` reducer belongs to a particular query
- So the searchblock uses the "old" querystring reducer state, which is old and a new one hasn't been triggered.

I'll try to fix that in the search block, but that basically means just making the listing body bug a feature

src/components/manage/Blocks/Search/layout/RightColumnFacets.jsx
```diff
@@ -75,16 +75,19 @@ const RightColumnFacets = (props) => {
...
-            {...props}
-            isEditMode={isEditMode}
-            setFacets={(f) => {
-              flushSync(() => {
-                setFacets(f);
-                onTriggerSearch(searchedText || '', f);
-              });
-            }}
-          />
+          <div>
```
tiberiuichim, line 78: You should add the div to LeftColumnFacets as well.

tisto (approved):
@tiberiuichim @sneridagh I haven't tested everything again. I guess we are good to merge this and then continue to iterate and refine. Fantastic works folks (sorry for repeating that over and over :))!

Minor thing that shouldn't prevent the merge: I was wondering if we should make "multiple choices" checked by default under the facets. I guess multiple choices are more common than single choices. I have to admit I was confused when I choosed "checkbox" and then saw radio buttons (because single choices are the default).

PR #2625, opened 2021-08-18
This Enable us to change the  default content type view

### LINKED ISSUE #1998: LayoutViews: if is definied a View for a content-type, is not possibile to change the layout view
this, because in src/components/theme/View/View.jsx the view is calculated as : 
```
const RenderedView =
      this.getViewByType() || this.getViewByLayout() || this.getViewDefault();
```
I think it's correct to invert first two condition, in this way you could apply a layout view if is definied a different default TypeView:

```
const RenderedView =
      this.getViewByLayout() || this.getViewByType() || this.getViewDefault();
```

iRohitSingh:
Fix #1998


tiberiuichim (approved):
I agree with the change and it's obviously a bug fix, but it may represent something to test on with existing sites. Though I expect that the number of breaking possibilities is limited, as there needs to be:

- a content type with a registered Volto view
- that content type also sometimes gets assigned a layout, so the layout property is stored in the context.

giuliaghisini:
> I agree with the change and it's obviously a bug fix, but it may represent something to test on with existing sites. Though I expect that the number of breaking possibilities is limited, as there needs to be:
> 
> * a content type with a registered Volto view
> * that content type also sometimes gets assigned a layout, so the layout property is stored in the context.
> 
> @plone/volto-team thoughts on this?

Completely agree. I'm scared what would be happen in existing sites... we need to test it before merging

sneridagh:
I agree with you both, I would label it as breaking and targeted for a major for sure.

tiberiuichim (approved):

PR #2634, opened 2021-08-23
Showing version in History view

iFlameing:
@tisto screenshot of implementation

Screenshot 2021-08-23 at 6 00 46 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Breaking
 
 ### Feature
+- Showing Version in History View @iFlameing
 
 ### Bugfix
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 ### Breaking
 
 ### Feature
-- Showing Version in History View @iFlameing
+- Show version in history view @iFlameing
 
 ### Bugfix
 
--- a/src/components/manage/History/History.jsx
+++ b/src/components/manage/History/History.jsx
@@ -145,16 +145,19 @@ class History extends Component {
           <Table selectable compact singleLine attached>
             <Table.Header>
               <Table.Row>
-                <Table.HeaderCell width={4}>
+                <Table.HeaderCell width={3}>
                   <FormattedMessage id="What" defaultMessage="What" />
                 </Table.HeaderCell>
-                <Table.HeaderCell width={4}>
+                <Table.HeaderCell width={3}>
+                  <FormattedMessage id="Version" defaultMessage="Version" />
+                </Table.HeaderCell>
+                <Table.HeaderCell width={3}>
                   <FormattedMessage id="Who" defaultMessage="Who" />
                 </Table.HeaderCell>
-                <Table.HeaderCell width={4}>
+                <Table.HeaderCell width={3}>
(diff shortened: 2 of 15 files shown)


CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 ### Breaking
 
 ### Feature
+- Showing Version in History View @iFlameing
```
tisto, line 8: @iFlameing "Show version number in history view" (changelog entries should be: lower case, present tense)

src/components/manage/History/History.jsx
```diff
@@ -145,16 +145,19 @@ class History extends Component {
           <Table selectable compact singleLine attached>
             <Table.Header>
               <Table.Row>
-                <Table.HeaderCell width={4}>
+                <Table.HeaderCell width={3}>
                   <FormattedMessage id="What" defaultMessage="What" />
                 </Table.HeaderCell>
-                <Table.HeaderCell width={4}>
+                <Table.HeaderCell width={3}>
+                  <FormattedMessage id="Version" defaultMessage="Version" />
```
tisto, line 152: @iFlameing Maybe this should be "Version Number" of "Version #". Just thinking out loud here. I am undecided...other opinions? @tiberiuichim @avoinea @jackahl?

tisto (approved):
LGTM. Since Victor is on holiday this week, maybe someone from the @plone/volto-team can do a second review here.

src/components/manage/History/History.jsx
pnicolli: I think "Version" is clear enough, because there are actual number below it so I guess it should be easy for users to understand. I would be ok with "Version #" if you prefer having some hint. I wouldn't make the column header wider than that with a longer string, since the rest of the column will always be filled by small numbers.

On the other hand, I would suggest using a more verbose id for the translation string, in order to avoid possible clashes in the future. For example:

* version_number
* Version Number
* History Version
* History Version Number

pnicolli: In fact, being the version column so small, it could also be arguable to reduce its width to 2 and increase the What column to 4.

tiberiuichim:
I think "Version" is too much of a comitment. It implies that there's a widespread support for these versions, which I don't think we have right now.

My proposal is like this: 

- move the Version column to first position
- remove its title, so the column has no entry in the thead. This way occupies minimal space
- Move the "WHO" column to first position. 

With this change the table becomes readable like "Admin edited 5 minutes ago" and the version numbers can still be referenced in human communication.

tisto:
@tiberiuichim we could just use "#" as a title. This uses minimal space and has some meaning for some people. :)

I guess I'd prefer to keep the "WHAT" column as second because it contains the clickable link to the actual version. We should make the version number clickable as well I guess.

src/components/manage/History/History.jsx
tisto: +1 to the better translation strings.

tiberiuichim:
What about using the Feed component? https://react.semantic-ui.com/views/feed/

sneridagh (approved):

tiberiuichim (approved):

PR #2639, opened 2021-08-27
Fix reactSelectAsyncPaginate in ArrayWidget and SelectWidget, and fix…

giuliaghisini:
… required prop of SearchWidget.

This fixes loading error in ArrayWidget and SelectWidget introduced with this merge: https://github.com/plone/volto/pull/2563/. 

It also fixes required prop 'pathname' in SearchWidget, because it's not always required

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 
 - Footer: Point to plone.org instead of plone.com @ericof
 
+- Fix reactSelectAsyncPaginate in ArrayWidget and SelectWidget, and fix required prop of SearchWidget. @giuliaghisini
 
 ## 13.12.0 (2021-08-20)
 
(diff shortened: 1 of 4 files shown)


giuliaghisini:
> @giuliaghisini Actually it's an interesting issue... I've also encountered the same type of problem while working on the loadables. I'm curious of more details about how you "fixed" this problem.

i had problems running when running volto, but not when running a project with volto as a dependency.. 
My fix doesn't work in 'a child' project..

tiberiuichim:
@giuliaghisini So the behaviour is different then? Volto vs project? Can you confirm?

giuliaghisini:
i saw this.. but when i removed node_modules and reinstalled with >yarn install
everithing was ok

tiberiuichim:
oh, the joy and wonders of JS packaging :)

Issue #2615, opened 2021-08-16
Tags (subjects) management in contents view partially works


cekk:
When you select one or more contents to tag, the popup does not know which tags the contents have (and potentially can be removed) and which ones are available to add.

Both fields only allows to add new terms.

**To Reproduce**
1. create one page (doc-a) with "foo" tag
2. create another page without tags (doc-b)
3. go to /contents view, select doc-a and click on "tags"
4. tags to remove is an empty list where we can add new terms (it should be listed "foo")
5. tags to add is an empty list where we can add new terms

**Expected behavior**
Probably the easiest thing should be using a single field populated with content's tags where the user can remove unwanted, select some from Subjects list and/or add new ones.

I don't know how to expect when i select more than one content with different subjects.

PR #2641, opened 2021-08-30
Fix content tags modal

nzambello:
Fixes: #2615 

Adds:

- `array` and `token` as named widgets in the config
   (this because I had conflicts in the widget retrieval due to choices attribute)

Fixes: 
- fixes correct behavior of the modal to change tags

Now you can correctly set or remove tags to one or more contents.
Retrieves Subject vocabulary to show options to set while you can also create a new one.
Shows contents' actual tags to remove.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,12 +8,13 @@
 
 ### Bugfix
 
+- Fix for Contents tag modal @nzambello
+
 ### Internal
 
 - Updated Brazilian Portuguese translations @ericof
-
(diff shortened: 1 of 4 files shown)


tiberiuichim (approved):

nzambello:
Update: I added a fix in ContentsItem to show also array type indexes (eg: Subject).
Now tags are shown as comma separated values in theirs column (if that column is set, not default).

sneridagh (approved):

avoinea (approved):

Issue #2528, opened 2021-06-17
Client-side form validation should validate only touched fields


tiberiuichim:
If you have a content type with multiple required fields, in the edit form the validation error triggers for all fields once you blur the first required field. I think it should only validate fields that are dirty, then rely on server-side validation errors to trigger error for all fields (in case I only fill-in one required field, out of many)

tisto:
@tiberiuichim +1. It is best practice in form validation to only client-side validate a form field when the user interacted with that particular field (mouse click, type, etc).

PR #2642, opened 2021-08-30
Validate touched-only `required` fields in Forms

nileshgulia1:
Fixes https://github.com/plone/volto/issues/2528
The problem with "required fields" validation is they are all validated in one go, unlike other widget specific validations.
This is only required to be applied on `onBlur` method leaving `onSubmit` logic intact.
Before:

https://user-images.githubusercontent.com/22280901/131368199-7adaa1c7-622c-4d3b-8ad1-e31ee2afcfc9.mov

After:

https://user-images.githubusercontent.com/22280901/131368260-58b0c99c-add6-40ed-8299-1e13571e8869.mov

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,12 +8,13 @@
 
 ### Bugfix
 
+- Validate `required` touched-only fields in Form everywhere @nileshgulia1
+
 ### Internal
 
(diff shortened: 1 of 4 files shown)


tiberiuichim (approved):
nice, solves really annoying issue

sneridagh (approved):

PR #2654, opened 2021-09-03
Change the batch size of folder content

iFlameing:
@tisto screenshot

Screenshot 2021-09-03 at 4 05 54 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Breaking
 
 ### Feature
+
 - Change the batch size of folder content @iFlameing
 
 ### Bugfix
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Breaking
 
 ### Feature
+- Change the batch size of folder content @iFlameing
 
 ### Bugfix
 
--- a/src/components/manage/Contents/Contents.jsx
+++ b/src/components/manage/Contents/Contents.jsx
@@ -402,7 +402,7 @@ class Contents extends Component {
       items: this.props.items,
       filter: '',
       currentPage: 0,
-      pageSize: 15,
+      pageSize: 50,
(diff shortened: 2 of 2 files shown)


tisto (approved):

tisto:
@plone/volto-team we would like to change the default batch size in the folder_contents to 50 and remove the 15 and 30 options. Are you folks ok with that? We'd like another review before merging.

avoinea (approved):

nzambello:
@tisto I think changing the default can be ok, but why removing other options?

tisto:
@nzambello batching options of 15 and 30 just do not make any sense IMHO. Why would you want to switch to 15/30 when you see all the items in 50? This is just a leftover from ancient times of the web, they do not make any sense today.

tisto:
@nzambello in the long run I'd like to get rid of this entirely. I see the reason for an "all" button but batching does not make much sense. Maybe we can lazy load content or just add a "more" button at the end which is more user-friendly IMHO.

nileshgulia1 (approved):

tisto:
@avoinea with 10k objects, it does not make any difference. The default batch size is 50 that always works. If you have 10k objects and click on "all" you are screwed, no matter if we stick with the current approach or improve it.

avoinea:
@tisto True, for the `All` button. I was referring to this statement `but batching does not make much sense.`

But indeed, maybe with a smart filter and a sorting functionality, this may be dropped.

The use case I have in mind is when you have 10k items and you want to go to the last page and remove old items. Or, I don't know see all items published 5 years ago.

tisto:
@avoinea I guess we fully agree here. I'd like to replace batching if possible. Though, we have to take all the use cases into account and make sure they will still work when we change things. The current PR is a good compromise. We can move in the right direction and start a discussion about where we might want to go. :)

Issue #2657, opened 2021-09-13
SelectWidget / ArrayWidget cache search results and on the second load they display an incomplete vocabulary


avoinea:
Update: Try on Chrome if Firefox doesn't show any interactions within the video playback

https://user-images.githubusercontent.com/323385/133121773-c5be5bc9-708e-478a-9e64-54795a4b336d.mp4

avoinea:
Someone did it on purpose:

https://github.com/plone/volto/blob/ec386e1b276138e6be0730b2aea1f798af2a2a4c/src/components/manage/Widgets/SelectWidget.jsx#L308-L313

nileshgulia1:
Not connected to this but might be good to take care also #2540

PR #2659, opened 2021-09-13
Reset vocabulary on SelectWidget / ArrayWidget component unmount

avoinea:
See #2657


avoinea:
99% works, but if you mix a multi-choice widget with a select widget on the same page with the same vocabulary it behaves strange.

avoinea:
I'll try to add a cypress for it.

PR #2661, opened 2021-09-14
(re) Fix auth unified login

sneridagh:
Opening a new PR for the new unified auth feature, #2535 for reference.

These are the issues:

- [ ] Toolbar flashes when not logged in (#2687)
- [x] No way to logout effectively from the UI (already known, but turned out essential for this PR) #2677 
- [ ] It seems that when logged in with Zope admin it also behaves strange


rpatterson:
I believe everything here has been rolled into [the subsequent PR](https://github.com/plone/volto/pull/2687).  LMK and/or re-open if I'm wrong about that.

PR #2669, opened 2021-09-18
Only externalize node built-ins

tiberiuichim:
The purpose of this PR is to make it possible to `yarn build` a production build then just copy the public + build folders on a server and be able to completely run Volto with just that.

With the changes here I was able to do that, but there's a strangeness in the `formidable` package that requires a patch.

To replicate:

- pull this PR
- apply formidable patch
- yarn build
- copy build + public folder to another location
- in that location, run `RAZZLE_API_PATH=http://localhost:8080/Plone NODE_ENV=production node build/server.js`

### DIFF
--- a/razzle.config.js
+++ b/razzle.config.js
@@ -251,23 +251,8 @@ const defaultModify = ({
     });
   }
 
-  config.externals =
-    target === 'node'
-      ? [
-          nodeExternals({
-            whitelist: [
-              dev ? 'webpack/hot/poll?300' : null,
-              /\.(eot|woff|woff2|ttf|otf)$/,
-              /\.(svg|png|jpg|jpeg|gif|ico)$/,
-              /\.(mp4|mp3|ogg|swf|webp)$/,
-              /\.(css|scss|sass|sss|less)$/,
-              // Add support for whitelist external (ie. node_modules npm published packages)
-              ...addonsAsExternals,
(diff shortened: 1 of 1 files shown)


patches/patchit.sh
```diff
@@ -1,2 +1,3 @@
 #!/bin/bash
 patch --quiet -p0 -N node_modules/razzle/config/createJestConfig.js < patches/razzle-jest.patch
+git apply --directory=node_modules/formidable/ patches/fixformidable.diff
```
sneridagh, line 3: Why git and not patch?

sneridagh (approved):

sneridagh:
Well, the tests didn't like it at all 😂

patches/patchit.sh
tiberiuichim: I don't know the patch magic incantation to apply to multiple files.

tiberiuichim:
```
Treating warnings as errors because process.env.CI = true.
Most CI servers set it automatically.

Failed to compile.

./node_modules/express/lib/view.js
Critical dependency: the request of a dependency is an expression
```

Webpack spits out warnings and CI treats them as errors. I could make a list of these special dependencies and put them in a subfolder with a package.json, for easy unified instal.

davisagli:
Closing as abandoned

PR #2670, opened 2021-09-18
Proper EN i18n defaults

sneridagh:
I might be wrong, but I have a hunch that this is the cause of the unchained `Cannot format message: "X", using message id as fallback` message.

It was a difficult one, to decide the behaviour. So I opted for not touching the JSON, but act on the po, because of the add-on use case. So we do not get JSON language files from addons, but POs. So, we need to act at PO level.

Also, there was the alternative to not act on anything, just translate manually every time in the EN PO, but I would have sucked. Like this is automatic and it's aware of custom translations that might happen there anyways.

Problem could be that if we ever, change defaults in JS, then we need to act on the EN PO.

Not an easy one...

### DIFF
--- a/src/i18n.js
+++ b/src/i18n.js
@@ -167,13 +167,19 @@ function poToJson() {
       items = [...libItems, ...items];
     }
 
-    // Write it
+    // Write the corresponding language JSON, cover the special EN use case for including
+    // defaults if not present
     fs.writeFileSync(
       `locales/${lang}.json`,
       JSON.stringify(
         zipObject(
           map(items, (item) => item.msgid),
-          map(items, (item) => item.msgstr[0]),
+          map(items, (item) =>
+            lang === 'en'
+              ? item.msgstr[0] ||
+                item.comments[0].replace('defaultMessage: ', '')
+              : item.msgstr[0],
+          ),
         ),
       ),
     );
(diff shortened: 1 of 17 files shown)


sneridagh:
I need to transfer it to the addons `i18n.js` script. @tiberiuichim one reason more to separate it from them in their own package.

sneridagh:
@tiberiuichim @erral not 100% on this one tbh, might be trying the other alternative (hack the en.json instead) if I could add the default message as a comment in the po, grab it from there.

src/i18n.js
```diff
@@ -223,6 +223,33 @@ ${map(pot.items, (item) => {
...
+    filename,
+    `${formatHeader(po.comments, po.headers)}
+${map(keys(getMessages()).sort(), (key) => {
+  const poItem = find(po.items, { msgid: key });
+  return [
+    `${map(poItem.references, (ref) => `#: ${ref}`).join('\n')}`,
+    `msgid "${poItem.msgid}"`,
+    `msgstr "${
+      poItem?.msgstr?.length && poItem?.msgstr[0] !== ''
+        ? poItem.msgstr[0]
```
tiberiuichim, line 245: If I understand the code well, you don't overwrite the English translation? I think we should take the Volto default messages as the canonical and always overwrite. I understand that there's a use case where somebody might want to overwrite the default messages. In that case a custom integration/workflow should be implemented (by those that want it).

sneridagh:
OK, I changed the approach and I'm more satisfied with this one.

So, added the default to all po files, hacking it only into the JSON EN if the value is not present.

sneridagh:
Superseded by: #2672

PR #2672, opened 2021-09-18
I18n own package

sneridagh:


tiberiuichim (approved):
I like this approach.

sneridagh:
Projects can't work on "local path" Volto deps... we need to make a release before it can work.

docs/source/upgrade-guide/index.md
```diff
@@ -15,6 +15,44 @@ This upgrade guide lists all breaking changes in Volto and explains the
...
+-    "i18n": "NODE_ENV=production node src/i18n.js"
++    "i18n": "rm -rf build/messages && NODE_ENV=production i18n --addon"
++  },
+```
+
+and add this to the `dependencies` list:
+
+```diff
++  "dependencies": {
++    "@plone/scripts": "latest"
```
tiberiuichim, line 42: maybe it should be "@plone/scripts": "*" ?

sneridagh: I was hesitant... even tempted to pin it. But then we might want an updater script...

sneridagh: Is there a way to mark "whatever one you have already inherited from the environment"? Yesterday trying, * it just got the latest one... :(

tiberiuichim: I think it's not an issue if you get the latest one, but I'm thinking that maybe someone, at one day, will want to pin it at their project level.

tisto (approved):
LGTM

PR #2676, opened 2021-09-20
Do not show lead image block when the content type does not have the …

sneridagh:
…behavior enabled


sneridagh:
A one liner. A good example that we have strong foundations in our framework.

tiberiuichim (approved):

tisto (approved):

PR #2677, opened 2021-09-20
Fix logout action, removing the  cookie as well, if present.

sneridagh:


tisto (approved):

tiberiuichim (approved):

rpatterson (commented):
Thanks for catching what I missed here.

PR #2680, opened 2021-09-20
Add i18n coverage gh action

tisto:


tisto:
@erral Erico told me you are working on something similar (for Jenkins). Just FYI: I am looking for a gh actions check that tells me when German translations are not 100%. I am committed to keeping the German translation coverage to 100% (all the time).

erral:
I asked @gforcada to prepare a jenkins job for volto that runs regularly (once a day), like the existing one for Plone translations, to check the status:

https://jenkins.plone.org/view/Translations/

I runs i18ndude against the PO files to extract stats for both plone.app.locales and Volto.

Currently the plone.app.locales build is failing, but I want to have a look at it before branching plone.app.locales for Plone 6.

tisto:
@erral do you plan to come to Sorrento this year? It would be great to have you there to be able to push the i18n story of Plone 6!

PR #2684, opened 2021-09-21
Use the block `title` as the source of the translation `id` instead o…

sneridagh:
…f using the `id` of the block

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Breaking
 
+- Use the block's `title` as the source of the translation `id` instead of using the `id` of the block. See upgrade guide for more information. @sneridagh
+
 ### Feature
 
 ### Bugfix
--- a/docs/source/upgrade-guide/index.md
+++ b/docs/source/upgrade-guide/index.md
@@ -43,6 +43,11 @@ Not really a breaking change, but it's worth noting it. By default, Volto 14 com
     />
 ```
 
+### Blocks chooser now looks uses the title instead of the id of the block as translation source
+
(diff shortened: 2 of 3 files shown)


avoinea (approved):

docs/source/upgrade-guide/index.md
```diff
@@ -43,6 +43,11 @@ Not really a breaking change, but it's worth noting it. By default, Volto 14 com
     />
 ```
 
+### Blocks chooser now looks uses the title instead of the id of the block as translation source
```
tiberiuichim, line 46: extra "looks" word, should be removed

tiberiuichim (changes requested):

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 ### Breaking
 
+- Use the block's `title` as the source of the translation `id` instead of using the `id` of the block. See upgrade guide for more information. @sneridagh
```
ksuess, line 7: Glad that you found this bug!

Correct me if I'm wrong, but the first `id`is too much.
"Use the block's `title` as the source of the translation instead of using the `id` of the block. See upgrade guide for more information. @sneridagh"

ksuess (changes requested):

CHANGELOG.md
sneridagh: Done! Thanks!

docs/source/upgrade-guide/index.md
sneridagh: Done

PR #2687, opened 2021-09-21
Fix flashy toolbar in 2661

sneridagh:
This try to fix the problems in #2661 with the flashy toolbar if the user is not logged in, which where a little bit deeper than I expected, and no, we cannot live with the flash of the toolbar in a live site.


tiberiuichim (approved):

sneridagh:
So, once I've delved into the real problem, tried to adjust Volto to it... 

Cypress now is clunky and flaky because having to make the types endpoint depend on the actions response (we need to know in advance if we are really authenticated from the backend, before we can issue the @types call or not). A bunch of components that relied on the token to be present (which was immediate), now have to rely the backend response for it as well. All these provokes timing issues in Cypress (could be addressed probably, but hard). Also, relying on the backend call to know if we are *really* logged in, entails other problems, like, you don't have the username upfront, nor token, which should also be taken care.

These are my 5c: the changes this PR # 26621 introduces are far too substantial in how Volto behaves. Don't know if we (and the projects we have to maintain) are ready for it now, or if we could address it differently. 

BTW, help is needed to fix Cypress (and probably other problems could all the changes provoke)...

@plone/volto-team Thoughts?

tisto:
@sneridagh I agree that the risk is too high ATM. We would need a company that uses Volto extensively on a daily basis to step up and test this for a while in prod and is willing to fix all the issues that come up before we can consider trying to merge this again.

avoinea (approved):

src/components/theme/App/App.jsx
```diff
@@ -196,6 +197,14 @@ export const __test__ = connect(
 
 export default compose(
   asyncConnect([
+    {
```
rpatterson, line 200: Can you elaborate on how this is related to to the `getTypes` changes above?

src/components/manage/Toolbar/Toolbar.jsx
```diff
@@ -194,9 +194,10 @@ class Toolbar extends Component {
    * @returns {undefined}
    */
   componentDidMount() {
-    this.props.listActions(getBaseUrl(this.props.pathname));
-    this.props.getTypes(getBaseUrl(this.props.pathname));
-    this.props.setExpandedToolbar(this.state.expanded);
+    this.props
+      .listActions(getBaseUrl(this.props.pathname))
+      .then(() => this.props.getTypes(getBaseUrl(this.props.pathname)));
```
rpatterson, line 199: Can you elaborate more on this than "fix @types missing"?  Again, I'm sure it's obvious to you but I at least could use some edification.  How is `getTypes` related to `listActions`?  I tried reading the definition of the `getTypes` function but neither the documentation comments nor the code says much about what or why it is.

Also again, is this related to my previous changes or cleanup you found along the way?

src/selectors/userSession/userSession.js
```diff
@@ -15,7 +15,13 @@ export function loggedIn(state) {
   /* The user may be authenticated by different means, including outside the UI.  Defer
    * to the response from Plone, sepcifically whether Plone presents an option to log
    * in. */
-  return !state.actions.actionsById.user.login;
+
+  if (
+    typeof state.actions.actionsById.user.login === 'undefined' &&
+    typeof state.actions.actionsById.user.logout === 'undefined'
+  )
+    return false;
```
rpatterson, line 23: Thanks for fixing my mistake!  :-)

src/selectors/userSession/userSession.js
```diff
@@ -15,7 +15,13 @@ export function loggedIn(state) {
...
+  return !!state.actions.actionsById.user?.logout;
```
rpatterson, line 24: Can you elaborate on why you switched from `!...login` to `!!...logout`?  Maybe even add a comment on that if the reason is something someone else reading the code should know?

src/selectors/userSession/useSession.test.js
```diff
@@ -0,0 +1,75 @@
+import { loggedIn } from './userSession';
+import { arrayWIdsToObject } from '@plone/volto/helpers/Utils/Utils';
+
+describe('loggedIn selector', () => {
```
rpatterson, line 4: More than anything, thank you *so* much for adding tests for *my* changes!

rpatterson (commented):
Thanks much for fixing my mistakes!  IIRC, I noticed that flash when I was making my changes but didn't understand how to address it.  I've commented on the individual commits and I look forward to understanding further.  In the meantime, I'm firing up the test bed I was using before to confirm your fixes.

rpatterson:
> Cypress now is clunky and flaky because having to make the types endpoint depend on the actions response

I'm guessing this means that [the integration tests that run using Cypress](https://docs.voltocms.com/developer-guidelines/testing/#developing-cypress-tests) are now having issues related to my changes?  Can you elaborate?  Are there specific failures?  Errors?  Intermittent behavior?

> you don't have the username upfront

Yeah, my plan for that when I was last working on this was to add a new API endpoint to return the user data that isn't included from the actions.  I know I wrote that up before but I can't find it quickly ATM, so please forgive the omission.  I'll try to remember to link back here when I do find it.

sneridagh:
@rpatterson answering to all here. 

I had to refresh myself on the issue, and the problem, sorry for not getting to you earlier. 

First of all, I recreated the PR (from the reverted PR) in 
https://github.com/plone/volto/pull/2661
and this PR is trying to solve the problems in there, so you can continue from here or there, whatever you want.

## Core problem

First, let's define the core problem here in a TL;DR fashion: Volto relies in something "tangible" as is the JWT token. As tangible I mean that can be checked immediately, and lives in the user browser. Changing to anything async backend-based will be *hard* and difficult. Since now the JWT token it's the cornerstone where all the architecture relies on. So shouldn't be taken lightly and should be a thoughtful decision, as should be the implementation.

As you can see, I tried hard to succeed and spent already quite some time in this PR (given my experience in Volto) and I miserably failed. My take: as it is now, it can't be done in the way we would like to. Sorry for being that hard on this, but if someone else is able to get it done right given the current Volto needs and specifications, please show us. 

So, I think as main outcome, that the whole architecture of the solution should be rethought. After you take an initial look at the depth of the problem and how we desire to fix it, we can have a meeting with the Volto Team and discuss about it all.

Also, since Eau de Web is pushing for it, I would like someone from the EdW Volto Team to help you during all the implementation and act as a liaison to the Volto Team.

## Requirements

- The toolbar should be permission aware, and dependant on it
- Some other components are permission aware too, so it should be taken into account too
- The @types endpoint should only be hit if we have to (egg. permissions to add)
- The username (which we get now from the JWT token) should be known
- The JWT token should be issued and put in place, if there's none and we are authenticated by other means
- The "unified login" should be implemented as opt-in, if possible
- We should test it thoroughly, both in Cypress and in real projects before merging

## My personal take on this

To be honest, right now, I don't see this feature worth the effort nor the hassle. Also we are moving to a recommended way of deploying Plone 6 in a way that classic is completely hidden (using the new JSON-only traversal). So I don't see it as a priority right now. Also, I think that if really needed, a more lightweight solution could be more suitable/sustainable (like, eg. unifying cookies? we did that in the past, and although one could argue that it's not elegant, it's effective and works. We are anyways issuing them in the backend).

Also, we have now in our plate challenges far more interesting and valuable to solve than this one. For me the wins if this succeeds are very scarce, and poorly valuable comparing to other features. /cc @alecghica 

## Permissions handling in Volto

I would like to shed some history on this one too. So this has been a recurrent item in the Volto's to-do list. Effectively we don't have a way to know specific permissions of a user in an specific context. We always relied in inner implementation details for endpoints (and that worked so far). However, we can't do a thing that Plone does all the time when showing/hiding features in UI (or access to those features): checking for specific user permissions. Also another more trivial things (eg. show the registration link or not) that is registry setting powered.

In the future, we would like to get an unified way of doing this. I do not know if it's via the @actions endpoint or not. But we will need a way to handle these more effectively and performant way.

One of the reasons that we still haven't managed to handle it yet, is that we still do not know what's the silver bullet for this.

## Questions

> Can you elaborate on why this should be a thunk?

If everything relies on the response from the backend to the `@actions` endpoint, then some actions should fire (or not) after knowing if we are or not logged in, right? Then, this is used later in the Toolbar component, one as a result of the other. So it needed to be a thunk. I might not be aware of the latest Redux best practices, but how you make one action dependent on another? This is the way we did always.

> Can you elaborate more on this than "fix @types missing"? 

Same as above...

> Can you elaborate on why you switched from !...login to !!...logout? 

I do not remember now, but I had a problem of "nullish" value handling here (eg. null !== false), and of naming as far as I remember. I wrote the tests for getting it right. If I'm not mistaken it revolved around the fact that I don't want to know if I'm logged in, but I want I really want to know is if I'm logged out.

>I'm guessing this means that the integration tests that run using Cypress are now having issues related to my changes?  Can you elaborate? Are there specific failures? Errors? Intermittent behavior?

You can see for yourself, and run cypress locally if you want (see the setup in the `package.json` scripts), take a look at the PR checks:
https://github.com/plone/volto/runs/3673409172?check_suite_focus=true
https://github.com/plone/volto/runs/3673416890?check_suite_focus=true

> Yeah, my plan for that when I was last working on this was to add a new API endpoint to return the user data that isn't included from the actions. I know I wrote that up before but I can't find it quickly ATM, so please forgive the omission. I'll try to remember to link back here when I do find it.

We need all the things in "Requirements" as a whole, and working successfully in the resultant PR.

rpatterson:
> First, let's define the core problem here in a TL;DR fashion: Volto relies in something "tangible" as is the JWT token. As tangible I mean that can be checked immediately, and lives in the user browser. Changing to anything async backend-based will be _hard_ and difficult. Since now the JWT token it's the cornerstone where all the architecture relies on. So shouldn't be taken lightly and should be a thoughtful decision, as should be the implementation.

Ah, this is very clearly stated and an important requirement to have well defined.  I'm not necessarily convinced by the rationalizations but I'll take this as direction. Thanks!

> if someone else is able to get it done right given the current Volto needs and specifications, please show us.

I the requirement is that the Volto Redux UI state be based on, and only on, the JWT token, then to my mind that re-frames the core issue here to be one that all the involved PAS authentication plugins be kept in sync.  I don't have any reason to think that isn't achievable on the API/back-end side.  On the Volto Redux side the only additional handling required would be to recognize a JWT cookie when the authenticated session was initiated outside of the Volto login component.  This doesn't seem too invasive or hard to maintain given the benefit of being able to be agnostic as to how authentication was established.  This probably also removes the need for a new user data API endpoint, at lease not for this.

> Also, since Eau de Web is pushing for it, I would like someone from the EdW Volto Team to help you during all the implementation and act as a liaison to the Volto Team.

I'm certainly happy to talk/pair/work with anyone.  If you're feeling that I'm directing too many of my questions or too much of my discussion at you, then please feel free to consider comments in PRs from me as directed at no one in particular and I'll refrain from @'ing you.  ;-)

> ## Requirements
> * The toolbar should be permission aware, and dependant on it
> * Some other components are permission aware too, so it should be taken into account too
> * The @types endpoint should only be hit if we have to (egg. permissions to add)

IMO, these 3 requirements are outside of the scope of this specific concern, namely unifying authentication across the 3 existing ways of authenticating to Plone/Zope. Agreed?

BTW, this permissions stuff a problem I'd love to be a part of solving!  Zope's ACL/permissions system is one of the things I most love from Zope and tackling the ways it meshes and clashes with modern web UI seems like it would be a fun challenge.

> > Can you elaborate on why this should be a thunk?
> 
> If everything relies on the response from the backend to the `@actions` endpoint, then some actions should fire (or not) after knowing if we are or not logged in, right? Then, this is used later in the Toolbar component, one as a result of the other. So it needed to be a thunk. I might not be aware of the latest Redux best practices, but how you make one action dependent on another? This is the way we did always.

Ah, gotcha, so thunks are for "chaining"" actions.  If anyone cares to elaborate on *why* a different thing is needed for this in Redux and *how* thunks are the right thing, I feel like understanding that would make me a better Redux developer.

> I'm guessing this means that the integration tests that run using Cypress are now having issues related to my changes?  Can you elaborate? Are there specific failures? Errors? Intermittent behavior?
> 
> You can see for yourself, and run cypress locally if you want (see the setup in the `package.json` scripts), take a look at the PR checks:
> https://github.com/plone/volto/runs/3673409172?check_suite_focus=true
> https://github.com/plone/volto/runs/3673416890?check_suite_focus=true

Thanks!  I haven't decided, given how the requirements are changing, if that will be next, but I'll dig into those failures at some point.

> We need all the things in "Requirements" as a whole, and working successfully in the resultant PR.

Well, I think doing all this in one PR stacks a whole lot on top of it and makes for a large, hard to review PR.  I'm not convinced there aren't ways we could do this in smaller chunks.  And I'm saying that as someone who's PRs are way too often too large for the teams I'm working on.  ;-) That said, I'll take that as direction and make it work as one PR.

Thanks much for your time!

sneridagh:
> I the requirement is that the Volto Redux UI state be based on, and only on, the JWT token, then to my mind that re-frames the core issue here to be one that all the involved PAS authentication plugins be kept in sync. I don't have any reason to think that isn't achievable on the API/back-end side. On the Volto Redux side the only additional handling required would be to recognize a JWT cookie when the authenticated session was initiated outside of the Volto login component. This doesn't seem too invasive or hard to maintain given the benefit of being able to be agnostic as to how authentication was established. This probably also removes the need for a new user data API endpoint, at lease not for this.
> 

Unfortunately, Plone 6 UI is more than a simple client-side Redux application. It does have its own requirements as a whole. 
 
> I'm certainly happy to talk/pair/work with anyone. If you're feeling that I'm directing too many of my questions or too much of my discussion at you, then please feel free to consider comments in PRs from me as directed at no one in particular and I'll refrain from @'ing you. ;-)

I'm more than happy to help on this one, but my time is limited too, and since this is effort sponsored by a Volto company, I think it's better that someone from the company that are part of the Volto Team this pairs with you and answers your questions about the app architecture and behaviour. I am open to meet and have a discussion on how we could design the solution in a way that is workable.
 
> > ## Requirements
> > 
> > * The toolbar should be permission aware, and dependant on it
> > * Some other components are permission aware too, so it should be taken into account too
> > * The @types endpoint should only be hit if we have to (egg. permissions to add)
> 
> IMO, these 3 requirements are outside of the scope of this specific concern, namely unifying authentication across the 3 existing ways of authenticating to Plone/Zope. Agreed?

To be honest I can't see now if these requirements are not met, we could merge the feature... If you break the Toolbar behaviour, but your feature works well, how could we merge it? Plone 6 is bigger than a single feature. We can't release a product whose quality is below the previous release, right? This feature moves the corner stone of Plone 6 UI. If after you move it, you don't left things in the same place, then better not move it.

> BTW, this permissions stuff a problem I'd love to be a part of solving! Zope's ACL/permissions system is one of the things I most love from Zope and tackling the ways it meshes and clashes with modern web UI seems like it would be a fun challenge.

That would be great!

rpatterson:
Oops, I didn't notice ths PR was against my previous/old branch which I just deleted.  I thought it was obseleted by this branch.  At any rate, I think these changes should be reviewed against `master` anyways.  I don't see a way to edit what branch a PR is against, so I'll open a new PR.

PR #2693, opened 2021-09-24
Defaults in blocks form

sneridagh:


sneridagh:
I do not like the fact that we will have to issue a breaking change because the new addition of `onChangeBlock` in `InlineForm`... of course, the feature just won't work without it (I added a guard there), but I guess it counts as breaking, for the mandatory update of all the instantiations.

tiberiuichim (approved):
I've create a PR to this... I'm not 100% sure that it fullproof, but I don't like disabling that hooks warning.

sneridagh:
It's a "componentDidMount".

sneridagh:
This last commit will fix the undefined when editing, going back to the default value :)  Defaults, only should be applied once. I'm afraid that by doing this:

https://github.com/plone/volto/pull/2693/files#diff-152a160265a543b914f56865c5c6b77c87866f006303b0682d087b90e2b675daL95-L100

we are breaking this fix:
https://github.com/plone/volto/pull/2299
for checkboxes. @razvanMiu @tiberiuichim thoughts?

src/components/manage/Form/InlineForm.test.jsx
```diff
@@ -11,14 +11,33 @@ const mockStore = configureStore();
 function NewBaseWidget(name) {
   return (props) => (
     <div id={`mocked-field-${props.id}`} className={`mocked-${name}-widget`}>
-      {props.title || 'No title'} - {props.description || 'No description'}
-      {props.value}
+      {props.title || 'No title'} - {props.description || 'No description'} -
```
tiberiuichim, line 14: needs i18n

src/components/manage/Form/InlineForm.jsx
```diff
@@ -92,11 +129,7 @@ const InlineForm = ({
               id={field}
               fieldSet={defaultFieldset.title.toLowerCase()}
               focus={index === focusIndex}
-              value={
-                'default' in schema.properties[field]
-                  ? formData[field] ?? schema.properties[field].default
-                  : formData[field]
-              }
+              value={formData[field]}
```
tiberiuichim, line 132: I think (but this needs to be checked) that this should be `formData[field] || null`, otherwise we're creating input components initially with 'undefined', then switching them to controlled inputs.

PR #2700, opened 2021-09-27
Refactor Image block: make it schema extensible

sneridagh:
The idea is to refactor all the Volto stock blocks to be schema drive, so they allow enhancing and extending with variations and the future style wrapper.


sneridagh:
@nileshgulia1 I took a look at your latest commits, some questions:
- We cannot change the field names... otherwise we are forcing people to make a migration of their blocks, and that's bad. Why did you do that? I saw the bbb code but still... Imagine the people that shadow the Image component.
- The Edit block does have the inline chooser everywhere in Pastanaga and in Quanta... and the Volto users are used to that. We can hide it in the sidebar then so there is only one place to set it.

sneridagh:
We also had the click in the first place, but it was nasty and people clicked on it by mistake (opening the upload selector all the time) so we removed it.

nileshgulia1:
> We cannot change the field names... otherwise we are forcing people to make a migration of their blocks, and that's bad. Why did you do that? I saw the bbb code but still... Imagine the people that shadow the Image component.

@sneridagh I was aware of that in the first place. I stumbled upon a few "deserialization" issues when using URLs from objectBrowser and manual. In restapi it expects URL in the form of [string](https://github.com/plone/plone.restapi/blob/d6f327ba320922a4e88a684a45730ed557e0ea04/src/plone/restapi/deserializer/utils.py#L20) thus it stucks when supplied scattered URLs (list).

I saw your hack proposed as convert URL as "string" before `onChangeBlock`, but then we have to re-convert it to objectBrowser data structure in Edit like [here](https://github.com/plone/volto/pull/2700/files#diff-998a9a97123da23e5c6831ee82fed7db452ff87aa4b6f96d8c770060dc2af15cR95)

Although, we are doing that successfully on the listing block https://github.com/plone/volto/blob/master/src/components/manage/Blocks/Listing/DefaultTemplate.jsx#L10

I propose we can have that check in the [restapi](https://github.com/plone/plone.restapi/blob/master/src/plone/restapi/deserializer/blocks.py#L187) itself, or else use `TextWidget` with custom actions instead of `objectBrowserWidget` in the sidebar(This is what we currently have). I will have to re-inspect the implementation though.

>We also had the click in the first place, but it was nasty and people clicked on it by mistake (opening the upload selector all the time) so we removed it.

Agree. If we have to stick with inline input( and remove objectBrowserWidget ), then it make sense to leave the "upload" button intact.

Thanks for feedback!

sneridagh:
@nileshgulia1 also, we can't rename the field, since the transforms in the backend for the resolveuid expect them to be in `url`. :( If you could revert that, it would be great. We can find a way to deal with that, Rob created a prop for it in the ObjectBrowser, that do not return an array.

nileshgulia1:
@sneridagh Done. I kept the `url` field in the schema, so that it isn't breaking now. I just noticed we can use `urlWidget` instead of `objectBrowser` in the sidebar field and allow users to add a new image using inline input( with objectBrowser ). 

In addition to this, I also added a storybook for the new image_sizes widget.

Screenshot 2022-05-09 at 4 38 37 PM

sneridagh:
It already looks good, two cents:

- The upper part of the image sidebar always had a miniature of the image itself. We could bring it back, since even Quanta has it.
- It's a pity that the Image Block, which is one of the most important blocks remains with a widget that is so ugly and the UX sucks. Can't we make the `objectBrowser` support the image same as the `multiple` mode variant has? (the one with the tokenised item) /cc @giuliaghisini 

I know that the image block was the first `objectBrowser` implementation and that it evolved (at some point we would like to refactor it, since right now it's a mess due to all the incarnations that it had over the years). Couldn't we improve it?

sneridagh:
@tiberiuichim @giuliaghisini @robgietema @pnicolli ^^

tiberiuichim:
There's somewhere an "AttachedImageWidget" https://github.com/eea/volto-block-image-cards/blob/master/src/ImageCards/widgets/AttachedImageWidget.jsx

giuliaghisini:
> * It's a pity that the Image Block, which is one of the most important blocks remains with a widget that is so ugly and the UX sucks. Can't we make the `objectBrowser` support the image same as the `multiple` mode variant has? (the one with the tokenised item) Retaining backwards compatibility, of course. /cc @giuliaghisini

Object browser already handles images when 'mode' is 'image': 
https://github.com/plone/volto/blob/866b160032329175f6b571f36daf69e1bc0c2765/src/components/manage/Sidebar/ObjectBrowserBody.jsx#L265

https://github.com/plone/volto/blob/866b160032329175f6b571f36daf69e1bc0c2765/src/components/manage/Widgets/ObjectBrowserWidget.jsx#L66

nileshgulia1:
@giuliaghisini  What difference does objectBrowserWidget have in respect of its [data](https://github.com/plone/volto/blob/master/src/components/manage/Widgets/ObjectBrowserWidget.jsx#L191) when in mode `image` vs in mode `multiple`? 
(In the context of Image block ) we don't have proper deserializer in restapi to account for object or array `url` fields. It should be handled in either frontend(image mode) or backend.

Also we can show preview of the image when in mode `image` just like what @tiberiuichim  suggested.

sneridagh:
LGTM to me!! (I can't approve a PR that I created :) )

giuliaghisini (approved):

PR #2701, opened 2021-09-27
Use debug library to optionally show i18n and shadowing errors

sneridagh:


tiberiuichim (approved):

tiberiuichim (commented):
I don't think it's needed to prefix the var with RAZZLE. I'd go for "VOLTO_DEBUG", so we can reuse the var, maybe?

tiberiuichim:
Ah, or rather, we should use that debugger library so we can do 

```
DEBUG="volto" yarn start
```

I think it's being used by superagent, I always do like `DEBUG="superagent"`. Also, we use it in our yo generators.

sneridagh:
This? https://www.npmjs.com/package/debug

I like it!

sneridagh:
DEBUG=volto:i18n
DEBUG=volto:shadowing

and then DEBUG=volto:* ?

tiberiuichim:
@sneridagh Yes yes, that would be the best

PR #2716, opened 2021-10-02
Documentation cleanup

ichim-david:
Bloc's anatomy.md and extensions.md have received some cleanup.
More important is the proofreading of the upgrade guide as it contained many
hard to read sentences due to grammatical inconsistencies


ichim-david:
@sneridagh I don't know if you would like for someone else from the docs team from plone to have a look at my changes.
They should be safe to merge without any extra checking as I used a grammatical checker as well but you never know :) 
Still, I think that they would be better after the merge than before.

tiberiuichim (approved):

sneridagh (approved):

sneridagh:
@ichim-david no need gfor that now, I guess that review will happen when we merge the docs with the upcoming doc revamp that the docs team are working on.

Thanks for taking care of this!

PR #2723, opened 2021-10-06
Improved mobile navigation

sneridagh:
I did this for the conference site. What do you think?

https://user-images.githubusercontent.com/486927/136258226-e751dd50-23c2-4c85-8eeb-8e860e7944e9.mov

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,11 +4,14 @@
 
 ### Breaking
 
+- Improve mobile navigation menu with a nicer interaction and a fixed overlay with a drawer (customizable via CSSTransitionGroup) animation @sneridagh
+
 ### Feature
 
 ### Bugfix
 
 ### Internal
+
 - Improved developer documentation. Proof read several chapters, most importantly the upgrade guide @ichim-david
 
 ## 14.0.0-alpha.14 (2021-10-01)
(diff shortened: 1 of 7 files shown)


tisto:
@sneridagh looks great!

tiberiuichim (approved):

pnicolli:
Looks good! 🎉 
But I would not merge until the scroll issue is fixed, it would be a regression.

tisto:
@ichim-david awesome!

ichim-david:
@sneridagh I've merged master into your branch
This was the mobile menu before my changes:
before

And after my changes:
after

I've added main.variables where I've added several variables that give the mobile menu the current look.

I would really prefer that we use the Semantic UI theme pattern as much as possible where we have stylistic CSS added
through variables, this way we can just override the variables and therefore we compile the CSS without adding any extra styles .

I know it's hard to do and takes a long time but it would make it that much easier than for the next person to turn the Pastanaga theme into their own without having to resort to CSS overrides that need to be even more specific.

ichim-david:
@sneridagh  I've merged master into your branch
This was the mobile menu before my changes:
before

And after my changes:
after

I've added main.variables where I've added several variables that give the mobile menu the current look.

I would really prefer that we use the Semantic UI theme pattern as much as possible where we have stylistic CSS added
through variables, this way we can just override the variables and therefore we compile the CSS without adding any extra styles .

I know it's hard to do and takes a long time but it would make it that much easier than for the next person to turn the Pastanaga theme into their own without having to resort to CSS overrides that need to be even more specific.

PR #2727, opened 2021-10-08
#2728: add cypress test that checks if pages are shown when not effective

ThomasKindermann:


ThomasKindermann:
#2728

davisagli:
Closing as outdated. In the meantime timezone issues with the effective date field were fixed, so maybe this isn't a problem now.

PR #2731, opened 2021-10-11
Listings do not use fullobjects

sneridagh:


ionlizarazu:
In which cases does the listing get the extra fields of a content type? Would be nice if we could choose when do we want to get the extra fields, mostly when we have a customized content type and template or variation

ionlizarazu:
I understand the options, but the developer will have to make another query, right? Why not make it easier and have a config to give the developer the usual template or variation develop?

tiberiuichim (commented):
What about adding a new setting in each listing block variation, which is to set either fullObjects or a list of metadata fields? Then we can avoid the bbb setting, as whoever needs it can just patch the variations.

sneridagh:
@tiberiuichim won't that be too much? Like in "I have to go through all my variations and decide if that one meets requirements or not"? I mean, it would be handy, of course to have this flexibility, but the migration can be tedious.

sneridagh:
Tests are still failing because we need to merge https://github.com/plone/plone.volto/pull/20 first.

sneridagh:
@tisto when things are green in https://github.com/plone/plone.volto/pull/20 we need a release.

tisto:
@sneridagh released!

sneridagh:
and, next one will be green as a broccoli :)

tisto:
@sneridagh for the plone.volto compatibility matrix: Volto 14 requires plone.volto 3.1.0 (which requires plone.restapi 8.13.0). Is that correct?

sneridagh:
@tisto yes!

sneridagh:
@tiberiuichim would you be ok if we merge this as it is? Or do you think the ability of choose the use of fullobjects in the variations is a big deal? We can hold it until Tuesday meeting, then decide. But after this one Volto 14 is good to go.

docs/source/upgrade-guide/index.md
```diff
@@ -132,6 +132,13 @@ the block title. Before, it took the `id` of the block, which is utterly wrong a
 Following the same convention as the above change, `Variation` field coming from the block enhancers now uses the `title` of the block as source for translating
 the variation title. Before, it took the `id` of the block, which as stated before, is wrong and missleading. There is a chance that  this change will trigger untranslated variation titles in your projects and add-ons.
 
+### Listing block no longer retrieve fullobjects
```
tisto, line 135: @sneridagh can we add a section to the regular docs that explains the differences between fullobjects and brains? Or is that something that is already fully documented on plone.restapi?

tiberiuichim:
@sneridagh I'm fine with that

PR #2732, opened 2021-10-12
Get widget by tagged values

ksuess:
Goal: Apply special widget for a field by 

`config.js`
`config.widgets.widget.specialwidget = SpecialWidget;`

`plone.schema`
```
directives.widget(
    'fieldname',
    frontendOptions={
        "widget": 'specialwidget',
        "version": 'extra'
    })
```

Coming back to the discussion of https://github.com/plone/plone.restapi/pull/1172#issuecomment-877766270 I would like to suggest to make no changes at all on plone.restapi.
This is my point of view now that I played with several options. I realised that all needed information is already there in restapi response. So picking up widgetOptions?.frontendOptions?.widget would get the backend tagged value for frontend options. I understand that this means to rely on a convention to use the tagged value 'frontendOptions'. Which then would have to be very clearly documented.

### DIFF
--- a/docs/source/recipes/widget.md
+++ b/docs/source/recipes/widget.md
@@ -1,5 +1,30 @@
 # Forms and widgets
 
+## Set frontend widget
+
+If you want to register a frontend widget for your field, you can define your field such as:
+
+```python
+directives.widget(
+    'specialfield',
+    frontendOptions={
+        "widget": 'specialwidget',
+        "version": 'extra'
+    })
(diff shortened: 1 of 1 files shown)


sneridagh:
I just have to say that I love it. The fact that we change nothing from p.restapi is a plus.

Thanks to tinker and work to push this forward! I think to have this in place will open a lot of improvements to be done in Plone for it to be more expressive on specify which widget should be used given a field.

PR #2740, opened 2021-10-14
Improve search block

tiberiuichim:
Fix many aspects of the search block. Each commit should be pretty atomic, so at the end I'll gather a list of what has been changed.


tisto:
@tiberiuichim man, you rock!!!

Showing initial results works.

Restricting the "sort on" values works like a charm!

Bildschirmfoto 2021-10-17 um 10 37 45

Bildschirmfoto 2021-10-17 um 10 40 44

Issue #2742, opened 2021-10-15
SelectWidget: in form-edit mode, SelectWidget with vocabulary lost selected value when changing tab


giuliaghisini:
**To Reproduce**

https://user-images.githubusercontent.com/51911425/137442717-9768118b-6db5-4921-9a01-dc7505cb253f.mov

PR #2743, opened 2021-10-15
Fix: Fixed SelectWidget: when there was a selected value, the selecti…

giuliaghisini:
…on was lost when the tab was changed

Fixes: https://github.com/plone/volto/issues/2742

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@
 
 ### Bugfix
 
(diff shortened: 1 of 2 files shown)


giuliaghisini:
> It seems to me that we need a ton of Cypress tests around the Selects/Array/Token widgets :/
> 
> For the Fanzone list for sure.
> 
> btw, @giuliaghisini did you check if this flaw is also present in Array/Token?

no Victor, i didn't check Array and Token widgets. We can't at the moment. If someone else would like to check it will be very apprecieted.

Issue #2721, opened 2021-10-06
Private top-level folder don't appear in the top navigation after logging in


tiberiuichim:
If there's any private top-level sections in the website, you have to refresh the page before they appear in the main top-level navigation.

davisagli:
Can't reproduce this currently.

PR #2751, opened 2021-10-17
Use only App.jsx asyncConnect for getNavigation dispatching

tiberiuichim:
This fixes the issue of having to reload the page when logging in, to be able to see private top level folders, see https://github.com/plone/volto/issues/2721

Also fixes the issue when deleting a top-level page in Contents (but you have to step out of contents view, as Contents is also asyncConnected. The `__SERVER__` condition should be removed from there, as well.

Looking for feedback on this... should we pursue this path?

### DIFF
--- a/src/components/theme/App/App.jsx
+++ b/src/components/theme/App/App.jsx
@@ -206,7 +206,6 @@ export default compose(
     {
       key: 'navigation',
       promise: ({ location, store: { dispatch } }) =>
-        __SERVER__ &&
         dispatch(
           getNavigation(
             getBaseUrl(location.pathname),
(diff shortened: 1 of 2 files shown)


tiberiuichim:
The tests would have to be updated... the `@navigation` endpoint is no longer (re)loaded on everything. I'd call that an desirable optimization.

avoinea (approved):

tiberiuichim:
As a workaround, in case this PR doesn't get through, here's how I did it in my project:


```
  config.settings.asyncPropsExtenders = [
    ...(config.settings.asyncPropsExtenders || []),
    {
      key: 'navigation',
      extend: (dispatchActions) => {
        const action = {
          key: 'navigation',
          promise: ({ location, store: { dispatch } }) => {
            // Customized, we don't want SERVER as a condition
            return dispatch(
              getNavigation(
                getBaseUrl(location.pathname),
                config.settings.navDepth,
              ),
            );
          },
        };
        return [
          ...dispatchActions.filter(({ key }) => key !== 'navigation'),
          action,
        ];
      },
    },
];
```

davisagli:
Outdated (Navigation component was refactored to a functional component). This might still be relevant, but it's usually not needed since navigation is loaded as an expander by default.

PR #2755, opened 2021-10-18
add a custom schema to the listing block default template

ionlizarazu:
I think that with the listing block and the use of the templates we are creating a very powerful tool, so I add this feature that I think can be very powerful.

I've added a schema to the listing block DefaultTemplate, so we can customize the fields wich are comming the now strict "title" and "description". This will be useful for content_types where we don't necessarily want to use the default fields and the default design is perfect for us. I have not added it, but we can also modify some other characteristics in this way such as the number of columns, show image or not...

Here a gif with the functionality added:
listing block template schema

@sneridagh @tiberiuichim

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ### Feature
 
+- add a custom schema to the listing block default template @ionlizarazu
+
 ### Bugfix
 
 ### Internal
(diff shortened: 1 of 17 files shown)


tiberiuichim:
If I remember well, Timo wanted the "headline" property added in the block, so we want those 2 extra fields.

PR #2757, opened 2021-10-19
Updated Volto production sites list.

giuliaghisini:


tisto:
@giuliaghisini w00t! Wow, that's quite an impressive list Giulia! Red Turtle rocks! Thanks for sharing!

PR #2763, opened 2021-10-20
Fix loading of cookie on SSR for certain requests, revert slight chan…

sneridagh:
…ge in how they are loaded introduced in alpha 16

@tisto weirdly enough, I cannot make it fail in Cypress (but out of Cypress, under the testing server conditions, it fails). So I couldn't write a suitable test. I guess it has to do with async-ish-ness...

### DIFF
--- a/src/helpers/Api/Api.js
+++ b/src/helpers/Api/Api.js
@@ -56,6 +56,7 @@ class Api {
           const authToken = cookie.load('auth_token');
           if (authToken) {
             request.set('Authorization', `Bearer ${authToken}`);
+            if (__SERVER__) request.set('Cookie', `auth_token=${authToken}`);
           }
 
           request.set('Accept', 'application/json');
--- a/src/helpers/AuthToken/AuthToken.js
+++ b/src/helpers/AuthToken/AuthToken.js
@@ -53,10 +53,12 @@ export function persistAuthToken(store) {
           cookie.remove('auth_token', { path: '/' });
         }
       } else {
-        cookie.save('auth_token', currentValue, {
-          path: '/',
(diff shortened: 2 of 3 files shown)


tiberiuichim:
When the SSR text/html calls comes in to Volto, it converts it to several backend api json calls (main "getContent", `@actions`, etc). The problem is that these calls should have the Authorization Bearer header set, but they don't. So we need to make sure that these calls get that header set.

tiberiuichim:
The flow of info on /edit:

- App.jsx getContent executes, it has the proper auth token
- Then Edit.jsx's getContent (and `getActions`) executes, they don't have the auth token

App.jsx's getContent when executed goes through the HPM:

```
authtoken eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJzdWIiOiJhZG1pbiIsImV4cCI6MTYzNDc5NTY5NCwiZnVsbG5hbWUiOm51bGx9.Ts_5AonWrnaieawjckAVeDgbDBrQ2V-6fuii803l-Gs /test-private-image
  superagent GET http://localhost:3000/++api++/test-private-image +2m
[HPM] Router new target: undefined -> "http://localhost:8080"
[HPM] GET /++api++/test-private-image ≈> http://localhost:8080
  superagent GET http://localhost:3000/++api++/test-private-image -> 200 +36ms



edit getContent
authtoken undefined /test-private-image
  superagent GET http://localhost:3000/++api++/test-private-image +3ms
[HPM] Router new target: undefined -> "http://localhost:8080"
[HPM] GET /++api++/test-private-image ≈> http://localhost:8080
  superagent GET http://localhost:3000/++api++/test-private-image -> 401 +10ms
```

PR #2785, opened 2021-10-27
Dedupe all requests when they are still pending

### LINKED ISSUE #2781: Duplicated requests
**Describe the bug**
Volto makes duplicated queries to the backend during the same render.

**To Reproduce**
1. Launch a brand new Plone 6.0.0a1 backend
2. Launch current Volto
3. Check network tab of the browser inspector
4. There are 2 ``@actions`` requests

**Expected behavior**
Only 1 ``@actions`` requests. In general, it should not allow duplicated requests to happen if multiple components use the same action with the same arguments.

pnicolli:
Fixes #2781


pnicolli:
@sneridagh 
Updated with latest master. You told me you had concerns with using a module scoped variable for storing pending requests, we can talk about it if you have any improvement ideas or any pointers to continue improving the PR.

This is, as said before, a first implementation that only removes duplicate queries that happen simultaneously. We could try to implement some client-side caching feature where we track requests and responses and serve cache-first replies and can decide on a fetch policy and so on.... but I would rather consider moving completely to react-query or swr something similar that already handles this situation and other cases instead. I'm open for discussion.

wesleybl:
Today, when I access Volto anonymously, I see two requests to `@actions`:

volto_requests


And it looks like the second request is pending, until the request to `++api++/` finishes. So this PR is welcome.

@pnicolli can you resolve the conflict that is occurring?

sneridagh:
@pnicolli reviewing what we have in store before final... I'm sorry that this rooted here :( 
I think we share now the vision of going towards a full react-query integration, which will solve these problems and others.
I really don't know if this is worth of merging and having it in the current Volto.
Maybe something we can talk during the conference?

pnicolli:
Yeah we can do that for sure :)

pnicolli:
@sneridagh I think we should close this and just focus on `@plone/client`, feel free to close if you agree

PR #2786, opened 2021-10-27
Micromanage and customize SemanticUI bundle. Move theme import to top.

sneridagh:


avoinea (approved):

ichim-david (approved):
@sneridagh  as previously talked, I would apply only the theme import after the config
and leave the micromanaging for another ticket.
Since we import theme.js after the config we could have a conditional import
where if you enable some configuration you get the minimal semantic.less from volto
otherwise you get the normal semantic.less.

If the minimal is not enabled by default it could even be done as part of this ticket,
what do you think about this proposal?

avoinea (approved):

nileshgulia1 (approved):

sneridagh:
For the sake of completeness, we did move the Theme import to the top already. The rest is on the fridge, but we would want that at some point.

sneridagh:
Closing as I think it's no longer relevant, and someone could already customize it in their projects fairly easily.

Issue #2707, opened 2021-09-30
Groups Control Panel sends request to show all groups and displays them in batches.


tisto:
**Describe the bug**

The Groups Control Panel ("controlpanel/groups") batches the available groups in the UI. Though, the frontend sends a request to retrieve all groups:

https://localhost:8080/Plone/@groups

This way we get the worse of two worlds:

a) we batch the results and require the user to click on the batches, even though this is not necessary
b) we send a request that takes long and might even time out dependent on the number of groups

**To Reproduce**

1. Fire up Plone backend and Volto frontend
2. Login as admin
3. Go to the groups control panel
4. Check the dev tool for the request

**Expected behavior**

We either drop the batching altogether and add a loading spinner (to load all groups). I guess for the groups that should be fine since there won't be thousands of them. Or we do real batching and send a request that restricts the number of the result set and allows fast response times.

**Screenshots**

<img width="2407" alt="Screenshot 2021-09-30 at 16 16 59" src="https://user-images.githubusercontent.com/632105/135472500-7cb2c4a0-3fa5-4f89-8f7f-341061a2da6d.png">

nileshgulia1:
I propose having a "user controlpanel settings" behaviour in plone.restapi and then corresponding frontend to change these settings right from volto. We have this in Plone 6 alpha.

PR #2792, opened 2021-10-29
add batching to groups controlpanel

nileshgulia1:
See #2707 
We might also need a PR in plone.restapi.


nileshgulia1:
obsolete.

PR #2795, opened 2021-10-30
Enable typescript compatibility

pnicolli:
This is to enable volto sites to use typescript. Nothing changes for who doesn't want to use TS.

I will add a section in the documentation with instructions on how to use typescript before merging.


tiberiuichim (approved):

nzambello (approved):
Add an entry in the changelog and document how to use typescript in projects or addons and you have green light from me 🍻

nileshgulia1 (approved):

pnicolli:
Ok, I tested things and read some documentation about typescript and razzle and I'm currently writing the needed documentation. It should be done by tomorrow.

nzambello (approved):
LGTM, good job @pnicolli 🍻

sneridagh (commented):
Thanks for this one! I

LGTM, however, I would like to be more explicit in the documentation. Stating explicitly that TS in core is not supported, but this enables the ability to provide TS compatibility to projects and add-ons. Like it is now, it's unclear if TS in core is allowed or not. 

I know that the changes to be done are only in the documentation, and thus, in core Volto they are not applied (for a reason), but it's not clear the files that you have to amend are the project ones.

Could we be more explicit? Small changes, here and there, and a big disclaimer stating the big fact.

sneridagh:
@pnicolli ^^ I hope it's fine for you!

pnicolli:
> @pnicolli ^^ I hope it's fine for you!

Yes, absolutely. Thanks for taking care of that, I had no time in the previous days and I was catching up right now.

PR #2796, opened 2021-10-30
controlpanels: multi select users/groups and ability to delete, improve styling on `checkboxes` (*the Plone 6 way*)

nileshgulia1:
A bit more improvement to the designs in controlpanels, Ability to select multiple users/groups and apply actions such as `"delete"`.  I think we should also unify the use of checkboxes and other widgets to all over sites. 

For instance,  checkbox is using different designs when used as widget and as standalone( according to prototypes/designs). Maybe we should refactor them....?

Screenshot 2021-10-31 at 12 36 03 AM

### DIFF
--- a/src/components/manage/Controlpanels/Groups/GroupsControlpanel.jsx
+++ b/src/components/manage/Controlpanels/Groups/GroupsControlpanel.jsx
@@ -21,11 +21,7 @@ import {
 } from '@plone/volto/components';
 import { Link } from 'react-router-dom';
 import { Helmet, messages } from '@plone/volto/helpers';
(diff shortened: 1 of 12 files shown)


nileshgulia1:
Related PR: #2800

nileshgulia1:
Also Depends on https://github.com/plone/plone.restapi/pull/1325

nileshgulia1:
@ksuess We can also plan to merge this along with your other PRs.

nileshgulia1:
Tested and works fine with latest plone.restapi and #3349

PR #2800, opened 2021-11-01
style checkboxes

nileshgulia1:
We use `<Icon/>` svgs for rendering this type of checkboxes. The older way also doesn't look as per the quanta/pastanaga designs. We could also do this on project level, but almost no client asks for it. It can be a nice idea to have this in core.

older:
Screenshot 2021-11-01 at 5 00 03 PM

newer:
Screenshot 2021-11-01 at 4 59 46 PM


theme/themes/pastanaga/extras/main.variables
```diff
@@ -11,3 +11,8 @@
 /* Mobile Menu Nav Item */
 @mobileMenuNavItemJustifyContent: center;
 @mobileMenuNavItemLineHeight: 1.25;
+
+/* Checkbox */
+@checkboxBorderUnchecked: 1px solid #826a6a;
```
nileshgulia1, line 16: @sneridagh Any better suggestion for color? I copied this from pastanaga designs mockups from zeplin.

sneridagh: Pastanaga is ok, since it's the old stuff. In your screenshot is it possible that it has a shadow? 
I can't see any in Pastanaga mocks. Could you please check?

image

nileshgulia1: @sneridagh Done! Also I think shadows looks best with lighter borders🤔?

nileshgulia1:
Tests are failing due to latest release in https://github.com/plone/plone.volto/blob/7dd17bb8b070631a77352e6e0fab437e21b39a02/setup.py#L59

tisto:
@nileshgulia1 @sneridagh don't we pin all version dependencies in Volto? We shouldn't have any checkouts or unpinned versions in any CI build.

nileshgulia1:
@sneridagh Would you also consider this as a breaking change? if yes, do you think we can also let it in volto 14?

sneridagh (approved):

PR #2801, opened 2021-11-01
Overlay on ObjectBrowser is only shown when opening the objectbrowser the second time

### LINKED ISSUE #2497: Overlay on ObjectBrowser is only shown when opening the objectbrowser the second time.
^

iRohitSingh:
Fix #2497

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Bugfix
 
+- Fix opening of overlay on first click @iRohitSingh
+
 ### Internal
 
 ## 14.0.0-alpha.26 (2021-11-01)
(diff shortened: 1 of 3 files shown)


iRohitSingh:
@sneridagh i don't know why the cypress test is failing .It gives me time-out error.

tiberiuichim (commented):
Do we actually need the overlay condition? I didn't get to look at the surrounding code, but it seems strange.

iRohitSingh:
@tiberiuichim I think that we don't want to show the overlay when opening object browser maybe that's why we have overlay condition.

davisagli:
Outdated PR. The fix might still be relevant, but it would need to be updated to the current repository structure.

PR #2809, opened 2021-11-02
Fix gap of image block in Paragraph styling that shows in chrome

### LINKED ISSUE #2768: Paragraph styling: paragraph shows gap when used with an image block
**Describe the bug**

**To Reproduce**
1. As admin
2. Create a page with multiple paragraphs and images, e.g. https://volto.kitconcept.dev/images
3. There is a gap before text paragraphs after an image

**Expected behavior**

Paragraphs and images are always on the same level.

**Screenshots**

Images (6)

iRohitSingh:
Fixed #2768

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Bugfix
 
+- Fix gap of image block in Paragraph styling that shows in chrome @iRohitSingh
+
 ### Internal
 
 ## 14.0.0-alpha.26 (2021-11-01)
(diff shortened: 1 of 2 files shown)


ichim-david (changes requested):
@iRohitSingh Although my proposal quickly fixed the visual issue, it doesn't address situations where you want to have margins on the bottom.

My proposal would be to put these values behind less variables set to 0.
This way, if other people need other values, then can simply override the variable value.

davisagli:
Closing as abandoned

PR #2811, opened 2021-11-02
Fix default in block variation select; avoid block API in InlineForm

tiberiuichim:
…me imports


sneridagh (approved):

sneridagh:
@tiberiuichim So instructions are outdated, and it's no longer a breaking, right?

PR #2822, opened 2021-11-08
fixed downloadableObjects default value

giuliaghisini:
fixed default value from 'Link' to 'File'.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Bugfix
 
+- Fix downloadableObjects default value @giuliaghisini
+
 ### Internal
 
 - Remove bundlesize @tisto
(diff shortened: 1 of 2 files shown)


tiberiuichim (approved):
I wish we'd have one-line comment to some of the settings that are not obvious. It makes it easy to read the config module as a self-referencing source, so that we wouldn't have to cross-reference the docs. Something along https://github.com/plone/volto/blob/c49b2bd77f67193402dfc8e4f5d6e4115ba19efe/src/config/index.js#L121-L122

tiberiuichim:
In this case the comment would be something like "internal links for this will lead to download". Or maybe it's obvious enough without the comment? Anyway, consider it just a suggestion.

giuliaghisini:
> In this case the comment would be something like "internal links for this will lead to download". Or maybe it's obvious enough without the comment? Anyway, consider it just a suggestion.

i've updated the comment. What do you think?

nileshgulia1 (approved):

Issue #2824, opened 2021-11-09
Search block live search shouldn't be the default


tiberiuichim:
Also add a "clear" for the text input.

Issue #2745, opened 2021-10-15
Search Block: Clicking the "x" inside the facet dropdown field does not remove the "x".


tisto:
**Describe the bug**

Clicking the "x" inside the facet dropdown field does not remove the "x".

**To Reproduce**
1. Go to https://volto.kitconcept.dev/search-block
2. Choose any type facet
3. Click on the "x" within the type facet field
4. "x" does not go away

**Expected behavior**

"x" is removed.

**Screenshots**

![Search (2)](https://user-images.githubusercontent.com/632105/137537481-b5625544-36b3-4f2e-99c8-1a4fb0884201.png)

tisto:
@tiberiuichim ^^^

tiberiuichim:
@tisto I couldn't reproduce the issue with this PR: https://github.com/plone/volto/pull/2740

Could you give that one a try?

ichim-david:
@tisto this was fixed sometime as I am looking at your demo page and clicking on a selected value removes it and the x disappears.
I've tried with FF, Edge, and Safari and it works with all
https://volto.kitconcept.dev/blocks-examples/search-block#query=%5B%7B%22i%22%3A%22path%22%2C%22o%22%3A%22paqo.string.absolutePath%22%2C%22v%22%3A%22%2Fexamples%22%7D%2C%7B%22i%22%3A%22portal_type%22%2C%22o%22%3A%22paqo.selection.is%22%2C%22v%22%3A%22Collection%22%7D%2C%7B%22i%22%3A%22review_state%22%2C%22o%22%3A%22paqo.selection.is%22%2C%22v%22%3A%22spam%22%7D%2C%7B%22i%22%3A%22show_inactive%22%2C%22o%22%3A%22paqo.boolean.isTrue%22%2C%22v%22%3A%22%22%7D%5D&sort_order=descending

One thing I might complain perhaps as a new issue is that you see the x turn dark as if it's active but you cannot **Tab** to it and then press **Enter** to remove that selection.
You can remove the selection using **Backspace** but I wonder how many would know that and not try instead to operate with that x.
If you think it's worth it we can add an issue and see when it will be fixed.

ichim-david:
Closing as fixed in https://github.com/plone/volto/pull/2828

tisto:
@ichim-david @tiberiuichim I can not reproduce this problem any longer:

[...]

PR #2828, opened 2021-11-11
Search block improvements

kreafox:
Fixed some issues of the search block:

- Disable default live search and added a clear button for search input box (#2824)
- Fixed facet dropdown clear button issue (#2745)
- Removed sorting label customization option in order to keep the search block simplified
- Small layout improvements, CSS polishments

### DIFF
--- a/locales/de/LC_MESSAGES/volto.po
+++ b/locales/de/LC_MESSAGES/volto.po
@@ -2407,11 +2407,6 @@ msgid "Sort on"
 msgstr "Sortieren nach"
 
 #: components/manage/Blocks/Search/schema
-# defaultMessage: Sort on label
-msgid "Sort on label"
-msgstr "Label für Sortieren"
-
-#: components/manage/Blocks/Search/schema
 # defaultMessage: Sort on options
 msgid "Sort on options"
 msgstr "Sortieroptionen"
--- a/src/components/manage/Blocks/Search/components/SearchDetails.jsx
+++ b/src/components/manage/Blocks/Search/components/SearchDetails.jsx
@@ -17,7 +17,11 @@ const SearchDetails = ({ total, text, as = 'h4' }) => {
   const intl = useIntl();
   return (
     <El className="search-details">
-      {text ? `${intl.formatMessage(messages.searchedFor)}: ${text}. ` : ''}
+      {text && (
+        <>
+          {intl.formatMessage(messages.searchedFor)}: <em>{text}</em>.{' '}
+        </>
+      )}
       {intl.formatMessage(messages.searchResults)}: {total}
     </El>
   );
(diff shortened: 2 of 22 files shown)


src/components/manage/Blocks/Search/components/SearchDetails.jsx
```diff
@@ -17,7 +17,11 @@ const SearchDetails = ({ total, text, as = 'h4' }) => {
   const intl = useIntl();
   return (
     <El className="search-details">
-      {text ? `${intl.formatMessage(messages.searchedFor)}: ${text}. ` : ''}
+      {text && (
+        <>
+          {intl.formatMessage(messages.searchedFor)}: <em>{text}</em>.{' '}
```
tiberiuichim, line 22: I think `text` position should not be hardcoded. Some language could translate it as "*blabla* was searched". Check the syntax for intl.formatMessage.

tiberiuichim (changes requested):

src/components/manage/Blocks/Search/components/SearchDetails.jsx
kreafox: @tiberiuichim Yes, I agree. It's fixed.

tiberiuichim (approved):

src/components/manage/Blocks/Search/layout/LeftColumnFacets.jsx
```diff
@@ -96,18 +96,16 @@ const LeftColumnFacets = (props) => {
             </div>
           )}
 
-          <div>
```
sneridagh, line 99: I think this div was needed for the flex layout to work well. Can we really remove it?

kreafox: @sneridagh Yes, we can safely remove that div.

locales/de/LC_MESSAGES/volto.po
```diff
@@ -2197,9 +2197,9 @@ msgid "Search users…"
 msgstr "Nach Nutzern suchen…"
 
 #: components/manage/Blocks/Search/components/SearchDetails
-# defaultMessage: Searched for
-msgid "Searched for"
-msgstr "Suche nach"
+# defaultMessage: Searched for: <em>{searchedtext}</em>.
```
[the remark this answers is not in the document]
tiberiuichim: I think it will be fine, the message is formatted like this: https://github.com/plone/volto/blob/e1f789a6e66dddbf554955b9de90686d0028dee1/src/components/manage/Blocks/Search/components/SearchDetails.jsx#L19-L26

tiberiuichim:
@sneridagh I think this should be good to go.

sneridagh:
@kreafox @tiberiuichim taking a look at the remaining open PRs... what's the status of this one? 
What I was afraid was to introduce too many changes that could be breaking, however, identifying them and adding them to the upgrade guide would be enough. Let's make this happen.

kreafox:
@sneridagh I've updated this PR, it's good to go. There is one thing though, currently the view switcher is called "Select view", but we both agreed with @tiberiuichim that "Display" is more appropriate. See my screenshot. Let me know what you think.

view-switcher

sneridagh (approved):

PR #2832, opened 2021-11-14
Remove python 2.7 from readme

fredvd:
& bump to python3.8.  We don't mention node LTS 12 or 14 here either, even though it's technically still possible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,7 +57,7 @@ First get all the requirements installed on your system.
 ### Prerequisites
 
 - [Node.js LTS (16.x)](https://nodejs.org/)
-- [Python 3.7.x / 2.7.x](https://python.org/) or
+- [Python 3.8.x](https://python.org/) or
 - [Docker](https://www.docker.com/get-started) (if using the Plone/Guillotina docker images)
 
 ### Create a Volto project using the generator


tisto:
@fredvd Volto 14 works fine on Plone 5.2 and Python 2.7 so far. Question here if we list minimal dependencies or the current best practice. Maybe we should say: Python 3.8 and Node 14 (16 for Volto 14) is recommended but we support down to Python 2.7 and Node 12.

fredvd:
Back from vaction and cathing up with saved up work. 

> question here if we list minimal dependencies or the current best practice.

The question is also which target groups interpret the information in which way. This README is kind of 'developer' marketing. 

I think Python2.7 has been bashed/end-of-lifed enough by the Python Community at large to have a negative connotation now.  So I wouldn't mention it near the top of the document. List  best practice / currently active maintained Python versions.  

Further down the same main README is a section on suported Node versions. That section also describe the the support for previous Python versions there where there is room to mention the caveats.

sneridagh:
It's fine for me if we mention the recommended, but somewhere (in the docs) we should mention all the range and combination possibles (even Plone 4! although in a limited mode)

sneridagh (approved):

PR #2833, opened 2021-11-14
Inject block data defaults from BlocksForm and RenderBlocks

tiberiuichim:


tiberiuichim:
@sneridagh The problem with the current implementation of the form defaults is that it overwrites data, not all of the defaults get applied in the block data. 

I see two solutions to this problem. The one presented here doesn't need any migration and has as bonus that it solidifies the concept of schema for blocks.

I'll try another PR with onChangeFormData implemented.

tiberiuichim:
had a discussion with Victor and Alin, we've settled on renaming existing `schema` to 'layoutSettingsSchema` and add the new `blockSchema` field.

sneridagh (approved):

sneridagh:
@tiberiuichim So we need:

- Documentation for the new config setting
- Another PR for changing the layoutSettingsSchema naming (and related docs and upgrade guide)

Question: Is it still required the inner `InlineForm` code for handling defaults or should we remove it?

tiberiuichim:
@sneridagh no, the InlineForm useEffect code is not needed anymore.

tiberiuichim:
@sneridagh  I think it's the perfect time to introduce some additional features:

- in the schema, you can set "primary:true" to a field...
- which would allow providing an automated default fallback form in the view (just like the Image blocks have now, as we're doing it in the Volto Addons tutorial)...
- which would allow us to completely eliminate the need for most custom Edit block components, maybe?

sneridagh:
+1000 to all!

tiberiuichim:
@sneridagh This one is ready (hopefully), the rest will come step by step.

PR #2834, opened 2021-11-14
Form defaults: onChangeFormData in InlineForm

tiberiuichim:
@sneridagh This is the other variant, with onChangeFormData in InlineForm. The problem with this approach is that there needs to be code updates everywhere (pass onChangeFormData). We can move that call into the BlockDataForm, but then we'd have to pass `{...props}` in all BlockDataForms.


tiberiuichim:
It was decided to go with #2833

Issue #2810, opened 2021-11-02
The Big List of Small Volto Rules


tiberiuichim:
I think we need a big list of small volto rules. Things that only need one line, just the rule. These are very much context dependent rules, maybe not known, obscure, whatever. Let's try to document them here, maybe? And then we can better organize our docs.

tiberiuichim:
A field in a client-side schema can have a `default` property and it would represent the initial value in the rendered schema.

tiberiuichim:
When you write a "blocks in a single block" type of thing, make sure to properly pass selected as false (or true) to the inner blocks to avoid focusing problems. You'll probably also want to set `blockHasOwnFocusManagement` to true.

tiberiuichim:
Volto's Pastanaga theme doesn't include all SemanticUI icons because it loads the fonts from the "basic" semantic-ui theme, not the "default" semantic-ui theme. You have to copy those into your theme.

tiberiuichim:
If you see a console warning from a component and you don't know which is that component, use the React Developer Tools browser extension, the Components tab, as it has quick links at the top to the components with warnings.

tiberiuichim:
Every time you add a new customized file you have to restart Volto.

tiberiuichim:
`getContent(getBaseUrl(item['@id']))` won't be ok with Volto 14 ++api++ calls. It needs to be `getContent(flattenToAppURL(getBaseUrl(item['@id'])))`

tiberiuichim:
Q: Can I use another component library with Volto? Yes! React integration is pretty straight forward, the CSS/styling part is probably less straightforward, but there's no issues there. You could be thinking of using a completely separate set of components + styling for the public site (and your blocks), while Volto uses the semantic-ui-react component for the CMS part (editing interface, block edit wrappers, sidebars, etc).

tiberiuichim:
The `key` param to a React component can be used even outside `somelist.map()` iterations, as a "cache busting" , in case the particular component needs to be recreated, if it keeps some internal state and refuses to respect outside props changing. For example:

```
<Dropdown defaultValue={x} options={y} />
```

If `x` changes, the semantic-ui Dropdown won't acknowledge it, so you can force the recreation of the dropdown with:

```
<Dropdown key={x} defaultValue={x} options={y} />
```

sneridagh:
[...]

PR #2836, opened 2021-11-15
Fix import path in App.jsx

jimbiscuit:
Fix Import to fix component shadowing App.jsx


nileshgulia1 (approved):

nzambello (approved):
LGTM

tiberiuichim:
The thing is, I'm baffled that this is needed. We use a special webpack resolver plugin that "rewrites" all the imports, so this sort of thing shouldn't be needed.

@jimbiscuit Can I have a confirmation that you've tried to shadow those files and they didn't work? Maybe, if you have time, you can try poking around https://github.com/plone/volto/blob/master/webpack-relative-resolver.js to see if you can get some info on what could be wrong?

tiberiuichim:
Or maybe it's just a problem of "if I shadow App.jsx, then I'll have to rewrite those imports in the shadowed file" ?

nileshgulia1:
I'm +1 on using absolute imports at least in the component level in all volto-core components. @tiberiuichim This can also go into one of https://github.com/plone/volto/issues/2810 :)

tiberiuichim:
@nileshgulia1 I'm conflicted on this one. Personally, as a developer writing Volto code I like being able to reference components and modules with relative paths. It's quicker to write the imports (they're usually shorter), avoids having to add them to the big `components/index.js` index, etc. 

Now, one year ago, Volto code that was written with this style of imports was impossible to customize. For example, if App.jsx imports '../SomeComponent', the `SomeComponent.jsx` file would be impossible to shadow without also shadowing App.jsx. So I wrote the webpack-relative-resolver plugin that fixes the problem, so now we are free to use whatever the code style we want, as we're guaranteed that shadowing will work.

But there's still a problem, though it's an easy fix: when I copy a Volto file in my project, to shadow, if the file contains code with relative imports, I'll have to tweak my copy to fix those imports. Indeed, not the best.

tiberiuichim (approved):

PR #2840, opened 2021-11-16
Fix isInternalUrl when settings.internalApiPath is empty string

tiberiuichim:
For some reason my internalApiPath appears as empty string and it trips this edge case.


nzambello (approved):
LGTM

sneridagh (approved):

PR #2842, opened 2021-11-18
Use @package instead of ~ tilda imports for root

tiberiuichim:


nzambello:
Wasn't it different because in a project `~` would fallback to Volto if not in the project and `@package` alias would not?
If I remember wrong or it isn't anything actual today, then it's completely fine to me

tiberiuichim:
@nzambello no, the `~` and `@package` are the same thing, basically. https://github.com/plone/volto/blob/3fde00f1322e22c449ff269b853bc3ea70fe5486/razzle.config.js#L212

nzambello (approved):
All right then, LGTM.

nileshgulia1 (approved):

sneridagh:
As a plus, we could remove the babel plugin for root, right? As a con, the naming gets a bit confusing :/ (mainly because of the duality). Maybe in the future, we should call it `@root` ? :)

tiberiuichim:
@sneridagh Yes, the root plugin wouldn't be needed anymore. Maybe we need a formal definition of Volto terms. At EEA we call the Volto-derived projects as [frontend](https://github.com/eea/?q=frontend&type=&language=&sort=), but I think `@project` could also be a good match in the internal imports.

Actually, looking at the documentation, the "Volto as project" is the default. "Installing Volto" generates a Volto project, so perhaps running standalone Volto is the "odd one".

tiberiuichim:
I like `@root`. It's both explicit and correct for both scenarios.

sneridagh:
@tiberiuichim could you please check https://github.com/plone/volto/pull/2842/commits/d58eeb4a9630ce7b05587b4e4f519f56fc8508f8 ? Thanks!

tiberiuichim:
LGTM

PR #2845, opened 2021-11-20
Add some widget stories for storybook

tiberiuichim:


tiberiuichim:
@sneridagh I've added some more stories. We could avoid the ugly tree navigation Storybook generates if we rename each named export from Default to whatever the navtitle is defined in the story. For example, for the email widget story, if we rename Default to Email, it would generate a single Email entry in the nav, instead of having `Email > Default` as nav tree. I thought to leave it like it is, with the Default, to encourage adding additional stories.

The Query widget story doesn't work, I'll leave it to be fixed another time. The array and select widgets don't work, but I think you're covering them in another PR.

sneridagh:
The selects are fixed in Giulia's PR. Indeed if a story just have one, we could go with a single item in the tree. Specially if the do not bring anything valuable (like the connected, renderer, etc)

sneridagh (approved):
😱😱😱 Thanks a lot @tiberiuichim, Way to go!!

sneridagh:
I'll merge this and port the fixes from the selects PR.

Issue #2865, opened 2021-11-26
[Safari] SyntaxError: Invalid regular expression: invalid group specifier name


nzambello:
**Describe the bug**
I just upgraded a Volto project from 14 alpha 9 to latest alpha 36.
With any browser it's working correctly, but on Safari I get:
```SyntaxError: Invalid regular expression: invalid group specifier name safari```
and no JS is loaded, so I'm getting the page without JS (hopefully we have SSR and critical CSS)

The issue is in a main chunk and seems to be a Safari issue on regular expressions.

<img width="500" alt="Schermata 2021-11-26 alle 11 52 13" src="https://user-images.githubusercontent.com/21101435/143569700-dbaf7f86-1931-456b-8ce4-efaf924e4982.png">
<img width="362" alt="Schermata 2021-11-26 alle 11 23 54" src="https://user-images.githubusercontent.com/21101435/143569676-26221ef4-ffa2-46aa-a8f6-4c8401aa7aa7.png">

https://stackoverflow.com/questions/51568821/works-in-chrome-but-breaks-in-safari-invalid-regular-expression-invalid-group
https://caniuse.com/js-regexp-lookbehind

**Software (please complete the following information):**
 - OS: iOS 15, macOS 12
 - Browser: Safari

**Additional context**
With other browsers it works fine

nzambello:
@plone/volto-team do anybody remember any particular regex in the last months?
I'm looking at the diff, it's huge but I'm not finding anything suspect.

nzambello:
To reproduce: Running volto itself on [master](7a0ffcfcf55a4fc4bcd621a4caec676e8c50540d) is happening locally, too.
Just load it in Safari

nzambello:
It's working in alpha 35 and the problem seems to be:
https://github.com/plone/volto/blob/7a0ffcfcf55a4fc4bcd621a4caec676e8c50540d/src/actions/vocabularies/vocabularies.js#L21

PR #2866, opened 2021-11-26
fix: retrieve vocabulary name in action without regex look behind

nzambello:
Fixes: #2865 

Changes #2862 with a different solution avoiding regex look behind since it's not supported by Safari


tiberiuichim (approved):

nzambello:
I confirm it's working in production now

sneridagh (approved):

PR #2875, opened 2021-11-29
fix: timewidget positioning

nzambello:
...don't ask why 84px.
To have the same behavior on 12 and 24 hrs, if in the current case it's positioned on 24px from right. If it's in "narrow" mode (so 24h) I need 60px more to have the picker dropdown to be aligned correctly.

### DIFF
--- a/theme/themes/pastanaga/extras/time-picker-overrides.less
+++ b/theme/themes/pastanaga/extras/time-picker-overrides.less
@@ -35,7 +35,7 @@
 
     .rc-time-picker-panel-inner {
       width: @timePickerWidthNarrow;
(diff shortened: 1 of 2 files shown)


theme/themes/pastanaga/extras/time-picker-overrides.less
```diff
@@ -35,6 +35,7 @@
 
     .rc-time-picker-panel-inner {
       width: @timePickerWidthNarrow;
+      right: 74px !important;
```
nzambello, line 38:
```suggestion
      right: 84px;
```

sneridagh (approved):

PR #2884, opened 2021-12-01
Fixed drag-and-drop list placeholder issues

reebalazs:
- Fix placeholder position to work with subtree margins.
  The original calculation assumes that there are no margins. The update
  works for all cases included when the subtree contains margins that
  extend to the node's parent.
- Remove placeholder when dropping in all cases.
- Make placeholder disappear when the element is moved outside the
  parent (thus, causes the place to be removed).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 - In search block, read SearchableText search param, to use it as search text input
   @tiberiuichim
+- Fixed drag-and-drop list placeholder issues @reebalazs
 
 ### Internal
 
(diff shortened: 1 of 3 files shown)


tiberiuichim:
@reebalazs Awesome work! 

Just in case you're not familiar with jest, you have to update the jest snapshots for the tests to pass. Run `yarn test` and hit `u` to update the snapshots (after you check the diff, of course).

reebalazs:
@tiberiuichim thank you, snapshot fixed :)

sneridagh (approved):

sneridagh (approved):

PR #2886, opened 2021-12-03
Update to plone.restapi 8.16.2 (revert missing_value PR)

sneridagh:


sneridagh:
@pbauer @ksuess I rethought this a bit, and setup all the boilerplate for this (testing, release, CI, etc) was too much. Since we already agreed that it would be great to have it on core, I transferred it already to `plone.volto`:

https://github.com/plone/plone.volto/pull/30

So we can release it from there and use it right away. Right now it's an optional profile and ZCML. I added a test for p.restapi `@types` schema rendering too. Volto side is coming :)

PR #2888, opened 2021-12-04
Add PreviewImage component

tiberiuichim:
There's a bunch of this logic that gets repeated in listings. This component makes it easier to reuse that logic.

There's a 'cover' option, which sets the preview image as background image. I thought it would be interesting to have it, there's listing scenarios where it could be useful.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ### Feature
 
+- Added a new component, PreviewImage. It renders a preview image for a catalog brain (based on the `image_field` prop). @tiberiuichim
+
 ### Bugfix
 
 ### Internal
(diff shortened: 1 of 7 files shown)


nzambello:
I think I like it, I'll have a deeper look into it

sneridagh (commented):
Didn't fit-object support make the background approach obsolete?

tiberiuichim:
@sneridagh right. No need for old school tricks. I'll remove that option

sneridagh:
@tiberiuichim @nzambello how this one relates to the uber Image component?

nileshgulia1:
Not related to this PR: I sometimes find the preview_image not shown when used in the summary template of the listing block without any criteria set. Dunno if it's a real bug, but then the listing block without criteria should not be supposed to save!

tiberiuichim:
@nileshgulia1 That's this ticket: https://github.com/plone/volto/issues/2891

The listing block without criteria works in "folder contents mode", listing the children of the current folder.

nileshgulia1:
This PR should be brought up to date.

PR #2892, opened 2021-12-06
The collective.folderishtypes package will not be installed by default

ale-rt:
Checks out https://github.com/plone/plone.volto/pull/29

### DIFF
--- a/.github/workflows/core.yml
+++ b/.github/workflows/core.yml
@@ -28,6 +28,30 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-yarn-
 
+      # python setup (temporary, while p.a.iterate changes are in a PR)
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v1
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      # python cache
+      - uses: actions/cache@v1
+        with:
+          path: ~/.cache/pip
+          key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements.txt') }}
+          restore-keys: |
+            ${{ runner.os }}-pip-
+      # python install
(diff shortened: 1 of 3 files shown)


sneridagh:
@ale-rt the builds are using docker with a fixed set of releases, so it's not enough by doing this. You could trigger it anyways and use the local robot server setup if modifying the Makefile. Let me do it for you.

Anyways, this test won't give us anywhere, since Volto assumes you are always in a all-folderish environment, and we want it to be like this. 

Do you have experience with class swapping migration scripts, or the right and accepted way to do it? I did it once, but I had to search for it. I might work on this today.

ale-rt:
I do not think class swapping by fiddling with `__class__` trick is of any use here.

It seems to me that the situation here is more complicated.

I will disregard the the fact that you need to fix the FTIs (that seems the easiest part).

I am more worried about the migration of objects that have content inside them.

Think about this starting point:
1. You have a folderish *News Item* 1/Plone/news/breaking`
2. `/Plone/news/breaking` contains some images

I figure (but stop me if I am wrong) that the state you want to reach using the current default Plone types is:
1. You have a *Folder* `/Plone/news/breaking`
2. "/Plone/news/breaking" contains the same contents that were first in the news **+** a newly created non folderish *News Item* (let's say `/Plone/news/breaking/main`)
3. `/Plone/news/breaking/main` is the default view for `/Plone/news/breaking`

Swapping the class to the one currently used by the "regular" (AKA non folderish) news type might cause really nasty effects.

Also I will avoid adding to the discussion the possibility of objects that are the default view of a container and they themselves have an object as a default view.

sneridagh:
superseded by https://github.com/plone/plone.volto/pull/36, plone.volto 4.0.0a already is c.folderishtypes free

PR #2898, opened 2021-12-07
Listing block fullobjects option per variation

### LINKED ISSUE #2899: listing block: enable fullobjects per block variation to fetch items blocks
I want a listing block variation with all blocks of listing items. And I do not want all listing block variations to do the same and fetch fullobjects.

ksuess:
fix #2899

### DIFF
--- a/cypress/tests/core-sandbox/listingblock.js
+++ b/cypress/tests/core-sandbox/listingblock.js
@@ -1,5 +1,5 @@
-context('Special fields Acceptance Tests', () => {
-  describe('ObjectListWidget', () => {
+context('Listing block tests', () => {
+  describe('Listing block variation', () => {
     beforeEach(() => {
       // given a logged in editor and a page in edit mode
       cy.visit('/');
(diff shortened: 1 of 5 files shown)


cypress/tests/core-sandbox/listingblock.js
```diff
@@ -0,0 +1,129 @@
...
+
+      cy.get('#toolbar-save').click();
+
+      cy.waitForResourceToLoad('@navigation');
+      cy.waitForResourceToLoad('@breadcrumbs');
+      cy.waitForResourceToLoad('@actions');
+      cy.waitForResourceToLoad('@types');
+
+      // Test after save
+      // Text of listing block is visible
```
ksuess, line 123: // Text of listing block is not visible

tiberiuichim:
@ksuess Looks good! I've been debating with myself if we should pass the variation and make the withQuerystringResults completely a "block thing", or just pass the boolean prop from the block view... but we're already using the `block` field from the data... so we're already assuming that. In any case, if somebody wants to use withQuerystringResults for something else other then a block, they can construct a `variation` object to fit it.

tiberiuichim (approved):

sneridagh (approved):

wesleybl:
@tiberiuichim @ksuess I don't see any variation using fullobjects. Can't we remove this option since Volto is in alpha?

PR #2901, opened 2021-12-09
Various local development build improvements

rpatterson:
I'm baaack!  And as usual capturing little build fixes as I encounter them while refreshing my local development environment.  Should be pretty much no risk of regression in the actual software and little to no risk of impact on other developers, CI/CD or other tooling.

See the individual commit messages for detail and elaboration on the individual changes.


rpatterson:
Regarding the failing `Changelog verifier` check, I don't see a `./news/` directory anywhere in this repo.  What am I missing here?  At any rate, none of these changes should have a changelog entry to my mind.

tiberiuichim (approved):

tiberiuichim:
@rpatterson Volto uses the CHANGELOG.md file in the repo root. Manually fill it in, in the unreleased section at the top.

rpatterson:
> Volto uses the CHANGELOG.md file in the repo root

Hmm, [the details link in the GitHub UI for that
check](https://jenkins.plone.org/roboto/missing-changelog) is the standard Plone message
about [towncrier](https://pip.pypa.io/en/latest/development/contributing/#news-entries).
In the interests of having both code quality checks and reducing hurdles to
contributions it would be good to correct that message.

In the meantime, I'll add an entry as described.

rpatterson:
> it would be good to correct that message.

I don't see anything in `./.github/workflows/` or elsewhere in the repo that defines or
configures that check.  Where is that defined?

rpatterson:
Can you refresh my memory, @tiberiuichim, am I as the contributor supposed to merge my own PRs once approved and all checks are green, or do I leave that to others?

tiberiuichim:
@rpatterson We usually leave the git merges up to the maintainer, @sneridagh. Sometimes (and I'm very guilty of this) we use the review as a method of saying "I'm ok with this going into the master, as long as all checks are passed", as we don't have other internal PLIP approval mechanism.

sneridagh:
@rpatterson Glad you're back!

> > it would be good to correct that message.
> 
> I don't see anything in `./.github/workflows/` or elsewhere in the repo that defines or configures that check. Where is that defined?

That is check "from above" by Mr.Roboto. The nature of the package (not being Python) and the difference of the ecosystem makes Volto has its own things different than the rest of the Plone core packages.

The stablished workflow is also a bit different, as @tiberiuichim pointed out, due to the nature of the project. That needs more "love" and in-advance careful selection of what goes in and when. This is Plone 6 UI, remember, not plone.app.discussion and a lot of things are at stake if we introduce something not expected or carefully tested that could potentially break people's project.

We are also in the verge of releasing Volto 14, and the only accepted things "in" are the ones already approved by the Volto Team: https://github.com/plone/volto/milestone/19

Other things can go in also, but not disruptive, and blessed by the Volto Team. I changed the build a bit in https://github.com/plone/volto/pull/2805 , so we must check that nothing is breaking there also. 

BTW, in which subject are you planning to work?

rpatterson:
> a lot of things are at stake if we introduce something not expected or carefully tested that could potentially break people's project

Ah, makes sense.  Perhaps then this project should switch to `towncrier` and adding entries
in `./news/*` files.  In the meantime I'll submit another PR adding this detail to the
contributor documentation.

> That is check "from above" by Mr.Roboto

For my edification, can you clarify where/how this is hooked up to the GitHub repo?

> BTW, in which subject are you planning to work?

I'm resuming work on the authentication issues between classic Plone, Volto, and ZMI.
From there, I'll be working on what @alecghica tells me to.  :-)

rpatterson:
AFAIK, this is ready for merge.  LMK if I'm missing something and more attention is needed.

api/buildout.cfg
```diff
@@ -77,7 +88,7 @@ plone.rest = 2.0.0a1
 
 plone.volto =
 robotframework =
-robotframework-debuglibrary =
+robotframework-debuglibrary = <2.2
```
sneridagh, line 91: Could you elaborate this? Did you find any problem?

rpatterson: A version incompatibility pops up when adding the `[test]` extra requirements for those three mr.developer checkouts.  I tried to reproduce it to get you the specific details but now I'm getting a different version conflict.  I'm digging into that, but in the meantime I took the version of `robotframework-debuglibrary` that was conflicting and decremented it's minor version and that worked.

rpatterson: See the commit messages of updated version pins from the `master` merge and subsequent commits I just pushed for more details.

api/buildout.cfg
```diff
@@ -72,14 +83,15 @@ eggs = ${instance:eggs}
 [versions]
 setuptools =
 zc.buildout =
-plone.restapi = 8.16.2 # force to latest p.restapi
-plone.rest = 2.0.0a1 # new JSON only traversal
+plone.restapi = >=8.13.0 # force to latest p.restapi
```
tisto, line 86: @rpatterson I'd prefer exact version pins here. CI would fail without notice when something changes in newer versions that we fetch.

rpatterson: All the rest of the version overrides in that buildout configuration are explicitly *clearing* version pins.  There are 14 version overrides there, only 2 of them were pins before my changes and those 2 are probably out of date.  So the CI failure risk you talk about is already there in *most* of those version overrides and unrelated to the changes in this PR, no?

IMO, it's important to accomplish *both* goals in one way or another:
1. Version constraints that define the *requirements* (E.g, we need the feature in this library that was introduced in v1.1 -> `foo.bar >= 1.1`)
2. Version pins for reproducability

I'd be happy to submit a separate PR to deal with this separate issue if that's what others want but I don't think it should be stacked on top of this PR.

rpatterson: [Submitted a separate PR](https://github.com/plone/volto/pull/2922#issue-804005730)

rpatterson:
Can someone merge this so I can use these improvements as I work on actual application code?

sneridagh:
@rpatterson We are in the verge of releasing Volto 14. We haven't yet been able to do so because we are busy with communication and release preparations. @tisto stated that he wants to take a closer look to both PRs. BTW, I have the same opinion about version pins in here.

The CI does work with docker images right now, not using this buildout, which is only for reference and development purposes, so we are quite safe. Also, although a useful addition, the test setup is not that important for the project.

Could you merge your changes into your PRs in the meanwhile?

sneridagh:
@rpatterson https://github.com/plone/volto/pull/2922 supersedes this one?

rpatterson:
> @rpatterson #2922 supersedes this one?

It builds on this one, yes.

Issue #2902, opened 2021-12-10
Missing layoutView for document_view


MarcoCouto:
**Describe the bug**
The default config object doesn't have a view for document_view in `views.layoutViews`. Volto will use the default view defined in `views.defaultView` in this scenario, but if we try to create a new layout view for a page, and later use that view in a certain page, it will no longer be possible to go back and use the document_view.

One possible solution would be to create a new entry in `views.layoutViews` for the document_view, where it's value is the defined default view:

```
config.views = {
    ...config.views,
    config.layoutViews: {
      ...config.views.layoutViews, 
      document_view: config.views.DefaultView,
    }
}
```

**To Reproduce**
1. Create a new view component (`WidepageView.jsx`, for instance).
2. Register the new layout view in config.js.
```
config.views = {
    ...config.views,
    layoutViews: {
      ...config.views.layoutViews,
      widepage_view: WidepageView,
    },
  };
```
3. Add the new layout view as an available view method in the ZMI.
![image](https://user-images.githubusercontent.com/2512508/145547824-315f5f4d-3226-4a35-bb33-a48f57e47340.png)
4. Use the new layout view in a page:
![image](https://user-images.githubusercontent.com/2512508/145548487-765bacda-3f73-4bc5-b6c1-c2e1b7974065.png)
5. It's no longer possible to go back to the document_view layout:
![image](https://user-images.githubusercontent.com/2512508/145548617-d9078be0-4481-4456-ae16-2cced94371f8.png)

**Software (please complete the following information):**
 - Volto version: 14.0.0-alpha.37
 - Plone version: 6.0.0a1
 - Plone.volto addon version: 3.1.0a6
 - Plone REST API version: 8.15.2

tiberiuichim:
@MarcoCouto Hi Marco! Do you think you could do a PR on this? I think the fix you've proposed is fine.

MarcoCouto:
Sure! To which branch?

tiberiuichim:
@MarcoCouto make a separate branch based on the master. Then push that branch to this repo, then create a PR out of that branch. In case you haven't signed the Plone Contributer Agreement, you have to do that.  https://plone.org/foundation/contributors-agreement

PR #2903, opened 2021-12-10
Add a default view layout do document_view

MarcoCouto:
Fix to #2902


MarcoCouto:
Yeah, ok. I'll do that.

PR #2908, opened 2021-12-12
Add form undo toolbar

tiberiuichim:
It works. Needs tests ~~and to fix the draftjs block, it refuses to update when external state changes~~.


tiberiuichim:
Current status: there is a strange warning strictly related to the draftjs title block. Doesn't seem to affect things too much. Maybe could be fixed.

tiberiuichim:
@sneridagh the current select widgets are really bad, they're keeping internal state and not following props data update. But I didn't look at their code, maybe there's something obvious that could be fixed.

src/components/manage/Widgets/ArrayWidget.jsx
```diff
@@ -119,18 +137,33 @@ class ArrayWidget extends Component {
     super(props);
 
     this.handleChange = this.handleChange.bind(this);
+  }
+
+  getValue() {
```
sneridagh, line 142: watch out! it's not binded!

tiberiuichim: There's more wrong with that ArrayWidget... I'll fix it and then I think the PR is good for review.

tiberiuichim: Ah... I've asked Silviu for some Cypress tests for this... I didn't get them yet. We need to wait for them.

tiberiuichim:
@sneridagh I think we need another feature from the backend for the vocabularies, if we want to support paging async vocabs, and that is to retrieve the terms (token + label) from a set of tokens (the value).

Imagine that I have a large vocab and I'm getting paged results from any request. If I have a field with an array select filled in already, the widget should be able to lookup directly the label+tokens for that value.

tiberiuichim:
@sneridagh We still need more cypress tests on this, but it's ready to be reviewed. I'm not too sure about the `UndoToolbar` name... it's just a bunch of buttons. On the other hand, it matches the `BlockToolbar` name, which is the same (just buttons), of which I'm guilty as well.

tiberiuichim:
Continued in #2942

PR #2910, opened 2021-12-13
Look ma, no `momentjs`!

sneridagh:
Still in development, all the "easy" ones are done. The new FormattedI18nDate component does the magic.

@nzambello next are the "utility" functions in helpers and the Event/Recurrent ones. These are a bit more "tricky". Do we have proper tests for them? Specially the utility ones. If not, we could implement the tests, then the refactoring.

I also need someone to take over, kind of exhausted on this one, but I think it's worth the effort, the bundle size will thank us, any takers? :) @plone/volto-team 

https://caniuse.com/?search=Intl.DateTimeFormat

image


tiberiuichim:
We'll not be able to completely remove momentjs, as it's a dependency of our Datetime widget. But we can make it lazyload. https://github.com/airbnb/react-dates#localization

sneridagh:
@tiberiuichim yeah a pity indeed. But as you said, if the component that uses react-dates is lazy loaded, then we are fine.

sneridagh:
@tiberiuichim didn't Node 13 and above have full-ICU support by default?

https://github.com/nodejs/node/issues/36057

I assumed that the full-ICU package is not required. Or is it because we also support 12?

tiberiuichim:
@sneridagh I'm not sure... we can try to remove the dependency and see what happens.

Note that the PR is not ready... I think there's incomplete tests. Those changes you did in the Contents component, for example, need to be taken care of. I've had to adjust the snapshots for some of the tests, but that's because we're now testing "the proper thing". In those tests momentjs was mocked without actually need it. Now they're better, they test the real thing.

tiberiuichim:
@sneridagh I gave up on trying to remove momentjs. We can't get rid of it from the bundle. Ideally we wouldn't need it to display event date information, but in that area it gets tricky, as it uses rrule for recurrence and so on. Maybe this could be the subject of a sprint... but as you've noticed, it's tiring, more so since it's not really covered by unit tests.

So I've lazyloaded all occurances of momentjs, and for the date helper (which could be removed, except that there's a "format" param that uses momentjs formatting, but which is not used in Volto) we need to pass it an instance of momentjs, like `parseDateTime(local, value, format, moment)`.

sneridagh:
I'm more than fine with the current shape of this PR!

tiberiuichim (approved):

tiberiuichim:
@sneridagh We should think about merging this soon, otherwise it will bit rot

sneridagh:
@tiberiuichim below 14 it should fail :( I think it's from 13 that it's included by default

tiberiuichim:
@sneridagh Confirmed, it fails one node 12. Advice on how to proceed?

sneridagh:
12 eol is 2022-04-30. I guess we can live with it in the meanwhile... any downside on the bundle size?

tiberiuichim:
@sneridagh I don't think it would make a difference. It would bring v12 in line with the rest of versions. I've watched the bundle size for v14 and it went down (due to offloading momentjs).

PR #2915, opened 2021-12-13
Fix select widget infinite loop

reebalazs:
If an attempt is made to update a selection item that is not in the
choices, an infinite loop occurs. This can be caused by rare conditions
of a stale browser state.

This infinite loop has to be guarded by a condition in setState to
avoid triggering another component render and thus breaking the infinite
loop.


sneridagh:
@reebalazs I just merged the long running PR with the improvements of the selects. Could you please adapt it and test it with the current code?

Also, could be that the problem does not happen in the new code.

Thanks!

reebalazs:
I've checked and the new release does contain the condition to break up the infinite loop which this fix was meant to correct. So I close this PR. Thanks!

PR #2744, opened 2021-10-15
SelectWidget - added customOption and fixed usage of isMulti prop

giuliaghisini:
- Added customOption to select widget to render options in a different mode, for example:

Schermata 2021-10-15 alle 09 42 16

- Fixed usage of isMulti prop https://github.com/plone/volto/issues/2717

PR #2942, opened 2021-12-20
Form undo + Giulia's Select Custom Option + isMulti

tiberiuichim:
This PR merges #2744 and #2908 to make them easier to integrate. In addition I've added two new SelectWidget stories, for the custom option and isMulti.

@sneridagh You can merge @giuliaghisini's branch to the master, then I can merge again with the master, then we can merge this branch.


giuliaghisini (approved):

sneridagh:
@tiberiuichim @giuliaghisini green!

Anything else to do/pending?

sneridagh (approved):

src/components/manage/Widgets/story.jsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
```
sneridagh, line 1: @tiberiuichim only a last thing: Isn't it this name too generic? ;) Also, shouldn't the file name be Capitalised?

tiberiuichim: @sneridagh I thought to keep only the Widget modules upper-cased, otherwise people reading Volto's source code will open the story utils file when it's not actually interesting. The name could be changed, maybe storybook.jsx ? storyutils?

src/components/manage/Widgets/SelectAutoComplete.jsx
```diff
@@ -160,6 +159,9 @@ class SelectAutoComplete extends Component {
       this.props.id,
       selectedOption ? selectedOption.map((item) => item.value) : null,
     );
+    this.setState((state) => ({
+      termsPairsCache: [...state.termsPairsCache, ...selectedOption],
```
tiberiuichim, line 163: I think we need to make sure we push only unique terms in the termsPairCache

tiberiuichim: Nevermind, not relevant.

tiberiuichim:
Good to go. Maybe the naming.

PR #2947, opened 2021-12-22
typo paralel vs. parallel

mamico:
this is just a typo "paralel" vs "parallel" it seems to me that nothing changes, because it is the default behavior


nileshgulia1 (approved):

tiberiuichim (approved):

tisto:
@mamico good catch! Thanks for the fix!

sneridagh (approved):

Issue #2949, opened 2021-12-22
Deleting time entry leads to error.


iRohitSingh:
**Describe the bug**

**To Reproduce**
1. Go to event view
2. Click on time widget and clear the time
3. See error

**Expected behavior**
**Screenshots**
We should see the error

PR #2950, opened 2021-12-22
Add Failing Cypress test for datetimeWidget

iRohitSingh:
#2949  Add Failing Cypress test for datetimeWidget


tiberiuichim:
it passes?

PR #2963, opened 2021-12-30
Fix query data in listing blocks ssr async call

cekk:
Without this fix, all querystring-search calls made from the ssr were wrong because they pass an overstructure of data (we only need **querystring** object in queries).

Plus: with a wrong query, the endpoint always use an empty query, so the block will be rendered as "show me all current context's children"


tiberiuichim (approved):
Must have been overlooked during refactoring.

sneridagh (approved):

PR #2970, opened 2022-01-08
Enable `components` property in Volto config registry. Add resolve and register methods. Add `Component` component.

sneridagh:


sneridagh:
@tiberiuichim  Back at Quanta, we will need this :) What do you think?

tiberiuichim:
I've built something along those lines here: https://github.com/eea/searchlib/blob/standalone-split/packages/searchlib/src/registry.js#L80 and using it with a `<Component factoryName="something">` from here: https://github.com/eea/searchlib/blob/standalone-split/packages/searchlib/src/components/Component/Component.jsx

It's a pretty low-tech solution but I think it's fit only for "closed systems". For flexibility we'd need adaptors or dependency injection, and for that we need types, which leads to Typescript.

sneridagh:
@tiberiuichim Quite amazing! I had in mind something much more simple, like the kind of registry we have now for views,  blocks, widgets and following the same rules. True that we still need the "watcher" in case you override two things from two different places (say two different add-ons). But that was not really an issue in the last years.

I had in mind something like, how to completely replace all `Button` components, without shadowing it.

Yes, at some point we'll have to touch the code everywhere, but I think it's worthy and valuable, in order to support different component libraries other than Semantic.

tiberiuichim (approved):

Issue #2839, opened 2021-11-16
'Cannot set headers after they are sent to the client' when user logged and Volto try to user loginRenew


giuliaghisini:
When you are loggedin in Volto, and periodically Volto try to renew the login, an error 
'Cannot set headers after they are sent to the client'
is throw

tiberiuichim:
@giuliaghisini Do you have a traceback? It would help with the hunt for this bug.

giuliaghisini:
@tiberiuichim this is the traceback

```

(node:88518) UnhandledPromiseRejectionWarning: Error [ERR_HTTP_HEADERS_SENT]: Cannot set headers after they are sent to the client
    at ServerResponse.setHeader (_http_outgoing.js:485:11)
    at ServerResponse.header (/Users/giuliaghisini/progetti/io-visit/volto-iovisit/node_modules/express/lib/response.js:771:10)
    at ServerResponse.json (/Users/giuliaghisini/progetti/io-visit/volto-iovisit/node_modules/express/lib/response.js:264:10)
    at ServerResponse.send (/Users/giuliaghisini/progetti/io-visit/volto-iovisit/node_modules/express/lib/response.js:158:21)
    at /Users/giuliaghisini/progetti/io-visit/volto-iovisit/node_modules/volto-middleware-static/src/index.js:43:1
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
(node:88518) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 24)
(node:88518) UnhandledPromiseRejectionWarning: Error [ERR_HTTP_HEADERS_SENT]: Cannot set headers after they are sent to the client
    at ServerResponse.setHeader (_http_outgoing.js:485:11)
    at ServerResponse.header (/Users/giuliaghisini/progetti/io-visit/volto-iovisit/node_modules/express/lib/response.js:771:10)
    at ServerResponse.send (/Users/giuliaghisini/progetti/io-visit/volto-iovisit/node_modules/express/lib/response.js:170:12)
    at ServerResponse.json (/Users/giuliaghisini/progetti/io-visit/volto-iovisit/node_modules/express/lib/response.js:267:15)
    at ServerResponse.send (/Users/giuliaghisini/progetti/io-visit/volto-iovisit/node_modules/express/lib/response.js:158:21)
[...]

PR #2975, opened 2022-01-11
Server dying because of a bunch of 404 in proxied content calls (@@images, etc)

sneridagh:
This fixes https://github.com/plone/volto/issues/2839 and all related problems with the node server "dying" because a bunch of proxied @@images or other proxied content throw a 404.

The error was: "Cannot set headers once the content has being sent"

The solution seems to let the middleware to handle the error itself, as well as a better SuperAgent call handling, using then/catch too in `getAPIResourceWithAuth`.


sneridagh:
@tiberiuichim 
I tried this too, but it does not work... it definitely doesn't like the closure solution :( My hunch continues to be that the requests get messed up and the resp/res being used are not from the same thread.

(server.js L75)
```
server.use(function (err, req, res, next) {
  if (err) {
    const { errorHandler } = req.app.locals;
    errorHandler(err);
  }
});
```

giuliaghisini (approved):

PR #2987, opened 2022-01-18
Small structure fixes

sneridagh:


stevepiercy:
This needs to be backported to `plone6-docs` with a `git cherrypick`.

Please get into the habit of backporting the commit when it includes docs. If that is not possible, let's figure out another way to ensure doc changes get into Plone 6 Docs.

PR #2994, opened 2022-01-20
Add support for language independent fields

sneridagh:


giuliaghisini (approved):

tiberiuichim (approved):
Looks good! 

On a side note, we're using less and less of lodash, it seems. Let's try to push it out, or at least move to direct import of lodash submodules. I think the only one that we might need to keep is isEqual (and deepClone). For isEqual we may is fast-deep-equal

tiberiuichim (approved):

mtoepfl:
> isEqual (and deepClone).

FYI, I red about structuredClone shortly. It's nativly supported in latest browser versions and there are polyfills for older ones.

https://web.dev/structured-clone/

PR #3000, opened 2022-01-22
Improve docs on environment variables, add recipes

sneridagh:
@fredvd trying to improve a bit the situation with this. It would be great if we could have a feedback session with you, so we can know all the issues you've encountered with this. I'm aware it's not perfect yet, but trying to get there.

CORS is hard, and the concepts have to be quite in mind when dealing with it. So explaining all involving CORS concept and how we bypass/deal with it in Volto and document it properly is essential for Plone developers.

Maybe we have to improve too the console messages, which might mislead instead of help. Regarding naming, or change the intent of who does what is hard, if we don't want to face breaking changes. 

Let me know what's the outcome!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,8 +11,8 @@
 ### Internal
 
 - Lazyload react-beautiful-dnd @tiberiuichim
-
 - Lazyload react-dnd @tiberiuichim
+- Improve docs on environment variables, add recipes @sneridagh
 
 ## 14.4.0 (2022-01-21)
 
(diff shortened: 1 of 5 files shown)


tiberiuichim:
Maybe we need a section on CORS? 

Something like: CORS is you browser's mechanism to protect you against sending data to unauthorized websites. On any AJAX call, it will do a handshake with the server, asking that server: "my user is using your data with this domain name (http://example.com), are you ok with this? what are your domains that are allowed". Or something like that

sneridagh:
I'm merging it right now, just for having the docs in place. Agreed, CORS should have its own place in the docs. However, frankly I don't know where to place it in the current structure. I guess we need a "Glossary" or "Concepts" section, and link the docs there accordingly. I would leave it for the docs overhaul that is coming.

PR #3009, opened 2022-01-27
Fix ObjectListWidget story

tiberiuichim:
This is a bug caused by the lazyloading of the DND component

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@
 
 ### Internal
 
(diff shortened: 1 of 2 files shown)


tiberiuichim:
@sneridagh I think we should get rid of the simple Wrapper in storybook.js and just use the RealStoreWrapper everywhere, to avoid situations like this one.

sneridagh:
But then there are some other stories that break... I never could figure why

sneridagh (approved):

PR #3015, opened 2022-01-30
Optimize lodash in production bundle

tiberiuichim:


nileshgulia1 (commented):
@tiberiuichim `TypeError: memoize is not a function` comes from react-dnd. We might also want to upgrade it.

tiberiuichim:
Maybe the alias needs to go the other way, lodash-es -> lodash.

PR #3032, opened 2022-02-04
Add myst and sphinx config

stevepiercy:
# IMPORTANT

Merge https://github.com/plone/volto/pull/3035 before merging this PR!

----------------

I am now able to do `make docs-html` or `make docs-livehtml` to build the docs using Sphinx and MyST. w00t!

This is the first step toward rapid development of the docs.

The intent here is to allow parallel development of Volto documentation. Whenever a change is made on the main branch in docs, it should be backported to `plone6-docs` (not merged) with a separate PR for peer review.

We also need to ask @ericof to set up automatic CI/CD for `plone6-docs` under plone.org. Please suggest a good hostname.

PR #3035, opened 2022-02-05
Fix myst sphinx warnings

stevepiercy:
This PR uses the branch `add-myst-sphinx-config` to build the docs for testing. This PR should be merged into https://github.com/plone/volto/pull/3032 before the other is merged into [`plone6-docs`](https://github.com/plone/volto/tree/plone6-docs)


stevepiercy:
Quick syntax question for y'all. I'm working on fixing the `jsx` Sphinx highlighting, and the parser is super picky. This does not parse:

```jsx
<SidebarPortal selected={this.props.selected}>
  <InlineForm
    icon={<Icon size="24px" name={nameSVG} />}
    schema={schema}
    title={schema.title}
    headerActions={<button onClick={() => {}}>Action</button>}
    footer={<div>I'm footer</div>}
    onChangeField={(id, value) => {
      this.props.onChangeBlock(this.props.block, {
        ...this.props.data,
        [id]: value,
      });
    }}
    formData={this.props.data}
  />
</SidebarPortal>;
```

...but if I quote the HTML values, then it does parse:

```jsx
<SidebarPortal selected={this.props.selected}>
  <InlineForm
    icon={"<Icon size="24px" name={nameSVG} />"}
    schema={schema}
    title={schema.title}
    headerActions={"<button onClick={() => {}}>Action</button>"}
    footer={"<div>I'm footer</div>"}
    onChangeField={(id, value) => {
      this.props.onChangeBlock(this.props.block, {
        ...this.props.data,
        [id]: value,
      });
    }}
    formData={this.props.data}
  />
</SidebarPortal>;
```

My question is whether the latter form is valid syntax when actually used as code? It would be great if someone could run a quick test for me, perhaps a simple `return <h1>hello</h1>` and `return "<h1>hello</h1>"` to compare?

To go along with this, we need to make a choice about sytnax highlighting for jsx.

1. Quote HTML values
2. Leave as is, and ignore lack of syntax highlighting for many jsx code blocks
3. [Submit a PR to fix the parser](https://github.com/fcurella/jsx-lexer)

tiberiuichim:
@stevepiercy no, `something={"myfunc()"}` binds the `something` parameter to a string, it doesn't execute `myfunc()`. Maybe we can use simple js syntax highlighting? Would it work? IMHO we could even live without highlighting, if there's no other way.

stevepiercy:
I reverted those incorrect "fixes".

I opened an issue here https://github.com/plone/volto/issues/3038 and https://github.com/fcurella/jsx-lexer/issues/15

`js` as the lexer also fails.

We can do without syntax highlighting for now, but someone who understands the lexer and jsx syntax should submit a PR to the project.

stevepiercy:
It turns out that `jsx-lexer` does not like `'` within HTML tags. sheesh!

stevepiercy:
This PR is now ready for final review. Any takers?

sneridagh (approved):

PR #3041, opened 2022-02-06
Show addons installed in control panel

sneridagh:
image

This week I also missed a way to check the internal config using a SSR path... (everything else failed) however, that could have some security implications...


sneridagh:
@tiberiuichim regarding the versions, we could improve and pass the whole Addon Registry to the loader... Right now it's the minimum.

ksuess:
Please guide me. I ever wondered how to check a pull request with a project.

tiberiuichim:
@ksuess Here's my version, others may have a different way:

In you local clone of plone/volto clone, checkout the desired feature branch. Then run

```
yalc publish
```

Yalc is a global utility, similar to mrs-developer. You have to install it with:

```
npm install -g yalc
```

Then, in your Volto project, you can now run:

```
yalc add @plone/volto --no-pure 
```

This "installs" the "locally published copy of `@plone/volto`" in your project. It makes a hard copy of it and it changes package.json, so beware not to push that in your sourcecode repo. Now you can test Volto with your project. Once you're done, you can remove the local copy of volto with 

```
yalc retreat
```

Notice that the workflow is pretty cumbersome if you want to do development of Volto, as you have to always do the dance of publish+add on every change in your original Volto clone.

tiberiuichim:
@sneridagh Maybe read [here](https://github.com/plone/volto/blob/7b519862c82e85a530ca0e7b52c09880d7c8ea78/addon-registry.js#L164) and [here](https://github.com/plone/volto/blob/7b519862c82e85a530ca0e7b52c09880d7c8ea78/addon-registry.js#L198) the package.json of each addon. You can then pass those versions (or even the whole addon registry [here](https://github.com/plone/volto/blob/7b519862c82e85a530ca0e7b52c09880d7c8ea78/razzle.config.js#L189)

giuliaghisini:
👏👏👏

sneridagh:
@tiberiuichim Now I think it's satisfactory :) 

image

image

It even works for testing add-ons :)

image

sneridagh:
> Please guide me. I ever wondered how to check a pull request with a project.

@ksuess I use always @tiberiuichim approach with `yalc`. It has some drawbacks, since you can't auto reload and the "link" is not updatable... The "old" way used to work... that's true, but that's back when Nathan made it possible very long before Volto was a thing. Although with all the add-ons story in place, I don't know if the old approach is fully functional. I know you documented it some time ago. Do you know if that works still?

ksuess:
Thanks @tiberiuichim  and @sneridagh for the hint to yalc. I think it's important that everyone knows about this and how to test a project with branch. Will try yalc just after lunch. @tiberiuichim already added his elaborated guideline to the list of Volto rules, that's wonderful! This list should be spreaded…

ksuess:
BTW It's always saying "debug mode", no matter which mode.

sneridagh:
@ksuess I took a look to https://github.com/plone/volto/pull/2194 again and I just tested it out in my current project... and I have to say I can't see any glitch/downside! More testing would be required, for sure...

There's the points that @ichim-david pointed out... but we already have a list of wishes for mrs-developer, maybe this could be one. I'm +1 of even the downsides, let's say we point people to use `addons` folder (whatever) and add it to the docs too along with the `yalc`approach.

So, `yalc` for quick testing (less intrusive) and the mrs-developer one for more in-depth debugging, "improve upstream as you develop for a project".

sneridagh:
> BTW It's always saying "debug mode", no matter which mode.

@ksuess fixed!

tiberiuichim (approved):

locales/de/LC_MESSAGES/volto.po
```diff
@@ -1616,6 +1616,11 @@ msgstr "Nachricht"
 msgid "No"
 msgstr "Nein"
 
+#: components/manage/Controlpanels/VersionOverview
+# defaultMessage: No addons found
+msgid "No addons found"
+msgstr ""
```
ksuess, line 1622:
```suggestion
msgstr "Keine Add-ons installiert."
```

src/components/manage/Controlpanels/VersionOverview.jsx
```diff
@@ -6,37 +6,66 @@
...
-      <p>
-        <FormattedMessage
-          id="Warning Regarding debug mode"
-          defaultMessage="You are running in 'debug mode'. This mode is intended for sites that are under development. This allows many configuration changes to be immediately visible, but will make your site run more slowly. To turn off debug mode, stop the server, set 'debug-mode=off' in your buildout.cfg, re-run bin/buildout and then restart the server process."
-        />
-      </p>
+
+      <h3>Add-ons</h3>
+      {isEmpty(addonsInfo) && <p>{intl.formatMessage(messages.no_addons)}</p>}
+      <ul style={{ fontSize: '16px', fontFamily: 'Monospace' }}>
```
ksuess, line 56: Same list type as for versions: SemanticUI 'List'.

ksuess (changes requested):

src/components/manage/Controlpanels/VersionOverview.jsx
sneridagh: @ksuess thing is that at some point we need to drop this kind of usage of components that can be modelled with plain HTML... Specially thinking when we'll move to Quanta and agnostic use of CSS framework/design system. So I try to not over-use SemanticUI components.

ksuess: Then maybe the other way round, <ul> for both? Looks nicer, more harmonious.

sneridagh: Done!

ksuess: and now a quick view on a netlify pr preview…

ksuess (approved):

PR #3054, opened 2022-02-09
Fix the `null` error in SelectAutoComplete Widget

iFlameing:
@reebalazs I added this optional chaining because what happens props has the choices but its value is null. Then we don't get the array initialization.

### DIFF
--- a/src/components/manage/Widgets/SelectAutoComplete.jsx
+++ b/src/components/manage/Widgets/SelectAutoComplete.jsx
@@ -135,7 +135,7 @@ class SelectAutoComplete extends Component {
   }
 
   componentDidUpdate(prevProps, prevState) {
-    const { value, choices = [] } = this.props;
+    const { value, choices } = this.props;
     if (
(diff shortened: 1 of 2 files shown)


src/components/manage/Widgets/SelectAutoComplete.jsx
```diff
@@ -139,7 +139,7 @@ class SelectAutoComplete extends Component {
     if (
       this.state.termsPairsCache.length === 0 &&
       value?.length > 0 &&
-      choices.length > 0
+      choices?.length > 0
```
reebalazs, line 142: ok so if we want to allow null then please remove the default value  = [] as it's not necessary any more.

Also adding a test for the choices=null case would be nice... but might not be trivial, provided that currently there is only a single render test case.

iFlameing: done

reebalazs (approved):

PR #3055, opened 2022-02-09
Protect against ghost content loading and scroll to top

reebalazs:
This fix introduces two improvements on loading:

1. When we switch from a non-content view to a content view, prevent the 
   previously loaded content appearing as a "ghost".

   It works by resetting the content before the content for the new
   route is fetched, thus preventing from the previous content to appear
   as a ghost.

2. When we switch from any view to a content view, delay scrolling to
   the top to the moment when the content for the route has been
   fetched. 
  
   It works by storing a delayed location for ScrollToTop.jsx, and
   utilizing this delayed location to sense when the scrolling to the
   top should take place.


src/storeProtectLoadUtils.js
```diff
@@ -0,0 +1,155 @@
...
+      return next(action);
+    default:
+      return next(action);
+  }
+};
+
+// Note that there is a bit of heuristics here. We assume that every action
+// like this is beginning/ending an action. If this logic fails then the counting
+// will be off, resulting either a premature or an indefinitely missing state change.
+const isRequestAction = (action) => action.type.endsWith('_PENDING');
```
tiberiuichim, line 63: Test for `action.request` presence. The api middleware doesn't produce any SUCCES/FAIL/PENDING action for non-request actions. See https://github.com/plone/volto/blob/712e7a6bd207e0da2fa2e54f0952f11d4a78bfdf/src/middleware/api.js#L114-L116

tiberiuichim: Nevermind, I think the use case is different.

sneridagh:
@reebalazs green! Please add a changelog.

sneridagh:
@reebalazs could you please take a look at this and try resurrect this? Also add the latests improvements in the projects.

reebalazs:
Recently fixed an issue that we found. Based on the deployment history at the client project, this is now good to go.

tiberiuichim (approved):

sneridagh:
@reebalazs sync'd with master! hoping for green!

sneridagh:
@reebalazs seems that all is good now, but Plone 6 test backend complains hard. I'd like to take a closer look at this during the sprint.

Issue #3047, opened 2022-02-07
Image uploads in contents view broken (Safari)


iFlameing:
**Describe the bug**
Image upload in content view is broken

**To Reproduce**
1. Go to /contents
2. Click on Upload button
3. Select any image

**Expected behavior**

I am able to upload the image without any problem

**Screenshots**

https://user-images.githubusercontent.com/33936987/152829042-22fbbe3c-9d4f-47ed-a064-2e1c6fd10ebe.mp4

# My findings
Safari doesn't provide the lastModifiedDate properties (ContentsUploadModel)

Safari
<img width="1429" alt="Screenshot 2022-02-07 at 9 52 06 PM" src="https://user-images.githubusercontent.com/33936987/152829489-f1aaf324-6785-4af5-9315-0a047a95fc7f.png">

Chrome

![Screenshot 2022-02-07 at 9 51 17 PM](https://user-images.githubusercontent.com/33936987/152829705-d0b63ea7-4819-4989-822f-eb796608b16b.png)

tiberiuichim:
To be fixed upstream, right? :)

iFlameing:
@tiberiuichim I didn't understand what do you mean by upstream.

tiberiuichim:
@iFlameing I meant it should be fixed in Safari and Firefox. It was meant as a joke.

The problem, as I understand it, is that the old code that used momentjs accepted junk as an entry value. So the old code was buggy (it didn't display the reality), it just didn't crash. As a possible solution, some checks need to be added in that part of the code, so that, in case the property doesn't exist on the uploaded file, don't display anything.

PR #3058, opened 2022-02-10
Fix the upload image in contents view

iFlameing:
Fixes https://github.com/plone/volto/issues/3047

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Bugfix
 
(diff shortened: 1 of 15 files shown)


tiberiuichim (commented):
I don't think we should show anything in that cell, there. Instead of "Not found", just leave empty space.

iFlameing:
@tiberiuichim @sneridagh done.

tisto (approved):

Issue #3036, opened 2022-02-05
Docs: WARNING: document isn't included in any toctree


stevepiercy:
Here is a complete list of the Sphinx warning `WARNING: document isn't included in any toctree`.

- docs/source/cheatsheet.md
- docs/source/configuration/locking.md
- docs/source/configuration/volto-config-js.md
- docs/source/contributing/volips.md
- docs/source/developer-guidelines/design-principles.md
- docs/source/developer-guidelines/style-guide.md
- docs/source/developer-guidelines/typescript.md
- docs/source/ie11compat.md
- docs/source/index.md
- docs/source/introduction/introduction.md
- ~~docs/source/legacy/introduction.md~~
- ~~docs/source/legacy/manage.md~~
- ~~docs/source/legacy/mosaic.md~~
- ~~docs/source/legacy/theme.md~~
- docs/source/recipes/actions-reducers.md
- docs/source/recipes/environment-variables.md
- docs/source/recipes/fetch-data.md
- docs/source/releases.md
- docs/source/theming/integrate-third-party.md

I added these two in https://github.com/plone/volto/pull/3035 because they were moved around in another PR.

- docs/source/developer-guidelines/design-principles.md
- docs/source/developer-guidelines/style-guide.md

Please advise whether the remaining items should be included in the documentation, deleted, or kept around for some other reason.

See also https://github.com/plone/volto/issues/3034

stevepiercy:
I also merged `plone6index` into `index`, and made necessary adjustments in 795a458ed760e5a5f1cb1fef0b7d9a9820d6615b

stevepiercy:
docs/source/legacy/* was removed in https://github.com/plone/volto/pull/3076

stevepiercy:
We're down to just 11 files to decide whether they should be included in the documentation, be deleted, or just kept around and not displayed.

```
/cheatsheet.md
/configuration/locking.md
/configuration/volto-config-js.md
/contributing/volips.md
/ie11compat.md
/introduction/introduction.md
/recipes/actions-reducers.md
/recipes/environment-variables.md
/recipes/fetch-data.md
/releases.md
/theming/integrate-third-party.md
```

Please advise. This issue is blocking #3001.

sneridagh:
> My take on this:
> 
> * /cheatsheet.md - no?

+ 1 I always take a look at it... it's just handy to have it around.

> * /configuration/locking.md - yes

+1 

> * /configuration/volto-config-js.md - yes

+1

[...]

PR #3069, opened 2022-02-14
Fixed missing typescript docs link in navigation

pnicolli:
I forgot it when I added compatibility in #2795


stevepiercy:
This also needs to be backported to `plone6-docs` branch. See https://github.com/plone/volto/issues/3036

stevepiercy (approved):

pnicolli:
Thanks @stevepiercy for taking care of creating #3080. I'm really late with some of the work I would like to do here :(

Issue #3038, opened 2022-02-05
Improve jsx syntax highlighting in MyST/Sphinx/Pygments


stevepiercy:
I'm working on fixing the `jsx` Sphinx highlighting, and the parser is super picky. This does not parse:

```jsx
<SidebarPortal selected={this.props.selected}>
  <InlineForm
    icon={<Icon size="24px" name={nameSVG} />}
    schema={schema}
    title={schema.title}
    headerActions={<button onClick={() => {}}>Action</button>}
    footer={<div>I'm footer</div>}
    onChangeField={(id, value) => {
      this.props.onChangeBlock(this.props.block, {
        ...this.props.data,
        [id]: value,
      });
    }}
    formData={this.props.data}
  />
</SidebarPortal>;
```

...but if I quote the HTML values, then it does parse:

```jsx
<SidebarPortal selected={this.props.selected}>
  <InlineForm
    icon={"<Icon size="24px" name={nameSVG} />"}
    schema={schema}
    title={schema.title}
    headerActions={"<button onClick={() => {}}>Action</button>"}
    footer={"<div>I'm footer</div>"}
    onChangeField={(id, value) => {
      this.props.onChangeBlock(this.props.block, {
        ...this.props.data,
        [id]: value,
      });
    }}
    formData={this.props.data}
  />
</SidebarPortal>;
```

My question is whether the latter form is valid syntax when actually used as code? It would be great if someone could run a quick test for me, perhaps a simple `return <h1>hello</h1>` and `return "<h1>hello</h1>"` to compare?

To go along with this, we need to make a choice about sytnax highlighting for jsx.

1. Quote HTML values
2. Leave as is, and ignore lack of syntax highlighting for many jsx code blocks
3. [Submit a PR to fix the parser](https://github.com/fcurella/jsx-lexer)

_Originally posted by @stevepiercy in https://github.com/plone/volto/issues/3035#issuecomment-1030588839_

stevepiercy:
Here's another one. If I remove `aria-label="Cancel"`, then it gets lexed.

````
```jsx
<button className="cancel" aria-label="Cancel" onClick={() => this.onCancel()}>
  <Icon
    name={clearSVG}
    className="circled"
    size="30px"
    title={this.props.intl.formatMessage(messages.cancel)}
  />
</button>
```
````

stevepiercy:
@tiberiuichim thank you! Fixed in #3077.

sneridagh:
[...]

PR #3077, opened 2022-02-17
fix jsx syntax

stevepiercy:
Per https://github.com/plone/volto/issues/3038#issuecomment-1041465786


polyester (approved):

sneridagh (commented):
I think not... in fact I had problems in the past if used the CamelCase:

https://reactjs.org/docs/accessibility.html#wai-aria

Should be a bug in the lexer...

stevepiercy:
> I think not... in fact I had problems in the past if used the CamelCase:
> 
> https://reactjs.org/docs/accessibility.html#wai-aria
> 
> Should be a bug in the lexer...

@sneridagh thanks for catching that. We should go with the ReactJS docs. I will update that issue, and close this PR.

PR #3081, opened 2022-02-18
WIP: Searchblock daterange widget

robgietema:


src/components/manage/Blocks/Search/components/DateRangeFacet.jsx
```diff
@@ -0,0 +1,118 @@
...
+    defaultMessage: 'Start Date',
+  },
+  endDate: {
+    id: 'End Date',
+    defaultMessage: 'End Date',
+  },
+});
+
+const PrevIcon = () => (
+  <div
```
tiberiuichim, line 27: Just it case somebody wants to style these controls, it will be easier for them if this style is assigned to a class.

robgietema: Good call, I'll update it.

PR #3087, opened 2022-02-19
Make build dirs parameterizable via BUILD_DIR new runtime env var

sneridagh:
Since Volto is parameterizable on runtime, it's been a while since I wanted to do this.

So, on release, it will create a plain build and upload it to the repo in the `dist` folder.

The build supposed before this PR that it was going to be in `build` always. Now you can parameterize with: `BUILD_DIR` runtime env var. In case you want to run from somewhere else.

```
yarn
BUILD_DIR=dist node dist/server.js
```

or 

```
yarn
yarn start:dist
```

Unfortunately, the initial yarn install is required still... :(

/cc @avoinea @ericof I hope this eases even more the Docker image builds.


tiberiuichim (approved):

docs/source/configuration/environmentvariables.md
```diff
@@ -171,21 +171,22 @@ As a result, your app will load the add-ons in the following order:
 - `@eeacms/volto-accordion-block`
 - `volto-slate`
 
-!!! important
-    `ADDONS` does *not* work for development packages, which are *always* enabled if defined in your `jsconfig.json` or via `mrs.developer.json`.
+```{important}
+`ADDONS` does *not* work for development packages, which are *always* enabled if defined in your `jsconfig.json` or via `mrs.developer.json`.
```
tiberiuichim, line 175: I find this line misleading. ADDONS shouldn't care if an addon comes from node_modules or is aliased via jsconfig. 

I think the message is "The ADDONS key is a Volto specific configuration. Simply setting ADDONS doesn't download the javascript package. This has to be covered by another way, either installing the addon package (with `yarn add`) or loading it as a development package with mrs-developer.

sneridagh: I wanted to state clearly that a developer add-on is always enabled, if declared in mrs-developer and thus, in jsconfig.js. I'll try to merge both ideas.

sneridagh: but maybe that clarification has to be done elsewhere, not referred to ADDONS env var...

PR #3088, opened 2022-02-20
Clean up Netlify configuration

stevepiercy:
Remove runtime.txt and instead use global setting for netlify. See https://docs.netlify.com/configure-builds/environment-variables/


stevepiercy:
I'm going ahead and merging this PR without a review, as it has no effect on the rest of the project, and it still builds on Netlify.

PR #3099, opened 2022-02-21
Better implementation of the add-on load coming from the environment …

sneridagh:
…variable ADDONS


addon-registry.js
```diff
@@ -111,6 +111,7 @@ class AddonConfigurationRegistry {
     this.resultantMergedAddons = [
       ...(packageJson.addons || []),
       ...(this.voltoConfigJS.addons || []),
+      ...(process.env.ADDONS || []).split(','),
```
tiberiuichim, line 114: A bit funky here. `[].split()` will fail. Also, what's the addon separator? The comma is already used as a separator for an addon's "profiles", like `volto-slate:minimalDefault,tableButton`. The `ADDONS` key should also consider that it's supposed to include also the additional addon loaders, not just the single package name. I propose that we use the `;` as addons separators. So it would be like `ADDONS=volto-slate:minimalDefault,tableButton;@kitconcept/volto-grid-block`

tiberiuichim: Also, at some point we'll have to decide how to go about merging the "addon profiles". What happens to the additional addon profiles? Are they all loaded? Some of them could be incompatible with each other.

sneridagh:
Total success it seems :) 

@tiberiuichim You are correct, I'll take a look.

sneridagh:
@tiberiuichim I think I finally got the gist of it.

tiberiuichim (approved):

stevepiercy (changes requested):
Specify the separator as `;`

docs/source/configuration/environmentvariables.md
```diff
@@ -165,6 +165,13 @@ yarn add volto-slate
 ADDONS=volto-slate:asDefault yarn start
 ```
 
+If you need to specify several of them:
```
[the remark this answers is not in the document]
sneridagh: Fixed!

stevepiercy (approved):

PR #3100, opened 2022-02-24
Redirect querystring fix.

robgietema:


src/helpers/Url/Url.js
```diff
@@ -102,6 +102,15 @@ export function flattenToAppURL(url) {
...
   );
 }
+/**
+ * Given a URL it remove the querystring from the URL.
+ * @method stripQuerystring
+ * @param {string} url URL of the object
+ * @returns {string} URL without querystring
+ */
+export function stripQuerystring(url) {
+  return url.replace(/\?.*$/, '');
```
tiberiuichim, line 112: If I understand correctly, it strips all the query part of the URl. Using the regex feels dirty, though using `new URL()` and then reconstructing without the search part can be quite cumbersome.

robgietema: That is indeed what it does. We're using the same regexp method in other functions in the url helper already. I agree using new URL() might be a clean way but will also be slower. This is a quick and easy solution.

PR #3103, opened 2022-02-25
Refactor language syncronizer. Remove it from the React tree, and han…

sneridagh:
…dling it on the `getContent` action level as a Redux Thunk.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ### Feature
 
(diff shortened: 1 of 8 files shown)


sneridagh:
Meh, just realised that this forces the change of the signature of the `getContent` action, so we would have to update it everywhere... So not good...

sneridagh:
superseded by: https://github.com/plone/volto/pull/3103

Issue #3050, opened 2022-02-08
Search block variations restricted to internal use


ksuess:
Use case: Bulk management of content attributes  (site administrator)

The search block with a custom variation is used to provide an admistration view.
![image](https://user-images.githubusercontent.com/1144737/152996761-e93b4cf6-28fc-46b8-a795-10949e4d3402.png)

"Problem" is, that the variation now is also available for editors that are using search/listing blocks. It's not a serious problem, the content is shown only for authorized users.
But, it would be nice to be able to register variations for "internal use": as a variation only for an admin view. This admin view is something like this:

view, trallalala, …

```
      <SearchBlockView
        {...subscribersblock_props}
        path={getBaseUrl(subscribersblock_props.path)}
        mode="view"
      />
```

So one way would be to allow 

```
  config.blocks = {
    ...config.blocks,
    blocksConfig: {
      ...config.blocks.blocksConfig,
      listing: {
        ...config.blocks.blocksConfig.listing,
        variations: [],
        variations_internal: [
          ...config.blocks.blocksConfig.listing.variations,
          {
            id: 'subscriber',
            isDefault: true,
            title: 'Subscribers',
            template: ListingSubscriber,
            fullobjects: true,
          },
          {
            id: 'subsriptions',
            isDefault: true,
            title: 'Subscriptions',
            template: ListingSubscriptions,
            fullobjects: true,
          },
        ],
      },
    },
  };
```
With a change in 
https://github.com/plone/volto/blob/39b434627b7e2d65994492204083bd562dc15198/src/components/manage/Blocks/Search/SearchBlockView.jsx#L13-L14
pulling in the internal variations just for administrative views, but not for editors.

ksuess:
Obsolete: Immitate "SearchBlockView" with component "ListingBody" with appropriate props. And compose with "withQueryString" and "withSearch". No registration of config.blocks.blocksConfig.listing.variations needed.

```

const SearchAndManageView = (props) => {
  const { data, searchData, variation } = props;

  const Layout = variation.view;

  const listingBodyVariation = config.settings.breakingnews
    .search_and_manage_variations
    ? config.settings.breakingnews.search_and_manage_variations.find(
        ({ id }) => id === data.listingBodyTemplate,
      )
    : config.blocks.blocksConfig.listing.variations.find(
        ({ isDefault }) => isDefault,
      );
  const listingBodyData = searchData;

  return (
    <div className="block search">
      <Layout {...props} isEditMode={false}>
[...]

PR #3107, opened 2022-02-26
Allow search block variations to be restricted to internal use

ksuess:
Fix https://github.com/plone/volto/issues/3050

And it adds a first search block test!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ### Feature
 
+- Add search block variations source 'variations_internal' for internal use of search block, not exposed to editor. @ksuess
+
 ### Bugfix
 
 - Fix redirect bug with URLs containing querystrings @robgietema
--- a/cypress/tests/coresandbox/listingblock.js
+++ b/cypress/tests/coresandbox/listingblock.js
@@ -70,6 +70,7 @@ context('Listing block tests', () => {
       cy.waitForResourceToLoad('@breadcrumbs');
       cy.waitForResourceToLoad('@actions');
       cy.waitForResourceToLoad('@types');
+      cy.waitForResourceToLoad('document');
 
       // Test after save
       // Text of listing block is visible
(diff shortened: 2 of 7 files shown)


tiberiuichim:
@ksuess The addition of the test is really nice.

I wonder about the `variations_internal`. What is the use case? Is this something that's provided by Volto? Can it be formalized, in one way or another?

ksuess:
> I wonder about the `variations_internal`. What is the use case? Is this something that's provided by Volto? Can it be formalized, in one way or another?

The use case is a search block in an administrative view for bulk editing. 
Problem is, that a variation is always in the search block variation field options for the editor. To achieve that a variation is not listed in this options i need to patch / shadow the SearchBlockView component or propose a change for Volto like this PR.
 
Instead of introducing a new `config.blocks.blocksConfig.listing.variations_internal`, an alternative way to distinguish between variations would be to introduce a new option of the variation itself `notforeditors`.

```
        config.blocks.blocksConfig.listing.variations: [
          ...config.blocks.blocksConfig.listing.variations,
          {
            id: 'subsriptions',
            isDefault: true,
            title: 'Subscriptions',
            template: ListingSubscriptions,
            fullobjects: true,
            notforeditors: true,
          },
        ],
```

tiberiuichim:
One way to do this would be something like this:

- The end goal is for you to be able to write a custom schemaEnhancer for the search block which can enable/disable the variations in the select, based on the role of the user
- I can see that [here](https://github.com/plone/volto/blob/ce0657fd2357608dce34a9cfdaed2e222e1941b9/src/helpers/Extensions/withBlockSchemaEnhancer.js#L119-L142), if we switch the order of "apply schema enhancer" and "add the select field to the schema", you'll have the chance to tweak the variations select options, from your own custom schema enhancer. The only thing that's missing would be that variable or property that could be used, from the schema enhancer, to understand if the variations select options should be tweaked or not. I don't know how you discriminate between the "special editors" and "regular editors", maybe there's some easily accessible information in the system that can be grabbed by the HOC and passed to the schema enhancers?

ksuess:
Obsolete: Immitate "SearchBlockView" with component "ListingBody" with appropriate props. And compose with "withQueryString" and "withSearch". No registration of config.blocks.blocksConfig.listing.variations needed. See https://github.com/plone/volto/issues/3050#issuecomment-1146857684

PR #3108, opened 2022-02-26
Debugrequests

sneridagh:


src/helpers/Api/Api.js
```diff
@@ -55,6 +56,8 @@ class Api {
         let promise = new Promise((resolve, reject) => {
           request = superagent[method](formatUrl(path));
 
+          debug('requests')(formatUrl(path));
```
tiberiuichim, line 59: I usually use `DEBUG=superagent yarn start`. Would that be enough?

sneridagh: you are right... I didn't remember this one! This one was a quick, compulsive PR :)

PR #3110, opened 2022-02-28
Fix overflowing issue in the toolbar

kreafox:
The toolbar menu height is restricted by setting a max-height and the content list can be larger than the space that the element provides. There is an overflow-y: auto defined in CSS, but we have some menu styling in the Toolbar component which causes that scroll to not be applied. I don't know why this override was needed, but I think we can safely remove it.
Toolbar menu with scroll: 
toolbar_after


sneridagh:
This had a reason, I'm sure it has to do with the mobile view.

kreafox:
@sneridagh I've checked the mobile view, I just forget to mention it, and everything looks good if we remove it.

sneridagh (approved):

PR #3115, opened 2022-03-02
Change prop `name` -> `componentName` in component `Component`

### LINKED ISSUE #3114: Component component name prop can possibly clash with other prop names
**Description**
The new ``Component`` component prop ``name`` is too generic.

It could lead to issues (prop name clashes) in registering components already available in the ecosystem, and also with the actual HTML attribute ``name``, even if it's not often used.

**Example**
I want to create a radio input component, which would look like this

```javascript
const MyRadio = (props) => (
  <input type="radio" {...props} />
);
```

I would use it like this

```html
<MyRadio id="id1" name="group-name" value="..." />
<MyRadio id="id2" name="group-name" value="..." />
```

I could not register this component and use it with ``Component``, if I understand this new registry correctly.

**Suggestion**
I'm not sure we can 100% avoid the clash without making the developer experience worse. I would suggest changing it to something less capable of clashing, like ``componentName``.

sneridagh:
Fixes #3114

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 ### Internal
 
+- Change prop `name` -> `componentName` in component `Component` @sneridagh
+
 ## 15.0.0-alpha.11 (2022-03-02)
 
 ### Bugfix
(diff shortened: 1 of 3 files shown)


tiberiuichim (approved):

pnicolli (approved):
Thanks! <3

PR #3116, opened 2022-03-04
allow extracting msgids from customizations

### LINKED ISSUE #1846: Translations in customizations
If I add a customization for a component, for example for Footer.jsx component in customizations/components/Footer/Footer.jsx
and i add some new strings that i want to translate, when running yarn i18n new strings ar not added to .po files, and i cannot translate theme

erral:
Fixes #1846


sneridagh:
@erral see https://github.com/plone/volto/issues/1846#issuecomment-1059006374

This is done on purpose, I'm afraid.

PR #3125, opened 2022-03-08
Sort the choices in Facets

iFlameing:


sneridagh (approved):
LGTM, @tiberiuichim do you have any objection to this?

tiberiuichim:
I still hope we'd have a way to sort options by count, but I don't think that's possible in the near future, unless we have a separate endpoint for qs information. I'm good with merging this PR.

tiberiuichim (approved):
Maybe a better changelog entry is needed, to mention that the change is about the search block.

PR #3132, opened 2022-03-09
Clear current block on insert new block.

robgietema:


sneridagh:
What's the purpose of this? This piece of code is shared with all the block-ish implementations (like the blocks-in-blocks blocks) and alike, so I want that @tiberiuichim and @avoinea take a look.

@tiberiuichim @avoinea do you remember if there's any drawback on this?

avoinea:
@sneridagh @robgietema  Try it in Dexterity Layout. Make a field required and see the impact.

robgietema:
> What's the reasoning of this? This piece of code is shared with all the block-ish implementations (like the blocks-in-blocks blocks) and alike.

@sneridagh This is needed for the "power user menu". For this feature you can type /image for example to enter an image block. But after you press enter this block should be empty again, hence this code. I don't see this method used anywhere else in Volto or in kitconcepts.volo-blocks. Do you have an example where this is used?

> Try it in Dexterity Layout. Make a field required and see the impact.

@avoinea I don't understand what you mean here. I assume you mean the "normal" dexterity form without the blocks layout but can't see any errors there and don't understand what this has to do with this code. Can you explain in more detail what you meant here?

avoinea:
@robgietema @sneridagh The side effect of this PR:

https://user-images.githubusercontent.com/323385/157629758-82a425a3-8c50-4bf3-90b7-e3501da530c0.mp4

As you can see in the end the editor managed to overcome the required field by layout and remove it.

PR #3140, opened 2022-03-11
Add the sorting order of the facet year

iRohitSingh:


src/components/manage/Blocks/Search/components/Facets.jsx
```diff
@@ -51,10 +51,15 @@ const Facets = (props) => {
                 ? query_to_values[field].includes(value)
                 : true,
             );
-
-          choices = choices.sort((a, b) =>
-            a.label.localeCompare(b.label, 'en', { sensitivity: 'base' }),
-          );
+          if (field === 'year') {
```
tiberiuichim, line 54: There are already options to rewrite the facet values, I don't think we need to hardcode for a specific type of field.

https://github.com/plone/volto/blob/97022b23162e2863aa6613952d3c53742b36a902/src/components/manage/Blocks/Search/components/Facets.jsx#L85

tiberiuichim (changes requested):

src/components/manage/Blocks/Search/components/Facets.jsx
tiberiuichim: And it's even documented :) https://docs.voltocms.com/blocks/settings/#facetwidgets-rewriteoptions-extension

tiberiuichim: Although, and maybe you can tackle this in this PR, you'd have to also pass the `field` to `rewriteOptions()`

Issue #3093, opened 2022-02-21
Move `docs/source/*` to `docs/*`?


stevepiercy:
Now that we've migrated to MyST and have Volto docs importing nicely into the main Plone docs, it occurred to me that we have slightly different docs folder structures:

- `plone/documentation/docs/*`
- `plone/volto/docs/source/*`

I propose to flatten the hierarchy in Volto docs by moving everything in `docs/source` up to `docs`. It would also mean one fewer directory to navigate, and I like not typing 6 more characters or saving a mouse click.

It would require a couple of configuration items to change, including `Makefile` targets.

PR #3144, opened 2022-03-14
Chaged docs folder structure

ktsrivastava29:
Earlier the folder structure was this `plone/volto/docs/source/`, one had to migrate one folder inside to access its files, I've know changes the folder structure to `plone/volto/docs/`.

fix: #3093


stevepiercy:
Volto 15 will be released soon, and after then, we can focus on merging this for Volto 16.

We would also need to update the repository `plone/documentation@6-dev`'s Makefile, because it imports Volto docs, in a PR in that repo.

PR #3142, opened 2022-03-12
Fixed development environment setup commands

ktsrivastava29:
I edited the command in the readme file; previously, there was a $ symbol before it, and when using direct copy, the $ symbol used to get copied as well; moreover, 
```
npm install -g yo @plone/generator-volto 
```
requires sudo access, so sudo was put before it.
Please look into it; if any other changes are needed, I'll be happy to make them.

PR #3152, opened 2022-03-15
Revert "Fixed development environment setup commands"

sneridagh:
Reverts plone/volto#3142


ktsrivastava29:
@sneridagh Not a problem, we want the best thing to remain, so can I at least erase the $ symbol before the commands because it's an outdated format and the $ symbol also gets copied when we try copy directly

Issue #3124, opened 2022-03-07
Clipped dropdown menu in Contents view


tiberiuichim:
![image](https://user-images.githubusercontent.com/141568/157092111-f8df8b82-06ee-47b3-a803-89bd36097500.png)

nileshgulia1:
Recently @ichim-david looked into these kinds of changes: https://github.com/plone/volto/pull/2819

tiberiuichim:
In this case it wasn't about small window, rather just limited number of items in the folder (just 2).

ichim-david:
Fixed by https://github.com/plone/volto/pull/3158 merging in master

PR #3158, opened 2022-03-15
[Bug #3124] - clipped dropdown menu in Contents view

mihaislobozeanu:


avoinea (changes requested):

theme/themes/pastanaga/modules/popup.overrides
```diff
@@ -7,3 +7,63 @@
     vertical-align: middle;
   }
 }
+
+
+.dropdown-popup-trigger{
```
tiberiuichim, line 12: I think this file needs prettier for formatting. Usually there's a space before curlies.

avoinea (approved):

src/helpers/index.js
```diff
@@ -94,3 +94,4 @@ export { useDetectClickOutside } from './Utils/useDetectClickOutside';
 export { usePrevious } from './Utils/usePrevious';
 export { usePagination } from './Utils/usePagination';
 export useUndoManager from './UndoManager/useUndoManager';
+export Popup from './Popup/Popup';
```
pnicolli, line 97: I thought about it after the team meeting today and I'm not a fan of having an actual component in the helpers folder. I would personally prefer moving it to the components folder. Just my 2 cents on this.

theme/themes/pastanaga/extras/contents.less
```diff
@@ -80,6 +80,7 @@
 
   .contents-table-wrapper {
     width: 100%;
+    //#3124 - dropdown menus are clipped when table has few or no records
```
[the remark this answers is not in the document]
avoinea: I think this is a leftover. @mihaislobozeanu

tiberiuichim:
@pnicolli We already have some components in helpers. https://github.com/plone/volto/blob/master/src/helpers/ScrollToTop/ScrollToTop.jsx and https://github.com/plone/volto/blob/master/src/helpers/BodyClass/BodyClass.jsx

I'm ok with moving it in another place. Maybe components/theme/PopupWrapper ?

sneridagh:
I think also that it's not a helper, but a component... I would re-place it, what do you think? 

Also, but this is another question, what when we get rid of SemanticUI (eg. projects that do not use it at all? :)

tiberiuichim:
> Also, but this is another question, what when we get rid of SemanticUI (eg. projects that do not use it at all? :)

when we get rid of semantic-ui we'll have other, bigger, things to worry about :-)

In any case, we'd begin with a refactoring of the Contents component and move it to react-table or something like that. I think that's something we should already start doing.

avoinea:
@pnicolli @tiberiuichim What we agreed on this one?

sneridagh:
If everybody agrees, we refactor it and move it to the components folder, then it's good to go?

pnicolli:
> If everybody agrees, we refactor it and move it to the components folder, then it's good to go?

Agreed

tiberiuichim:
I still don't think components/theme is the right place to put it, as it's a helper component that can be used even outside Volto. I can agree that src/helpers is not the best location, but there's no other existing place for it. Maybe a new `src/components/ui`. But I'm fine with placing it in src/components/theme, so +1 from me.

sneridagh:
@tiberiuichim I agree with you that it's a new kind of component, and that ideally should be "separated" from the structural ones. I also thought in having this new category in the components folder for Quanta. See

https://github.com/plone/volto/pull/3086/files#diff-9837e4844c73b95722bd37d81ab4ce85e00a07552905224cfd7d9041e62703daR1

I initially called that "core", but naming... again. We can start doing it that now, or reorder them later.

tiberiuichim:
In any case, we expose components through the `@plone/volto/components` index, so if we decide that we need to move it later, we can do that without too much breakage. Let's put it in components/theme.

sneridagh:
@mihaislobozeanu could you please take care of the decision? Thanks!

mihaislobozeanu:
> 

Resolved

giuliaghisini (approved):

src/components/manage/Contents/Contents.jsx
```diff
@@ -618,7 +619,7 @@ class Contents extends Component {
    * @returns {undefined}
    */
   onChangeSelected(event, { value }) {
-    event.stopPropagation();
+    //event.stopPropagation();
```
sneridagh, line 622: Are we sure about this one? and if so, let's remove it.

sneridagh (commented):
@mihaislobozeanu LGTM, sorry for the pickiness, but could you take a look at the last thing? then let's merge.

PR #3166, opened 2022-03-16
Remove non add-on names from `addonNames` list in Addons Registry. Up…

sneridagh:
…date the list in the `addonsInfo` for the addons loader as well.


sneridagh:
@tiberiuichim a bit puzzled now, the tests should have failed long long ago, when I introduced the version...

tiberiuichim (approved):

PR #3174, opened 2022-03-17
Add helper utils.

robgietema:


src/helpers/Utils/Utils.test.js
```diff
@@ -345,4 +349,31 @@ describe('Utils tests', () => {
...
+  });
+
+  describe('parseDateTime', () => {
+    it('reorders an array immutable-ish', () => {
+      const array = ['a', 'b', 'c'];
+      deepFreeze(array);
+      const result = reorderArray(array, 2, 0);
+      expect(result).toEqual(['c', 'a', 'b']);
+    });
+  });
```
davisagli, line 378: @robgietema this claims to be describing `parseDateTime`!

robgietema: Good catch, fixed

PR #3177, opened 2022-03-18
Fix addon registry

sneridagh:


avoinea:
@sneridagh It seems to be ok now but the addons are not listed in control-panel:

### Before:

before

### After

after

sneridagh:
@avoinea I know, I filtered only for "registered" === Declared in `package.json` `addons` key now. Do you think that it's useful that it shows all of them?

sneridagh:
So addons in addons are not showing now.

avoinea:
@sneridagh Would helpful to show them all. At least in our case :smile:

sneridagh:
Problem is that we can't differentiate "addons" from non addons (libs) in dev

avoinea:
@sneridagh Then I think we can leave with it :smile: As we have the kitkat version, we can easily track the included addons :wink:

avoinea (approved):

sneridagh:
Then we will be in square one again, since will be addons there that will not being really applied. (Eg. Ones coming from volto.config.js)

tiberiuichim (approved):

tiberiuichim:
I think this would be useful to show in the control panel: 

https://github.com/plone/volto/blob/74723c5c3500797bcc517ec5416b4f7889c3b782/addon-registry.js#L268-L272

sneridagh:
Nice! Will do!

PR #3180, opened 2022-03-19
Grid block in core + primitive Container block

sneridagh:
Backport of `@kitconcept/volto-blocks-grid` (7.x.x) features, using Volto's default blocks-in-block architecture. This change will render old grid blocks incompatible with this one, unless a migration is made. However, coexistence between old grid and new ones is possible.

**UPDATE 01.Oct.2022**: Teasers moved to a separate PR: #3706 so they can go on its own lifecycle. I left the old implementation here for testing purposes. They should be removed after #3706 is merged.

**UPDATE 09.June.2023**: After discussion in the Beethoven Sprint these decisions were made for short term (affecting this PR):
- This PR is about the porting of the current state of `@kitconcept/volto-blocks-grid` (in its version in 7.x.x series).
- The name of the PR is decided to continue being `Grid`, as it was before to be consistent in what we are porting.
- The internal name will be renamed to `gridBlock` to allow coexistence between both blocks (the core and the add-on one) at the same time, in order to avoid migrations.
- This PR won't be backported to 16.x.x series (for marketing reasons). People willing to use the grid in 16 should continue using the add-on version.

### Future:

An improved drag-and-drop system will be needed and the ability to "move" blocks from outside (main block engine column) to inside the Row/Column blocks, and from one row/column block to another.

### Minimum Viable (mergeable) Product

- [x] Main port for grid block
- [x] Cleanup internal naming (classnames, Component names, etc)
- [x] Uses Volto's default blocks-in-blocks architecture
- [x] Unified Style wrapper compatible
- [x] Unknown block problem, if block is "empty"
- [x] Fix DraggableOverlay properly https://github.com/plone/volto/issues/4405
- [x] Cypress tests

### Improvements

- [x] Inject self blocks config if present in block config
- [ ] Variations per Grid element, different than a non-grid element.

## Nice to have

- [ ] Solve Toolbar(s) quirks (Quanta/-ish?) opt-in?
- [ ] Improved drag-and-drop system for supporting moving blocks in/out (please someone look into it)


avoinea (dismissed):
Niiicee!!!

nileshgulia1:
> An improved drag-and-drop system will be needed and the ability to "move" blocks from outside (main block engine column) to inside the Row/Column blocks.

Looks promising🔥

djay:
@sneridagh can you replace the x icon with a trash icon as per https://github.com/plone/volto/issues/3816 ?

stevepiercy:
@sneridagh also consider removing the `X`, as it looks to be redundant to an existing trash can icon to the right, per https://github.com/plone/volto/issues/3816#issuecomment-1298047935

sneridagh:
@stevepiercy please could you take a look at the docs?

stevepiercy (changes requested):
I made a bunch of English syntax and grammar changes, and a wee bit of MyST stuff, in a commit and pushed. Please pull down the changes before editing further.

I have some confusion about this Thingie we call a "Container block primitive" that is not a block. See comments for discussion.

docs/source/blocks/core/container.md
```diff
@@ -0,0 +1,94 @@
...
+# Container Block Primitive
+
+```{versionadded} Volto 17.0.0-alpha.10
+```
+
+Volto comes with a Container Block primitive.
+It is __not__ a block by itself, rather than that, it allows to build blocks that contains other blocks.
+The Volto Grid block is build using it.
+The whole idea of primitive is to use it as "template" and pattern to use it and its internal methods to build blocks that share the same feature: allow the blocks in block container and the management of them.
+The blocks build on this primitive will use schema properties (or styling wrapper ones) to provide layout to the containing blocks (horizontal/vertical/spacing/behavior/backgroundColor/...) along with other look and feel properties.
```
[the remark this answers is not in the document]
sneridagh: Yeah, let's find a moment and talk about this!

src/components/manage/Blocks/Container/Edit.jsx
```diff
@@ -0,0 +1,195 @@
...
+          setSelectedBlock(id);
+        }}
+        onChangeFormData={(newFormData) => {
+          onChangeBlock(block, {
+            ...data,
+            ...newFormData,
+          });
+        }}
+        onChangeField={(id, value) => {
+          if (['blocks', 'blocks_layout'].indexOf(id) > -1) {
```
davisagli, line 161:
```suggestion
          if (['blocks', 'blocks_layout'].includes(id)) {
```

src/components/manage/Blocks/Grid/adapter.js
```diff
@@ -0,0 +1,19 @@
...
+  value,
+}) => {
+  let dataSaved = {
+    ...data,
+    [id]: value,
+  };
+  if (id === 'direction') {
+    dataSaved = {
+      ...dataSaved,
+      '@type': value,
```
davisagli, line 15: I don't understand this part; why does updating the direction set the type?

src/components/theme/View/RenderBlocks.jsx
```diff
@@ -23,14 +25,15 @@ const messages = defineMessages({
...
+  const { content, location, metadata, blockWrapperTag } = props;
+  const intl = useIntl();
   const blocksFieldname = getBlocksFieldname(content);
   const blocksLayoutFieldname = getBlocksLayoutFieldname(content);
   const blocksConfig = props.blocksConfig || config.blocks.blocksConfig;
   const CustomTag = props.as || React.Fragment;
 
   return hasBlocksData(content) ? (
-    <CustomTag>
+    <MaybeWrap condition={CustomTag} as={CustomTag}>
```
davisagli, line 36: How would CustomTag not evaluate to true here, given that it is defaulted to React.Fragment on line 33?

src/components/theme/View/RenderBlocks.jsx
```diff
@@ -43,37 +46,55 @@ const RenderBlocks = (props) => {
...
-        ) : blockData ? (
-          <div key={block}>
-            {intl.formatMessage(messages.unknownBlock, {
-              block: content[blocksFieldname]?.[block]?.['@type'],
-            })}
-          </div>
-        ) : (
+        if (content[blocksFieldname]?.[block]?.['@type'] === 'empty') {
+          return (
+            <MaybeWrap condition={blockWrapperTag} as={blockWrapperTag}>
```
davisagli, line 51: Needs a `key`

src/components/theme/View/RenderBlocks.jsx
```diff
@@ -43,37 +46,55 @@ const RenderBlocks = (props) => {
...
+              <RenderEmptyBlock />
+            </MaybeWrap>
+          );
+        }
+
+        if (Block) {
+          return (
+            <MaybeWrap condition={blockWrapperTag} as={blockWrapperTag}>
```
davisagli, line 59: Needs a `key`

theme/themes/pastanaga/extras/grid.less
```diff
@@ -0,0 +1,405 @@
...
+      }
+
+      & > div > div {
+        height: 100%;
+      }
+    }
+  }
+}
+
+// Override handler TODO: do it better (only for grids)
```
davisagli, line 28: What's left to do here?

theme/themes/pastanaga/extras/grid.less
```diff
@@ -0,0 +1,405 @@
...
+    top: 0;
+    left: 50%;
+    transform: translate(-50%, 0);
+  }
+}
+
+// Required to overcome the droppable from b-d&d at the bottom
+// !! Required for ALL blocks that have b-d&d and a grid
+.block.gridBlock,
+[class$='Row'].block {
```
davisagli, line 127: Are the selectors for `[class$='Row']` still used?

theme/themes/pastanaga/extras/grid.less
```diff
@@ -0,0 +1,405 @@
...
+  .block.inner.slate {
+    cursor: text;
+  }
+
+  body.cms-ui .ui.basic.segment.content-area & {
+    cursor: grab;
+  }
+}
+
+// Aspect ratio on all images
```
davisagli, line 393: This might be too opinionated.

davisagli (changes requested):
I still want to do some testing and review the docs, but here's my first pass looking at the code.

docs/source/blocks/core/grid.md
```diff
@@ -0,0 +1,148 @@
...
+The Grid block uses a fixed horizontal layout with a configurable maximum items per block, defaulting to four elements.
+
+The Grid block can potentially contain any block type. However, the used blocks might need some styling work to adapt to the horizontal and small widths in the container. For this reason, only these four are included by default: `slate` (text), `image`, `listing` and `teaser`.
+
+## Settings
+
+The Grid has the following settings:
+
+Headline
+: The headline is a heading that appears above the title of the Grid Block.
```
davisagli, line 27:
```suggestion
: The headline is a heading that appears above the columns of the Grid Block.
```

docs/source/blocks/core/grid.md
```diff
@@ -0,0 +1,148 @@
...
+---
+
+(grid-block)=
+
+# Grid Block
+
+```{versionadded} Volto 17.0.0-alpha.10
+```
+
+Volto comes with a Grid block, building upon the primitive of a container block, it is a block that contains other blocks.
```
davisagli, line 17: I would not mention the container block in the first sentence; that's an implementation detail and not relevant to users of the block. You could mention it further down.

docs/source/blocks/core/grid.md
```diff
@@ -0,0 +1,148 @@
...
+
+Headline
+: The headline is a heading that appears above the title of the Grid Block.
+
+## Block configuration
+
+The Grid block has these configuration settings:
+
+templates
+: It's a function to provide a list of available templates.
```
davisagli, line 34: What is a template?

docs/source/blocks/core/grid.md
```diff
@@ -0,0 +1,148 @@
...
+  };
+};
+```
+
+maxLength
+: The maximum elements length allowed inside the grid block.
+By default is four elements.
+
+allowedBlocks
+: The internal `id` of the allowed blocks in the grid block.
```
davisagli, line 60:
```suggestion
: Array of block types allowed in the grid block.
```

docs/source/blocks/core/grid.md
```diff
@@ -0,0 +1,148 @@
...
+      head_title: value[0].head_title,
+    };
+  }
+  onChangeBlock(block, dataSaved);
+};
+```
+
+## Migration from `@kitconcept/volto-blocks-grid`
+
+From Volto 17.0.0-alpha.10 the Grid block, as it's in the `@kitconcept/volto-blocks-grid` add-on version 7.x.x, is included in Volto core.
```
davisagli, line 142:
```suggestion
The Grid block was included in Volto core starting from Volto 17.0.0-alpha.10. It is based on the `@kitconcept/volto-blocks-grid` add-on version 7.x.x.
```

davisagli (commented):
Done testing and reviewing the docs now. It seems to work well.

cypress/support/commands.js
```diff
@@ -705,6 +705,20 @@ Cypress.Commands.add('getSlate', (createNewSlate = false) => {
   return slate;
 });
 
+Cypress.Commands.add('getSlateSelector', (selector = SLATE_SELECTOR) => {
```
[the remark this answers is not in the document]
sneridagh: as in

```
Cypress.Commands.add('getSlate', (createNewSlate = false, selector = SLATE_SELECTOR) => {
```

I've thought about it, but then the call will be cumbersome, as in:

```
getSlate(null, '')
```

I thought the previous one would be more elegant :( it's the price for not using an object from the beginning.

docs/source/blocks/core/grid.md
sneridagh: ok, but remember that this is dev docs, not user docs.

src/components/manage/Blocks/Grid/adapter.js
sneridagh: Uh, this was a leftover from the column block experiments

docs/source/blocks/core/grid.md
sneridagh: I pushed a more detailed description. 

Now I'm thinking that the TemplateChooser maybe should be part of the particular block, not of the container... I will think about it.

src/components/manage/Blocks/Block/BlocksForm.jsx
```diff
@@ -42,6 +42,7 @@ const BlocksForm = (props) => {
     isMainForm = true,
     blocksConfig = config.blocks.blocksConfig,
     editable = true,
+    direction,
```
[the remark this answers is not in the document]
sneridagh: Done.

src/components/manage/Blocks/Grid/adapter.js
sneridagh: Removed.

src/components/theme/View/RenderBlocks.jsx
sneridagh: Fixed.

src/components/theme/View/RenderBlocks.jsx
sneridagh: Fixed

src/components/theme/View/RenderBlocks.jsx
sneridagh: Fixed

theme/themes/pastanaga/extras/grid.less
sneridagh: It was a marker for when we use the Quanta Toolbar... nothing important, I completed it.

theme/themes/pastanaga/extras/grid.less
sneridagh: It's configurable, maybe we should include docs for it.

theme/themes/pastanaga/extras/grid.less
sneridagh: Removed

docs/source/blocks/core/container.md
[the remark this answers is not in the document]
sneridagh: @stevepiercy in the meanwhile we don't talk yet, think in the primitive, as a base class from where to inherit from, only that in React, you "compose" components from other components so you can reuse props and methods.

theme/themes/pastanaga/extras/grid.less
sneridagh: Included.

cypress/support/commands.js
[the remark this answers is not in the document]
davisagli: @sneridagh oh yeah, that makes sense. I was thinking in Python. :)

docs/source/blocks/core/grid.md
davisagli: I know, I was just thinking there's a difference between a developer who wants to configure the existing block and a developer who wants to build one like it.

I would at least rephrase it slightly to:
Volto comes with a Grid block, which is a block that contains other blocks arranged in a horizontal row. It is implemented using the Container block primitive.

docs/source/blocks/core/grid.md
davisagli: @sneridagh I don't have a strong opinion on that. The new description is helpful and helped me realize how the templates could be useful for other kinds of containers.

docs/source/blocks/core/grid.md
```diff
@@ -0,0 +1,156 @@
...
+      },
+    ],
+  },
+};
+```
+
+## Block Image Styling customization
+
+By default, the images inside a Grid block are using a 16/9 aspect ratio.
+It is configurable by setting a LESS variable `@grid-images-aspect-ratio` or setting a CSS property somewhere in your project: `--grid-images-aspect-ratio`.
```
davisagli, line 96: And what if I don't want a fixed aspect ratio? Set it to `unset`?

docs/source/blocks/core/grid.md
```diff
@@ -0,0 +1,156 @@
...
+There is also another setting for the `object-position` CSS definitin:  `@grid-images-object-position` and the corresponding `--grid-images-object-position`.
```
davisagli, line 97:
```suggestion
There is also another variable for the `object-position` CSS property:  `@grid-images-object-position` and the corresponding `--grid-images-object-position`.
```

ksuess:
I think both is of interest: image as it is in a grid and images cropped to predefined aspect ratio in grid. How about offering both _full height_ and _aspect ratio_ to the user (via image alignment)?

image_options

---
cropped to aspect ratio
grid_image_aspect_ratio

image as it is, uncropped
grid_image_full_height

ksuess (changes requested):
I think, teaser with image alginment _full width_ should show image in full width of the column. 
grid_teaser_full_width

docs/source/blocks/core/grid.md
```diff
@@ -0,0 +1,169 @@
...
+However, due to its horizontal alignment and narrow widths of the blocks it contains, the blocks might need additional styling.
+For this reason, only four blocks are included by default, namely `slate` (text), `image`, `listing`, and `teaser`.
+
+
+## Settings
+
+The grid block has the following settings.
+
+`Headline`
+:   The headline is a heading that appears above the title of the grid block.
```
davisagli, line 32: @stevepiercy It's not "above the title". It is the title. We already fixed this before your latest edits. (Thanks for those, btw.)
```suggestion
:   The headline is a heading that appears above the columns of the grid block.
```

docs/source/blocks/core/grid.md
sneridagh: Yes, we can document this. But by doing it, then the cards look awful (every image with its own height) if you don't set one. I'd say let's be opinionated on this one, document the override.

sneridagh:
@ksuess @davisagli The policy in Volto core has always been "don't make the user think" for defaults, being the default the one he most likely would want, and looks better by default (even if it's opinionated). If the integrator wants to allow other approaches (like the one you proposed), then either provide a handler (widget) for it, or a variant as preset (even more straightforward, and don't make the user think what widget combination will provide the result he's looking for).

Defaulting to a fixed aspect ratio provides consistency (if you force it everywhere) to your site images and a streamlined and better look overall.

docs/source/blocks/core/grid.md
```diff
@@ -105,6 +105,8 @@ By default, the images inside a grid block are using a 16/9 aspect ratio.
 It is configurable by setting a LESS variable, `@grid-images-aspect-ratio`, or setting a CSS property somewhere in your project, `--grid-images-aspect-ratio`.
 There is another setting for the `object-position` CSS definition, `@grid-images-object-position`, and its corresponding `--grid-images-object-position`.
 
+If you want to disable this feature from the grid images, you should set both to `unset`.
```
stevepiercy, line 108: Does "both" refer to the aspect ratio and the position? If so, then:

```suggestion
If you want to disable a default setting, then set its value to `unset`.
```

stevepiercy (commented):
See comment.

ksuess:
> @ksuess @davisagli The policy in Volto core has always been "don't make the user think" for defaults, being the default the one he most likely would want, and looks better by default (even if it's opinionated). If the integrator wants to allow other approaches (like the one you proposed), then either provide a handler (widget) for it, or a variant as preset (even more straightforward, and don't make the user think what widget combination will provide the result he's looking for).
> 
> Defaulting to a fixed aspect ratio provides consistency (if you force it everywhere) to your site images and a streamlined and better look overall.

OK, I was just irritated by the fact, that in edit mode the image of an image block appears with a fixed aspect ratio of 16:9 and in view mode it does not but unchanged at all. I assume you want the image of an imge block to come per default with an aspect ratio 16:9?

edit mode:
grafik

view mode:
grafik

stevepiercy:
> OK, I was just irritated by the fact, that in edit mode the image of an image block appears with a fixed aspect ratio of 16:9 and in view mode it does not but unchanged at all. I assume you want the image of an imge block to come per default with an aspect ratio 16:9?

I would expect both the view and edit modes to be WYSIWYG with the same aspect ratio applied.

ksuess:
> OK, I was just irritated by the fact, that in edit mode the image of an image block appears with a fixed aspect ratio of 16:9 and in view mode it does not but unchanged at all. I assume you want the image of an imge block to come per default with an aspect ratio 16:9?
> 

I think it's just a CSS rule missing.
By now it seems to assume a .grid-image-wrapper arround images of image blocks?
I don't see this. Where does the wrapper come from? 
In plain vanilla Volto it looks like this:
grafik

sneridagh:
@ksuess let me see, they should match, for sure!

sneridagh:
@stevepiercy I added a couple of paragraphs more, I realised that the props apply to the container too, and added two more, in case you want to review.

stevepiercy (approved):

docs/source/blocks/core/container.md
```diff
@@ -28,6 +28,52 @@ When building a custom block that uses a container as its base, you must pass yo
...
 Currently only one property is supported, although others may be added in future releases.
 
-`Direction`
+`direction`
 :   The arrangement of blocks in the container, either `horizontal` or `vertical`.
     The default value is `horizontal`.
+
+`templateChooser`
+: To complete the full customization of the template choice experience, you can customize the component that shows the choices as well.
+So you can enhance the default component located in {file}`src/components/manage/TemplateChooser/TemplateChooser.jsx` by providing your own.
```
stevepiercy, line 37:
```suggestion
    You can enhance the default component located in {file}`src/components/manage/TemplateChooser/TemplateChooser.jsx` by providing your own component.
```

docs/source/blocks/core/container.md
```diff
@@ -28,6 +28,52 @@ When building a custom block that uses a container as its base, you must pass yo
...
+Remember that most likely you will have to provide also your own custom `templates` function that will play well with this one.
```
stevepiercy, line 38:
```suggestion
    You might also need to provide your own custom `templates` function that is compatible with your component.
```

docs/source/blocks/core/container.md
```diff
@@ -28,6 +28,52 @@ When building a custom block that uses a container as its base, you must pass yo
...
+
+`editBlockWrapper`
+: You can customize the container's `editBlockWrapper` by providing your own. This will be passed to the `BlocksForm` component used by the container. It is used to customize the block wrapper if the block that you are building needs to have one different for the user experience of your block.
```
stevepiercy, line 41:
```suggestion
:   You can customize the container's `editBlockWrapper` by providing your own.
    This will be passed to the `BlocksForm` component used by the container.
    It is used to customize the block wrapper if the block that you are building needs to have a different one for the user experience of your block.
```

docs/source/blocks/core/container.md
```diff
@@ -28,6 +28,52 @@ When building a custom block that uses a container as its base, you must pass yo
 The container has particular properties.
 Currently only one property is supported, although others may be added in future releases.
 
-`Direction`
+`direction`
 :   The arrangement of blocks in the container, either `horizontal` or `vertical`.
     The default value is `horizontal`.
+
+`templateChooser`
+: To complete the full customization of the template choice experience, you can customize the component that shows the choices as well.
```
stevepiercy, line 36:
```suggestion
:   To complete the full customization of the template choice experience, you can customize the component that shows the choices as well.
```

stevepiercy (changes requested):
Minor English grammar fixes, MyST indentation for definition list terms.

sneridagh:
After this one, it's ready for review and merge!!

src/config/finalizeConfiguration.jsx
```diff
@@ -0,0 +1,24 @@
...
+  // This is required in order to initialize the inner blocksConfig
+  // for the grid block, since we need to modify how the inner teaser
+  // block behave in it (= no schemaEnhancer fields for teasers inside a grid)
+  // Afterwards, it can be further customized in add-ons using the same technique.
+  config.blocks.blocksConfig.gridBlock.blocksConfig = {
+    ...config.blocks.blocksConfig,
+  };
+  config.blocks.blocksConfig.gridBlock.blocksConfig.teaser = {
+    ...config.blocks.blocksConfig.teaser,
+    schemaEnhancer: gridTeaserDisableStylingSchema,
```
davisagli, line 22: @sneridagh Doesn't this make it impossible for an addon to use custom schema enhancers for the teaser block inside a grid?

sneridagh: Yes, you are correct... I've been struggling with this one... because slate is an add-on, we need their config first, if not, then the slate block is not in the `blockConfig` when you override it in the Grid, then it's not available. A chicken-egg situation :( not sure how to deal with it. 

Maybe initializing it from the core? Any ideas? /cc @tiberiuichim

sneridagh: Ok, I think I got it, incoming commit...

docs/source/blocks/core/container.md
```diff
@@ -0,0 +1,81 @@
...
+
+## Building blocks using a container
+
+When building a custom block that uses a container as its base, you must pass your block's properties into the container.
+
+
+## Container properties
+
+The container has particular properties.
+Currently only one property is supported, although others may be added in future releases.
```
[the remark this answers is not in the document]
sneridagh: done.

sneridagh:
@davisagli I realised that if the slate add-on was responsible for adding its own config, why don't make it responsible to add the config for the grid too?

So here we go, last commit. Let me know what do you think. I removed the finalize thinggie too.

davisagli (approved):
LGTM now

stevepiercy (approved):

pnicolli:
> @ksuess @davisagli The policy in Volto core has always been "don't make the user think" for defaults, being the default the one he most likely would want, and looks better by default (even if it's opinionated). If the integrator wants to allow other approaches (like the one you proposed), then either provide a handler (widget) for it, or a variant as preset (even more straightforward, and don't make the user think what widget combination will provide the result he's looking for).
> 
> Defaulting to a fixed aspect ratio provides consistency (if you force it everywhere) to your site images and a streamlined and better look overall.

Couple days late to the party here probably, but I don't fully agree here. I agree on having sane and good looking defaults, therefore I am totally ok with having an aspect ratio as the default and having it configurable with a css property. What I would want though is for the user to be able to choose to unset the aspect ratio. Just unset, not configure. I have seen use cases for both set and unset ratio in the same site, I think we should allow that.
I just read the comments here and could not try the PR so I am answering based on what I understood. If the option is actually there, then I'm fine :)

stevepiercy:
For convenience, new docs preview:

- [Core Blocks developer notes](https://deploy-preview-3180--volto.netlify.app/blocks/core/index.html#)
- [Container](https://deploy-preview-3180--volto.netlify.app/blocks/core/container.html)
- [Grid block](https://deploy-preview-3180--volto.netlify.app/blocks/core/grid.html)

avoinea:
Add a grid with an image and a teaser block, the make the image full width:

Screenshot from 2023-06-20 13-39-39

avoinea:
Add a grid block with an image and text block inside, the copy some text with multiple paragraphs from https://www.lipsum.com/feed/html and paste it within the text block, no more 4 block limit:

Screenshot from 2023-06-20 13-44-40

avoinea:
The same behavior if you un-bullet bulleted text:

Screenshot from 2023-06-20 13-55-11

Screenshot from 2023-06-20 13-55-21

davisagli:
@pnicolli Yes, it's simple to unset the aspect ratio in CSS, and that is mentioned in the docs now.

pnicolli:
@davisagli Yeah but I meant unsetting via Sidebar as a user, not via config.

sneridagh:
@avoinea it should be fixed now. Give it a try!

djay:
@sneridagh I think the UX might need a little more work. If you want to create a set of columns you might use a vertical container inside a grid block.

This screenshot is a grid block inside a grid block but you see the problem? the control bars are going to overlap. And I think it's going to get confusing which control bar relates to which container.

Screen Shot 2023-06-21 at 5 13 50 pm

I can give it some thought on what a more sustainable UX solution might be for add, delete and settings for multi-containment scenario might be, or else will you ask @albertcasado?

sneridagh:
@djay we agreed on these terms during the sprint:

UPDATE 09.June.2023: After discussion in the Beethoven Sprint these decisions were made for short term (affecting this PR):

This PR is about the porting of the current state of @kitconcept/volto-blocks-grid (in its version in 7.x.x series).
The name of the PR is decided to continue being Grid, as it was before to be consistent in what we are porting.
The internal name will be renamed to gridBlock to allow coexistence between both blocks (the core and the add-on one) at the same time, in order to avoid migrations.
This PR won't be backported to 16.x.x series (for marketing reasons). People willing to use the grid in 16 should continue using the add-on version.

> @sneridagh I think the UX might need a little more work. If you want to create a set of columns you might use a vertical container inside a grid block.
> 

We discussed a lot of scenarios where we can go with the current building blocks, as a community we should explore them from now on. Personally I'm not much interested in the containers in containers use case, as I see it as a pitfall where you open and endless set of issues of all kind, but I will give a try to more powerful ideas that came up from the discussions. I personally expect that this leads to a new breed of container blocks that enhances the current Volto ecosystem.

> This screenshot is a grid block inside a grid block but the problem is going to be the same right? the control bars are going to overlap. And I think it's going to get confusing which control bar relates to which container. Also if you are dragging, are you dragging the inner most block or one of the containers? and then all of this has to work on mobile too.
> 
> Screen Shot 2023-06-21 at 5 13 50 pm
> 
> I can give it some thought on what a more sustainable UX solution might be for add, delete and settings for multi-containment scenario might be, or else will you ask @albertcasado?

As said, the containers in containers is just another use case that should be explored. The current PR only fixes the use case for the Grid block as it was in the add-on. You are correct that then we will have to put in place a nicest "toolbar" for the container in place, only making available the current one you have selected. However this is a major problem that we should improve in the future, hopefully with the Quanta block toolbar in place (no this poor man replacement) and for this we will need to evolve and improve the current solution. I don't see now how we can do that, except to provide a pluggable toolbar, when you can replace it completely with your own control toolbar (or extend the current one).

I can give a quick try on this last one as an improvement to the current primitive, although could also be an improvement to the primitive to be made in the future, no breaking change involved...

We briefly went through the possible UX for it, all pointing out to the Quanta blocks toolbar, and that should be enabled when you select the inner blocks (for the inner block) very much the same as WP or Elementor does nowadays... For the rest, we didn't had time to explore much more the use case.

djay:
> Personally I'm not much interested in the containers in containers use case, as I see it as a pitfall where you open and endless set of issues of all kind

Sure and I totally get why some sites might prevent columns inside columns scenario. But, a columns block is already containers inside containers and a columns block is necessary. By columns block I mean horizontal container containing vertical containers. This is the basic layout block in WP and its really what should be in plone core also. So containers in containers in unavoidable no matter what even if you if you ban column blocks inside column blocks

I get the point you want to move forward and let people experiment. However what is the goal of the container primative.
1. make it easy for developers to create new blocks that contain other blocks
2. give the user a consistent UI. e.g. how form-block allows adding sub items is the same as eea.columns adding subitems and managing containment.

So letting block developers come up with their own UI to deal with container inside container scenarios is both making it harder for them and resulting in inconsistent UI for the end users.
So really it's a question of if there is anything we can add in the near term to get something that does work well enough?

> We briefly went through the possible UX for it, all pointing out to the Quanta blocks toolbar, and that should be enabled when you select the inner blocks (for the inner block) very much the same as WP or Elementor does nowadays...

I don't think the quanta toolbar solves the problem. Lets be clear about the problem
1. You need a way to navigate up and down the containment heirachy in a clear and obvious way. esp there are properties on parent containers you need to get at.
2. It needs to be clear what an add button does and how to go about adding something. when you have containers then you have both adding after the current block as well as how do I add something after any of the parent containers.
3. you need that to work on mobile also

In the current container UI the problem is the container add and settings buttons being on screen the whole time. It has to change.

I think containment is always complicated. you see that figma and other drawing programs. there is no magic bullet. 
But I did look at the latest gutenburg and I can see why they have evolved to make the choices they did. maybe not perfect but I think it solves the problems above so is not the worst thing to follow

Screenshot_2023-06-21-19-06-25-921_com android chrome

- adding
   - the blue top left add button constantly adds a block after the current block. so if its a column selected it adds a column after it. if its a paragraph in a column it adds a new block under.
   - the black container add button will add something inside the current container (if its a container). not sure about having 2 add buttons on screen but at least it's consistent and colour coded.
   - in addition you can hover the mouse (or click on mobile) inbetween blocks and an inbetween insert add appears
      - image
   - in addition there is a block toolbar menu to insert after and before
- selecting containers
   - on desktop they have container breadcrumbs on the button of the screen which is nice. and they have a figma like tree you can enable on the side. Without those you just have to click carefully on the edges to try and select the right container
      - image
   - on mobile it works differently which is interesting. Here touching a block will select the top level container. another touch the next level down etc etc until you can edit the leaf block. click elsewhere and it resets it and you have to drill down again to get to edit some text in a columns. I can see why they did this. it's a tradeoff on extra clciks but you don't have the screen space for breadcrumbs etc. it's slower but it workable.
   - in addition they have a select parent container on their block context menu
 
anyone can try on https://wordpress.org/gutenberg/

if we had the global add after button and the containment breadcrumbs and the mobile container drill down clicks. that would be a min to solve the problem I suspect. all the rest is bonus.
Or maybe there is a way to do better than what they have?

djay:
@sneridagh More specifically these four things are needed IIMO that make the container primitive much more useful.

Screen Shot 2023-06-23 at 1 09 44 pm


1. have a very clear indication of selection. 
   - Currently if the grid is selected rather than the block, this isn't clear. This should be built into the primitive.
2. Get rid of all buttons on the screen that aren't related to what is currently selected. 
   - The grid add/settings toolbar and the add button at the bottom (which doesn't seem to know about containment).
3. Have an add button that will "insert after" the selected block. 
   - In this case I've shown it consistent with the add button @davisagli added just in case thats easier to impliment. but it could also be in a quanta floating toolbar above the block, or like gutenburg, in the main toolbar not floating over the content at all. Personally I think the gutenburg's way has a lot going for it. It's going to work better on mobile and it's less likely to overlap something inside the block. It's also always in the same place so never have to look for it.
4. Have a way to get to containers in case it's hard to click on them. 
   - In this case I've put container breadcrumbs in the settings sidebar. But I actually think what gutenburg did with a bottom bar is better. That would give more space if the containment is complex and means you don't need to have the settings open.
   
 How hard would that be and what can we do to help?

sneridagh:
> @sneridagh More specifically these four things are needed to be built into the container primative IIMO that would make the it much more useful.
> 

This is the first draft of the container, and I'm assuming that we will continue iterate over it in the future. I totally agree that the toolbar should be configurable as well, this is one thing I can take care of in this PR.

> Screen Shot 2023-06-23 at 1 09 44 pm
> 1. have a very clear indication of selection.
>    
>    * Currently if a block is selected in a grid then both the block and the grid have an outline showing they are both selected which is confusing.

We are so perusing the nesting ability (using the same core components in a nested way). The outter block is selected as well, thus it has outline. You could remove it via CSS only for your container use case or even for the Grid? Yes, however, our users never complained. I think it's informative and useful and gives context. Could be that in other use cases, it's redundant and it gets in the middle? could be.

> 2. Get rid of all buttons on the screen that aren't related to what is currently selected.
>    
>    * The grid add/settings toolbar and the add button at the bottom (which doesn't seem to know about containment).

It doesn't. TBH, I haven't used the "new add button" feature while developing. I will take a look and remove it if a containment is present.

> 3. Have an add button that will "insert after" the selected block.
>    
>    * In this case I've shown it consistent with the add button @davisagli added just in case thats easier to impliment. but it could also be in a quanta floating toolbar above the block, or like gutenburg, in the main toolbar not floating over the content at all. Personally I think the gutenburg's way has a lot going for it. It's going to work better on mobile and it's less likely to overlap something inside the block. It's also always in the same place so never have to look for it.

If you think your block is better with this, then implement it with a custom `EditBlockWrapper.jsx` the container allows to configure it for a reason. If I make the toolbar configurable you could also push for it easily. 

The grid block works with the current toolbar, and it did all this time. We decided in the sprint not to change the grid block (in this PR) with new UX, so it will be the same as in `@kitconcept/volto-blocks-grid`.

> 4. Have a way to get to containers in case it's hard to click on them.
>    
>    * In this case I've put container breadcrumbs in the settings sidebar. But I actually think what gutenburg did with a bottom bar is better. That would give more space if the containment is complex and means you don't need to have the settings open.
> 

The breadcrumb/icon in the inner block title looks ok, but could be hard to implement... The sidebar knows nothing about its parent or if it's contained. That's why the toolbar has a "cog" icon handler in order to select it. Guttenberg has such a handler as well always visible (in the current selected inner block).

> How hard would that be and what can we do to help?

So, as I see it now, you want to push for a better user experience for contained blocks and I share this, and I guess all the community. However, it's incredibly hard to get right, as example, the Guttenberg and Elementor use cases, where they changed their minds a couple of times.

We need some stability and because of that, we agreed that the core grid was 1:1 to the add-on. That's settled.

As scape hatch, we thought into this primitive container, so the community can explore and build their own contained blocks out of it in an unified way, also with the aim to help people building such a type of block. It's the first draft, and I expect that we evolve it and improve it over time. I myself committed to work in a better version of the container paradigm while I quite like the Elementor approach (without the bloated UX/widgets etc...). So I plan to continue working on improving it. 

As I said, the containers in containers use case I'm not that interested because it just opens a Pandora box in all possible subjects. If you are interested in exploring it, let's do it.

I personally see the grid block as the minimum feature as in a horizontal container for Volto, having the very basics. But I do expect a new breed of add-ons with container blocks as a result. 

So I would ask you to be able to move forward and be able to merge this PR as soon as possible. Then continue the containers discussion (and feedback from developing such containers) in another place? Maybe the forum? I'm very looking forward for it, really.

djay:
@sneridagh 

Consider the following layout. Multiple vertical containers that can hold any of the basic block types. Ie a Grid > Column > Block.

unselected


My clients need to create things like this. EEA do too, that’s why they created eea volto-columns-block.
It’s a fairly basic requirement I would guess. 

One thing you haven’t answered is how the container primitive would work with this common columns scenario. How the UI would work without block developers have to remove the primitive UI and replace it with their own inconsistent versions?

I can only see three options
Prevent any 2 level containment. Ie grid block is allowed but columns is not
Give up on WYSIWYG and allow extra spacing/chrome around containers all at the same time. This prevents two different containers toolbars from being on top of each other.
Go to a proper selection model where you only show overlays about the currently selected block and not it’s parents.

I remember the original discussions about containment in tokyo. I know albert thinks containment is an antipattern. You seem to be trying to stick to that idea of not dealing with containers yet wanting to create a universal container primitive at the same time. You can’t have it both ways. If you have a container then people will use it to make containers inside containers. This is unavoidable. Not dealing with this issue now will just create more problems later. You would be releasing a half baked solution and causing churn.

The current version on the container primitive means volto-columns-block would have to disable your container toolbar due to the overlapping toolbar issues I mentioned. Out of the box the container primitive would give them this.

container_overlap


Then they need to add in another way to insert and add in another way to navigate up the the parent container, like they had to do once already

Screen Shot 2023-06-21 at 5 15 21 pm


And lhen later, once you get around to trying to solve this, which you will have to do eventually, they will then have to upgrade their block again with container primitive 2?
We are going to have to do the same. And others too. You are creating a lot of work for other developers and an inconsistent user experience for end users.
So I think for this reason trying to restrict containers to one level is also a bad idea.

I’d guess giving up on WYSIWYG is also out of the question. I believe it’s a core volto principle. Putting spacing/chrome/buttons around containers that make the items not appear how they would in the final version I believe is not the way to go.

So that only leaves showing chrome and overlays only on the current selected block (like every other WYSIWYG container based component editor I’ve found so far). Gutenburg may have iterated but that’s a good thing. We can take advantage of all their testing and not make the same mistakes.

However you don’t have to get this perfect now. You just have to provide a way that works regardless of how the containers are used so that block authors don’t have to remove your UI and replace it with their own.

What I’m suggesting is very close to the ideas in quanta (depending how you interpret it because it doesn’t deal with containers) and doesn’t change the current UI much.

The grid block toolbar is converted to a quanta toolbar “lite” that follows the block that is selected
The key point is that it’s very clear what block is selected and the toolbar is contextual to the selected block. You aren’t showing a toolbar for the grid while you are working on a sub-block.

The toolbar doesn’t need slate merged in with yet or a context menu. Those can be done later. It doesn’t need the settings button as selecting the block will automatically do that. It just needs an add and delete and the  “add” will always “add after” the current selection so the user experience is consistent.

selectgrid

- “add” will add a block below this grid block

selecttext


- “add" will insert a new block after this text block. This is the most common thing you will want to do.
- I’ve left the outer containers having a border (but visually different) as you seem to prefer that (but I‘m personally not sure it adds much). 

selectcol


- “add” will insert a new column in the middle. (but you could also do this in the settings of the grid)

Because this is WYSIWYG and because containers might not have much padding to allow them to be clicked on superate from their contents there needs to be at least one other way of selecting a parent other than clicking. Gutenburg has 4 or 5 ways. We only need one to begin with. Again it doesn’t have to be perfect now.

You didn’t like having breadcrumbs in the side bar even though technically there is no reason it can’t be there but how about this instead?

Screen Shot 2023-06-27 at 8 30 36 pm


but the gutenberg solution at the bottom of the screen is also good and could be added later (so can it’s outline view)

Screen Shot 2023-06-27 at 8 32 36 pm


If you think the method to get to the parent is still too subtle another idea is the quanta toolbar could include a button that moves the selection to the parent. A kind of “back to parent” button. I’ve just noticed volto-columns-block has exactly this.
So for grid scenarios you get something like this (not sure on the icon yet).

selectcol_up


The add inserts a new column in the middle and “back” will select the parent grid block so you can adjust its settings. 

You will probably say the add button doesn’t make it clear where the insert will take place even if it is consistent and I agree with you. But there is no reason another overlay can’t be used to show where it will be added if desired.

selectcol_up_pos

And clicking “back” we get to the grid block and can then add a new block to the page.

selectgrid_up_pos

As you may have realised, the page is now a container so there is no inconsistency for how blocks work within the page vs how they work inside containers.

PR #3187, opened 2022-03-21
fixes in sitemap.xml.gz

mamico:
* generateSitemap use apiPath instead of preferring internalApiPath when it exists
* sitemap middleware returns wrong headers when errors occur in the sitemap generation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,8 @@
 
 - Fix addon registry regression @sneridagh
 - Fix `Bosnian` language @avoinea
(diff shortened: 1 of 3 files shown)


sneridagh:
@mamico `internalApiPath` is meant to be used only in docker deployments, when the SSR node server needs to reach by its own (not through the public URL) the backend, normally using docker link names. Thus, in these cases, this change would be wrong... (eg. URLs with backend:8080/bla)

sneridagh:
@mamico uh, then afterwards it DOES use `toPublicURL` to fix the URLs! sorry about that!

tiberiuichim (approved):

PR #3194, opened 2022-03-25
Remove clean-tmp-worktrees from Makefile

tisto:
This break `Makefile` compat for MacOS users, so until is properly adapted, it's worth removing so people does not have to mess up with their environments.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -78,29 +78,7 @@ dist:
 .PHONY: test
 test:
 	$(MAKE) -C "./api/" test
-.PHONY: test-clean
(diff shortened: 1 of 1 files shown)


sneridagh (approved):

rpatterson:
> This break `Makefile` compat for MacOS users, so until is properly adapted, it's worth removing so people does not have to mess up with their environments.

Please provide the details of how this breaks on OS X, @tisto, so that another contributor (me in this case) has at least a chance to fix it without having to fire up an OS X VM.  Thanks!

PR #3197, opened 2022-03-27
Fix/Improve the console logging when the server starts

sneridagh:
image

How about this?


src/start-server.js
```diff
@@ -27,7 +27,10 @@ export default () => {
         console.log(
           `Using internal proxy: ${app.publicURL} -> ${app.devProxyToApiPath}`,
         );
-      console.log(`🎭 Volto started at ${app.publicURL} 🚀`);
+      if (!process.env.RAZZLE_PUBLIC_URL)
+        console.log(`Current guessed public URL: ${app.publicURL}`);
```
tiberiuichim, line 31: I'm not a fan of "guessed", but I can't really come up with a better suggestion. I think the fact that "things are tricky/complicated here" is a smell.

sneridagh: Just `Current public URL`? No info at all? We could put additional info under a `debug` lib flag.

sneridagh: done in the last commits

sneridagh:
image

tiberiuichim (approved):

Issue #3214, opened 2022-03-31
Array widget allows keywords with same sequence of letters just different lower and upper case letters.


ksuess:
**Describe the bug**
It's not possible to add two keywords with same sequence of letters.

**To Reproduce**
Try to type in 'Beb' and 'BEB' in a keyword field. It's not possible.

![grafik](https://user-images.githubusercontent.com/1144737/160986586-9a177d9a-79a9-421a-8fe3-ae8af0f5a8b7.png)

Volto 15.2.2

PR #3215, opened 2022-03-31
Overwrite isValidNewOption of ArrayWidget. Ignore lower and upper case.

ksuess:
Fix https://github.com/plone/volto/issues/3214

https://github.com/JedWatson/react-select/blob/b0411ff46bc1ecf45d9bca4fb58fbce1e57f847c/packages/react-select/src/Creatable.js#L35-L37
https://react-select.com/props#creatable-props

Hi @giuliaghisini and @tiberiuichim would you please have a look on what is needed to restrict new keywords. You've seen way more use cases than me. My use case is variants of glossary terms, which obviously come along often as for example 'BEB' and 'BeB'.
It's for sure not OK to overwrite with a brutal `()=>true`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -81,6 +81,7 @@
 - Pass placeholder and isDisabled properties to EmailWidget and UrlWidget @mihaislobozeanu
 - Pass placeholder property to PasswordWidget and NumberWidget @mihaislobozeanu
 - Fix getVocabName when vocabNameOrURL is false @avoinea #2955, #2919
+- Overwrite isValidNewOption of ArrayWidget to allow variants @ksuess
 
 ### Internal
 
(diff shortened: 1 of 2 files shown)


tiberiuichim:
At first glance, there would be an edge case with defaulting to true, if the new option is already present as a value, would it cause duplicated entries?

ksuess:
No. I didn't dig in further, but that was my first UI test. It does NOT allow duplicates, definitely not, at least

tiberiuichim:
Then I guess we're good to go, right?

ksuess:
Well, my gut feeling is not optimal. On the other side, it's easy to correct later, as the prop 'isValidNewOption' has a speaking name.

tiberiuichim (approved):

giuliaghisini (commented):
I'm not totally sure to return true, 
this means that the check for existing options will not done, and the 'Add new option...' options is always present even if option already exists..
Default here:
https://github.com/JedWatson/react-select/blob/cfd5d9f632b4a8b699fba1028d3243735428e428/packages/react-select/src/useCreatable.ts#L83

ksuess:
Instead of a brute force 'all valid', 
now isValidNewOption like default, but comparing with considering upper and lower case.

giuliaghisini (approved):

ksuess:
@sneridagh, this seems to be ready.

sneridagh (approved):

PR #3223, opened 2022-04-03
Replace `connected-react-router` with `redux-first-history` (drop-in …

sneridagh:
…replacement, compatible with RR6)

This is first step for adopt RR6.


tiberiuichim (approved):

sneridagh:
It does not work well with RR6. I'd vote to remove the connected router feature in next Volto releases (when we upgrade to RR6). However, that will involve to remove AsyncConnect as well, since it kind of relies on it (on one single line).

Closing for now.

PR #3224, opened 2022-04-04
Make the parseDateTime function to handle only date also

iFlameing:


sneridagh (approved):

iFlameing:
@tiberiuichim @robgietema can you please review this?

src/helpers/Utils/Utils.test.js
```diff
@@ -348,6 +348,12 @@ describe('Utils tests', () => {
         parseDateTime('de', isoDate, undefined, moment).toISOString(),
       ).toBe(`${isoDate}Z`);
     });
+    it('Parses the Date only', () => {
+      const isoDate = '2022-01-16';
```
robdayz, line 352: Maybe is better to check against another date format, not only the expected (to check if regexp is working fine)...  like 2022-04-11T18:06:47.220571

Issue #3232, opened 2022-04-07
External link type object not supported by ContextNavigation component


UnaiEtxaburu:
**Describe the bug**
If an object is an external link, when using ContextNavigation block it redirects to localhost:3000/{external link}

**Expected behavior**
If the object is an external link, ContextNavigation should redirect to that link directly without starting with "localhost"

**Screenshots**
![imagen](https://user-images.githubusercontent.com/80053871/162147586-a6dcf502-5b83-4b34-ba2b-c3bdffaf1c8a.png)

PR #3234, opened 2022-04-07
ContextNavigatin checks if object is link or not

UnaiEtxaburu:
Fixes issue #3232


src/components/theme/Navigation/ContextNavigation.jsx
```diff
@@ -46,6 +47,9 @@ function renderNode(node, parentLevel) {
             ''
           )}
         </RouterLink>
+         ) : (
+          <a href={flattenToAppURL(node.href)}>{node.title}</a>
```
tiberiuichim, line 51: I wonder if UniversalLink would be a better fit. Though would be fine like this, as well.

tiberiuichim (approved):

src/components/theme/Navigation/ContextNavigation.jsx
UnaiEtxaburu: I changed the <a> tag with <UniversalLink> tag because it gives more options

PR #3236, opened 2022-04-07
Fix handling of single reference.

robgietema:


tiberiuichim (approved):
As far as I'm aware, the "single" mode is a new introduction to the ObjectBrowserWidget API (which I very much support). 

Maybe it would be a good idea to document this in the ObjectBrowserWidget stories: https://6.dev-docs.plone.org/storybook/?path=/story/widgets-object-browser--connected&globals=measureEnabled:false

Also, change the stories to show the real value, not the `@id` of the item that was picked.

src/components/manage/Form/Field.jsx
```diff
@@ -83,6 +84,30 @@ const getWidgetPropsFromTaggedValues = (widgetOptions) =>
...
+        "selectableTypes": [
+            "Image"
+        ]
+    })
+
+ */
+const getWidgetPropsFromPatternOptions = (widgetOptions) =>
+  isArray(widgetOptions?.pattern_options?.selectableTypes) &&
+  intersection(
+    widgetOptions.pattern_options.selectableType,
```
sneridagh, line 105: we use `frontend_options`, `pattern_options` is for classic. They *must* be different, we cannot assume to use the same. Plone 6 should stick to one, and `frontend_options` is the most sensible one.

So -1 to support this in Volto. Why do you need it?

src/components/manage/Form/Field.jsx
```diff
@@ -155,9 +177,9 @@ const UnconnectedField = (props, { intl }) => {
     getWidgetByFieldId(props.id) ||
     getWidgetFromTaggedValues(props.widgetOptions) ||
     getWidgetByName(props.widget) ||
-    getWidgetByChoices(props) ||
     getWidgetByVocabulary(props.vocabulary) ||
     getWidgetByVocabularyFromHint(props) ||
+    getWidgetByChoices(props) ||
```
sneridagh, line 182: This, unfortunately, it's a breaking change. What is the reason behind?

src/components/manage/Form/Field.jsx
```diff
@@ -170,6 +192,7 @@ const UnconnectedField = (props, { intl }) => {
   const widgetProps = {
     ...props,
     ...getWidgetPropsFromTaggedValues(props.widgetOptions),
+    ...getWidgetPropsFromPatternOptions(props.widgetOptions),
```
sneridagh, line 195: Same, I think we should not support that.

sneridagh (changes requested):
+1 to @tiberiuichim comment, we need docs about the new prop, also update storybook stories and add a storybook for the fixed widget.

src/components/manage/Form/Field.jsx
robgietema: That is how fields are defined currently, so that means the widget will not activate for images if you define them this way.

src/components/manage/Form/Field.jsx
robgietema: Check the getWidgetByChoices implementation; it checks if there is a vocabulary and defaults to a choices widget, even if the vocabulary is the catalog one. So that is why the vocabulary should be checked first.

src/components/manage/Form/Field.jsx
sneridagh: I'll stick to `frontend_options`, I'll open the Pandora's box on this one, before it's too late.

src/components/manage/Form/Field.jsx
sneridagh: Let's do it then, seems safe.

sneridagh (approved):

PR #3255, opened 2022-04-20
Docs: Add selectableTypes example to ObjectBrowserWidget storybook

### LINKED ISSUE #2993: More options for the object browser's selectable content types
The object browser currently only has one way to filter the selectable content types is to set the mode to 'image', which forces you to select an image. The object browser's image mode currently works by checking `settings.imageObjects` for a list of allowed image objects. It would be useful if it were possible to apply similar restrictions to other content types, such as for use in the video block or for a file picker.

## Options
- Have various modes to support a number of various out-the-box modes.
  - This would not be useful for custom content-types unless the list of modes was adjustable, such as through a config setting.
- Have the user pass in the short name of allowed content-types to `openObjectBrowser`.

JeffersonBledsoe:
Just a quick PR to add an example of the `selectableTypes` to the `ObjectBrowserWidget`'s storybook.

Closes #2993

### DIFF
--- a/docs/source/blocks/editcomponent.md
+++ b/docs/source/blocks/editcomponent.md
@@ -290,7 +290,7 @@ export const widgets = {
 
 #### Selectable types
 
-If **selectableTypes** is set in `widgetOptions.pattern_options`, then only items whose content type has a name that is defined in `widgetOptions.pattern_options.selectableTypes` will be selectable.
+If `selectableTypes` is set in `widgetOptions.pattern_options`, then only items whose content type has a name that is defined in `widgetOptions.pattern_options.selectableTypes` will be selectable.
 
 ```jsx
 <ObjectBrowserWidget ... widgetOptions={{pattern_options:{selectableTypes:['News Item','Event']}}}>
--- a/docs/source/blocks/editcomponent.md
+++ b/docs/source/blocks/editcomponent.md
@@ -290,7 +290,7 @@ export const widgets = {
 
 #### Selectable types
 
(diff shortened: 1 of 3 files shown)


docs/source/blocks/editcomponent.md
```diff
@@ -290,7 +290,7 @@ export const widgets = {
 
 #### Selectable types
 
-If **selectableTypes** is set in _widgetOptions.pattern_options_, widget allows to select only items that matches types defined in _widgetOptions.pattern_options.selectableTypes_.
+If **selectableTypes** is set in `widgetOptions.pattern_options`, then only items whose content type has a name that is defined in `widgetOptions.pattern_options.selectableTypes` will be selectable.
```
stevepiercy, line 293: I think that "selectableTypes" is an inline literal, but if not then please ignore this suggestion. This content is good. Thank you!

```suggestion
If `selectableTypes` is set in `widgetOptions.pattern_options`, then only items whose content type has a name that is defined in `widgetOptions.pattern_options.selectableTypes` will be selectable.
```
```suggestion
If **selectableTypes** is set in `widgetOptions.pattern_options`, then only items whose content type has a name that is defined in `widgetOptions.pattern_options.selectableTypes` will be selectable.
```

stevepiercy (approved):
LGTM. Let's get a core maintainer to review as well before merging.

sneridagh (approved):

sneridagh:
@JeffersonBledsoe awesome! thanks!

PR #3268, opened 2022-04-24
[WIP] Move to a centralized DnD Context for blocks

sneridagh:
Looks good, but the fancy placeholder we put in place not long ago, it has to be passed down to the Droppables... I haven't figured out how to do that... since bDnD uses render functions, you can't pass down the `placeholderProps` to it, for the placeholder to compute the size...

Worth to investigate during the sprint.


davisagli:
react-beautiful-dnd is no longer maintained, so we need a different approach

PR #3269, opened 2022-04-25
Core Blocks Style Wrapper

sneridagh:
Initial implementation, subject to change.

Please, feedback is needed!


docs/source/blocks/blockStyleWrapper.md
```diff
@@ -0,0 +1,62 @@
...
+  "keywords": "Volto, Plone, frontend, React, Blocks, Edit, Style, Wrapper, components"
+---
+
+# Blocks - Style Wrapper
+
+The block style wrapper part of a block anatomy, allows you to inject styles from the block schema into the block wrapper in form of class names. It's wrapping the block edit and the view components.
+
+## Enabling Style Wrapper in a block
+
+The wrapper is always present in the render of the view and the edit components, but if you want to add styles, you have to enhance the schema of your block:
```
stevepiercy, line 15:
```suggestion
The wrapper is always present in the render of the view and the edit components.
If you want to add styles, you have to enhance the schema of your block:
```

docs/source/blocks/blockStyleWrapper.md
```diff
@@ -0,0 +1,62 @@
...
+  // (in your block config object)
+  my_custom_block: {
+    // (more block settings)
+    schemaEnhancer: styleSchemaEnhancer,
+  }
+```
+
+This will add a new fieldset `Styling` to your block schema settings.
+By default, only a `backgroundColor` property is set, configured by: `defaultSchema` in `src/components/manage/Blocks/Block/StylesSchema.jsx`.
+The schema can be overriden, using the `stylesSchema` key in your block config object, like:
```
stevepiercy, line 29:
```suggestion
The schema can be overridden using the `stylesSchema` key in your block config object as follows:
```

docs/source/blocks/blockStyleWrapper.md
```diff
@@ -0,0 +1,62 @@
...
+    "backgroundColor": "#ee22ee",
+    "myCustomStyleField": "red",
+    "myCustom2StyleField": {
+      "color": "black",
+      "gradient": "MyGradient"
+  }
+}
+```
+
+The resultant class names injected into the Style Wrapper are built from this `style` field and its values, like:
```
stevepiercy, line 56:
```suggestion
The resultant class names injected into the Style Wrapper are built from this `style` field and its values as follows:
```

docs/source/blocks/blockStyleWrapper.md
```diff
@@ -0,0 +1,62 @@
...
+
+```
```
stevepiercy, line 58:
```suggestion
```html
```

docs/source/blocks/blockStyleWrapper.md
```diff
@@ -0,0 +1,62 @@
...
+<div className="has--backgroundColor--ee22ee has--myCustomStyleField--red has--myCustom2StyleField--color--black has--myCustom2StyleField--color--MyGradient">
+```
+
+Then, it's at your discretion how you define this CSS class names in your theme.
```
stevepiercy, line 62:
```suggestion
Then it's at your discretion how you define the CSS class names in your theme.
```

docs/source/blocks/index.md
```diff
@@ -15,6 +15,7 @@ introduction
 anatomy
 settings
 editcomponent
+blockStyleWrapper
```
stevepiercy, line 18: Let's use lower-case only. Also consider using dashes for SEO. Also rename the file to which this refers.

```suggestion
block-style-wrapper
```

stevepiercy (changes requested):
I reviewed only the docs part, and made suggestions.

src/helpers/Blocks/Blocks.js
```diff
@@ -405,3 +405,33 @@ export function applyBlockDefaults({ data, intl, ...rest }, blocksConfig) {
 
   return applySchemaDefaults({ data, schema });
 }
+
+export const buildStyleClassNamesFromData = (styles) => {
```
tiberiuichim, line 409: awesome!

src/components/manage/Widgets/StyleWidget.jsx
```diff
@@ -0,0 +1,17 @@
+import ObjectWidget from '@plone/volto/components/manage/Widgets/ObjectWidget';
+
+const StyleWidget = (props) => {
+  const { block, onChange, schema } = props;
+
+  return (
+    <div className="styles-widget">
```
tiberiuichim, line 7: Is there any chance to avoid this custom widget? If the classname is absolutely needed, maybe we can get the ObjectWidget to output a classname with the name of the field?

src/components/manage/Blocks/Block/StyleWrapper.jsx
```diff
@@ -0,0 +1,16 @@
+import cx from 'classnames';
+import { Container } from 'semantic-ui-react';
+import { buildStyleClassNamesFromData } from '@plone/volto/helpers';
+
+const StyleWrapper = ({ children, data }) => {
+  const styles = buildStyleClassNamesFromData(data?.styles);
+
+  return (
+    <div className={cx(styles)}>
```
tiberiuichim, line 9: We're missing "native integration", we don't want to wrap `<p>` in divs. Also, I don't think it's safe to use `<Container>` here, or we need, at least, a way to tell if we're already in a container or not. A "contextual container", something that provides a react context so that further down we know if we can use the container or not.

The style wrapper could be a primitive that could also be used for "other things", columns, sections of a block, etc.

pnicolli: At a first glance, I wouldn't add a Container either way here. It's not the purpose of this component, am I right?

pnicolli:
Nice addition. It this planned for Volto 16?
Imho as of today this is a potential breaking change, it changes the dom structure of pages.

docs/source/blocks/blockStyleWrapper.md
```diff
@@ -53,10 +56,10 @@ The `style` field is mapped to an `objectWidget`, and the `stylesSchema` adds th
 }
```
stevepiercy, line 56: I just noticed that this block is missing one `}`, which is why GitHub highlights the closing three backticks and the rest of the content in red.

Also the second to last `}` needs to be indented properly.

stevepiercy (changes requested):
I missed an issue with a code block.

Also the file `docs/source/blocks/blockStyleWrapper.md` needs to be renamed to `docs/source/blocks/block-style-wrapper.md`.

giuliaghisini:
I'm not very keen on adding a wrapper to existing blocks. 
I think it is better to add the class to the existing block wrappers, for example 
`<div class = "block image bg-gray full-width"> ... </div>`
or use the wrapper inside the block, for example:
```
<div class = "block image"> 
      <div class = "bg-gray full-width> </div> 
</div>
```

Even the forced use of the Container does not drive me crazy. There are several situations where you don't want to use the Container at that point (for example if in a full-width block I want to use an image as a background, etc.)

In RedTurtle we have already developed something similar that could be inspired by, and we have also implemented a widget for choosing colors from a predefined list. It would be nice to also introduce this widget together with the ColorPicker, in order to give two possible solutions to the developers based on the freedom they want to give to the editors.

We could take a look during the sprint.
For example this is what we do:

https://user-images.githubusercontent.com/51911425/165303778-297093d6-ea15-4970-946a-73479b44da75.mov

sneridagh:
@plone/volto-team 

We should discuss this matters during the sprint:

- Force the inner `Container` in the wrapper (as @pnicolli hinted) might be not required, but only if we change how we do things in the default layout. It would imply to remove the main `Container` in the `#page-document`, `#page-edit`, `#page-add` (main) block containers, so we control the flow in the block itself. BUT, that is a breaking.

So do we want to issue that kind of breaking change just before Plone 6 is shipped? Is it worth it? Add-ons and themes would have to adjust. Granted, it won't be a big one, but still...

- Should this include then a full re-work of the layout in core already? We were pushing for this kind of layout in our current and latests projects:

Frame 10

It follows the latest trends in websites, and it's quite appealing for our clients.

sneridagh:
We have to list all the use cases that we want to address then find if we can spare the container (I think we do still have it for some reason).

sneridagh:
@stevepiercy thanks for reviewing! I'll take a look at the rest of the things, the docs are still in flux and still need more love.

giuliaghisini:
> @plone/volto-team
> 
> We should discuss this matters during the sprint:
> 
> * Force the inner `Container` in the wrapper (as @pnicolli hinted) might be not required, but only if we change how we do things in the default layout. It would imply to remove the main `Container` in the `#page-document`, `#page-edit`, `#page-add` (main) block containers, so we control the flow in the block itself. BUT, that is a breaking.
> 
> So do we want to issue that kind of breaking change just before Plone 6 is shipped? Is it worth it? Add-ons and themes would have to adjust. Granted, it won't be a big one, but still...
> 
> * Should this include then a full re-work of the layout in core already? We were pushing for this kind of layout in our current and latests projects:
> 
> Frame 10
> 
> It follows the latest trends in websites, and it's quite appealing for our clients.

i think this will be a breaking change that needs to review all the previeous project when upgrading volto version..
Maybe, it will be an option or a template for the block 'Title' (that could become a combination of title and leadimage block', what do you think?

tiberiuichim:
I'm +1 to introduce the new layout for the Plone 6 release, to be worked as part of the Bethoven sprint. It is a major step forward to making Volto more flexible and easier to understand and control.

If needed, we can think about methods to keep backwards compatibility. This problem can be solved in the layout view component and the block style wrapper.

sneridagh:
@giuliaghisini oh, no. Sorry I did not express myself correctly. Ignore the first block, just take a look at the overall block widths and their disposition, depending on if they are floating or now. Also, note the "default" text block width is narrower, so we have this full/normal/narrow container types for blocks.

src/components/manage/Widgets/StyleWidget.jsx
sneridagh: Done

src/components/manage/Blocks/Block/StyleWrapper.jsx
sneridagh: Removed container.

docs/source/blocks/blockStyleWrapper.md
sneridagh: Fixed.

sneridagh:
@stevepiercy I amended a bit the docs, if you want to have a final look.

sneridagh:
@plone/volto-team I refactored it today, following all the feedback received, thanks a lot!
I think this is generic enough, and non breaking to merge it.

However, I'd wait since @giuliaghisini told me about the awesome RT style/color picker. Is it open sourced yet?

Also, we can work on the layout once this is in place.

If we merge it asap, let's issue a 16 alpha right away and let's maintain it during the sprint and the aftermath.

sneridagh:
Also: tempted to add the `'block blockType'` automatically too, since we do this all the time. Opinions?

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@
 
 ### Feature
 
+- Added new Block Style Wrapper. This implementation is marked as **experimental** during Volto 16 alpha period. The components, API and the styling are subject to change **without issuing a breaking change**. You can start using it in your projects and add-ons, but taking this into account. See documentation for more information. @sneridagh
```
stevepiercy, line 9: Is this still experimental, now that it is a non-breaking change?

docs/source/blocks/block-style-wrapper.md
```diff
@@ -0,0 +1,99 @@
+---
+html_meta:
+  "description": "The block style wrapper part of a block anatomy, allows you to inject styles from the block schema into the block wrapper in form of class names."
```
stevepiercy, line 3:
```suggestion
  "description": "The block style wrapper is part of a block anatomy. It allows you to inject styles from the block schema into the block wrapper in the form of class names."
```

docs/source/blocks/block-style-wrapper.md
```diff
@@ -0,0 +1,99 @@
...
+  "property=og:description": "The block style wrapper part of a block anatomy, allows you to inject styles from the block schema into the block wrapper in form of class names."
```
stevepiercy, line 4:
```suggestion
  "property=og:description": "The block style wrapper is part of a block anatomy. It allows you to inject styles from the block schema into the block wrapper in the form of class names."
```

docs/source/blocks/block-style-wrapper.md
```diff
@@ -0,0 +1,99 @@
...
+  "keywords": "Volto, Plone, frontend, React, Blocks, Edit, Style, Wrapper, components"
+---
+
+# Blocks - Style Wrapper
+
+The block style wrapper is part of a block anatomy.
+It allows you to inject styles from the block schema into the block wrapper in the form of class names.
+It wraps the block edit and the view components.
+
+It can be used mapping directly unique values to CSS properties, like `background-color` mapped to a single color.
```
stevepiercy, line 15:
```suggestion
It can be used by directly mapping unique values to CSS properties.
For example, `background-color` could be mapped to a single color.
```

docs/source/blocks/block-style-wrapper.md
```diff
@@ -0,0 +1,99 @@
...
+However, the real intention is to bring the ability to create a way to define complex CSS and a full set of styles mapped to a CSS class.
```
stevepiercy, line 16:
```suggestion
Although that is a simple use case, the real intent of style wrappers is to allow the definition of complex CSS through a full set of styles mapped to a CSS class.
```

docs/source/blocks/block-style-wrapper.md
```diff
@@ -0,0 +1,99 @@
...
+For example, very often, applying a background color is not enough, and you need to modify also the font color as well, in order to make it readable because of contrast issues.
```
stevepiercy, line 17:
```suggestion
For example, often applying a background color is not enough, and you need to also modify the font color to make it readable because of contrast issues.
```

docs/source/blocks/block-style-wrapper.md
```diff
@@ -0,0 +1,99 @@
...
+The style object field can be extended to hold any number of fields, depending on your needs.
+The resultant set of class names to be injected are built concatenating the key-value pairs of the `style` object field, separated by `--` and prefixing the result with the `has--` prefix.
```
stevepiercy, line 19:
```suggestion
The resultant set of class names to be injected are built by concatenating the key-value pairs of the `style` object field, separated by `--` and prefixing the result with the `has--` prefix.
```

docs/source/blocks/block-style-wrapper.md
```diff
@@ -0,0 +1,99 @@
...
+The class name generator algorithm supports up to one level of nested objects, constructing the concatenated CSS class from the nested structure as well.
+See below for an example.
+
+## Enabling Style Wrapper in a block
+
+The wrapper is always present in the render of the view and the edit components.
```
stevepiercy, line 25: This is a little unclear. Is the intent the following?

"The wrapper is always present in the rendering of both the view and edit components."

docs/source/blocks/block-style-wrapper.md
```diff
@@ -0,0 +1,99 @@
...
+If you want to add the default set of styles, you need to enable with the following flag:
```
stevepiercy, line 26:
```suggestion
If you want to add the default set of styles, you need to enable them with the following flag:
```

docs/source/blocks/block-style-wrapper.md
```diff
@@ -0,0 +1,99 @@
...
+```js
+  // (in your block config object)
+  my_custom_block: {
+    // (more block settings)
+    stylesEnabled: true,
+  }
+```
+
+```{note}
+This will work if your block uses `BlocksForm` component to define an schema driven block configuration settings.
```
stevepiercy, line 37:
```suggestion
This will work if your block uses the `BlocksForm` component to define schema-driven block configuration settings.
```

docs/source/blocks/block-style-wrapper.md
```diff
@@ -0,0 +1,99 @@
...
+```
+
+This will add a new fieldset `Styling` to your block schema settings.
+By default, only a `backgroundColor` property is set, configured by: `defaultSchema` in `src/components/manage/Blocks/Block/StylesSchema.jsx`.
```
stevepiercy, line 41:
```suggestion
By default, only a `backgroundColor` property is set, configured by `defaultSchema` in `src/components/manage/Blocks/Block/StylesSchema.jsx`.
```

docs/source/blocks/block-style-wrapper.md
```diff
@@ -0,0 +1,99 @@
...
+
+## Extending the default styling field
+
+You can modify the default set of styles, by using a `schemaEnhancer` function, in the same way that you would do it for any block schema enhancer.
```
stevepiercy, line 45:
```suggestion
You can modify the default set of styles by using a `schemaEnhancer` function in the same way that you would for any block schema enhancer.
```

docs/source/blocks/block-style-wrapper.md
```diff
@@ -0,0 +1,99 @@
...
+Use the `stylesSchema` key in your block config object as follows:
```
stevepiercy, line 46:
```suggestion
Use the `stylesSchema` key in your block configuration object as follows:
```

docs/source/blocks/block-style-wrapper.md
```diff
@@ -0,0 +1,99 @@
...
+  }
+```
+
+```{note}
+The signature for a `schemaEnhancer` is `({schema, formData, intl})`.
+```
+
+## The `styles` field
+
+The `style` field is mapped to an `objectWidget`, and the `stylesSchema` adds the fields into this field, creating an object that is the sum of all of the widgets assigned to it, and its values.
```
stevepiercy, line 65: The use of `style` here does not align with the preceding heading's use of `styles`. Is it singular or plural? Please make sure they match.

```suggestion
The `style` field is mapped to an `objectWidget`.
The `stylesSchema` adds the fields into this field, creating an object that is the sum of all of the widgets assigned to it and its values.
```

docs/source/blocks/block-style-wrapper.md
```diff
@@ -0,0 +1,99 @@
...
+    "myCustomStyleField": "red",
+    "myCustom2StyleField": {
+      "color": "black",
+      "gradient": "MyGradient"
+    }
+  }
+}
+```
+
+## Using the classNames in your block
```
stevepiercy, line 80: The plural or singular form of this should align with the subsequent content.

```suggestion
## Using `className` in your block
```

docs/source/blocks/block-style-wrapper.md
```diff
@@ -0,0 +1,99 @@
...
+
+The resultant class names injected as a `className` prop into the wrapped Block. So you can use it in the root component of your block view and edit components, as follows:
```
stevepiercy, line 82:
```suggestion
The resultant class names are injected as a `className` prop into the wrapped block.
Thus you can use it in the root component of your block view and edit components as follows:
```

docs/source/blocks/block-style-wrapper.md
```diff
@@ -0,0 +1,99 @@
...
+```jsx
+const BlockView = (props)=> (
+  <div className={props.className}>
+    // Block's code
+  </div>
+)
+```
+
+Same for the block edit component.
+The resultant HTML would be like:
```
stevepiercy, line 93:
```suggestion
The resultant HTML would be the following:
```

stevepiercy (changes requested):
This is looking great! I've added my review of the docs portion only.

giuliaghisini (approved):

ksuess:
I do not understand the pros of this concept.

An editor selects a background color 'orange'.
The block gets the class 'has--backgroundColor--orange'
Only if the developer provides a CSS rule for this class, something is styled.
As soon as the editor changes to another color. There is no styling as long as the developer does provide a CSS rule for the new color.

Another topic: presets.
I'm using Eau de Webs block styler with presets: three sets of CSS rules (background color, padding, font color)
That's to offer the editor just three options and not a full bunch of colors to ruin the layout.
edw block styler adds a class per selected preset. That's making live easy for the editor (he does not have to be creative, he selects from three options), also for the developer (if the client wants highlighted blocks green instead of blue: it's just a change in config.js.

tiberiuichim:
@ksuess One advantage of using explicit classes for colors is that they're "portable". If the website look changes, then you can adjust the class styling. The eeacms/volto-block-style is too blunt in its approach.

@sneridagh Maybe if we have the CSS props included, it would make more sense on why this approach was chosen?

ksuess:
> @ksuess One advantage of using explicit classes for colors is that they're "portable". If the website look changes, then you can adjust the class styling. The eeacms/volto-block-style is too blunt in its approach.
> 

Yes, I see the abstraction layer. That's why I like to offer just a set of options to the edior.

sneridagh:
@ksuess the color chooser is still in progress. It should be configurable, per block to allow you only add the presets per project for that block. 

I'll continue working on this during the next days, for now, I'd like to know what do you think about the flexibility that the foundations of this has. The idea is to have presets in core as well, so we have a set of well known set of styles (extensible) by default.

Do you think that it's valuable to set the CSS directly in the style? I think that mapping values in the settings directly to real CSS definitions is very limited, and will end in a dozen of properties in there, that only power users know their purpose. I think that map that to abstract styling is better, so you can (potentially) have control over other properties, depending on the preset. That's why I think that the style preset is better over mapping CSS properties alone. 

Of course, the wrapper allows you to map them in that way, if you feel that matches also your requirements. Using classes, makes them portable too (and configurable)... using custom CSS properties, like:

```
@headerFont: var(--custom-headers-font, @pageFont);
```

Then customise per project the CSS prop, without touching the CSS class definitions.

ksuess:
I assume that everyone wants both or at least the first:
1. presets per project (and additional to edw's styler: per block type)
2. individual fine granular options for the editor

I personally try to convince the customer to offer only the first.

1. 
image
2. 
image

Your increased configurability with a schema enhancement is of course welcome.
If understand right it would look like
```
  config.blocks = {
    ...config.blocks,
    blocksConfig: {
      ...config.blocks.blocksConfig,

      slate: {
        // (more block settings)
        ...config.blocks.blocksConfig.slate,
        stylesEnabled: true,
        stylesSchema: myCustomStyleSchema,
        availableColors: ['orange', 'green'],
        availableFonts: ['foo', 'bar'],
        availableSomething: ['trallalala']
      },
```
depending on the custom schema 'myCustomStyleSchema'.

I'm just puzzled how the CSS rules can be written for this big bunch of classes that are generated by the individual options (2. case above) for
```html
<div className="has--backgroundColor--ee22ee has--myCustomStyleField--red has--myCustom2StyleField--color--black has--myCustom2StyleField--color--MyGradient">
```
But I understand that this abstraction is need to achieve an abstraction that allows to change the mapping later without fiddling with explicit CSS rules in blog data.

CHANGELOG.md
sneridagh: This is an unmentioned fact, every feature added in the alpha period IS in fact experimental. I initially thought that it might be interesting to really note it and being explicit in this case, instead of being implicit... But it's true that kind of lead to confusion now...

docs/source/blocks/block-style-wrapper.md
sneridagh: yep

docs/source/blocks/block-style-wrapper.md
stevepiercy: The heading uses `styles` (plural), and the paragraph uses `style` (singular). Please use one or the other for both.

docs/source/blocks/block-style-wrapper.md
```diff
@@ -62,7 +63,8 @@ The signature for a `schemaEnhancer` is `({schema, formData, intl})`.
 
 ## The `styles` field
 
-The `style` field is mapped to an `objectWidget`, and the `stylesSchema` adds the fields into this field, creating an object that is the sum of all of the widgets assigned to it, and its values.
+The `style` field is mapped to an `objectWidget`.
```
stevepiercy, line 66: The heading above uses `styles` (plural), and the paragraph uses `style` (singular). Please use one or the other for both.

stevepiercy (changes requested):
Missed one. Once that is fixed, I approve. Thank you!

ichim-david:
> I'll continue working on this during the next days, for now, I'd like to know what do you think about the flexibility that the foundations of this has. The idea is to have presets in core as well, so we have a set of well known set of styles (extensible) by default.
> 
> Do you think that it's valuable to set the CSS directly in the style? I think that mapping values in the settings directly to real CSS definitions is very limited, and will end in a dozen of properties in there, that only power users know their purpose. I think that map that to abstract styling is better, so you can (potentially) have control over other properties, depending on the preset. That's why I think that the style preset is better over mapping CSS properties alone.
> 
I am aware of what **volto-block-style** does with style tags. One advantage that it has over the class approach is the specificity gains from inline styles vs potentially a single class styled which get's overridden by more specific styles.
Although this is a nice thing to have, you could potentially have changes to the class style if the styles were added in a class.

I don't think the current mapping approach is a good idea because it sets up the class name for failure anytime there is a change in the actual color value.
"has--backgroundColor--AABBCC" 
If you decide to tweak the color slightly that class will then have a false naming.
To me it seems Victor that you want to create a sort of object to CSS generation but instead of having a class created with all of the properties you have several classes named as inline style plus values, at least from the tests and current demo.
If we go in the abstraction domain where you don't add an inline style but you get a class that you need to add your style then it would be better to name them different from the inline style values.

If the idea is to move the inline styles to classes then I think that on yarn build there should be a step that goes through the config, reads these style objects and appends these classes to a CSS file with a configurable location.

This way if you have
const styles = {
        color: 'red',
        backgroundColor: '#AABBCC',
      };
and you output has-color-red and has-backgroundColor-AABBCC
You don't need to add by hand these classes as they are provided for you when you build the site.
When you have a change to the styles and  you rebuild the file is reconstructed.
This will fail though if the value changes as you will be left with orphan classes on the previously set blocks?

sneridagh:
@ichim-david 

I apologise for not having time to develop and properly write down the reasoning behind this approach before asking your feedback. I'll try to do it now, and continue the discussion during the sprint, because this always was the original idea.

# Top-Down styling approach vs editor's per block instance free will

One of the selling points in Volto and the idea behind its UX was always to lead the user/editor to a easy, safe path, that will result into styling consistency and unified look and feel across the site. This is true for everything UX related in Volto, eg. we do not simply put in front of the editor a myriad of choices and let them the free will to combine them to reach to a random result each an every single time the user configures a block.

We could have a system that allow user to choose between a number of styling properties mapped directly to CSS inline styles. Elementor does that for WordPress... and IMHO the result from the user point of view is a mess. The average editor does not have a clue what a `padding` is, or how it works and relates to the items around, just to start. Then, the styles that you apply in one block are hardly reproducible in others, after days, weeks, which style did you set in the other section sliders?

I know that this could sound useful for power users, and one could say: hey, if the client wants just this, let's give it to them. That's fine, but it's not Volto's USP. That's probably why `volto-style-block` is how it is, so I wanted to give the option for the foundation of this to allow it, that's why the shape of the `styles` fields is how it is, so you can enhance it later and bend it to your needs.

I think that we should stick to the Top-Down approach for Volto, where you have a set of variations over a block, and you get that from your site's designer. We must sell that as a Volto's feature. And yes, force our clients to think into this direction and make it one of our selling points.

The idea is that the user/editor, in fact, is not going to (or rather, should not to) have this free will. So the choices will be given, and the model behind it only can be set with classes. More on this below.

# Enter the block presets

So, ideally, you get a list and a mockup of all the blocks that the client is going to need in their future site, along with their variants. Let's call this presets. (We can find a better wording, doh naming!)

These are the grid block (with teaser) variations (or presets) of our last project:

image

You can see the slight differences that should be applied in elements inside (fonts, images, links, shadows, hovers) that can't be expressed with inline styles over a parent element. There are also some variants that need different font colors (only some of them, so you can't generalise), because of contrast with the background. 

The `styles` object and classname generation also helps with modelling a set of classes for that. The first nesting level? For namespacing the presets. Maybe I overengineered this, and we don't need such an abstraction, but... Also, you can put the color name on the values of the keys, but I would not do that... I'd better name presets and variants. Time will tell if we should improve or refactor the idea.

For that, we need a widget so you can choose the presets and it's enough powerful to showcase them properly (doesn't have to be precise, only model them in a way that a user can easily map the preset to the actual resultant block variant. I liked a lot the RT chooser, but we can further improve it to match the needs.

# Default block presets

One could say, that depending of the size of the project you won't have the luxury of such a detailed mockup set. I'd propose to have a default good known presets for that, that are easy overridable. Problem #1 is that we don't have a mockup of them, neither a set of colors that PastanagaUI defined back in the day. I can ping Albert, or rather, we find a set of colors that could work by default. 

These defaults could be generic enough (using abstract naming) and driven by CSS properties, so one could customise only the CSS property on the project side, without any other amendment.

Also, maintain them may become daunting at some point. So I would initially not do so.

# Where to go from here

As said, the idea is to set the foundation for something bigger and flexible. I'm ok to enable the inline styling so all possible cases are covered. I envisioned the `styles` object because of that as well. 

By any means I would go the CSS in JS rabbit hole, but basic inline styling should be also possible as it is in `volto-style-block`, but not by default or used by vanilla Volto. We don't want another Elementor.

Any after-build option is not an option, let's keep this simple and transparent. The fact that you have to provide your CSS as preset is unavoidable, and a good thing.

sneridagh:
@plone/volto-team I think this is ready now. It sets the foundation for the future of block styling, and new layout. I'd say let's merge it since it's not a breaking change, let's start build upon it and get as much feedback as possible from testing it.

sneridagh:
@stevepiercy I added some paragraphs more to the docs :P

src/components/manage/Blocks/Block/StyleWrapper.jsx
```diff
@@ -0,0 +1,30 @@
...
+
+export default StyleWrapper;
+
+// For some reason the HOC alternative is not working, and ends in an infinite updateloop
+// const withStyleWrapper = (Component) => ({ ...props }) => {
+//   const { data } = props;
+//   const styles = buildStyleClassNamesFromData(data?.styles);
+
+//   return <Component {...props} className={cx(styles)} />;
+// };
```
pnicolli, line 30: I like the current implementation, I think this comment can be removed.

src/components/manage/Blocks/Listing/Edit.jsx
```diff
@@ -48,7 +48,7 @@ const Edit = React.memo(
         : intl.formatMessage(messages.items));
 
     return (
-      <>
+      <div className={props.className}>
```
pnicolli, line 51: If we wanted to be super strict, this would be a breaking change, so it would need to be tracked, but it would be ok to release it since this is going into 16 alpha.

src/components/manage/Sidebar/AlignBlock.jsx
```diff
@@ -20,19 +20,26 @@ const messages = defineMessages({
     id: 'Center',
     defaultMessage: 'Center',
   },
+  wide: {
+    id: 'Full',
+    defaultMessage: 'Full',
+  },
```
pnicolli, line 26: These need a different id and defaultMessage, it is currently equal to `full`, the one right below

pnicolli (changes requested):

docs/source/blocks/block-style-wrapper.md
```diff
@@ -39,7 +39,8 @@ This will work if your block uses the `BlocksForm` component to define schema-dr
 ```
 
 This will add a new fieldset `Styling` to your block schema settings.
-By default, only a `backgroundColor` property is set, configured by `defaultSchema` in `src/components/manage/Blocks/Block/StylesSchema.jsx`.
+By default, this fieldset has two fields.
```
stevepiercy, line 42:
```suggestion
By default, this fieldset has two fields: `backgroundColor` and `align`. They are both configured by `defaultSchema` in `src/components/manage/Blocks/Block/StylesSchema.jsx`.
```

docs/source/blocks/block-style-wrapper.md
```diff
@@ -39,7 +39,8 @@ This will work if your block uses the `BlocksForm` component to define schema-dr
...
+A `backgroundColor` field, and the `align` field, both configured by `defaultSchema` in `src/components/manage/Blocks/Block/StylesSchema.jsx`.
```
stevepiercy, line 43: Combined with previous comment.

```suggestion
```

docs/source/blocks/block-style-wrapper.md
```diff
@@ -96,7 +97,22 @@ Same for the block edit component.
...
-<div className="has--backgroundColor--ee22ee has--myCustomStyleField--red has--myCustom2StyleField--color--black has--myCustom2StyleField--color--MyGradient">
+<div class="has--backgroundColor--ee22ee has--myCustomStyleField--red has--myCustom2StyleField--color--black has--myCustom2StyleField--color--MyGradient">
 ```
 
 Then it's at your discretion how you define the CSS class names in your theme.
+
+## Main edit wrapper class injection
+
+Under the hood, there is yet another class injection happening in the main Block Engine Wrapper.
+This is in place to help positioning properly the block in the current layout.
```
stevepiercy, line 108:
```suggestion
This is in place to help properly position the block in the current layout.
```

docs/source/blocks/block-style-wrapper.md
```diff
@@ -96,7 +97,22 @@ Same for the block edit component.
...
+
+Each block in the Block Engine has a main wrapper with an automatic class name `block-editor-<block_id> <block_align>`, as the following:
```
stevepiercy, line 110:
```suggestion
Each block in the Block Engine has a main wrapper with an automatic class name `block-editor-<block_id> <block_align>`, as shown in the following example:
```

stevepiercy (changes requested):
Just a few tweaks.

stevepiercy (approved):

src/components/manage/Blocks/Block/EditBlockWrapper.jsx
```diff
@@ -34,7 +35,13 @@ const EditBlockWrapper = (props) => {
     <div
       ref={draginfo.innerRef}
       {...draginfo.draggableProps}
-      className={`block-editor-${data['@type']}`}
+      // Right now, we can have the alignment information in the styles property or in the
+      // block data root, we inject the classname here for having control over the whole
+      // Block Edit wrapper
+      className={cx(`block-editor-${data['@type']}`, {
+        [data.align]: data.align,
+        [data?.styles?.align]: data?.styles?.align,
```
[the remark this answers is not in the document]
sneridagh: Done!

src/components/manage/Blocks/Block/StyleWrapper.jsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import cx from 'classnames';
+import { buildStyleClassNamesFromData } from '@plone/volto/helpers';
+
+const StyleWrapper = ({ ...props }) => {
```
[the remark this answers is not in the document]
sneridagh: Done!

src/components/manage/Blocks/Block/StyleWrapper.jsx
```diff
@@ -0,0 +1,30 @@
...
+  const { children, data } = props;
+  const styles = buildStyleClassNamesFromData(data?.styles);
```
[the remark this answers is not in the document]
sneridagh: Done!

src/components/manage/Blocks/Block/StyleWrapper.jsx
sneridagh: Removed!

src/components/manage/Blocks/Listing/Edit.jsx
sneridagh: True, let's leave it unchanged now.

src/components/manage/Sidebar/AlignBlock.jsx
sneridagh: Copy paste issues :)

sneridagh:
@ichim-david waiting for your last ack!

PR #3242, opened 2022-04-09
Cypress commands and runs improvements

sneridagh:

PR #3271, opened 2022-04-26
Changed Cypress command generation to use the Cypress Commands from Volto core

JeffersonBledsoe:
This PR intends to centralise the common commands and 'best-practise' for cypress tests. While this can be done by developers already, it isn't immediately obvious as the generator creates all the commands inline, putting the burden of maintaining them on the developer. This PRs approach is similar to how bobtemplates includes the Robot keywords from plone.app.robotframework.


JeffersonBledsoe:
@sneridagh I saw you are working on the Cypress tests a bit in #3242. Any opinions on this?

sneridagh:
@JeffersonBledsoe +1 for this. So please go ahead. Since both are not related, just complete it and let's merge it.

cypress/support/commands.js
```diff
@@ -1,14 +1,20 @@
 import '@testing-library/cypress/add-commands';
 import { getIfExists } from '../helpers';
+
+const HOSTNAME = Cypress.env('BACKEND_HOST') || 'localhost';
+const GUILLOTINA_API_URL = `http://${HOSTNAME}:8081/db/web`;
+const PLONE_SITE_ID = Cypress.env('SITE_ID') || 'plone';
+const PLONE_API_URL = `http://${HOSTNAME}:55001/${PLONE_SITE_ID}`;
+
```
sneridagh, line 8: Awesome! This will come in handy when configuring the Cypress GH Action in parallel mode, in a container infrastructure.

Since this change is related to Volto itself, can we have a CHANGELOG for this change in main Volto too? Thanks!

sneridagh (changes requested):

cypress/support/commands.js
JeffersonBledsoe: Done!

cypress/support/commands.js
```diff
@@ -1,14 +1,20 @@
 import '@testing-library/cypress/add-commands';
 import { getIfExists } from '../helpers';
+
+const HOSTNAME = Cypress.env('BACKEND_HOST') || 'localhost';
+const GUILLOTINA_API_URL = `http://${HOSTNAME}:8081/db/web`;
+const PLONE_SITE_ID = Cypress.env('SITE_ID') || 'plone';
+const PLONE_API_URL = `http://${HOSTNAME}:55001/${PLONE_SITE_ID}`;
```
sneridagh, line 7: Can we introduce another fallback using another env var? like
```suggestion
const PLONE_API_URL = Cypress.env('API_PATH') || `http://${HOSTNAME}:55001/${PLONE_SITE_ID}`;
```

Since matches the others env vars naming and I introduced it in https://github.com/plone/volto/pull/3242/files#diff-70c7e133120d2e4b9afe435052f4988b27e00f9423e2b22e5b4d09f6d58e04d3R2

Thanks!

JeffersonBledsoe:
Thanks for finishing this off @sneridagh!

PR #3274, opened 2022-04-27
@@display-file support

cekk:
Plone has a view called **@@display-file** that is similar to **@@download** but only exposes the file and do not add additional header to download it.
It could be useful if we want to show files and not directly download them.

With these changes Volto can handle that endpoint too.


stevepiercy:
This sounds like a good feature to add to Volto's documentation, too. I'm not sure where, though.

giuliaghisini (approved):

nileshgulia1 (approved):

sneridagh (approved):

stevepiercy:
@sneridagh I found no reference to either `@@download` or `@@display-file` in `plone.api` or `plone.restapi`. At the very least, I think something should be in the Volto docs, even if it is something like "There is no documentation in the backend about how to use this endpoint". Documenting that it is not documented is better than no documentation at all.

sneridagh:
@stevepiercy ^^ :)

sneridagh:
@stevepiercy merging this, I need to have things settled for pushing for the first 16alpha. I hope is fairly ok! We can amend it later.

PR #3291, opened 2022-04-30
Make yarn 3 work

sneridagh:
We decided it's too risky to merge this one before the conference, given the trainings, talks, etc
A change too harsh, having the conference only a few days away.


stevepiercy:
Will this change require an update to any documentation that references "yarn", and specifically "yarn classic"? While working on the main Plone 6 Install docs, I found this one:

https://github.com/plone/volto/blame/master/docs/source/getting-started/install.md#L105

...which also now in the main docs in an open PR.

sneridagh:
Don't give much attention to this now, since I'm not sure about it. My intention was to give it a try and see if it was feasible.

Let's also discussion it in the next days.

stevepiercy:
Got it!

FYI, the [Classic UI team mentions a couple of uses of yarn](https://github.com/plone/documentation/blame/d5322f8b1da693e9e347495a6b0612ae613cb8ab/docs/classic-ui/theming/from-scratch.md#L104-L112). That's the only other mention I found in the main docs. I would say that whatever course you take with yarn, it will have zero impact on other core packages.

Also FYI, I will be participating remotely in the two upcoming sprints.

.yarnrc.yml
```diff
@@ -0,0 +1,4 @@
+nodeLinker: node-modules
```
tiberiuichim, line 1: It would be nice to make it also compatible with anything else (like `nodeLinker: pnpm`).

In principle there's nothing stopping us from being neutral to the node_modules layout. I think there's just one place that hardcodes a node_modules path for resolution, and that can be solved.

.github/workflows/acceptance.yml
```diff
@@ -20,8 +20,9 @@ jobs:
       # node cache
       - name: Get yarn cache directory path
         id: yarn-cache-dir-path
-        run: echo "::set-output name=dir::$(yarn cache dir)"
-      - uses: actions/cache@v1
+        run: echo "::set-output name=dir::$(yarn config get cacheFolder)"
```
davisagli, line 23: This can happen separately, but github deprecated this syntax for setting output. Now it should be:
```suggestion
        run: echo "dir=$(yarn config get cacheFolder)" >> $GITHUB_OUTPUT
```

.yarnrc.yml
davisagli: I'm interested in trying that too, but we should do it as a separate step after merging this, just so it's easier to identify the cause of any problems.

package.json
```diff
@@ -421,5 +421,6 @@
   "volta": {
     "node": "16.15.0",
     "yarn": "1.22.18"
```
davisagli, line 423: Surely this needs to be updated or removed.

davisagli: Also, we should be able to remove yarn-deduplicate from `dependencies`

davisagli (commented):
@sneridagh I finally took some time to read about yarn 3 and try this out. It looks close and seems to work fine.

What needs to happen to update yarn.lock so it's ready to merge?

There are 2 other packages with yarn.lock files which aren't touched here (the volto-testing and scripts packages). Should they be updated as well?

.yarnrc.yml
tiberiuichim: Yes, it should be a separate step. I actually started to test with nodeLinker: pnpm and got into problems with the plone scripts chalk dependency. I think, even if we don't use it by default, working on pnpm compatibility will result in an overall cleaner stack.

.github/workflows/acceptance.yml
sneridagh: I planned to remove them all (just pushed the change) in favour of using the cache support of setup-x actions.

package.json
sneridagh: Done and done.

davisagli (approved):

PR #3299, opened 2022-05-03
Changed identifier field for password reset to 'username'

giuliaghisini:
Added config.settings.useEmailAsLogin (default:false) to know which field display on password reset


giuliaghisini:
if it's possibile we'd love to have this pr in volto 15 (and not 16)

davisagli:
I think it depends on the `plone.use_email_as_login` registry setting.

sneridagh:
@davisagli right, thanks! 
@giuliaghisini I would follow the same policy here as we did with the other settings from the registry. I would create a setting in the config and tie that to it.

giuliaghisini:
> @davisagli right, thanks! @giuliaghisini I would follow the same policy here as we did with the other settings from the registry. I would create a setting in the config and tie that to it.

ok!

giuliaghisini:
@sneridagh @davisagli 
added 'useEmailAsLogin' config option in volto for now. 
We will implement the control panel for plone.use_email_as_login registry setting and use it later.

giuliaghisini:
@sneridagh @davisagli 
added 'useEmailAsLogin' config option in volto for now. 
We will implement the control panel for plone.use_email_as_login registry setting and use it later.

sneridagh (approved):

PR #3304, opened 2022-05-03
return a 302 Redirect when visiting the link view

### LINKED ISSUE #3302: Link view does not issue a 302 redirect
**Describe the bug**
In Classic Plone the link_redirect_view issues a HTTP 302 Redirect, in Volto the view just changes the history of the browser, which does not issue the redirect:

erral:
Fixes #3302 

This way search engines get proper HTTP 302 Redirects.


sneridagh:
@erral we are removing code that I think fixed some similar issues with the redirects in SSR. Do the `<Redirect>` component works in SSR?

erral:
Anyway I tried to add a test to this and I got an InvariantError saying that I should not be using the `<Redirect>` component out of a `<Router>` component, so I don't know whether this is a correct approach to return a proper HTTP Redirect :confused:

sneridagh:
@erral try wrap it in a <MemoryRouter>, there are some examples in the current tests.

erral:
Yes I tried that in the other PR, the RedirectBlock one, and it was OK, so now I have to take the time to bring that test here

PR #3315, opened 2022-05-04
add a slugify function and use it when creating the field id

### LINKED ISSUE #3314: Field names created through dexterity control panel are not properly slugified
**Describe the bug**
In Classic UI when one creates a new content types and adds a field to it, the field id is computed from the field title and it is properly slugified, like:

- Field Title: This is a crapy field?
- Field id: `this_is_a_crapy_field`

When one does the same in Volto the Title is not lowercased and only one whitespace is replaced by _:

- Field Title: This iis a crapy field?
- Field id: `This_is a crapy field?`

erral:
... from the Dexterity control panel.

Fixes #3314


avoinea (approved):

src/helpers/Utils/Utils.js
```diff
@@ -250,3 +250,15 @@ export const reorderArray = (array, origin, target) => {
 
   return result;
 };
+
+/**
+ * Slugify a string: remove whitespaces, special chars and replace with _
+ * @param {string} string String to be slugified
+ * @returns {string} Slugified string
+ */
+export const slugify = (string) => {
```
sneridagh, line 259: Can we have a quick test for this? Thankssss

sneridagh (changes requested):

src/helpers/Utils/Utils.js
erral: The tests is ready

sneridagh (approved):

Issue #2396, opened 2021-05-03
Link Integrity for Content Objects


tisto:
**Is your feature request related to a problem? Please describe.**

Plone Classic shows a link integrity warning when an editor deletes a page that is referenced by another page. This is an essential CMS feature that Volto currently lacks.

**Describe the solution you'd like**

* Create page "My Page"
* Create page "My Second Page" with a link to "My Page" in DraftJS
* Delete "My Page"
* After hitting the "Delete" Button, Volto shows a warning message that lists all the references this delete action would break and the option to cancel or finish the delete operation

cekk:
related to this: https://github.com/plone/plone.restapi/pull/953

ThomasKindermann:
@steffenri yes i worked on the frontend part in Sorrento

davisagli:
This was added in #3317

PR #2783, opened 2021-10-27
add linkintegrity check on deleting items #2396

ThomasKindermann:
add linkintegrity action that checks for Items that link to the item to delete.
add check for linkintegrity on /delete view of components.
add check for linkintegrity on /contents view for multiple components.

volto issue: #2396
restapi PR: https://github.com/plone/plone.restapi/pull/953

PR #3317, opened 2022-05-04
Link Integrity

danielamormocea:
Solves #2396

Requires plone.restapi (https://github.com/plone/plone.restapi/pull/953)

image


src/components/manage/Contents/Contents.jsx
```diff
@@ -1141,6 +1159,42 @@ class Contents extends Component {
                             </li>
                           ))}
                         </ul>
+                        {this.state.linkIntegrityBreakages.length > 0 ? (
+                          <div>
+                            <h3>Potential link breakage</h3>
+                            <p>
+                              By deleting this item, you will break links that
```
avoinea, line 1166: @danielamormocea Please use i18n messages here.

src/actions/content/content.js
```diff
@@ -13,6 +13,7 @@ import {
   UPDATECOLUMNS_CONTENT,
   LOCK_CONTENT,
   UNLOCK_CONTENT,
+  FC_DELETE,
```
avoinea, line 16: @danielamormocea Maybe we can find a better name, more self-explenatory, for this action.

avoinea (changes requested):

sneridagh:
It seems we already had a rotten PR going on: https://github.com/plone/volto/pull/2783 😫 that did the same thing.

@danielamormocea could you take a look and try to merge them (if there's something valuable, that I'm not aware now, i18n maybe?) there that we could reuse in this one? I'm biassed to use this one though, make sure that the i18n is also in sync.

As soon as we have the PR merged in p.restapi we can merge this one.

avoinea (approved):

cekk:
@danielamormocea @avoinea we changed the endpoint method from POST to GET, so i think that you have to change something in the frontend call

sneridagh:
@danielamormocea I think they are changing the backend PR a bit. Please talk to @cekk to adjust

sneridagh (approved):
LGTM, when the restapi is merged we can merge this.

tisto:
@danielamormocea awesome work! I just added the plone.restapi checkout to this PR. I would love to see a Cypress test that covers the basic use cases:

- delete object that is referenced by a text block link
- delete object that is referenced by an attribute reference
- delete parent object of an object that is referenced by a link (either text or attribute or maybe even both)

tisto:
@danielamormocea the link integrity feature works very well. I'd just have a few suggestions for the UX:

Contents (1)

tisto:
@danielamormocea nested folder structures are taken into account, which is good. Though, the system currently does not show the right breaches:

Contents (2)

danielamormocea:
> @danielamormocea the link integrity feature works very well. I'd just have a few suggestions for the UX:
> 
> Contents (1)

In terms of UI, I have no better idea than to simply mention again before listing the items, that those are going to be deleted (although it seems redundant), or to add a separator again after the first list of items (but for me that looks uglier).

Furthermore, regarding the breaches, if there is a page inside a page that is referenced somewhere, the response won't give me that information, so the main page (the one selected) is the only "reference" I get, therefore the only page I can show to the user.

danielamormocea:
@tisto I also added the test for the reference, but unfortunately the linkintegrity api part does not detect objects referenced by a text block link

tisto:
@danielamormocea I was suspecting that Plone and the restapi are the issue here. Then I guess we have to specify in plone.restapi how the backend is supposed to work and then work on the implementation there.

tisto:
@danielamormocea would you be able to create a PR in plone.restapi documenting the missing functionality in a Python unit test?

FYI: Acceptance tests and code analysis are failing for this PR.

danielamormocea:
@tisto  Acceptance tests are failing because there is no linkintegrity api for deleting items from contents (the failing tests are in content.js and folder-contents.js)

rohnsha0 (approved):

sneridagh:
@danielamormocea I hope that now that we have the backend side, we are green and ready to merge.

sneridagh:
@danielamormocea we have a test that fails, could you please take a look? Thanks!

sneridagh:
@danielamormocea I fixed the Cypress test (it was using draftJS still). 
However, I can't see the @linkintegrity call happening... Could you please take a look?

I'd like to have this for Volto 16 which will happen next week for sure.

/cc @avoinea @alecghica

sneridagh:
@avoinea I still couldn't take a look at this, it should work now but it keeps failing. Should be a silly thing.

Issue #1178, opened 2020-02-10
plone.app.redirector control panel


tisto:
* https://github.com/plone/plone.restapi/pull/1398
* https://github.com/plone/volto/pull/3323

davisagli:
The aliases control panel was added in volto 16.0.0-alpha.34

PR #3323, opened 2022-05-05
Alias controlpanel

andreiggr:
Alias/Url Management interface from control panel & object
See: https://github.com/plone/volto/issues/1178


ksuess:
The new control panel looks pretty good! 
May I suggest to give a confirmation via toast after adding an alias?

andreiggr:
> The new control panel looks pretty good! May I suggest to give a confirmation via toast after adding an alias?

Agree. Toast notifications would look good for this case.

giuliaghisini:
agree with @ksuess

avoinea (approved):
LGTM.

avoinea:
@sneridagh As the Plone RestAPI part https://github.com/plone/plone.restapi/pull/1398 was merged and release I think it's safe to go on with this one also.

PR #3329, opened 2022-05-05
Fix changing Volto locale to match the content language

### LINKED ISSUE #3297: Volto 15.4.x no longer supports Language Variants on the backend (en_US, nl_BE, etc.)
When you create a new API backend for a Volto site in Plone 6 Alpha 4 the  Language dropdown contains both Languages (English, Dutch) and Language Variants, like English (USA), Dutch (Belgium).  The former have codes like EN and NL, the latter are encoded with en_US and nl_BE . 

The Default selected language depends on your locale and which browser you use. Some uses with their local and browser will have a language variant selected by default. This explains why not everybody might see this issue. 

Screenshot 2022-05-02 at 16 06 20

Volto 15.4.1 will throw an error if you created the backedn with any Language Variant:

```
"Unhandled Rejection (Error): Cannot find module './en_US.json'
./locales lazy recursive ^./..json$/webpackAsyncContext/<
/home/Code/Plone/Plone-6/frontend/locales lazy ^/./.json$ namespace object:50"
```

This appears to be a regression, as the documentation explicitly has a language variant as an example and Plone 5.2 also has language variants enabled and listed in the Site language dropdown when you create a site.

davisagli:
If content is loaded in a language other than the current locale, Volto tries to switch to the new locale, but it breaks if it's a locale with no Volto translation available. This includes language-specific variants like en_US when only the en locale is available.

To avoid that error, this takes a bit more care to only change to the new language if it's supported.

Fixes #3297

(This is my first Volto PR, so please let me know if I'm missing anything.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 - Fix toolbar bottom opacity on edit page when toolbar is collapsed @ichim-david
 - Fixed secure cookie option. @giuliaghisini
 - Changed addon order in addon controlpanel to mimic Classic UI @erral
+- Fixed error when loading content in a language for which a Volto translation is not available. @davisagli
 
 ### Internal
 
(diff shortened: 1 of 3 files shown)


sneridagh (approved):
Woohoo! :) LGTM! Thanks a lot!

PR #3332, opened 2022-05-06
change history route name

### LINKED ISSUE #2692: Impossible to edit a document called "history"
**Describe the bug**

When an editor creates a new object with the name "history", it is impossible for the editor to edit that object.

**To Reproduce**

1. As admin
2. Go to https://volto.kitconcept.com/
2. Create a new Page with the title "History"
3. Page "https://volto.kitconcept.com/history" shows the front-page
4. Clicking on edit leads to the edit view of the front-page

**Expected behavior**

I'd expect that either the system prevents the user from creating content with reserved words (show a validation error, rename the content object automatically), or that we make this "just" work.

danielamormocea:
Fixes #2692


sneridagh:
@plone/volto-team I guess we should get to an agreement on naming.

Personally, pageHistory is far too specific, because could be the history of other content (eg. event). What about `content-history`?

Also, what about going back to the `@@` notation?

avoinea:
@sneridagh The funny thing is that if you add a page called Sharing it's renamed to `sharing-1`. 

@danielamormocea Can you please investigate what is different within the `sharing` case and why it doesn't behave the same for `history`?

avoinea:
Because the issue is more deeper and it doesn't resumo only to History. Try adding a Page called `controlpanel`

sneridagh:
During lunch we decided to go ahwad with the @@ namespace _leaving_ the current route for v16, deprecating it in next version. We change all references to @@ . Does that makes sense to you?

avoinea:
@sneridagh `@@historyview`

danielamormocea:
@sneridagh  I don't exactly understand what you mean by "We change all references to @@". Should I add some middleware, or I just change all the route names, or maybe there is another way I am not aware of? Some example would really help :) Thanks!

danielamormocea:
@sneridagh  I don't exactly understand what you mean by "We change all references to @@". Should I add some middleware, or I just change all the route names, or maybe there is another way I am not aware of? Some example would really help :) Thanks!

sneridagh:
Since the naming of the views is controversial, I'd say we use the same name that Plone has: `historyview`, and mark it as a Breaking change.

@danielamormocea could you please take care?

Issue #2794, opened 2021-10-30
Listing Block shows different querystring filters


tisto:
**Describe the bug**

The listing block shows different querystring filters dependent on if you start fresh or have already chosen filter criteria.

**To Reproduce**
1. As admin
2. Create a new page
3. Add a listing block
4. Choose a criterion in the listing block (any)
5. Click on the dropdown of the criterion that you choosed
6. -> you will see additional options such as "order in folder", "related to", "default page"

**Expected behavior**

The dropdown list will always return the same list of criteria.

**Screenshots**

Default criteria options:

<img width="374" alt="Bildschirmfoto 2021-10-30 um 07 18 17" src="https://user-images.githubusercontent.com/632105/139521346-b4394b25-0c0c-474e-a207-07c4e85e7cac.png">

Criteria options after choosing a criteria:

<img width="374" alt="Bildschirmfoto 2021-10-30 um 07 18 30" src="https://user-images.githubusercontent.com/632105/139521357-e122baa5-e183-4406-ac04-af41d0ca0f79.png">

**Software (please complete the following information):**
 - Volto Version [14]
 - Plone Version [5.2.5]

PR #3335, opened 2022-05-06
Fix different querystring filters in the querystring widget

kreafox:
If you add more than one criteria in the querystring widget those options are different from the first, default criteria options. 

Options that are disabled in the `@querystring` endpoint should not be included in the criteria dropdown. This PR filters these options in the first Criteria list as well. 

Solves #2794


sneridagh:
@kreafox while you are on it :) Do you think that it's easy to remove from the list an already added Criteria? eg. You add a `Type` criteria, then you can add it again... only the last one work. We can merge this one in the meanwhile, of course.

fredvd:
> @kreafox while you are on it :) Do you think that it's easy to remove from the list an already added Criteria? eg. You add a `Type` criteria, then you can add it again... only the last one work. We can merge this one in the meanwhile, of course.

@sneridagh Are you sure this is the most UX friendly behavior? IMHO you shouldn't be able to add the second (duplicate) and criterions in use should be greyed out.

By removing the first already configured criterion, user can unintentionally alter the existing query without noticing. Worse: the previous query cannot be restored, say there was a type criterion with 4 content types set: these are dropped and the current editor cannot remember what they were...

sneridagh:
@fredvd I expressed myself poorly, I meant what you say. We should grey out the criteria already chosen in the dropdown.

@kreafox do you think that you could take a look or should we just merge this one?

kreafox:
@sneridagh Just a small issue: If you select the first option, which in our case is the "Creator" and select another criteria or you open an existing one, the first option "Creator" it doesn't look like a disabled option at first with grey color until you hover it. It is a very strange bug and it is only happening with the first option (you can see in my demo) I couldn't identify yet what's causing it. But other than that it is working as expected.

https://user-images.githubusercontent.com/20109479/167846427-c700bc85-ff89-4847-a30e-123a672931cd.mp4

sneridagh:
@kreafox caught it! :) 🎉 

I will merge on green.

Issue #1395, opened 2020-04-20
Support srcset in Volto


tisto:
Volto should deliver images with srcset to reduce the download sizes of images.

nzambello:
IMHO this is related to #1163

davisagli:
Duplicate of #1428

PR #3337, opened 2022-05-06
Image rendering

pnicolli:
Related: #2103 #1428 #1395 

~This is mainly documentation about how to properly render images, at this time.
This is meant to be improved still, in order to document how we want to do this and then implement it.~

This now contains the new Image component implementation.

Depends on https://github.com/plone/plone.restapi/pull/1642


giuliaghisini:
good. This morning i found out also that defining the aspect ratio in css could help the browser if you don't know the real size of the rendered image.. 
So,  in https://github.com/plone/volto/pull/2103 i've added the inline-style with aspect-ratio value taked from real image sizes..
What do you think?

src/components/theme/PreviewImage/PreviewImage.jsx
```diff
@@ -11,11 +11,32 @@ import DefaultImageSVG from '@plone/volto/components/manage/Blocks/Listing/defau
...
-  const src = item.image_field
-    ? flattenToAppURL(`${item['@id']}/@@images/${item.image_field}/${size}`)
-    : DefaultImageSVG;
+  if (item.image_field) {
+    const { width, height } =
+      item.image_scales?.[item.image_field]?.[0]?.scales[size] ?? {};
 
-  return <img src={src} alt={alt ?? item.title} {...rest} />;
+    return (
+      <img
```
giuliaghisini, line 19: Why not using <Image> component? 
Could it switch to render image from image object if exists and use image composite url as a fallback?

src/components/manage/Blocks/Listing/SummaryTemplate.jsx
```diff
@@ -25,9 +25,9 @@ const SummaryTemplate = ({ items, linkTitle, linkHref, isEditMode }) => {
         {items.map((item) => (
           <div className="listing-item" key={item['@id']}>
             <ConditionalLink item={item} condition={!isEditMode}>
-              <Component componentName="PreviewImage" item={item} alt="" />
+              <Component componentName="PreviewImage" item={item} />
```
pnicolli, line 28: Using PreviewImage for backwards compatibility at this time. This shoule be changed to the new Image component, but that will be a breaking change.

src/components/theme/Image/Image.jsx
```diff
@@ -0,0 +1,140 @@
...
+  //           }
+  //         }, 10);
+  //       },
+  //       { threshold: [0], rootMargin: '100px' },
+  //     );
+  //     observer.observe(imageRef.current);
+  //   } else if (srcSet?.length > 0) {
+  //     applySrcSet();
+  //   }
+  // }, [imageRef, applySrcSet, imageHasLoaded, srcSet, srcset]);
```
pnicolli, line 83: This comment can be ignored for now, it is IntersectionObserver stuff that is still work in progress.

src/components/theme/Image/Image.jsx
```diff
@@ -0,0 +1,140 @@
...
+ */
+const Image = ({
+  image,
+  alt = '',
+  className,
+  pictureClassName,
+  size = 'full',
+  responsive = true,
+  role = 'img',
+  loading = 'eager',
```
pnicolli, line 25: The default here is up for discussion. 'eager' for now imho because it's the default web behavior, where you would add `loading="lazy"` where needed, otherwise it's eager.

src/components/theme/Image/Image.jsx
```diff
@@ -0,0 +1,140 @@
...
+ * @param {string} role - img role attribute
+ * @param {boolean} loading - (default: eager) if set to `lazy`, the srcset is loaded when the image is in the viewport
+ */
+const Image = ({
+  image,
+  alt = '',
+  className,
+  pictureClassName,
+  size = 'full',
+  responsive = true,
```
pnicolli, line 23: I'm not sure if the correct name for this is `responsive`, like frameworks (e.g. Bootstrap) usually say, or maybe `fluid`, which is the actual behavior that this gives to the image. Default to true because it is more backwards compatible and is the more common option imho.

src/components/theme/Image/Image.jsx
```diff
@@ -0,0 +1,140 @@
...
+    <>
+      <picture
+        className={cx('volto-image', {
+          pictureClassName,
+          responsive,
+        })}
+      >
+        {loading === 'eager' && srcSet.length > 0 && (
+          <source srcSet={srcSet} sizes="100vw" />
+        )}
```
pnicolli, line 95: This needs to be improved with the intersection observer above.

src/components/theme/Image/Image.jsx
```diff
@@ -0,0 +1,140 @@
...
+                  class="${className || ''}"
+                  role="${role}"
+                  ${width ? `width="${width}` : ''}
+                  ${height ? `height="${height}` : ''}
+                  loading="lazy"
+                />
+              `,
+            }}
+          />
+        )} */}
```
pnicolli, line 123: Commented for now since it will be mainly used as the fallback for the intersection observer.

src/components/theme/Image/utils.js
```diff
@@ -0,0 +1,88 @@
+import { flattenToAppURL } from '@plone/volto/helpers';
+
+const SIZE_MAPPING = {
+  small: 'preview',
+  medium: 'large',
+  large: 'great',
+};
```
pnicolli, line 7: I believe this should be somewhere where it is more configurable. Volto config?

src/components/theme/PreviewImage/PreviewImage.jsx
pnicolli: I thought about it, the best thing as you say would be to use the new Image component here. This would break all of the current uses of PreviewImage though.
It's still up for discussion if this PR should introduce a hard breaking change or try to break the smallest possible set of things.
For example, if you change this to the Image component, the listing block SummaryTemplate renders in a very different way.
I implemented it like in order to have a PoC of a change that is working but I'm happy to discuss this point, it is one of the most important ones.
On the other hand, this PreviewImage component could become obsolete if the backend change regarding having image scale metadata in the catalog gets fixed.

docs/source/developer-guidelines/images.md
```diff
@@ -0,0 +1,62 @@
...
+
+### Other discussion points
+
+_Should we render in the `srcset` the sizes that are bigger than the chosen one?_
+
+_In fluid layouts, the alignment decision made by the user will impact the actual width of the image, based on CSS rules. Should this be taken into account for `srcset` rendering?_
+
+Sizes are defined in the Volto config, or in the Tinymce controlpanel if using Classic UI.
+
+## Lazy loading
```
mamico, line 58: Nowadays, lazy loading image is well implemented by the `loading="lazy"` attribute which is widely supported by all browsers https://caniuse.com/loading-lazy-attr (pay attention that for some browser lazy attribute is not supported for iframes, but it is for images). 

IMHO adding `IntersetionObserver` may not be necessary or have a negative impact.

pnicolli: Only the very latest versions of Safari support this attribute by default: https://caniuse.com/?search=loading%20lazy
This feature has an overall 80% support globally among browsers, while [IntersectionObserver is way higher with 94%](https://caniuse.com/?search=intersectionobserver).

In addition to this, I'm not sure if we want to lazy load the tiny resolution placeholder that we use before the srcset is applied by the IntersectionObserver, more thought needs to be given on this decision imho.

Moreover, @MrTango was reporting an issue in Chrome support for `loading="lazy"`, regarding the pixels threshold that the browser tracks for loading the image lazily. @MrTango if you could paste a pointer to the issue it would be great, unless it has been solved already.

src/components/theme/Image/Image.jsx
```diff
@@ -0,0 +1,83 @@
...
+/**
+ * Image component
+ * @param {object | string} image - Plone image as object or url
+ * @param {string} baseUrl - Base URL of the image item (the part before /@@images)
+ * @param {string} alt - Alternative text for the image
+ * @param {boolean} loading - (default: eager) set to `lazy` to lazy load the image
+ */
+export default function Image({
+  image,
+  baseUrl = '',
```
pnicolli, line 13: I stuck with the image + baseUrl implementation, because it's probably more flexible than having a complete item object. For example, in the view of a custom content type I could have more than one image field and want to show more than one.

davisagli: @pnicolli It makes sense to pass in the specific image that should be rendered, rather than guessing the field name. But I think it is probably a good idea to also include the item anyway, for context in case a customized version of the Image component needs to render the image differently based on some other property of the item.

I will check more carefully whether our current implementation actually depends on this. I think it does but I might be remembering details wrong.

pnicolli:
Note from Beethoven sprint discussion: in blocks that pick images (e.g. Hero and Image block), if we save image metadata like image_scales in the block data, we can end up having outdated information, because if the image changes, the scales change and the blocks data do not reflect that change.

Since the image selection is basically a `url` in the block data, and those are already being transformed by the resolveuid machinery, we could plug into that, check if the item is an image and update the block data accordingly, which likely also means deleting that data if the image is no longer there.

@davisagli feel free to correct me and/or add info.

davisagli:
@pnicolli yes, that's right. I have code from a project that can be adapted for this.

pnicolli:
Most of this is working. What is left is:
- styling the new reset button
- image upload is broken

Working on this in the next days

pnicolli:
@sneridagh @davisagli should be quite fine now, but I am having trouble with cypress tests. The tests here on github most likely fail because the backend PR is missing, but I wanted to test cypress locally. Any suggestion on how to test a custom restapi branch with cypress locally? I found out how to configure volto and cypress but I don't know much about the backend and I need help configuring a custom backend or docker container to work for e2e tests.

pnicolli:
I also still need to update the docs, will do asap.

davisagli:
@pnicolli There's probably a better way but this seems to work:

1. Create a file named Dockerfile.acceptance containing:
```
FROM plone/server-acceptance:latest
RUN pip install git+https://github.com/plone/plone.restapi.git@block-image-scales
```
(This extends the image built from https://github.com/plone/plone-backend/blob/6.0.x/Dockerfile.acceptance, which already includes the necessary changes for e2e tests like installing plone.app.robotframework and running on port 55001)

2. run `docker build -f Dockerfile.acceptance -t volto-acceptance .`
3. edit volto's Makefile to change DOCKER_IMAGE_ACCEPTANCE to `volto-acceptance:latest`
4. run `make start-test-acceptance-server`

pnicolli:
@davisagli Thanks, seems like it's working, I am currently fixing tests.

pnicolli:
All tests are now green on my machine with this restapi PR https://github.com/plone/plone.restapi/pull/1642

pnicolli:
Also added docs now, I believe this PR is now ready.

docs/source/recipes/images.md
```diff
@@ -0,0 +1,51 @@
+# Images
```
stevepiercy, line 1: All docs need HTML meta tags for SEO, and a label for creating links from cross-references.

```suggestion
---
myst:
  html_meta:
    "description": "Volto code recipe for images"
    "property=og:description": "Volto code recipe for images"
    "property=og:title": "Volto code recipe for images"
    "keywords": "Plone, Volto, code, recipe, images"
---

(images-label)=

# Images
```

docs/source/recipes/images.md
```diff
@@ -0,0 +1,51 @@
...
+
+Volto comes with an Image component that allows to load images in an optimal way.
```
stevepiercy, line 3:
```suggestion
Volto comes with an `Image` component that loads images in an optimal way.
```

docs/source/recipes/images.md
```diff
@@ -0,0 +1,51 @@
...
+
+Features of the Image component are:
```
stevepiercy, line 5:
```suggestion
Features of the `Image` component are the following.
```

docs/source/recipes/images.md
```diff
@@ -0,0 +1,51 @@
...
+
+- optimized lazy loading
+- automatic `srcset` generation, if given a content item or catalog brain
+- optional responsive width
+
+The component can accept all the regular html `<img>` attributes, as well as a few extra for the features described above.
```
stevepiercy, line 11:
```suggestion
The component can accept all the regular HTML `<img>` attributes and a few extra for the above features.
```

docs/source/recipes/images.md
```diff
@@ -0,0 +1,51 @@
+# Images
+
+Volto comes with an Image component that allows to load images in an optimal way.
+
+Features of the Image component are:
+
+- optimized lazy loading
```
stevepiercy, line 7:
```suggestion
- optimized [lazy loading](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/loading)
```

docs/source/recipes/images.md
```diff
@@ -0,0 +1,51 @@
...
+- automatic `srcset` generation, if given a content item or catalog brain
+- optional responsive width
+
+The component can accept all the regular html `<img>` attributes, as well as a few extra for the features described above.
+
+Setting the `loading="lazy"` attribute to the Image component also adds the `decoding="async"` attribute which is generally considered an improvement when loading images lazily.
```
stevepiercy, line 13:
```suggestion
Setting the `loading="lazy"` attribute to the `Image` component also adds the `decoding="async"` attribute.
The [`async`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/decoding) attribute decodes the image asynchronously to reduce delay in presenting other content.
```

docs/source/recipes/images.md
```diff
@@ -0,0 +1,51 @@
...
+
+In order to render an image from a content object, you would use the component like this:
```
stevepiercy, line 15:
```suggestion
To render an image from a content object, you can use the component as shown in the following code snippet.
```

docs/source/recipes/images.md
```diff
@@ -0,0 +1,51 @@
...
+
+```javascript
+import { Image } from '@plone/volto/components';
+
+<Image item={content} imageField="image" alt="" />;
+```
+
+This will generate the `src`, `srcset`, `width` and `height` attributes automatically for the given image.
```
stevepiercy, line 23:
```suggestion
This will generate the `src`, `srcset`, `width`, and `height` attributes automatically for the given image.
```

docs/source/recipes/images.md
```diff
@@ -0,0 +1,51 @@
...
+- automatic `srcset` generation, if given a content item or catalog brain
+- optional responsive width
+
+The component can accept all the regular html `<img>` attributes, as well as a few extra for the features described above.
+
+Setting the `loading="lazy"` attribute to the Image component also adds the `decoding="async"` attribute which is generally considered an improvement when loading images lazily.
+
+In order to render an image from a content object, you would use the component like this:
+
+```javascript
```
stevepiercy, line 17: The JavaScript lexer is OK, but let's try JSX for React components. It usually does a better job, except when it cannot parse the code.

````suggestion
```jsx
````

File: docs/source/recipes/images.md [hunk 20]
```diff
@@ -0,0 +1,51 @@
...
+import { Image } from '@plone/volto/components';
+
+<Image item={content} imageField="image" alt="" />;
+```
+
+This will generate the `src`, `srcset`, `width` and `height` attributes automatically for the given image.
+
+If the `responsive` prop is added, a class is applied that will make the image width responsive.
```
stevepiercy, line 25: Switch from passive to active voice, and address the reader.

Also please check these two sentences for accuracy. I cannot tell if the code inserts a provided named CSS class, or the literal `responsive` as the CSS class by default. IOW does it do the following?

``

or

``?

```suggestion
To make the image width responsive, you can add the `responsive` prop with the name of a CSS class, and its name will be applied to the HTML `image` tag's `class` attribute.
If you omit a name, `responsive` will be applied by default.
```

docs/source/recipes/images.md
```diff
@@ -0,0 +1,51 @@
...
+
+## Serving the right image size
+
+In order to serve the proper image size in every situation, a `sizes` attribute should be added to the Image component while rendering, this responsibility is left to the developer because it depends on where the image is rendered.
```
stevepiercy, line 29:
```suggestion
To serve the proper image size for the viewer's device, you should a `sizes` attribute to the `Image` component while rendering.
This responsibility is left to the developer because it depends on where the image is rendered.
```

docs/source/recipes/images.md
```diff
@@ -0,0 +1,51 @@
...
+
+For example, if the image is rendered in a container that is always half as wide as the page, the `sizes` attribute should reflect that in order to properly inform the browser.
```
stevepiercy, line 31:
```suggestion
For example, if the image is rendered in a container that is always half as wide as the page, the `sizes` attribute should reflect that to properly inform the browser which image size to request and render.
```

docs/source/recipes/images.md
```diff
@@ -0,0 +1,51 @@
...
+
+```javascript
```
stevepiercy, line 33:
````suggestion
```jsx
````

File: docs/source/recipes/images.md [hunk 24]
```diff
@@ -0,0 +1,51 @@
...
+<Image item={content} imageField="image" alt="" sizes="50vw" />
+```
+
+If the image is rendered at full viewport width when the viewport is less than 900px wide, though, you can add a media condition to the `sizes` attribute.
+
+```javascript
```
stevepiercy, line 39:
````suggestion
```jsx
````

File: docs/source/recipes/images.md [hunk 25]
```diff
@@ -0,0 +1,51 @@
...
+
+In order to serve the proper image size in every situation, a `sizes` attribute should be added to the Image component while rendering, this responsibility is left to the developer because it depends on where the image is rendered.
+
+For example, if the image is rendered in a container that is always half as wide as the page, the `sizes` attribute should reflect that in order to properly inform the browser.
+
+```javascript
+<Image item={content} imageField="image" alt="" sizes="50vw" />
+```
+
+If the image is rendered at full viewport width when the viewport is less than 900px wide, though, you can add a media condition to the `sizes` attribute.
```
stevepiercy, line 37:
```suggestion
If the image is rendered at full viewport width when the viewport is less than `900px` wide, though, you can add a media condition to the `sizes` attribute.
```

docs/source/recipes/images.md
```diff
@@ -0,0 +1,51 @@
...
+```javascript
+<Image
+  item={content}
+  imageField="image"
+  alt=""
+  sizes="(max-width: 900px) 100vw, 50vw"
+/>
+```
+
+More info on this subject can be found at:
```
stevepiercy, line 48:
```suggestion
More information on this subject can be found at the following resources.
```

docs/source/recipes/images.md
```diff
@@ -0,0 +1,51 @@
...
+
+- <https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images>
```
stevepiercy, line 50: A MyST plugin `linkify` supports bare URLs, so angle brackets are no longer necessary. 🎉 

```suggestion
-   https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images
```

docs/source/recipes/images.md
```diff
@@ -0,0 +1,51 @@
...
+- <https://www.builder.io/blog/fast-images>
```
stevepiercy, line 51:
```suggestion
-   https://www.builder.io/blog/fast-images
```

stevepiercy (changes requested):
Mostly style changes:

- Change from passive to active voice, and address the reader
- Use `jsx` lexer instead of `javascript` (which has the alias `js` if you feel lazy)
- Added some references to terms
- Some MyST syntax fixes

I have one question about the responsive prop. Please see comment.

docs/source/recipes/images.md
pnicolli: TIL about the oxford comma 😬

docs/source/recipes/images.md
pnicolli: I guess it 's a little bit of both. It adds the literal `responsive` as the CSS class, which also has custom CSS rules for making it responsive

docs/source/recipes/images.md
```diff
@@ -0,0 +1,64 @@
...
+  imageField="image"
+  alt=""
+  sizes="(max-width: 900px) 100vw, 50vw"
+/>
+```
+
+More information on this subject can be found at the following resources.
+
+- https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images
+- https://www.builder.io/blog/fast-images
```
pnicolli, line 64: @stevepiercy  Is it mandatory to avoid using angle brackets for bare links? I have markdownlint installed in vscode and it really complains about removing the angle brackets :D I can deal with it if needed, of course, I was just wondering

src/components/manage/Blocks/Image/View.jsx
```diff
@@ -32,15 +36,26 @@ export const View = ({ data, detached }) => {
...
+                        '@id': data.url,
+                        image_field: data.image_field,
+                        image_scales: data.image_scales,
+                      }
+                    : undefined
+                }
                 src={
-                  isInternalURL(data.url)
+                  data.image_scales
+                    ? undefined
```
ksuess, line 57: I appreciate the use of srcset and further modern image options. But…
Image block images aligend left or right and especially of smaller size than 'large' are fetched in viewport size. That's much more than necessary. 
Let's think about a way to give the browser a hint to fetch the appropriate size.

To reproduce the situation that causes my headache:
- Create an image block and upload a bigger than big image.
- Align left
- Size it to medium

Battery_and_NowPlaying_and_Item-0_and_Item-0_and_Item-0_and_AppleTimeMachineExtra-2

ksuess (changes requested):

src/components/manage/Blocks/Image/View.jsx
ksuess: > I appreciate the use of srcset and further modern image options. But… Image block images aligend left or right and especially of smaller size than 'large' are fetched in viewport size. That's much more than necessary. Let's think about a way to give the browser a hint to fetch the appropriate size.
> 
> To reproduce the situation that causes my headache:
> 
>     * Create an image block and upload a bigger than big image.
> 
>     * Align left
> 
>     * Size it to medium
> 
> 
> Battery_and_NowPlaying_and_Item-0_and_Item-0_and_Item-0_and_AppleTimeMachineExtra-2

before on 17a14

Cursor_and_image_rendering

pnicolli: Yeah you are right, I even documented how to do this ad then did not do it here :) I will fix this

pnicolli: @ksuess added the sizes prop to the image and leadimage blocks. Should now work way better, it does for me. Using this approach also considers the pixel density of the screen, which is a nice improvement.

docs/source/recipes/images.md
stevepiercy: Got it. What do you think of this version?

```suggestion
To make the image width response, you can add the prop `responsive`, and the literal `responsive` will be added to the HTML attribute `class`.
```

Regarding other HTML attributes, I think an example of how the component would render would help clarify. I'll draft one up above for you to refine.

docs/source/recipes/images.md
```diff
@@ -0,0 +1,66 @@
...
+
+Volto comes with an `Image` component that loads images in an optimal way.
+
+Features of the `Image` component are the following.
+
+- optimized [lazy loading](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/loading)
+- automatic `srcset` generation, if given a content item or catalog brain
+- optional responsive width
+
+The component can accept all the regular HTML `<img>` attributes and a few extra for the above features.
```
stevepiercy, line 22:
````suggestion
The component can accept all the regular HTML `<img>` attributes and a few extra for the above features.
The following example demonstrates how the code with attributes will render to HTML.

```jsx

const Image = config.getComponent({ name: 'Image' }).component;

<Image
  item={content}
  imageField="image"
  alt="Alt text for accessibility"
  class="myClass"
/>;
```

```html
<img src="my-image.png" alt="Alt text for accessibility" class="myClass" />
```
````

File: docs/source/recipes/images.md [hunk 32]
```diff
@@ -0,0 +1,66 @@
...
+The [`async`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/decoding) attribute decodes the image asynchronously to reduce delay in presenting other content.
+
+To render an image from a content object, you can use the component as shown in the following code snippet.
+
+```jsx
+import config from '@plone/volto/registry';
+
+const Image = config.getComponent({ name: 'Image' }).component;
+
+<Image item={content} imageField="image" alt="" />;
```
stevepiercy, line 34: I just realized we should set a good example. Empty `alt` makes me sad. 😢 

```suggestion
<Image item={content} imageField="image" alt="Alt text for accessibility" />;
```

docs/source/recipes/images.md
```diff
@@ -0,0 +1,66 @@
...
+<Image item={content} imageField="image" alt="" sizes="50vw" />
+```
+
+If the image is rendered at full viewport width when the viewport is less than `900px` wide, though, you can add a media condition to the `sizes` attribute.
+
+```jsx
+<Image
+  item={content}
+  imageField="image"
+  alt=""
```
stevepiercy, line 58:
```suggestion
  alt="Alt text for accessibility"
```

docs/source/recipes/images.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Serving the right image size
+
+To serve the proper image size for the viewer's device, you should add a `sizes` attribute to the `Image` component while rendering.
+This responsibility is left to the developer because it depends on where the image is rendered.
+
+For example, if the image is rendered in a container that is always half as wide as the page, the `sizes` attribute should reflect that to properly inform the browser which image size to request and render.
+
+```jsx
+<Image item={content} imageField="image" alt="" sizes="50vw" />
```
stevepiercy, line 49:
```suggestion
<Image item={content} imageField="image" alt="Alt text for accessibility" sizes="50vw" />
```

docs/source/recipes/images.md
stevepiercy: It is an optional style guide. Please do whatever makes your job easier or scratches your itch.

However, I would suggest the [MyST-Markdown VS Code extension](https://marketplace.visualstudio.com/items?itemName=ExecutableBookProject.myst-highlight) for better parsing of MyST. MyST syntax is extended CommonMark markdown syntax. We also use several MyST and Sphinx extensions which might get flagged by code editors.

stevepiercy (changes requested):
- Revised responsive docs sentence.
- Added example for HTML attributes, should be reviewed.
- Set a good example for no empty `alt` tags.

src/components/manage/Blocks/Image/View.jsx
ksuess: @pnicolli great solution!

ksuess (approved):
Big improvement. 
I also like the coherent use of the Image component from the registry.

ksuess:
@pnicolli would you add to the PR description, that reviewing needs https://github.com/plone/plone.restapi/pull/1642?

docs/source/recipes/images.md
davisagli: @stevepiercy should be className, not class

davisagli: And I think we need either this example or the one below, but not both.

theme/themes/pastanaga/extras/userscontrolpanel.less
davisagli, line 1: These changes don't look related to the image component.

src/components/theme/View/SummaryView.jsx
```diff
@@ -34,11 +33,13 @@ const SummaryView = ({ content }) => (
               </UniversalLink>
             </h2>
             {item.image_field && (
-              <PreviewImage
+              <Component
+                componentName="PreviewImage"
                 item={item}
-                alt={item.image_caption ? item.image_caption : item.title}
```
davisagli, line 39: We don't want to lose the `alt`, right?

src/components/theme/View/NewsItemView.jsx
```diff
@@ -41,15 +38,12 @@ const NewsItemView = ({ content }) => {
       )}
       {content.image && (
         <Image
-          className="documentImage"
+          className="documentImage ui right floated image"
           alt={content.title}
           title={content.title}
-          src={
-            content.image['content-type'] === 'image/svg+xml'
```
davisagli, line 48: Does the new component need special handling for SVGs?

src/components/theme/Image/Image.jsx
```diff
@@ -0,0 +1,95 @@
...
+}) {
+  if (!item && !src) return null;
+
+  // TypeScript hints for editor autocomplete :)
+  /** @type {React.ImgHTMLAttributes<HTMLImageElement>} */
+  const attrs = {};
+
+  if (!item && src) {
+    attrs.src = src;
+    attrs.alt = alt;
```
davisagli, line 33:
```suggestion
```

(not used)

src/components/manage/Blocks/Teaser/utils.js
```diff
@@ -1,44 +0,0 @@
-import { isInternalURL } from '@plone/volto/helpers';
-import config from '@plone/volto/registry';
-
-export function getTeaserImageURL({ href, image, align }) {
```
davisagli, line 4: We should mention in the upgrade guide that this is no longer used, in case someone has shadowed it.

src/components/manage/Blocks/Teaser/DefaultBody.jsx
```diff
@@ -60,9 +53,11 @@ const TeaserDefaultTemplate = (props) => {
               {(href.hasPreviewImage || href.image_field || image) && (
                 <div className="image-wrapper">
                   <Image
-                    src={hasImageComponent ? href : defaultImageSrc}
+                    item={image || href}
+                    imageField={image ? image.image_field : href.image_field}
                     alt=""
                     loading="lazy"
+                    responsive={true}
```
davisagli, line 60: Ideally, we would set `sizes` here based on the number of columns in the grid containing the teaser.

davisagli (commented):
@pnicolli Thanks for your hard work on this! I have some comments, but they are all pretty minor and I think this is close.

I also tested it in the context of our project where we also customized image rendering heavily, and didn't see any obvious regressions there. /cc @sneridagh

docs/source/recipes/images.md
stevepiercy: @davisagli thanks, corrected. I would prefer the above complete example, too. It's the first one and illustrates the main concepts.

I think the (now) second example is still useful, but may have its `import` and `const` removed for brevity.

pnicolli: What about using only basic HTML attributes here and an item `content` in the example below to show specifically the extra feature of using the content object?

This first example would become

```jsx

const Image = config.getComponent({ name: 'Image' }).component;

<Image
  src="https://picsum.photos/200/300"
  alt="Alt text for accessibility"
  className="myClass"
  width="200"
  height="300"
/>
```

```html
<img
  src="https://picsum.photos/200/300"
  alt="Alt text for accessibility"
  class="myClass"
  width="200"
  height="300"
/>
```

docs/source/recipes/images.md
pnicolli: Generally speaking, empty `alt`s are not necessarily bad :)
I agree we should emphasize that usage is important for a11y though, so we should add it here.

docs/source/recipes/images.md
pnicolli: I did not know it existed. I will gladly try switching extensions and see how it goes :)

src/components/manage/Blocks/Teaser/DefaultBody.jsx
pnicolli: Yeah, I thought about it but isn't the teaser block being reworked? I didn't want to waste time.
Also, the teaser could be used without grid or with a different kind of grid block.
Maybe setting this sizes should be part of a separate work.

docs/source/recipes/images.md
stevepiercy: That works for me. @davisagli?

docs/source/recipes/images.md
stevepiercy: Agreed, however I would assume that these are images used in content blocks that add meaning to the page, and are not purely decorative images. For posterity, this is a good guide to determine whether to use `alt`: https://www.w3.org/WAI/tutorials/images/decision-tree/

src/components/theme/Image/Image.jsx
pnicolli: You are right. Fixing it

src/components/theme/View/NewsItemView.jsx
pnicolli: Actually, probably it does. Adding it

src/components/theme/View/ListingView.jsx
```diff
@@ -28,11 +28,13 @@ const ListingView = ({ content }) => (
             {item.description && <p>{item.description}</p>}
           </Container>
           {item.image_field && (
-            <PreviewImage
+            <Component
+              componentName="PreviewImage"
               item={item}
               size="thumb"
-              alt={item.image_caption ? item.image_caption : item.title}
```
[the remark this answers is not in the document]
pnicolli: We recently had an a11y review from a specialized company that, among other things, suggested that we don't add the image title as alt because it's already there, right before or after the image, so an empty alt is better.
On the other hand, as you suggested in other comments, we should keep the `image_caption` as alt when provided, so I am adding it back, but I would not use the title as fallback.

docs/source/recipes/images.md
davisagli: Sure!

src/components/theme/View/ListingView.jsx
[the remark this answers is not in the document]
davisagli: @pnicolli Okay, I understand the logic now: don't reuse text for the `alt` that is already displayed nearby. Makes sense.

docs/source/recipes/images.md
pnicolli: Done

src/components/manage/Blocks/Teaser/utils.js
pnicolli: Done.
Also tagging @stevepiercy for this.
My formatter worked hard on some indentation and other formatting in the upgrade guide. I went through all of it, removed something but everything else looks quite legit. I know it's polluting the PR diff but I actually feel like they are good modifications. Let me know what you both think about it.

src/components/theme/Image/Image.jsx
pnicolli: Done

src/components/theme/View/NewsItemView.jsx
pnicolli: Done

src/components/theme/View/SummaryView.jsx
pnicolli: Fixed

theme/themes/pastanaga/extras/userscontrolpanel.less
pnicolli: You're right, removed. It was only formatting, but it was useless here.

src/components/theme/View/AlbumView.jsx
```diff
@@ -78,19 +78,19 @@ class AlbumView extends Component {
...
                             <Segment className="imageborder">
-                              <PreviewImage
+                              <Component
+                                componentName="PreviewImage"
                                 item={item}
-                                alt={
-                                  item.image_caption
-                                    ? item.image_caption
-                                    : item.title
-                                }
```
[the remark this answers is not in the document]
pnicolli: Fixed

src/components/theme/View/ListingView.jsx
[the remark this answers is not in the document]
pnicolli: Fixed

src/components/manage/Blocks/Image/View.jsx
```diff
@@ -57,8 +75,23 @@ export const View = ({ data, detached }) => {
...
+                    if (data.size === 'm') return '50vw';
+                    if (data.size === 's') return '25vw';
+                  }
+                  if (data.align === 'left' || data.align === 'right') {
+                    if (data.size === 'l') return '50vw';
+                    if (data.size === 'm') return '25vw';
+                    if (data.size === 's') return '15vw';
+                  }
+                  return undefined;
+                })()}
```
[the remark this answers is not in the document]
pnicolli: Done. It is now configurable in the blocks config AND through shadowing.

stevepiercy (approved):

docs/source/upgrade-guide/index.md
```diff
@@ -2316,7 +2322,7 @@ Modify the parent element of your block making these changes:
 - Add a proper role for it
 
 ```jsx
-  role="presentation"
+  role='presentation';
 ```
```
stevepiercy, line 2326: I am OK with the MyST and English syntax and grammar changes.

I will leave the JSX linting to others, as I am not familiar with the automatic linting tool's results. But this does raise an interesting question: How should we format and lint code snippets in documentation, and check that they conform to our guidelines? Is there such a tool that can check this? I use PyCharm which lacks a decent MyST syntax checker, and I haven't tried VS Code's extension for MyST.

stevepiercy (approved):
Looks good to me, with a couple of follow up questions. Any insight would be appreciated. Thank you!

docs/source/upgrade-guide/index.md
pnicolli: In Volto it is quite easy because prettier handles that, so it has the same configuration as the codebase. That's what happened here and I am actually surprised it did not happen earlier. I don't know about other Plone repos though.

docs/source/recipes/images.md
```diff
@@ -0,0 +1,87 @@
...
+
+To render an image from a content object, you can use the component as shown in the following code snippet.
+
+```jsx
+<Image item={content} imageField="image" alt="Alt text for accessibility" />;
+```
+
+This will generate the `src`, `srcset`, `width`, and `height` attributes automatically for the given image.
+
+To make the image width response, you can add the prop `responsive`, and the literal `responsive` will be added to the HTML attribute `class`.
```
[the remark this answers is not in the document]
pnicolli: Fixed

src/components/manage/Blocks/Teaser/utils.js
pnicolli: Removed all extra alterations to this file, left only the added lines

src/components/theme/View/ListingView.jsx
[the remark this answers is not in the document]
pnicolli: We don't really have an `alt` thingie here from the block configuration or the item, we just have the optional `image_caption` field that we use here.
I am not sure I fully understand your question. The `alt` will be read together with the image, always, so the caption will always be in the same context. The item title, on the other hand, can be before or after based on the listing variation, but in any case the screen reader will read the elements in the order they are rendered and they will always be listed in the same order in this block, so it should be fairly easy to grasp.

src/components/theme/View/ListingView.jsx
```diff
@@ -28,11 +28,13 @@ const ListingView = ({ content }) => (
             {item.description && <p>{item.description}</p>}
           </Container>
           {item.image_field && (
-            <PreviewImage
+            <Component
+              componentName="PreviewImage"
               item={item}
-              size="thumb"
-              alt={item.image_caption ? item.image_caption : item.title}
+              alt={item.image_caption}
```
pnicolli, line 34: I am testing this and I found out we don't really receive the image_caption from restapi :S
And the reason why I was testing it is that I realized that if we were to receive caption data from restapi in listing blocks, we would not be considering we could have a preview image instead of the lead image. This means that just using image_caption, given that we receive it, could mean using the wrong caption.
I would suggest we just remove this alt from here and go with the empty one here, since it's not a regression to do so.

I tested creating a page in demo.plone.org with a listing block which shows a news item and an event.

docs/source/upgrade-guide/index.md
stevepiercy: I am surprised as well. Perhaps prettier only checks files that are edited? And why does GitHub Actions run prettier, but not on documentation code snippets?

docs/source/recipes/images.md
```diff
@@ -1,10 +1,10 @@
...
   html_meta:
-    "description": "Volto code recipe for images"
-    "property=og:description": "Volto code recipe for images"
-    "property=og:title": "Volto code recipe for images"
-    "keywords": "Plone, Volto, code, recipe, images"
+    'description': 'Volto code recipe for images'
+    'property=og:description': 'Volto code recipe for images'
+    'property=og:title': 'Volto code recipe for images'
+    'keywords': 'Plone, Volto, code, recipe, images'
 ---
```
stevepiercy, line 8: Please revert conversion of " to '.

docs/source/upgrade-guide/index.md
```diff
@@ -848,22 +850,22 @@ the block title. Before, it took the `id` of the block, which is utterly wrong a
 ### Variation field now uses the title instead of the id of the variation as translation source
 
 Following the same convention as the above change, `Variation` field coming from the block enhancers now uses the `title` of the block as source for translating
-the variation title. Before, it took the `id` of the block, which as stated before, is wrong and missleading. There is a chance that this change will trigger untranslated variation titles in your projects and add-ons.
+the variation title. Before, it took the `id` of the block, which as stated before, is wrong and missleading. There is a chance that  this change will trigger untranslated variation titles in your projects and add-ons.
```
stevepiercy: Revert.

```suggestion
the variation title. Before, it took the `id` of the block, which as stated before, is wrong and missleading. There is a chance that this change will trigger untranslated variation titles in your projects and add-ons.
```

docs/source/upgrade-guide/index.md
```diff
@@ -848,22 +850,22 @@ the block title. Before, it took the `id` of the block, which is utterly wrong a
...
+    variations: [
+      {
+        id: 'default',
+        isDefault: true,
+        title: 'Default',
+        template: DefaultListingBlockTemplate,
+        fullobjects: true
+      },
+    ]
 ```
```
stevepiercy: Revert. Extra indentation is not needed.

docs/source/upgrade-guide/index.md
```diff
@@ -884,7 +886,7 @@ This is not a strict breaking change, but it's worth mentioning it as it might b
 
 ### `getVocabulary` action changed its signature
 
-The `getVocabulary` action has changed API. Before, it used separate positional arguments, but now it uses named arguments by passing a single object as the argument. You'll have to adjust any call you do if you are using this action in custom code to the new API.
+The `getVocabulary` action has changed API. Before, it used separate positional arguments, but now it uses named arguments by passing a single object as the argument.  You'll have to adjust any call you do if you are using this action in custom code to the new API.
```
stevepiercy: Revert.

```suggestion
The `getVocabulary` action has changed API. Before, it used separate positional arguments, but now it uses named arguments by passing a single object as the argument. You'll have to adjust any call you do if you are using this action in custom code to the new API.
```

docs/source/upgrade-guide/index.md
```diff
@@ -1010,22 +1012,22 @@ config.blocks.blocksConfig.listing.templates = {
...
-];
+  config.blocks.blocksConfig.listing.variations = [
+    ...config.blocks.blocksConfig.listing.variations,
+    {
+      id: 'mycustomvariationid',
+      isDefault: false,
+      title: 'My custom listing variation',
+      template: MyCustomListingBlockTemplate,
+    }
+  ]
```
stevepiercy: Revert. Extra indentation is not needed.

docs/source/upgrade-guide/index.md
```diff
@@ -1237,14 +1239,14 @@ export default function applyConfig(config) {
...
-````{warning}
+```{warning}
 The add-ons you might be using might need to migrate to use the new
 configuration registry too. Make sure all of them are already migrated to Volto 12.
 
 ```{tip}
 Although this might be daunting, the migration is quite straightforward, and the refactoring
 of the required code can be undergone through a series of "search and replace" in your IDE of choice.
-````
+```
```
stevepiercy: The tip should be nested within the warning.

`````
````{warning}
The add-ons you might be using might need to migrate to use the new
configuration registry too. Make sure all of them are already migrated to Volto 12.

```{tip}
Although this might be daunting, the migration is quite straightforward, and the refactoring
of the required code can be undergone through a series of "search and replace" in your IDE of choice.
```
````
`````

File: docs/source/upgrade-guide/index.md [hunk 52]
```diff
@@ -1606,20 +1608,19 @@ theme folder. So, in your project's `theme.config` file, you can replace:
...
 ```less
 @themesFolder: '~volto-themes';
 @siteFolder: '~@package/../theme';
-@fontPath: '~volto-themes/@{theme}/assets/fonts';
+@fontPath: "~volto-themes/@{theme}/assets/fonts";
 ```
 
 You might consider moving your theme files to a subfolder called `site`, to
-prepare for the arrival of add-ons theming and their overrides. In that case,
+prepare for the arrival of add-ons theming and their overrides.  In that case,
```
stevepiercy: Revert.

```suggestion
prepare for the arrival of add-ons theming and their overrides. In that case,
```

docs/source/upgrade-guide/index.md
```diff
@@ -1684,7 +1685,7 @@ Add this key to the `jest.moduleNameMapper`:
   }
 ```
 
-because the new version of Jest is a bit more picky when importing externals. Attention, this mapping needs to be the first, it needs to come before the `@plone/volto/(.*)$` key.
+because the new version of Jest is a bit more picky when importing externals. Attention, this mapping needs to be the first, it needs to come before the ``@plone/volto/(.*)$`` key.
```
stevepiercy: Revert.

```suggestion
because the new version of Jest is a bit more picky when importing externals. Attention, this mapping needs to be the first, it needs to come before the `@plone/volto/(.*)$` key.
```

docs/source/upgrade-guide/index.md
```diff
@@ -1904,7 +1905,7 @@ Not a breaking change, but now there's a new setting in Blocks, `initialBlocks`
 
 ```js
 const initialBlocks = {
-  Document: ['leadimage', 'title', 'text', 'listing'],
+    Document: ['leadimage', 'title', 'text', 'listing' ]
```
stevepiercy: This looks odd to me.

docs/source/upgrade-guide/index.md
```diff
@@ -2097,7 +2098,7 @@ import '@plone/volto/../theme/themes/pastanaga/extras/extras.less';
 
 Enzyme has been removed, in favor of `@testing-library/react`, and the configuration should be removed in `package.json`:
 
-```diff
+``` diff
```
stevepiercy: Revert.

````suggestion
```diff
````

File: docs/source/upgrade-guide/index.md [hunk 56]
```diff
@@ -2159,7 +2160,7 @@ The edit block wrapper boilerplate was quite big, and for bootstraping an edit b
 
 In order to upgrade your blocks you should simplify the outer `<div>` (took as an example the Title block):
 
-```diff
+``` diff
```
```suggestion
```diff
```

stevepiercy (changes requested):
Linting went crazy! Please see comments and revert.

docs/source/recipes/images.md
pnicolli: Done

pnicolli:
> Linting went crazy! Please see comments and revert.

It is the other way around actually. All of those changes you saw there, it was me reverting my linter going crazy on the upgrade guide. In other words, I removed all the linting fixes that should not have been there because it's better if they go in a different PR. If you check the PR diff of all commits, you will see that the only change I am making to the upgrade guide in the master branch is this one:

image

stevepiercy:
> > Linting went crazy! Please see comments and revert.
> 
> @stevepiercy It is the other way around actually.

OK, I was looking at GitHub's "View changes" view.

stevepiercy (approved):

pnicolli:
@sneridagh I don't understand this PR https://github.com/plone/volto/pull/4964/files
I am inclined to resolve the conflicts by just removing all the changes done there for the teaser block DefaultBody component, because we don't need to handle a case where we don't have an Image component registered in the component registry. Am I missing something?

sneridagh:
@pnicolli a part of that was fixing the use case of having a local override of the teaser image. If you are referring to amend the instantiation of the registered component, then go ahead, by all means.

sneridagh:
@pnicolli btw, master now have the latest p.restapi version (but I guess you already realised).

pnicolli:
> @pnicolli a part of that was fixing the use case of having a local override of the teaser image. If you are referring to amend the instantiation of the registered component using the "new one" used in the core component, then go ahead, by all means.

@sneridagh The current implementation in this branch does the same exact thing you did in that PR, exept this does not fallback to a native `` tag. I would remove the fallback and all the code, basically restoring what you can see in this PR in that component, because we now have a component registered with the name `Image` by default in core volto.

pnicolli:
I really don't understand why the unit tests are failing here.

https://github.com/plone/volto/actions/runs/5589278409

This is the error:

```
PASS src/components/manage/Blocks/LeadImage/Edit.test.jsx
PASS src/components/theme/Widgets/SelectWidget.test.js
PASS src/components/manage/Blocks/LeadImage/LeadImageSidebar.test.jsx
PASS src/components/manage/Blocks/Video/Edit.test.jsx
PASS src/components/manage/Blocks/Maps/View.test.jsx
PASS src/helpers/Api/Api.test.js
node:internal/process/promises:279
            triggerUncaughtException(err, true /* fromPromise */);
            ^

Error: connect ECONNREFUSED 127.0.0.1:8080
    at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1278:16) {
  errno: -111,
  code: 'ECONNREFUSED',
  syscall: 'connect',
  address: '127.0.0.1',
  port: 8080,
  response: undefined
}
```

I don't get these errors on my local laptop, I actually get other errors, which are related to the Html SSR component and to auth token verifications - two things I did not touch at all anyway.

@plone/volto-team any pointers by any chance?

sneridagh:
@davisagli @pnicolli 🟢

sneridagh (approved):

sneridagh:
@stevepiercy I added some last minute docs regarding upgrade guide.

stevepiercy (changes requested):
Minor tweak, then it's good to go. Thank you!

stevepiercy (approved):

tisto:
@sneridagh do we have a PLIP for this feature? Anything else that is left to do here? This is something we definitely want to mention when talking about Plone 6.1.

sneridagh:
I thought we had one, but not. I think there's a plain issue: https://github.com/plone/volto/issues/1428

Issue #3238, opened 2022-04-08
image not visible in new created content


priyansh71:
**Describe the bug**
On creating a new event/news page, the image uploaded is not displayed.

**To Reproduce**
- Go to the https://6.demo.plone.org/
- After logging in, select the Add icon from the sidebar.
- Create a new event/ news.
- Add title, description and upload an image.
- The image uploaded is not visible. Neither on the list of events, nor in the detailed description of the content.

**Expected behavior**
The uploaded image from the admin panel must be visible.

**Screenshots**

<img width="1429" alt="Screenshot 2022-04-09 at 02 52 26" src="https://user-images.githubusercontent.com/77532581/162534712-d82ffd29-cb16-4da0-8279-127bdc4f6666.png">

**Software (please complete the following information):**

- OS: macOS 
- Browser: Chrome
- Volto Version : Current hosted version.
- Plone Version : Current hosted version.
- Plone REST API Version : Current hosted version.

**Additional context**

tisto:
IMHO the lead image should go and be replaced by the new preview_image. The lead image is a concept from Classic that does not make much sense in Volto. cc @plone/volto-team

giuliaghisini:
i think in content detail we have to display 'image' field, and in listing or other summary block type, we have to use PreviewImage component that displays 'preview_image' field if it's setted, or 'image' field

PR #3339, opened 2022-05-06
Show lead image in event view

danielamormocea:
Related to  #3238

### DIFF
--- a/src/components/theme/View/EventView.jsx
+++ b/src/components/theme/View/EventView.jsx
@@ -47,7 +47,15 @@ const EventView = (props) => {
       <Grid>
         <Grid.Column width={7} className="mobile hidden">
           {hasBlocksData(content) ? (
-            <RenderBlocks {...props} />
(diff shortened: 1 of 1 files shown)


sneridagh:
@danielamormocea could you add a changelog?

src/components/theme/View/EventView.jsx
```diff
@@ -47,7 +47,15 @@ const EventView = (props) => {
       <Grid>
         <Grid.Column width={7} className="mobile hidden">
           {hasBlocksData(content) ? (
-            <RenderBlocks {...props} />
+            <>
+              {content.image && (
+                <Image
```
nileshgulia1, line 52: @danielamormocea Couldn't we use `PreviewImage` component here?

davisagli:
Outdated. The change might still be relevant but it would need to be updated to the current repository structure, and follow current patterns for including image scales.

PR #3345, opened 2022-05-07
Enable Plone 6 acceptance tests, with fix for the p.a.robotframework …

sneridagh:
…fixture


datakurre:
@sneridagh Still working on this and continuing on Monday. Just `pip install` did not initially upgrade plone.app.robotframework to the checkout version. Then it seems zodb_setup/teardown did not trigger WSGI layer testSetUp testTearDown as I expected. Need to re-learn how robot-server really works with that and maybe add some fixes just to see if the fix works (so, for now, we still have not been restarting WSGI server between tests).

datakurre (approved):
LGTM

PR #3351, opened 2022-05-10
Add action to get configuration registry record.

ksuess:
Reason for this is to get the user group setting (many_users, many_groups).

An alternative would be to use the getControlpanel action. But for me it's in general easier to address the record, but a built in or add-on control panel.


ksuess:
It seems to be better to get 'many_users' and 'many_groups' via 'getControlpanel' action which checks permission.

PR #2890, opened 2021-12-05
Add storybook setup in scaffolding

tiberiuichim:
Based on https://github.com/plone/volto-quanta-project/tree/master/.storybook

I've added integration with addons `razzle.extend.js` main webpack modifier.

I think this is a simplistic setup right now, I think with a bit of effort we can change the main.js to load the addons from the package.json and then run storybook on all of them. I don't know. Thoughts on this?

CC @plone/volto-team

PR #3357, opened 2022-05-13
Fixed Storybook configuration for add-ons

pnicolli:
As promised in #2890, this is our configuration to use storybook in add-ons. It works in a project that is a few months old, can't remember the exact version but I can explore it further if this doesn't work.

I am creating this as a Draft for now, even if it could be final, because it needs a few tests with the generator that I can't do today.

### DIFF
--- a/.storybook/main.js
+++ b/.storybook/main.js
@@ -1,4 +1,5 @@
 const webpack = require('webpack');
+const fs = require('fs');
 const path = require('path');
 const makeLoaderFinder = require('razzle-dev-utils/makeLoaderFinder');
 const fileLoaderFinder = makeLoaderFinder('file-loader');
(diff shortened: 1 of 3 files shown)


.storybook/main.js
```diff
@@ -83,7 +84,15 @@ module.exports = {
...
-    // console.dir(resultConfig, { depth: null });
+    // Addons have to be loaded with babel
+    const addonPaths = registry.addonNames.map((addon) =>
+      fs.realpathSync(registry.packages[addon].modulePath),
+    );
+    resultConfig.module.rules[1].exclude = (input) =>
+      // exclude every input from node_modules except from @plone/volto
+      /node_modules\/(?!(@plone\/volto)\/)/.test(input) &&
+      // If input is in an addon, DON'T exclude it
+      !addonPaths.some((p) => input.includes(p));
```
pnicolli, line 95: I copied this over from our project, it is also added in the generator (see below), I'm not entirely sure it is needed in the core configuration as well. Needs testing.

sneridagh:
Tried it out, works like a charm... @pnicolli Thanks a lot for sharing!

PR #3362, opened 2022-05-16
Remove `div` as default if `as` prop from `RenderBlocks`.

sneridagh:


sneridagh:
@tiberiuichim @avoinea
We started to use our own dog food in core since the merge of the StyleWrapper.
Please take a look, I will merge it then.

tiberiuichim (approved):

avoinea (approved):

PR #3370, opened 2022-05-18
Extensions to the React18 PR

pnicolli:
This PR targets the react18 branch.

It aims to upgrade dependencies to the latest ones and to use the new SSR renderer available in React 18, `renderToPipeableStream`.

This is still a draft, the new renderer implementation still needs polishing and testing, and tests need fixes.

A few of this PR changes could be copied over to the original PR if the new renderer configuration takes long to be polished.


pnicolli:
> @pnicolli something from this one that we want to rescue for Volto 18? I will close the original issue for now.

Nah I don't think so. The main feature that I wanted to add here was renderToPipeableStream but it was way harder than expected :D I'd say let's close this and start over if and when possible.

PR #3373, opened 2022-05-19
Fix withStylingSchemaEnhancer enhancer mechanism

sneridagh:


tiberiuichim (commented):
@sneridagh I think we should do a hoc (per your initial design) instead of the component hacks, here https://github.com/plone/volto/blob/4a1df9aa85d7c0080b8fdab7bd4d71bda4a9a8b2/src/components/manage/Blocks/Block/StyleWrapper.jsx#L5

sneridagh:
I tried, but it breaks badly, because of the default values implementation, somehow 😞. Infinite updates I think... Didn't manage to debug it.

tiberiuichim:
@sneridagh then we should avoid avoid doing default values in component useEffects. We need to do it as a proper schema + formlib implementation, with default values computed with proper methods in the schema (even async). Maybe we need to think on using a form lib implementation, such as formik or whatever.

sneridagh:
I couldn't agree more... we should go a step back and evaluate again the defaults inplementation. 

However the including such a lib might be challenging and time consuming 😩

sneridagh:
@tiberiuichim let's merge this and take the suggestions into account.

PR #3384, opened 2022-05-22
Added preconnect link tag to speed up google font retrieval

### LINKED ISSUE #3328: Preconnect to the font origin
Related to #3310 

Preconnect to the font origin: https://web.dev/uses-rel-preconnect/?utm_source=lighthouse&utm_medium=devtools
It should be enough adding ``<link rel="preconnect" href="https://fonts.gstatic.com">`` in the head.

pnicolli:
Fixes #3328 

Add preconnect link tag as suggested by lighthouse. The current SemanticUI configuration requests 4 different font files from this origin, this preconnect should help optimize the connection.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ### Feature
 
+- Added preconnect link tag to speed up google font retrieval. @pnicolli
+
 ### Bugfix
 
 ### Internal
@@ -21,7 +23,7 @@
 
 ### Feature
 
-- Added viewableInBrowserObjects setting to use in alternative to downloadableObjects, if you want to view file in browser intstead downloading. @giuliaghisini
+- Added viewableInBrowserObjects setting to use in alternative to downloadableObjects, if you want to view file in browser instead downloading. @giuliaghisini
 - Disable already chosen criteria in querystring widget @kreafox
 - Added X-Forwarded-* headers to superagent requests. @mamico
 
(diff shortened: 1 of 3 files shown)


ichim-david:
@pnicolli @sneridagh although this improves the current google fonts font loading I wish we could have  a pluggable section in the head to add any needed pre-fetching without having to customize HTML.jsx.
We miss the slots from plone classic main_template where you could of added many things without resorting to
any overrides.

sneridagh (approved):

sneridagh:
@ichim-david agreed. We tried it in the past, but we can't use Pluggables here (they are not SSR-able). 
I could imagine a setting with an array or `Map()` of components, then render them all in order? So multiple add-ons could subscribe to it?

/cc @tiberiuichim ?

tiberiuichim:
We could use a similar model as the AppExtras, where we register in configuration components that will be rendered in that "slot".

sneridagh:
@tiberiuichim oh sure, I didn't recall that we refactored the `appExtras` in such a way. 😅

The question that I have now is that if @pnicolli addition should go into such a slot, or has sense to be always there by its own.

However, I would like to have such a slot in place soon, tbh.

mamico (changes requested):

tiberiuichim:
Slots can have "static" components as well. Maybe it's the moment to introduce them, at least the configuration and the static part? https://github.com/plone/volto/blob/slots_quanta_split_form_undo_prefix_path/src/config/slots.js#L4

pnicolli:
This could be superseded by #3492 and #3493 
I will either close or move this forward based on the result of those two.

PR #3386, opened 2022-05-23
Hero Block: blockExtensions support

danielamormocea:
Related to #2769


erral:
It would be nice, I don't know if part of this PR or in a new one, to have variations support in Hero block, this way this block can be easily reused to build any kind of block using a a title, text and an image (something like Teasers but without getting the information from a content of the site) to use it to build home-pages

nileshgulia1 (approved):

nileshgulia1:
@erral Variations support would be nice to have. As the current one is Hero-image left, it can be extended further to support any layout and dimensions. @danielamormocea  you have time to tackle this, may be in a separate PR?

src/components/manage/Blocks/Hero/schema.js
```diff
@@ -13,17 +13,25 @@ const messages = defineMessages({
...
+const schemaHero = ({ intl, variation }) => {
   return {
     title: 'Block settings',
     required: [],
     fieldsets: [
       {
         id: 'default',
         title: intl.formatMessage(messages.more),
-        fields: ['linkTitle', 'linkHref'],
+        fields: [
```
nileshgulia1, line 30: @danielamormocea We also need a Image Sizes widget, the current image renders smaller than those in image Block.

nileshgulia1: For the Background Image template, I could also imagine a `opacity` slider to alter the transparency for bg image....🤔

nileshgulia1 (changes requested):
LGTM apart from image sizes and few more options on bg image template

src/components/manage/Blocks/Hero/schema.js
```diff
@@ -13,17 +13,25 @@ const messages = defineMessages({
...
+          'linkTitle',
+          'linkHref',
```
nileshgulia1, line 32: Also I can't find where the `linkHref` is pointing to, at least it didn't worked as I tried.

src/components/manage/Blocks/Hero/Hero.jsx
```diff
@@ -0,0 +1,35 @@
+/**
+ * View image block.
```
nileshgulia1, line 2: Also please ensure the JS doc comments for functions are relevant.

nileshgulia1:
Also after adding the image, I'm unable to Edit `Title and description`, maybe we should provide fixed width to images.
Screenshot 2022-06-10 at 2 23 17 PM

src/components/manage/Blocks/Hero/schema.js
```diff
@@ -13,17 +13,31 @@ const messages = defineMessages({
...
+  Opacity: {
+    id: 'Opacity',
+    defaultMessage: 'Opacity',
+  },
 });
 
-const schemaHero = ({ intl }) => {
+const schemaHero = ({ intl, variation }) => {
   return {
     title: 'Block settings',
```
sneridagh, line 28: Watch out the title, is wrong. Please add a i18n string too.

src/components/manage/Widgets/NumberWidget.jsx
```diff
@@ -46,6 +47,7 @@ const NumberWidget = (props) => {
         max={maximum || null}
         value={value || defaultValue}
         placeholder={placeholder}
+        step={step || 1}
```
sneridagh, line 50: slipped change? Does this belong here?

sneridagh (changes requested):
What about a variation with the image on top?

Also, when the slate PR is merged, we will have to migrate the DraftJS fields to something else. I've been testing: `react-contenteditable` in  

https://github.com/kitconcept/volto-heading-block

Take a look.

/cc @tiberiuichim @ichim-david

src/components/manage/Widgets/NumberWidget.jsx
danielamormocea: I added this to have support for "opacity" filter, as it can go from 0 to 1 with a step of 0.1 (and in general the step is 1), and I also made use of this widget, as we do not have a slider one.

danielamormocea:
> What about a variation with the image on top?
> 
I don't think it is necessary, as a user can use "Image block" and then "text block" if they want a behaviour like this, the same with image on bottom.

tiberiuichim:
I would still see value in having a composite block of Title+Image, as it allows styling them together (for example a background color with padding)

nileshgulia1 (commented):
LGTM. Just one small change related to background-img variation. The text in the stretched image is to be centred. Thank you!

tiberiuichim:
I'll continue work on this PR

sneridagh:
@danielamormocea @tiberiuichim since we've removed the Hero block from core, I'd continue this as an add-on, if you are still interested. I will add it to the "rescuable during a sprint" project.

sneridagh:
This should be covered from an add-on, since we deprecated the Hero block in Volto 18. 
Closing for now.

PR #3389, opened 2022-05-23
Inject the style classNames in the main Block Editor wrapper.

sneridagh:
@ichim-david @tiberiuichim @pnicolli @giuliaghisini  

I've been developing a project with the new style wrapper and I really think that I missed the shot when deciding where to inject the classNames. 

I think that we should inject it at the very top of the Editor wrapper to be able to have enough flexibility and control over the CSS definitions in the Editor. I even think that we might save some definitions that work both for the editor and for the view mode. 

I also took a closer look at EEA's style wrapper, and I think that we need to replicate the inline styling in the current style wrapper. That's why I left it untouched, the idea is that we remove the class injection, but we leave a way to inject inline definitions (if any), what do you think?

From the merged PR thinkings:

@ichim-david Several "to-dos" from the other day talk:

I think we should include the current `size` and `align` data properties in the styling classname generator.

So now we have `has--align--` `has--size--` classnames. See questions below.

I'm convincing myself that when we talk about width containers, the wording is not "align"... although it's a widely used concept, but it Hink that "block/container width" would be more appropriate. What happens when, we have (fairly common) use case, that given a specific container width, you still want to "justify" the elements inside the block in a way (left, center, right). eg. A button block, or a heading block. 

Other use case that breaks the "align" wording is the block that has a background color, but the inner container is variable, let's say, a grid block that has a background color but the inner contents have a specific "inner width", that should be possible for a user to be able to configure.

So far, we have:
- "align/inline width" (images, videos, maps) blocks alignment (I think it's the only legit use case for that wording)
- "block/container width" (the outer container of a block)
- "inner width" (the inner container of a block, where the content lies)

Also, as legacy, we have these properties in the root of the `data` structure:
- size
- align 

all referring to image/video/maps but might be also present in custom blocks.

I think the style wrapper would have to deal with all these, and include the legacy size/align on it, generating a different set of CSS rules, so different classnames.

I do like a lot this approach, we need it for the next layout:

```
.addReducedTextElements() when (@reduceTextElementsSize = true) {
  @{textElementsParentSelector} {
    @{textElementsSelectors} {
      max-width: @textElementsMaxWidth;
      margin-right: @textElementsRightMargin;
      margin-left: @textElementsLeftMargin;
    }
  }
}
```

```
@textElementsSelectors: h2, h3, h4, h5, h6, p, ol, ul, blockquote;
```

However, as we discuss the other day, we would like more descriptive classnames, like

`container-lg`, etc...

and have the sizes be configurable. I agree with it, so they remain "neutral" and not tied to a name. In our last project we have:

```
:root {
  --layout-container-width: 1440px; // this is the header/footer container
  --default-container-width: 940px; // this is the "block inner width"
  --narrow-container-width: 620px; // this is the narrow (text elements) inner width
}
```

and the full-size (full viewport) in addition.

Questions:
- Does the prefix `has--` make any sense to you?
- Should we introduce these three types of container widths in the style wrapper?

### DIFF
--- a/src/components/manage/Blocks/Block/EditBlockWrapper.jsx
+++ b/src/components/manage/Blocks/Block/EditBlockWrapper.jsx
@@ -1,6 +1,9 @@
 import React from 'react';
 import { Icon } from '@plone/volto/components';
-import { blockHasValue } from '@plone/volto/helpers';
+import {
+  blockHasValue,
+  buildStyleClassNamesFromData,
+} from '@plone/volto/helpers';
 import dragSVG from '@plone/volto/icons/drag.svg';
 import { Button } from 'semantic-ui-react';
 import includes from 'lodash/includes';
@@ -31,6 +34,8 @@ const EditBlockWrapper = (props) => {
     ? data.required
     : includes(config.blocks.requiredBlocks, type);
 
+  const styles = buildStyleClassNamesFromData(data.styles);
+
   return (
     <div
       ref={draginfo.innerRef}
@@ -38,10 +43,7 @@ const EditBlockWrapper = (props) => {
       // Right now, we can have the alignment information in the styles property or in the
       // block data root, we inject the classname here for having control over the whole
       // Block Edit wrapper
-      className={cx(`block-editor-${data['@type']}`, {
-        [data.align]: data.align,
-        [data.styles?.align]: data.styles?.align,
-      })}
+      className={cx(`block-editor-${data['@type']}`, styles)}
     >
       <div style={{ position: 'relative' }}>
         <div
--- a/src/components/manage/Blocks/Block/EditBlockWrapper.jsx
+++ b/src/components/manage/Blocks/Block/EditBlockWrapper.jsx
@@ -34,7 +34,7 @@ const EditBlockWrapper = (props) => {
     ? data.required
     : includes(config.blocks.requiredBlocks, type);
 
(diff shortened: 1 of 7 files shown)


src/components/manage/Blocks/Block/EditBlockWrapper.jsx
```diff
@@ -31,17 +34,16 @@ const EditBlockWrapper = (props) => {
...
+
   return (
     <div
       ref={draginfo.innerRef}
       {...draginfo.draggableProps}
       // Right now, we can have the alignment information in the styles property or in the
       // block data root, we inject the classname here for having control over the whole
       // Block Edit wrapper
-      className={cx(`block-editor-${data['@type']}`, {
-        [data.align]: data.align,
```
ichim-david, line 42: @sneridagh  I understand the desire to unpack all of the styles from data.styles but I don't understand the removal of data.align from the  block-editor classes in the idea that if  someone will not enable the styling for that block they will not get the styling align and now it's own align field  classes will not be present.

sneridagh:
@ichim-david fixed. I will conduct more tests today with this shape.

sneridagh:
Let's let this rest for now, do more tests and play with it.

In the meantime: https://github.com/plone/volto/pull/3412

tiberiuichim:
Closing for now, we can revisit if needed

PR #3401, opened 2022-05-25
refactor table block

danielamormocea:
Related to #2769


src/components/manage/Blocks/Table/CellSidebar.jsx
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { BlockDataForm } from '@plone/volto/components';
+import { CellSchema } from './schema';
+import { useIntl } from 'react-intl';
+
+const CellSidebar = (props) => {
```
nileshgulia1, line 7: @danielamormocea We don't require two BlockDataForms. You can directly add multiple fieldsets (which have their own schema ) like in https://github.com/eea/volto-accordion-block/blob/master/src/components/manage/Blocks/Accordion/Schema.js#L72

nileshgulia1 (changes requested):
Please see my comments!

sneridagh:
Closing, as the current core Table block is the `@plone/volto-slate` one and it has already a schema.

PR #3406, opened 2022-05-29
Look ma, no AsyncConnect (useQuery, instead)

sneridagh:
I had this in my mind since some time ago...

It turns out it does work. PoC correct SSR working only for root route.

So, next steps, if we want to follow the rabbit hole:
- Make it work for all routes, detect Express route, pass it to react-query.
- Ask in a single SSR request for common expansions (breadcrumbs, navigation, etc)
- Put in place a way to extend that with route matching "a la AppExtras/slots" way
- Scale the useQuery "actions" as we do with the Redux-based ones now, app-wide architecture
- Think in a way to make coexist this with Redux API requests
- Eventually upgrade to RR6.4

### DIFF
--- a/package.json
+++ b/package.json
@@ -339,6 +339,7 @@
     "react-intl-redux": "2.2.0",
     "react-medium-image-zoom": "3.0.15",
     "react-portal": "4.2.1",
+    "react-query": "3.39.0",
     "react-redux": "7.2.4",
     "react-router": "5.2.0",
     "react-router-config": "5.1.1",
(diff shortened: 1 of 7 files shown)


tiberiuichim:
Nice! `withContent.js` is missing, and I think we should also be getting rid of the request triggers in the nav components (breadcrumbs, navigation, etc). I think we need to move completely to "static", introspectable methods of async data fetching. for backwards compatibility, we can keep the asyncExtenders configuration, as the function signature could be more or less the same (I don't know about redux store, but I feel the redux store may be needed in some of those functions).

sneridagh:
@tiberiuichim doh, pushed!

Indeed, I just added it to the to-do list. +1 to all.

davisagli:
Old proof of concept. I don't think it's useful to keep the PR open; we can still refer to it when it is closed.

PR #3407, opened 2022-05-30
Preview image widget for image fields

nileshgulia1:
Ready-only and unwrapped widget for showing previews of images in the sidebar pertaining to the field passed through block schema. Right now it works with the video and refactored-image block.


nileshgulia1:
> Unless I'm missing something, you don't need to generate this as part of the block sidebar form, it can be just a regular react component. See the example in the training

Yes, that's what I imagined in the first place. I noticed we have been doing this for a long time, But I'm trying to make this more generic in a way that it can be used directly like a "plug n play widget" similar to other widgets. This has been discussed in the last volto meeting. 
Now I see that there isn't any way to make volto widgets aware of block formData without explicitly passing formData through schema, which seems antipattern 🤔

nileshgulia1:
closing in favour of https://github.com/plone/volto/pull/2700/commits/adaebf4573434232afe1dab33c57d89759c77d4d

PR #3422, opened 2022-06-02
Add control panel 'User and Group Settings'

ksuess:


ksuess:
Obsolete with plone.restapi https://github.com/plone/plone.restapi/pull/1440

PR #3425, opened 2022-06-04
Bugfix: avoid rendering a '0' in the search block when no facets were added

tiberiuichim:
Fixes this little bugger. I can't really understand why something like this happens sometimes with the `&&` based conditional rendering

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Bugfix
 
+- Don't render junk when no facets are added to the search block @tiberiuichim
+
 ### Internal
 
 ### Documentation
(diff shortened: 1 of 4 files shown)


tiberiuichim:
Ah... maybe it should be "converted" to boolean, so a condition like `x.length > 0`

avoinea (approved):

tiberiuichim:
The ternary operator is not ideal, in that article Ken avoids the ugly looking code by showing JS code, but if you're dealing with jsx, it's silly (specially when the jsx tag is large).

```
{ something ? <Tag /> : null }
```
vs
```
{ !!something && <Tag /> }
```

sneridagh (approved):

PR #3430, opened 2022-06-06
Fix visibility of toolbar workflow dropdown for more states as fitting in .toolbar-content

ksuess:
all states should be visible or at least to be scrolled into view.


ksuess:
image

sneridagh (approved):

Issue #3414, opened 2022-05-31
BlockStyleWrapper CSS for multiple tabs


avoinea:
If `stylesSchema` has multiple `fieldsets` we should display it as an accordion and not as tabs:

![styling-with-multiple-tabs](https://user-images.githubusercontent.com/323385/171206610-ef3881b6-9b1f-4401-bcc7-04a738c4de38.png)

PR #3444, opened 2022-06-20
display style schema as accordion not tabs

danielamormocea:
related to  #3414

Screenshot 2022-06-20 at 17-27-22 Edit Page

### DIFF
--- a/.github/ISSUE_TEMPLATE/2_feature_request.md
+++ b/.github/ISSUE_TEMPLATE/2_feature_request.md
@@ -2,7 +2,7 @@
 name: '🦄 Feature request'
 about: 'Suggest an idea for Volto'
 title: ''
-labels: '04 type: enhancement'
+labels: ['04 type: enhancement', '30 needs: triage']
 assignees: ''
 ---
 
--- a/.github/actions/node_env_setup/action.yml
+++ b/.github/actions/node_env_setup/action.yml
@@ -3,7 +3,7 @@ name: Set up Node.js environment
 runs:
   using: "composite"
   steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v5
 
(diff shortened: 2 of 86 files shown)


danielamormocea:
Screenshot 2022-06-23 at 18-39-09 Edit Page

Screenshot 2022-06-23 at 18-38-47 Edit Page

I added a boolean option named "stylesSchemaTabsView" which can be passed to the Blocks. If it is true, we can have the Styling with tabs, and if it is false, then we have the accordions. 

I also talked with Madalina for a view on the UI situation, and she recommended to have the 'Styling' header just like the Block's name header, so it is cleaner and consistent. Also, in the accordions' case, we always have the 'Default' open when we first open the edit view sidebar, and there is just one accordion open at all times, so when we click on "Other" the "Default" closes.

tiberiuichim:
This would be a nice addition to the ObjectWidget, but it needs to be framed as an ObjectWidget configuration option, rather then something related to the schema.

nileshgulia1:
I see its targeted for 19.x. It's a nice feature to have. As Daniela is not working with us, I will assign it myself for the moment.

pnicolli:
@nileshgulia1 did you get any chance to go back to this? It doesn't look like breaking change but we are getting closer to a Volto 19 release and wanted to confirm.

nileshgulia1:
@pnicolli This has become outdated. I have to resurrect it. I will look at the conflicts and will update you with the current state of the PR.

sneridagh:
@nileshgulia1 please close this PR and create another one from the changes (copying code). To make sure the injection attempt with the forcepush is gone. Thanks.

nileshgulia1:
Sure, Victor. Thank you.

fredvd:
I have removed the referenced user who made the injection attempt from the Plone organisation

PR #3448, opened 2022-06-26
change passwordreset route

erral:
I am changing this here because we tried to do it in plone.volto patch of the password-reset function, and it breaks several Products.CMFPlone tests.

See:

https://github.com/plone/plone.volto/pull/61
https://github.com/plone/Products.CMFPlone/pull/3538
https://github.com/plone/plone.volto/pull/66

cc: @ionlizarazu @mauritsvanrees


mauritsvanrees (commented):
For the record, I agree with this, but I will refrain from approving Volto PRs usually.

sneridagh (changes requested):
I'm ok with this also. We might want to discuss it with the Team anyways, in case someone sees some other issues with existing deployments.

@erral also, there are a bunch of components where this route is used (Login, PasswordReset, etc)

cekk:
@erral @sneridagh i made a pr for this branch that fix other "passoword-reset" entries in the code.

I agree with this pr and this can keep compatibility with Plone.

erral:
I have included @cekk's additional changes here.

sneridagh (approved):

PR #3449, opened 2022-06-26
Change the password-reset route [15.x.x]

erral:
The same as in #3448 but for 15.x.x branch

cc: @ionlizarazu


sneridagh:
@erral we cannot introduce a breaking in the 15.x.x branch :(

PR #3456, opened 2022-07-02
Use type info instead of id type as icon title in the folder contents

mamico:
Use type info instead of id type as icon title in the folder contents. 
Type info is for humans and eventually translated.

before:
image

after:
image


tiberiuichim (approved):

pnicolli (approved):

tiberiuichim:
I think the "Pagina" title would be dependent on the current locale?

erral:
@tiberiuichim yes that response depends on the backend language, plone.restapi handles the translation of the portal_type and other items in the frontend

sneridagh (approved):
Does this works in p.restapi already, or it's due to merge?

sneridagh:
Oh yeah, it's brain serialization, merging!

Issue #3438, opened 2022-06-17
Properly internationalize the When component EventDatesInfo 


erral:
**Describe the bug**
The [When component](https://github.com/plone/volto/blob/master/src/components/theme/View/EventDatesInfo.jsx#L27) is not internationalized and has hard-coded `from` and `to` strings to signal event `from` and `to` dates:

https://github.com/plone/volto/blob/master/src/components/theme/View/EventDatesInfo.jsx#L90-L92

**Additional context**

This is not as easy as adding `formatMessage` calls in the `from` and `to` strings, we have to be aware that some languages do not work like that. For instance in basque the `from` and `to` strings are used like suffixes to the hour or date strings instead of prepositions, so this needs to be implemented using variables like `from ${start_date} to ${end_date}` or whatever.

erral:
For complete reference in classic Plone we are using `i18n:translate` and `i18n:name` tags to handle those translations: https://github.com/plone/plone.app.event/blob/master/plone/app/event/browser/formatted_date.pt#L13

PR #3458, opened 2022-07-04
i18n for EventDatesInfo

mbarde:
to fix https://github.com/plone/volto/issues/3438

### DIFF
--- a/src/components/theme/View/EventDatesInfo.jsx
+++ b/src/components/theme/View/EventDatesInfo.jsx
@@ -1,4 +1,5 @@
 import React from 'react';
+import { FormattedMessage } from 'react-intl';
 import PropTypes from 'prop-types';
 import { List } from 'semantic-ui-react';
 import cx from 'classnames';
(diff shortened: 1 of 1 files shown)


erral:
I think that this suggestion does not work. I think that we need something with parameters like in Classic UI:

https://github.com/collective/plone.app.locales/blob/master/plone/app/locales/locales/plone.pot#L8866-L8883

tiberiuichim:
@mbarde Yes, we need a full message with params, as @erral is suggesting. 

See the example in the docs:

```
function () {
  const messages = defineMessages({
    greeting: {
      id: 'app.greeting',
      defaultMessage: 'Hello, {name}!',
      description: 'Greeting to welcome the user to the app',
    },
  })

  return intl.formatMessage(messages.greeting, {name: <b>Eric</b>})
}
```

at https://formatjs.io/docs/react-intl/api/#formatmessage

erral:
Last week I tried unsuccessfully to build such messages, trying to keep existing HTML, but it should be possible like @tiberiuichim suggests

davisagli:
Closing since this is out of date and not our preferred solution.

Issue #3467, opened 2022-07-08
A11y : Maps Block


iRohitSingh:
**Describe the bug**
Frames should have a unique title attribute

**To Reproduce**

1. create "page "
2. Add two or more maps Block 
3. run "axeDevTool"
4. See error "Frames should have a unique title attribute"

**Screenshots**
![Screenshot from 2022-07-08 18-59-47](https://user-images.githubusercontent.com/61353484/178001718-efe824c9-ec9c-4b05-b641-839dce6d228d.png)

iRohitSingh:
@tisto I can fix it by putting the data.url in the title field but it doesn't make sense for the user. any feedback how i can fix it.

I don't know how to fix it.

data.url https://www.google.com/maps/embed?pb=!1m18!1m12!1m3!1d224345.83923100954!2d77.06889716447195!3d28.527582006607656!2m3!1f0!2f0!3f0!3m2!1i1024!2i768!4f13.1!3m3!1m2!1s0x390cfd5b347eb62d%3A0x52c2b7494e204dce!2sNew%20Delhi%2C%20Delhi!5e0!3m2!1sen!2sin!4v1657288342054!5m2!1sen!2sin

jackahl:
https://dequeuniversity.com/rules/axe/3.5/frame-title-unique

tisto:
Let's add a new field "alt-tag" to the maps block that can be used by the React component to generate a unique iframe title.

PR #3468, opened 2022-07-11
Fix a11y maps block

iRohitSingh:


iRohitSingh:
Screenshot from 2022-07-11 16-10-35

iRohitSingh:
fix : #3467

src/components/manage/Blocks/Maps/View.jsx
```diff
@@ -38,7 +38,9 @@ const View = ({ data, intl }) => (
       })}
     >
       <iframe
-        title={intl.formatMessage(messages.EmbededGoogleMaps)}
+        title={`${intl.formatMessage(messages.EmbededGoogleMaps)} ${
+          data.title
+        }`}
```
davisagli, line 43: @iRohitSingh I think we should simply show data.title as it was entered in the block schema, and not add "Embedded Google Maps of " as a prefix. Maybe the editor wants to provide a more detailed description of the map that does not fit grammatically with the prefix. Maybe some site customizes the view of this block to display maps from another source instead of Google.

(If we do keep the prefix, the spelling needs to be fixed: `Embedded` has 3 Ds.)

davisagli (approved):
LGTM

PR #3476, opened 2022-07-16
Pass full configuration registry object to add-ons on bootstrap time + component registry API refactor

sneridagh:
AKA named component lookup in Volto, FTW


src/registry.test.js
```diff
@@ -3,25 +3,41 @@ import config from './registry';
...
+  it('resolves unexistent component', () => {
+    expect(config.getComponent('Toolbar.Doh').component).toEqual(undefined);
+    expect(config.getComponent('Toolbar.Doh')).toEqual({});
   });
-  it('register component by name', () => {
-    config.register('Toolbar.Bar', { component: 'this is a Bar component' });
-    expect(config.resolve('Toolbar.Bar').component).toEqual(
+  it('registers a component by name', () => {
+    config.registerComponent('Toolbar.Bar', {
+      component: {
```
sneridagh, line 36: I don't like this... too ugly :(

src/registry.js
```diff
@@ -88,12 +88,21 @@ class Config {
     this._data.components = components;
   }
 
-  resolve(component) {
-    return this._data.components[component] || {};
+  getComponent(name, options = {}) {
```
sneridagh, line 91: Should we go to an all keywords arguments?

src/registry.js
```diff
@@ -88,12 +88,21 @@ class Config {
...
+    const { context = '' } = options;
+    const componentName = `${name}${context ? `|${context}` : ''}`;
```
sneridagh, line 93: I went for a string based composition for name... We could go for the key based one as well.

src/registry.test.js
tiberiuichim:
```
registerComponent({name: 'Toolbar', component: Toolbar, context: 'Bar'})
```
or:
```
registerComponent('Toolbar', Toolbar, 'Bar')
```
or:
```
registerComponent({name: 'Toolbar', component: Toolbar, dependencies: ['Bar'], options: {something: 'else'}})
```
I think I like the last?

tiberiuichim: We also use `dependencies` in the Pluggables implementation.

tiberiuichim (approved):

src/registry.test.js
sneridagh: Done

PR #3479, opened 2022-07-18
Remove `@unlock` endpoint reducer since it was having some bad intera…

sneridagh:
…ctions with the `RESET_CONTENT` action and the new `DefaultView`


sneridagh:
Mmm we can't do this because of the unlock button :( meh... closing

Issue #2927, opened 2021-12-16
Portrait/Avatar broken under Toolbar > PersonalTools panel


pigeonflight:
The user portrait returns as a broken link under the toolbar > personaltools.

**To Reproduce**
1.  Click on your avatar 
2. Click on 'Profile'
3. Upload a portrait 
![image](https://user-images.githubusercontent.com/31827/146381665-22b24d6e-de80-4828-8543-fffe4a0f69ca.png)

After it successfully uploads, the portrait will fail to load
For the admin user the output In the console looks like this:
```admin:1 GET http://localhost:3000/portal_memberdata/portraits/admin 404 (Not Found)```

**Expected behavior**
I expect the the portrait to load successfully and display in panel

**Software (please complete the following information):**
Volto 13.15.1
Plone 6.0.0a1
CMF 2.5.4
Zope 5.3
Python 3.9.8 (main, Nov 10 2021, 03:21:27) [GCC 8.3.0]
PIL 8.3.1 (Pillow)

pigeonflight:
I think it relates to the Plone restapi as the api returns the broken URL when you query the @users endpoint.
see the associated acton that makes the call to the api: https://github.com/plone/volto/blob/2b062ea2b69b4ab7bad1b2fa3c3da11a5d27c9ee/src/actions/users/users.js

giuliaghisini:
reproduced too

PR #3487, opened 2022-07-19
Issue #2927 Fix user avatars

jmevissen:


jmevissen:
The getNonAPIResourceWithAuth Function is questionable. 
But the User-Portrait is called like: http://localhost:8080/Plone/portal_memberdata/portraits/jm without '/++api++'.
Maybe the getAPIResourceWithAuth Function could be updated to conditionally not use the '/++api++' part.
Or an new Api resource for User-Portraits is needed.

sneridagh:
@jmevissen thanks for opening the PR! From my point of view, if the portrait URL is not accessible using the `++api++` traversal, it should be, and it's a backend issue. In short, all resources used by the frontend should be accessible using the api traversal, and if they do not, they should be fixed.

@davisagli @jensens any idea why it might not be traversable using the api traversal?

jmevissen:
@sneridagh thanks for your comment.
The User Portraits are called like: http://localhost:8080/Plone/portal_memberdata/portraits/jm
Where "portal_memberdata" is a MemberDataTool and "portraits" is a BTreeFolder2 that is an attribute of the MemberDataTool.

The ++api++ traversal does not work in this case and throws a KeyError quite early, 
when trying to call http://localhost:8080/Plone/++api++/portal_memberdata/portraits 
(https://github.com/plone/plone.rest/blob/master/src/plone/rest/traverse.py#L25)

jmevissen:
@sneridagh 
I think that options maybe:

- enable the ++api++ traversal to access attributes of Tools
- add a new rest-api service for user portraits

sneridagh:
@davisagli @jensens what do you think about this one?

My take is that we should make the ++api++ traversal work, but I don't like the idea to expose things inside a tool through that...

davisagli:
@sneridagh I need to think about it some more. I agree in principle that user portraits should be made available by the API. But on the backend member info is looked up via the pluggable auth service so we can't assume it is always stored in portal_memberdata; for example with membrane and Dexterity-based members it could come from a NamedBlobImage field on a content item.

jensens:
The whole portrait handling in the backend is a PITA and I would like to see `plone.api` and `plone.restapi` to abstract this away from the existing tools and to not depend Volto to traverse to `portal_memberdata` behavior at all . 

I.e. if it comes to `pas.plugins.ldap` there is still an ugly patch active https://github.com/collective/pas.plugins.ldap/blob/master/src/pas/plugins/ldap/monkey.py to get portraits into our old (IIRC GRUF based, or even earlier CMF based) approach. This never got lifted up to PAS. And IIRC other plugins do the same (so you can not use them together).

That said, I have no good idea why traversal fails.

sneridagh:
This gets superseded by: https://github.com/plone/volto/pull/3487

PR #3492, opened 2022-07-20
Staticize Poppins font to be compliant with EU privacy

giuliaghisini:


theme/themes/pastanaga/globals/site.overrides
```diff
@@ -1,3 +1,5 @@
 /*******************************
         Global Overrides
 *******************************/
+
+@import (multiple) '../extras/fonts';
```
nileshgulia1, line 5: @giuliaghisini Shouldn't we import `fonts.less` in https://github.com/plone/volto/blob/staticize_font/theme/themes/pastanaga/extras/main.less#L560 ?

sneridagh: @giuliaghisini I think @nileshgulia1 is correct, shouldn't we reference it from main.less?

giuliaghisini:
@nileshgulia1 @sneridagh @tiberiuichim moved fonts.less import in main.less

theme/themes/pastanaga/globals/site.overrides
pnicolli: To put it in site.overrides was my idea in the first place, sorry 😝 
We "removed" if from site.variables, so it felt like the most appropriate place. In addition, it is more connected with how semantic-ui works. I was not a fan of importing from a different folder though, so I'm ok with changing this as you suggest.

sneridagh (approved):

sneridagh:
Don't know why but the `cy.waitForResourceToLoad` helper has started to blow up with this PR. Pushing some fixes, but I don't get why...

nileshgulia1 (approved):

pnicolli (approved):

PR #3500, opened 2022-07-22
(docs)upgrade guide: @plone/volto-slate

nileshgulia1:
I think it's better if we move this to `upgrade guide` in Volto 16 section?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@
 
 ### Documentation
 
+- Migration guide: volto-slate -> @plone/volto-slate @nileshgulia1
+
 ## 16.0.0-alpha.15 (2022-07-21)
 
 ### Breaking
@@ -175,6 +177,7 @@
 - Fix RenderBlocks: path @ksuess
 - Fix field id creation in dexterity control panel to have slugified id @erral
 - Changed to get intl.locale always from state @ionlizarazu
+
 ### Feature
 
 - Updated Brazilian Portuguese translation @ericof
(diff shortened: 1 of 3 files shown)


stevepiercy:
From a higher level, we have a *lot* of [Plone upgrade guides](https://6.dev-docs.plone.org/upgrade/index.html). I think yet another upgrade guide, and developers' brains will explode. 🤯 

I agree that it belongs in the existing [Volto upgrade guide](https://6.dev-docs.plone.org/volto/upgrade-guide/index.html). I think that when people look to upgrade the frontend, they do not think only about Slate, even though that is a big part of the frontend upgrade.

stevepiercy:
FWIW, in the Plone 6 Coordination Team meeting yesterday, it was pointed out that this piece you are working on was needed to migrate from the old editor to Slate.

There is one more missing part, about how to transform JSON content from the old to the new editor. Sorry, I cannot recall technical details, but maybe @pbauer or @ksuess can fill that gap?

nileshgulia1:
@stevepiercy Yes, though the draftjs JSON content migration part is not clear to me as we have a [blocks-conversion-tool](https://github.com/plone/blocks-conversion-tool) which convert HTML into slate content, but not from draftjs. Maybe we require a plone addon for this, ?...to be discussed.

sneridagh:
@nileshgulia1 I agree with @stevepiercy , so far, what you wrote in this PR belongs to the main frontend upgrade guide.

The migration guide, on the other hand is another beast and we should find the correct place for it.

nileshgulia1:
#3508 looks more complete, Let's continue there.

PR #3503, opened 2022-07-24
Update generator dotfiles

ericof:
- Add .editorconfig to generated codebases
- Add .prettierignore to generated codebases


packages/generator-volto/generators/app/templates/.prettierignore
```diff
@@ -0,0 +1,2 @@
+src/**/CHANGELOG.md
```
sneridagh, line 1: I'd remove it from https://github.com/plone/volto/blob/02ccaba9044eac900a1982b88827e642edf5151c/packages/generator-volto/generators/app/templates/package.json.tpl#L16 as well (or instead).

PR #3507, opened 2022-07-25
Fix edge case in Cypress flaky tests "Cannot read properties of null …

sneridagh:
…(reading required)" when the schema action has not been fulfilled yet (for some reason) and the test tries to save.

I think I've found the cause of it... let's see. 

Another option would be that the reducer always return an empty object... @robgietema any reason why we opted to return null if the reducers are empty?

### DIFF
--- a/src/helpers/FormValidation/FormValidation.js
+++ b/src/helpers/FormValidation/FormValidation.js
@@ -184,8 +184,8 @@ const validateRequiredFields = (
 ) => {
   const errors = {};
   const fields = isEmpty(touchedField)
-    ? schema?.required
-    : intersection(schema?.required, keys(touchedField));
+    ? schema.required
+    : intersection(schema.required, keys(touchedField));
   map(fields, (requiredField) => {
     const type = schema.properties[requiredField]?.type;
     const widget = schema.properties[requiredField]?.widget;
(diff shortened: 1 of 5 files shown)


src/helpers/FormValidation/FormValidation.js
```diff
@@ -349,7 +349,7 @@ class FormValidation {
     touchedField = {},
   } = {}) {
     return validateFieldsPerFieldset(
-      schema,
+      schema || {}, // If schema is null for some reason (schema request haven't fulfilled yet) we need to normalize it
```
davisagli, line 352: @sneridagh so if the schema hasn't loaded yet, we effectively just skip validation. Is that okay? I think so, since the validation that counts should happen on the server anyway.

davisagli (approved):

davisagli:
Hmm, now there is a different error that sounds like it has the same cause:
```
CypressError: Timed out retrying after 5000ms: `cy.wait()` timed out waiting `5000ms` for the 1st request to the route: `schema`. No request ever occurred.`
```

So maybe the problem is that something is preventing requesting the schema in the first place?

sneridagh:
@davisagli I've seen this one too, schemas that do not load (or that take far too long in the test?) No clue why could that be happening.

src/helpers/FormValidation/FormValidation.js
sneridagh: yeah, maybe that is better.

davisagli (approved):

sneridagh:
LGTM

PR #3514, opened 2022-07-27
bring user-group-controlpanel into 15.x.x

erral:
I have cherry-picked the PR with your changes to master into 15.x.x

We have tested it and it works nicely, can you check it please @ksuess ?


ksuess (commented):
You may want to check https://github.com/plone/volto/blob/15.x.x/cypress/support/commands.js#L28-L49 to make tests green.

erral:
Thanks @ksuess this is green!

ksuess (changes requested):
I realised that I shouldn't alter the signature of action `listUsers`. I corrected in https://github.com/plone/volto/pull/3516/files You may want to bring this also to this PR in Volto 15. Unfortunatly it has unrelated test failures. Let's see if a re-run of the failing tests will be OK later. I started a re-run just now.

erral:
Thanks @ksuess cherry-picked that change too.

ksuess (approved):
LGTM, let's group!

sneridagh:
@plone/volto-team I'm +1 to merge this one, but keep in mind that we will have to maintain it too! Meaning, backporting all the fixes/updates of this feature.

I'm planning the product strategy for the upcoming years, so can have a clear policy on how to deal with things like this.

@erral thanks for taking care!
@ksuess thanks for the help!

erral:
@sneridagh yes, this is directly related to supported versions etc. We need a clear policy on that :)

sneridagh:
@erral @ksuess I was about to release it and I realised that it included a breaking change in the `listUsers` action (the arguments now are an object).

There's a way out though, how about to support both the object and the string arguments? detecting the kind of var type it is? So if it's string, we use an stripped down version of the action, working like before, and if it's called with an object, the full fledged one.

I can imagine someone calling the `listUsers` action from custom components and facing the breaking :( 

At first I'm opposed to revert and work on this, but what do you think? Could you take care? It should be done rather quick, since it's holding the release.

erral:
I can have a look at it later today, I will report here if I achieve something

erral:
@sneridagh @ksuess see https://github.com/plone/volto/pull/3587 pls.

PR #3524, opened 2022-07-28
Upgrade @plone/scripts

sneridagh:


yarn.lock
```diff
@@ -17522,6 +17552,24 @@ signal-exit@^3.0.0, signal-exit@^3.0.2, signal-exit@^3.0.3, signal-exit@^3.0.7:
...
+    debug "^4.3.3"
+
+simple-git@^3.3.0:
+  version "3.10.0"
+  resolved "https://registry.yarnpkg.com/simple-git/-/simple-git-3.10.0.tgz#f20031dd121d3c49e215ef0186a102bece3f89b2"
+  integrity sha512-2w35xrS5rVtAW0g67LqtxCZN5cdddz/woQRfS0OJXaljXEoTychZ4jnE+CQgra/wX4ZvHeiChTUMenCwfIYEYw==
+  dependencies:
+    "@kwsites/file-exists" "^1.1.1"
+    "@kwsites/promise-deferred" "^1.1.1"
+    debug "^4.3.4"
```
davisagli, line 17571: It looks like @plone/scripts and mrs-developer can both use simple-git 3.5.0, so it would be nice to avoid the two different versions added here

PR #3526, opened 2022-07-28
Updatebrowserlist

sneridagh:


davisagli:
I don't understand why it's failing to build. It looks like the problem is the nullish coalescing operator (`??`), but it is supported by Node 14 and I'm able to run `yarn build` locally with Node 14.

nileshgulia1:
> I don't understand why it's failing to build. It looks like the problem is the nullish coalescing operator (`??`), but it is supported by Node 14 and I'm able to run `yarn build` locally with Node 14.

@davisagli Maybe it's the yarn.lock that's bringing incorrect babel versions?

sneridagh:
@davisagli same here, locally works well... CI hiccup?

davisagli:
@sneridagh I was able to reproduce it locally after deleting node_modules. I still don't really get why it is necessary, but adding `@babel/plugin-proposal-nullish-coalescing-operator` to the babel configuration seems to help.

erral:
we had to add that babel plugin in a client's project with Node 14 in order to get the project compiled due to the ?? operator and upgrading it to Node 16 fixed also the issue

PR #3528, opened 2022-07-28
resolved more-always-show-root-contents issue

### LINKED ISSUE #3365: "more" actions always show root actions
**Describe the bug**

The "more" button (three dots on the sidebar) are not correctly updated if you start for example from /contents view in the site root and navigate.

**To Reproduce**

1. Go to homepage
2. Click on "Contents"
3. Click on one child (possibly a folderish content)
4. Click on "..." button (on the toolbar)
5. The opened menu has the title of the root and not the context. And it has the actions of the root (only sharing) and not the ones for the context

**Expected behavior**

See the current folder actions (state, view, history, sharing) and the title of the current context in the opened menu (in the toolbar)

Tested also here: https://volto.kitconcept.com/

MdSahil-oss:
fixes #3365

### DIFF
--- a/src/components/manage/Toolbar/Toolbar.jsx
+++ b/src/components/manage/Toolbar/Toolbar.jsx
@@ -27,6 +27,7 @@ import {
   listActions,
   setExpandedToolbar,
   unlockContent,
+  getContent,
 } from '@plone/volto/actions';
 import { Icon } from '@plone/volto/components';
 import { BodyClass, getBaseUrl, getCookieOptions } from '@plone/volto/helpers';
@@ -187,6 +188,7 @@ class Toolbar extends Component {
       menuComponents: [],
       loadedComponents: [],
       hideToolbarBody: false,
+      currentPathname: null,
     };
   }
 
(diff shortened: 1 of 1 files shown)


tiberiuichim (commented):
I don't think it's a good idea to trigger getContent from the toolbar... we need it under control, somewhere centralized.

src/components/manage/Toolbar/Toolbar.jsx
```diff
@@ -209,9 +211,11 @@ class Toolbar extends Component {
    * @returns {undefined}
    */
   UNSAFE_componentWillReceiveProps(nextProps) {
-    if (nextProps.pathname !== this.props.pathname) {
+    if (nextProps.pathname !== this.state.currentPathname) {
```
sneridagh, line 214: @MdSahil-oss are you sure we need to keep the state for this? I can see the pathname change in the contents view. So the route should be updating, if it's not, then it should in the parent component or how this component gets the pathname prop.

MdSahil-oss: @sneridagh Worked on this issue a long time ago, give me a while then I get back to you.

MdSahil-oss: @sneridagh I removed the state from the code and made changes optimal as per plone codebase. Now I'm sure its good to go.

sneridagh:
Seems that something is borked out in this PR, the GHA do not run...

stevepiercy:
> Seems that something is borked out in this PR, the GHA do not run...

@sneridagh I see all green now.

sneridagh:
I think that @tiberiuichim is right... we cannot place the getContent in the Toolbar... I'll test it later, but if I'm not mistaken, that will trigger a double content call (the one in App and this one.

I'll say we have to move it to the contents component, so it only triggers when it's present.

sneridagh:
Yeah, two contents: 
image

sneridagh:
@MdSahil-oss @tiberiuichim I've moved the re-fetch to Contents.
In the meanwhile ww've discovered a thinggy in the Volto routes definitions, I'll create another PR for fixing it.

sneridagh (approved):

PR #3508, opened 2022-07-26
Add volto-slate upgrade guide

sneridagh:

PR #3529, opened 2022-07-28
Volto slate upgrade guide cleanup

stevepiercy:
This addresses the missed review items from https://github.com/plone/volto/pull/3508

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,9 @@
 
 ### Documentation
 
+- Minor clean up of volto-slate upgrade guide. @stevepiercy
+
+
 ## 16.0.0-alpha.19 (2022-07-28)
 
 ### Breaking
--- a/docs/source/upgrade-guide/index.md
+++ b/docs/source/upgrade-guide/index.md
@@ -46,8 +46,8 @@ This is a major change and should be planned in advance before you install 16.0.
 
 ```{note}
 Volto 16 is intended to be the long-term support (LTS) version used when Plone 6 Final is released.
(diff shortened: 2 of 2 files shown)


stevepiercy:
Two unresolved questions from:
https://github.com/plone/volto/pull/3508

1. We say that `draftJS` will be removed in Volto 18, but we do not say to what version of Plone that might correspond. Is there a roadmap for removal of `draftJS` in a future version of Plone, such as Plone 6.2 or 7.0, or is it "whatever version number of Plone is at whenever we release Volto 18"?
2. Do we return a deprecation warning to the developer if they use draftJS? I don't know if that is even possible in Volto, but we do that in Pyramid when a deprecated method is called.

avoinea (approved):

sneridagh:
> Two unresolved questions from: #3508
> 
> 1. We say that `draftJS` will be removed in Volto 18, but we do not say to what version of Plone that might correspond. Is there a roadmap for removal of `draftJS` in a future version of Plone, such as Plone 6.2 or 7.0, or is it "whatever version number of Plone is at whenever we release Volto 18"? 

More the latter because, we do not know that yet. We will assign Volto releases to Plone versions but we can't tell upfront. I have a plan to write down the Plone's product strategy and discuss it during the conf with the Release team and people interested.

We did this "deprecated but will be removed two majors from now" before, it worked well. Allows people to adjust.

> 2. Do we return a deprecation warning to the developer if they use draftJS? I don't know if that is even possible in Volto, but we do that in Pyramid when a deprecated method is called.

We could throw a console message if the configuration is set to draftJS, but never did it.

sneridagh (approved):

PR #3536, opened 2022-07-31
Update to razzle 4 (still with webpack 4)

davisagli:
Based on #2351, but without trying to upgrade to webpack 5 at the same time.

What does razzle 4 give us?
- updated dependencies
- a path to using webpack 5
- an option to extract css in dev mode, which can avoid the flash of unstyled content

To do:
- [x] fix tests
  - [x] the generator acceptance test is failing because it can't find @plone/volto/start-server in build/server.js
- [x] determine and document any updates needed to addons --> looks like the old razzle plugin format (modify function) is still supported, with a warning
- [x] try it with a real project
- [x] document how to update projects
  - [x] update webpack config for compatibility with postcss 8
  - [x] update razzle plugin format
  - [x] add cache to .gitignore
  - [x] use --env=jsdom instead of --env=jest-environment-jsdom-sixteen
  - [x] replace jest-css-modules transform with identity-obj-proxy
- [ ] after merging this:
  - [ ] release @plone/scripts and update the version used by volto itself, so we don't require 2 different versions of razzle


davisagli:
@sneridagh @tiberiuichim I think this is ready for another look. How do you feel about including it in volto 16, or should we wait? I've done some testing of this with a project using yalc and I think the risk from the changes in Razzle itself is fairly low, but perhaps there are adjustments needed for updates to its underlying libraries that I haven't found yet.

sneridagh:
Sure! I'll give it a try, also @plone/volto-team if someone else can try it out also and provide feedback, it would be great!

davisagli:
@sneridagh @plone/volto-team Any feedback? I'm wondering if we can include this for Volto 16, or if I should put it on the backburner.

sneridagh:
We talked about it yesterday, all in favor to onclude it in 16. Some of us we will try it out and give feedback, if all ok, we can merge!

I'll try it out in our latest project too.

sneridagh:
@davisagli I tried in our last project and I can't make it work. 

I even went and followed what the GHA does (project generator tests) and no luck either:

```
❯ yarn start
yarn run v1.22.17
$ razzle start
 WAIT  Compiling...

/Users/sneridagh/Development/plone/volto/testproject/node_modules/@plone/volto/razzle.config.js:234
  const { include } = options.webpackOptions.babelRule;
          ^

TypeError: Cannot destructure property 'include' of 'options.webpackOptions.babelRule' as it is undefined.
    at defaultModify (/Users/sneridagh/Development/plone/volto/testproject/node_modules/@plone/volto/razzle.config.js:234:11)
    at modifyWebpackConfig (/Users/sneridagh/Development/plone/volto/testproject/node_modules/@plone/volto/razzle.config.js:317:27)
    at /Users/sneridagh/Development/plone/volto/testproject/node_modules/@plone/scripts/node_modules/razzle/config/createConfigAsync.js:1023:22
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

Any clue?

I'll try to rerun the test now (now shows green).

sneridagh:
@davisagli ok, I was missing to yalc the missing @plone/scripts Now it works, and the build is still green.

sneridagh:
@davisagli I guess that all boils down to have released an updated @plone/scripts version, since all add-ons rely on it.

sneridagh:
@davisagli I did some tests, and I guess we could remove the hard dependency on `razzle`, and just left the `babel-preset-razzle` one. It will force us to update all add-ons anyways, but at least they will be more lightweight on install. They will need to update because of the hoisting. We must make sure only one set of versions of babel plugins are in place. We had a lot of problems with that and `yarn` over the years.

I can't really recall why I introduced it, I think that it was from the days we tried to make the project inside an add-on idea... which we abandoned in favour of `npx @plone/scripts addon clone ...`.

davisagli:
@sneridagh Yeah, I think you're right

tiberiuichim:
anything that would reduce ongoing babel dependencies pain is worth taking the initial "hit"

package.json
```diff
@@ -398,11 +395,11 @@
     "@storybook/addon-essentials": "^6.3.0",
     "@storybook/addon-links": "^6.3.0",
     "@storybook/react": "^6.3.0",
-    "babel-loader": "^8.1.0",
+    "babel-loader": "8.2.2",
     "full-icu": "1.4.0",
     "identity-obj-proxy": "3.0.0",
-    "jest": "27.4.5",
-    "jest-environment-jsdom-sixteen": "1.0.3",
+    "jest": "26.6.3",
```
sneridagh, line 401: I did a final check, I think Tiberiu bumped this to 27 at some point for some reason. 
Is there any limitation that we should stick to 26?

package.json
```diff
@@ -308,24 +302,24 @@
     "lodash": "4.17.21",
     "lodash-move": "1.1.1",
     "lodash-webpack-plugin": "0.11.5",
-    "luxon": "1.22.0",
```
sneridagh, line 311: We had problems with this in the past, weird resolution at compilation. That's why we needed to lift it here. Probably yarn problems. We should keep an eye on this one if we remove it.

packages/generator-volto/generators/app/templates/package.json.tpl
```diff
@@ -142,7 +141,6 @@
     "eslint-plugin-prettier": "3.1.3",
     "jest-junit": "8.0.0",
     "mrs-developer": "*",
-    "postcss": "8.3.11",
```
sneridagh, line 145: This is necessary for add-ons, if they have stylesheets, for current stylelint to work well.

sneridagh (changes requested):
One thing only, and two comments.

package.json
sneridagh: ok, razzle breaks :) It seems that something is wrong in the API after 26, and the razzle test script:

```
TypeError: Cannot read properties of undefined (reading 'onlyChanged')
    at _default (/Users/sneridagh/Development/plone/volto/node_modules/razzle/node_modules/@jest/core/build/getChangedFilesPromise.js:49:20)
    at /Users/sneridagh/Development/plone/volto/node_modules/razzle/node_modules/@jest/core/build/cli/index.js:296:71
    at Generator.next (<anonymous>)
    at asyncGeneratorStep (/Users/sneridagh/Development/plone/volto/node_modules/razzle/node_modules/@jest/core/build/cli/index.js:108:24)
    at _next (/Users/sneridagh/Development/plone/volto/node_modules/razzle/node_modules/@jest/core/build/cli/index.js:128:9)
    at /Users/sneridagh/Development/plone/volto/node_modules/razzle/node_modules/@jest/core/build/cli/index.js:133:7
    at new Promise (<anonymous>)
    at /Users/sneridagh/Development/plone/volto/node_modules/razzle/node_modules/@jest/core/build/cli/index.js:125:12
    at _run (/Users/sneridagh/Development/plone/volto/node_modules/razzle/node_modules/@jest/core/build/cli/index.js:374:20)
    at /Users/sneridagh/Development/plone/volto/node_modules/razzle/node_modules/@jest/core/build/cli/index.js:178:13
```

@tiberiuichim what do you think? is it fine for you to get back to 26?

davisagli: I was just trying to match what Razzle is using. Yeah, let's see what @tiberiuichim thinks...

package.json
davisagli: I have a better idea now about how that could be a problem than I did when I removed it. We can add it back.

packages/generator-volto/generators/app/templates/package.json.tpl
davisagli: Yeah, I'm not sure what I was thinking here. Also I think razzle expects it to be a peer dependency, so needs to be specified at the top level.

package.json
sneridagh: Tiberiu said it was ok for now. We can push for having it working later if needed. I added the breaking notice as well.

sneridagh:
https://github.com/plone/volto/pull/3678 last one before merging.

PR #3537, opened 2022-07-31
Use the @navroot and @site controlpanels to render the <title> and the logo

### LINKED ISSUE #3530: Logo title and alt values should come from Plone site configuration and not be hardcoded in i18n messages
See:

https://github.com/plone/volto/blob/master/src/components/theme/Logo/Logo.jsx#L36-L45

### LINKED ISSUE #3531: Site (Navigation Root) title not included in the TITLE of the site
In Plone the `<title>` tag of the site is composed by the current object's `.Title()` and the navigation root's `.Title()`, right now it only shows the Title of the object.

We may need a new endpoint in restapi to expose the Site Title and the Navigation Root title.

erral:
To be tested with plone.restapi >= 8.40.0

- Use navroot information to add navigation root title as the second part of the <title> tag like it is done in Classic Plone.
- Use the site endpoint to render the Logo
- Provide a widget to set the logo in the site control panel

Fixes #3531 #3530


erral:
Mmm.... I don't know how to modify the `<Helmet />` component to always inject the navroot title when used with custom titles (for instance in custom views, but that should be an edge case anyway...

erral:
@ionlizarazu

erral:
After his PR, we will be always requesting to expand the `@navroot` component of the REST API. We are already doing that for the `@translations` endpoint in a multilingual site, so I think that's acceptable

erral:
I have reworked this to use explicit dispatch calls to @navroot and @site instead of relying on expansion

erral:
I am aware that these changes are quite big and need further discussion, but I think it is worth to have them in Volto:

- Build the <title> tag using the information coming from the brand new `@navroot` information a la Plone Classic
- Provide a widget to set the site logo and use the `@site` endpoint to render the logo in the site
- Modify the search box action using the `@navroot` url.
- To achieve that and in order to avoid extra requests to the REST API I have included the `navroot` expansion in all requests. This is an arguable decision but I'm happy to reconsider it.

I would be happy to discuss them, even treating them as breaking changes or splitting this PR in 2 (<title> changes and Logo changes).

In the past @JeffersonBledsoe and @avoinea were interested in such changes, so we can comment them here.

src/components/theme/ContentMetadataTags/ContentMetadataTags.jsx
```diff
@@ -45,10 +45,16 @@ const ContentMetadataTags = (props) => {
 
   const contentImageInfo = getContentImageInfo();
 
+  let title_tag_content = seo_title || title;
+  let nav_root_title = props.content['@components']?.navroot?.navroot?.title;
```
dobri1408, line 49: Hi @erral, I see that you hardcoded the introduction of the site title/language root title here. I was thinking that maybe it is better to work as this https://github.com/plone/volto/issues/4723, as it is more customizable. I already did this in my pr, which is now closed, could you include this? Check the getTitle() function from https://github.com/plone/volto/pull/4862/files

dobri1408: With this approach, you allow to change the separator character though settings. This might be better because other projects that use volto can avoid the customization of this.

erral: I like the approach @dobri1408

ksuess (approved):

src/actions/navroot/navroot.js
```diff
@@ -0,0 +1,16 @@
+import { GET_NAVROOT } from '@plone/volto/constants/ActionTypes';
+
+/**
+ * Set sidebar tab function.
+ * @function getSite
+ * @returns {Object} Get the Navroot information
+ */
```
ksuess, line 7:
```suggestion
/**
 * Get the navigation root information.
 * @function getNavroot
 * @returns {Object} navroot
 */
```

src/actions/site/site.js
```diff
@@ -0,0 +1,16 @@
+import { GET_SITE } from '@plone/volto/constants/ActionTypes';
+
+/**
+ * Set sidebar tab function.
+ * @function getSite
+ * @returns {Object} Get the Site information
+ */
```
ksuess, line 7:
```suggestion
/**
 * Get the Site information.
 * @function getSite
 * @returns {Object} site info
 */
```

ksuess (changes requested):

src/components/theme/ContentMetadataTags/ContentMetadataTags.jsx
erral: @dobri1408 I have reworked this with your suggestion. 

now the default setting is not to show the site title (and thus it's not a breaking change).

Moreover the site title is not shown if it is equal to the page title (like it is done in Classic)

src/components/theme/ContentMetadataTags/ContentMetadataTags.jsx
```diff
@@ -45,10 +45,28 @@ const ContentMetadataTags = (props) => {
...
 
+  const getTitle = () => {
+    const includeSiteTitle =
+      config?.settings?.siteTitleFormat?.includeSiteTitle || false;
+    const titleAndSiteTitleSeparator =
+      config?.settings?.titleAndSiteTitleSeparator || '-';
+
+    const navRootTitle = props.content['@components']?.navroot?.navroot?.title;
+    const siteRootTitle =
+      props.content['@components']?.site?.['plone.site_logo'];
```
dobri1408, line 56: I tested and it works, but I have one tiny question. Why plone.site_logo here?

erral: Sorry, it's a mistake :wink:

erral: fixed,

src/components/manage/Widgets/SiteLogoWidget.jsx
```diff
@@ -0,0 +1,208 @@
...
+    : '';
+
+  /**
+   * Drop handler
+   * @method onDrop
+   * @param {array} files File objects
+   * @returns {undefined}
+   */
+  const onDrop = (files) => {
+    const file = files[0];
```
davisagli, line 91: This should be updated to use the validateFileUploadSize helper I just added in https://github.com/plone/volto/pull/4868

src/components/manage/Widgets/SiteLogoWidget.jsx
```diff
@@ -0,0 +1,208 @@
...
+const SiteLogoWidget = (props) => {
+  const { id, value, onChange, isDisabled } = props;
+  const intl = useIntl();
+
+  const fileName = value?.split(';')[0];
+  const imgsrc = fileName
+    ? `${toPublicURL('/')}@@site-logo/${atob(
+        fileName.replace('filenameb64:', ''),
+      )}`
+    : '';
```
davisagli, line 82: Would it be possible to use the existing FileWidget and not duplicate so much code, if that widget could be configured with how to derive the src from the current value?

src/actions/content/content.js
```diff
@@ -150,7 +150,9 @@ export function getContent(
           b_size: settings.defaultPageSize,
         }
       : {},
-    settings.isMultilingual ? { expand: 'translations' } : {},
+    settings.isMultilingual
+      ? { expand: 'translations,navroot' }
+      : { expand: 'navroot' },
```
davisagli, line 155: Could it be added via the apiExpanders config setting, instead of here? That would make it easier to turn off for a large site that is trying to minimize the number of backend requests needed to render a page.

davisagli (commented):
@erral Nice work. I think this is an important usability improvement for small volto sites. I'm on board as long as it's not too hard to turn off for a large site that is optimizing performance and wants to use customized logo and metadata components that hardcode things rather than depending on the additional backend requests.

src/components/manage/Widgets/SiteLogoWidget.jsx
erral: I tried to do so. But the point is that the backend saves the logo info differently from what the FileWidget expects. The FileWidgets builds a JSON representation of the file like it is required by the REST API, but the logo is saved in the plone registry, in a bytes field in a specific format. That's why I had to write an register a specific Widget for it.

src/actions/content/content.js
erral: Yes, this is something I wanted to discuss, specifically regarding performance.

We have code spread through Volto doing the ```if !hasApiExpanders(...)``` and first I thought to write this like that. 

On the other hand I saw that this ```navroot``` call was being done in every request after implementing this PR, so I thought it was worth to request the navroot expander by default. Moreover, I initially thought that we should be doing the same for at least the breadcrumbs expander too.

Then I saw [this config.settings setting called ```apiExpanders```](https://github.com/plone/volto/blob/master/src/config/index.js#L78) which really does what I say (expand the query by default with given expanders), but it is not used at all in plain Volto, which makes me think...

I would say (but that's perhaps a biased opinion because of this PR), that Volto should by default use that apiExpanders settings, to expand by default breadcrumbs, navroot, site and navigation, which are used anyway to build a stock Volto site, but let disable them using the config. This way the code could be written defensively using the ```is !hasApiExpanders```...

I don't know... I am open to discuss.

src/components/manage/Widgets/SiteLogoWidget.jsx
davisagli: ok

src/actions/content/content.js
davisagli: Yes, I think we want to use apiExpanders by default for the expanders that are needed everywhere, but we were just proceeding cautiously. Right @sneridagh? Can we add to the default expanders for volto 17 or is there a reason to still be cautious?

erral:
The tests are broken because they rely on the endpoints that are not available yet in the backend images

src/components/manage/Widgets/SiteLogoWidget.jsx
erral: Added the validation, thanks!

src/actions/content/content.js
erral: I have reworked this. I have gone the `!hasApiExpanders` way and let as it is. We can come to the use of `apiExpanders` setting later :)

ksuess (approved):

src/reducers/navroot/navroot.js
```diff
@@ -0,0 +1,51 @@
+/**
```
sneridagh, line 1: These reducers need to write to the ['components'] expanders, if the expanders are defined. See breadcrumbs for more info. I'm also missing tests for them.

src/components/theme/SearchWidget/SearchWidget.jsx
```diff
@@ -95,7 +95,13 @@ class SearchWidget extends Component {
    */
   render() {
     return (
-      <Form action="/search" onSubmit={this.onSubmit}>
+      <Form
+        action={
+          `${this.props?.content?.['@components'].navroot?.navroot?.['@id']}/search` ||
```
sneridagh, line 100: This is using the ['@components'] expander without knowing if the expander is in place or not.

sneridagh: Instead of prop drilling the `content` from App then Header then here, it's not better to call the useSelector for content and get it only here?

sneridagh (changes requested):
Take a look at comments. 
I think we should start introducing the "all in" expanders in core. This might be a good time for do it. IMHO, introduce two more requests in core by default I think it's not an option.

Did you take an eye on SSR? It does behave well? It's something worth to be sure it does.

src/actions/content/content.js
sneridagh: I'd introduce the expanders "all in" right now. There's nothing that prevents us for doing it in 17.

src/reducers/navroot/navroot.js
erral: I wasn't aware of it. I will check it.

src/components/theme/View/View.jsx
```diff
@@ -123,6 +128,13 @@ class View extends Component {
     if (!hasApiExpander('actions', getBaseUrl(this.props.pathname))) {
       this.props.listActions(getBaseUrl(this.props.pathname));
     }
+    if (!hasApiExpander('site', getBaseUrl(this.props.pathname))) {
+      this.props.getSite(getBaseUrl(this.props.pathname));
+    }
+    if (!hasApiExpander('navroot', getBaseUrl(this.props.pathname))) {
+      this.props.getNavroot(getBaseUrl(this.props.pathname));
+    }
+
```
pnicolli, line 137: I am not a fan of this, especially for the `@site` endpoint, it's possibly a lot of data moving around on every navigation, data that will very likely not change. I wouldn't fetch the `@site` on every view render, I wouldn't even want it in expanders.

I would take a different approach here, I suggest fetching this once on the first SSR and saving it in redux and be done with it until next full page reload. At least for the `@site` endpoint, not sure about the `@navroot`, because of possible nested navroots in the future. And also I expect navroot data to be smaller.
My rationale here is: what are the chances that the logo and other site controlpanel data change while a user is navigating and what are the chances that they care about seeing the new one in their specific session? They will get the new config the next time they fully load a page of our site.

This is exactly the approach I was going for in my PoC last year and I really believe the users are happy with just that: https://github.com/collective/volto-site-settings/blob/main/src/index.js#L9
In this PoC I am using the extender in config but there might be a better spot in core Volto for SSR only queries.

src/components/theme/SearchWidget/SearchWidget.jsx
pnicolli: Agree against prop drilling. With expanders handled correctly, it should be possible to always get this info from the navroot redux store with useSelector

src/components/theme/View/View.jsx
erral: Yeah, I initially though about something like that, but somehow went for the *easy* way of querying the `@site`. I will look at the SSR call thingy, which makes total sense for the use-case.

erral:
I have reworked the PR following @sneridagh and @pnicolli's suggestions (use hasApiExpander helper and add conditionally the navroot content like it is done in the breadcrumbs reducer); added the `@site` information in the SSR phase (I did what @pnicolli suggested, I didn't find any other way to inject a custom call to the backend during the SSR); and added some more tests for the actions/redurects, the modified components and the newly added widget.

Anyway the a11y cypress test are failing, I suspect that it is because of plone.restapi 8.40.0 not being available in the acceptance tests docker image. I tried adding it like @sneridagh did for the standard docker image, but it didn't work.

Does anyone know how can I do it?

news/3537.feature
```diff
@@ -0,0 +1,5 @@
+- Use the @navroot endpoint to build the title tag @erral
```
stevepiercy, line 1: These should not be list items.

```suggestion
Use the `@navroot` endpoint to build the `title` tag. @erral
```

news/3537.feature
```diff
@@ -0,0 +1,5 @@
...
+
+- Use the @site endpoint to render the logo @erral
```
stevepiercy, line 3:
```suggestion
Use the `@site` endpoint to render the logo. @erral
```

news/3537.feature
```diff
@@ -0,0 +1,5 @@
...
+
+- Register a widget to set the logo in the site controlpanel @erral
```
stevepiercy, line 5:
```suggestion
Register a widget to set the logo in the site control panel. @erral
```

stevepiercy (changes requested):
Some MyST and English syntax and grammar fixes. Thank you!

I reviewed only the docs and news item. A maintainer should review the code.

docs/source/configuration/settings-reference.md
```diff
@@ -19,7 +19,7 @@ This list is still incomplete, contributions are welcomed!
 
 They are exposed in `config.settings`:
 
-```{glossary}
+````{glossary}
```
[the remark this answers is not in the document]
erral: ains, this was not wanted, but my VScode linter does very strange things with markdown files, sorry about that...

sneridagh:
@erral Don't forget to add the new expander to the list!

https://github.com/plone/volto/pull/3537/files#diff-1523e3000d64e09cd6eb8370a4568715090b6763fd9192f61ab6044ac6d86fa9R238

I've merged the Expanders PR yesterday.

stevepiercy (approved):
Docs look good now. Thank you!

sneridagh:
@erral just pushed a fix, after this, it should be ready.

erral:
At last this is green!!!! :sweat_smile:

sneridagh:
I for got to Approve it, but I do :)

PR #3540, opened 2022-07-31
Linked headlines

tiberiuichim:


tiberiuichim:
@sneridagh This works, but the cypress and a11y tests need to be fixed. I'll need assistance on this, also maybe the CSS needs to be redone.

avoinea:
@tiberiuichim 

Hmm. This breaks the **Table of contents** block as it is generated based on the blocks uids. 

Also, @razvanMiu confirmed that the block uids are used somewhere within the Link widget to link to sections within the page.

Thus, I think it's safer for now to go with the block UIDs in order not to introduce it as a breaking change and in a separate PR/PLIP we can talk about changing the heading ids to more friendly names (maybe the same way we do with the **Override the TOC entry** checkbox).

tiberiuichim:
"breaking the toc" is not a blocker, we can fix the toc block.

It would represent a major breaking change for Volto, though.

tisto:
@tiberiuichim I checked out this branch and I quite like it! I guess we always thought too hard about how an editor could enter the anchor ID in the UI. Though, this solution is both elegant and simple. The only possible downside that I see is that links can break easily when editors change a headline.

avoinea:
@tisto This feature is also available as an add-on until it hits Volto Core. See: https://github.com/eea/volto-anchors

nileshgulia1:
I was also thinking about the adaptability of this feature, whether make it available per content-type, or per block?

tiberiuichim:
> The only possible downside that I see is that links can break easily when editors change a headline.

Yes, it's a downside. But I would argue that we shouldn't focus too much on this type of scenario, it's counterproductive. Honestly, because this PR is modeled after the github readmes, it happened to me many times that I arive at a document where the link is no longer valid. And it's frustrating, yes, but in reality I could scan the destination document to find the info I was looking for. Or I could look at the URL fragment and understand what the new section would be. Or the document was fully refactored, split into multiple pages, so the link would no longer be valid anyway.

What we're gaining is simplicity for the "happy case".

tisto:
@tiberiuichim would it be possible to enable/disable this via a feature toggle? This would give us some flexibility in which Volto version we could ship this feature.

tiberiuichim:
@tisto sure!

Even in its current form, it's ultimately configurable, as we're replacing the way the hx elements are rendered, in the config:

image

tisto:
@tiberiuichim awesome!

nileshgulia1:
🎉

PR #3534, opened 2022-07-29
support for many_users and many_groups flag in /users controlpanel

nileshgulia1:
…panels

PR #3544, opened 2022-08-02
Test against Plone 5.2.9 and 6.0.0b1

davisagli:


nileshgulia1 (approved):

nileshgulia1:
@davisagli I also need this for https://github.com/plone/volto/pull/3534, Are the test failures related?

davisagli:
I assume the test failures are related and need a closer look. I won't have time this evening.

sneridagh:
I can see some tests that is requiring folders, and I think latest `plone.volto` removed it from the default set of available content types.

sneridagh:
Also, related:

https://github.com/plone/plone.volto/pull/79

This one is important, because the default profiles use slate too. Although tests might dodge this (by not testing on the DX root).

davisagli:
> I can see some tests that is requiring folders, and I think latest plone.volto removed it from the default set of available content types.

@sneridagh So I guess we need to update the tests to use folders or documents based on the Plone version?

@nileshgulia1 the empty KGS was intentional (though maybe not exactly correct yet). On Plone 5.2 we need some extra version pins to override the ones from dist.plone.org, but on Plone 6 the ones from dist.plone.org should be adequate. I hope.

nileshgulia1:
>@nileshgulia1 the empty KGS was intentional (though maybe not exactly correct yet). On Plone 5.2 we need some extra version pins to override the ones from dist.plone.org, but on Plone 6 the ones from dist.plone.org should be adequate. I hope.

@davisagli I added it back in [15a8e55](https://github.com/plone/volto/pull/3544/commits/15a8e5522573161696fe08beadd892740afae88a), sorry I was assuming this KGS will be overriding the default KGS version pins somewhere.

sneridagh:
@davisagli @nileshgulia1 I changed the tests, it seemed the most convenient thing to do, if default Plone 6 is not going to have them by default. So we can catch something wrong if Documents are used instead too.

CHANGELOG.md
```diff
@@ -20,6 +20,8 @@ See https://6.dev-docs.plone.org/volto/upgrade-guide/index.html for more informa
 
 ### Internal
 
+- Fix propTypes for Pagination component @davisagli
```
davisagli, line 23:
```suggestion
```

(bad merge)

cypress/tests/coresandbox/search.js
```diff
@@ -1,53 +0,0 @@
-context('Search action tests', () => {
-  describe('Search action with array', () => {
```
davisagli, line 2: This test seems to be for making sure that the folder view showed its contents. There are separate tests for the search view and the listing block, so I don't think it's needed when there are no folders. But let me know if I missed something.

tisto (approved):
LGTM!

sneridagh (approved):

sneridagh:
The next one should be flipping from using as base 5.x and use 6 instead, 5.x only essential.

sneridagh:
https://github.com/plone/volto/issues/3571

PR #3570, opened 2022-08-18
Better de translation for Site Setup

davisagli:


tisto (approved):
LGTM! The existing naming confuses me every single time, even after 15 years+.

PR #3579, opened 2022-08-23
Fix sharing page after using Pluggables

### LINKED ISSUE #3557: Sharing is broken
**To Reproduce**
- create content
- try to share it with a group
- no groups shown to share with

**Versions**
Volto 16

Plugs are breaking the sharing. I think https://github.com/plone/volto/pull/3372 should be reverted or fixed.

BTW Why does this view need to have plugs? IMHO if a customization of the sharing view is need, then shadow. It's an edge case, isn't it? The introduction of plugs here in sharing needs to be discussed. At least it should work correctly with any change.

/cc @JeffersonBledsoe

JeffersonBledsoe:
Fixes #3557

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 - Fix grouping of the "users and groups" control panels (plone-users category) @sneridagh
 - Improve `Display` and `Workflow` widgets in `More` menu. Fix alignments. @sneridagh
+- Fixed searching in the sharing page not showing any results @JeffersonBledsoe #3579
 
 ### Internal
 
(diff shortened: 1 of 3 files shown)


mauritsvanrees:
The title of this PR says WIP. It is indeed still work in progress?

JeffersonBledsoe:
> The title of this PR says WIP. It is indeed still work in progress?

Updated the title! I believe this fixes the issue, just wanted to check with the original issue reporter that everything is all good before requesting a merge :)

ksuess (approved):
Thank you for fixing the sharing!

ksuess:
@sneridagh, would say this is ready.

PR #3582, opened 2022-08-23
Change back button icon in folder contents.

robgietema:


sneridagh (approved):

sneridagh:
@robgietema it's a breaking in UX, can we have a proper entry in the changelog? It would be great if we explain briefly the rationale of this change as well in there too. Thanks!

PR #3583, opened 2022-08-23
Make `internalApiPath` client aware, since there are some corner case…

sneridagh:
…s when the client needs to know it to properly handle API server URLs.

Fixes: #3584

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
 
 ### Feature
 
+- Make `internalApiPath` client aware, since there are some corner cases when the client needs to know it to properly handle API server URLs @sneridagh
+
 ### Bugfix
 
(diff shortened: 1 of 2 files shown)


reekitconcept (approved):

ericof (approved):
LGTM

davisagli (commented):
It seems like it would be more ideal to make sure the hydrated page doesn't include internal API URLs, so that the internal hostname is not exposed. But, I think this is okay as a workaround.

sneridagh:
@davisagli indeed, I agree. However, it's a delicate story and should be thoroughfuly thought. Something we should bring to the Volto Team table. Could have some implications that we didn't have in mind right now. Also, something I wouldn't push to force for 16 either.

/cc @plone/volto-team

PR #3587, opened 2022-08-25
check for variable type not to introduce a breaking change

erral:
Would this work?

I am concerned about the treatment of default parameter in JS. I guess that if there is no parameter it should be treated like it was before do a plain call to `/@users` 

@sneridagh @ksuess

### DIFF
--- a/src/actions/users/users.js
+++ b/src/actions/users/users.js
@@ -76,41 +76,54 @@ export function getUser(id) {
  * @param {int} limit
  * @returns {Object} List users action
  */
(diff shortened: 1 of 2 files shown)


sneridagh:
To be sure, what about to have the same tests in place as before, and the new ones all along? This would help to spot errors and edge cases.

ksuess (approved):

sneridagh (approved):
@erral Thanks a lot for the quick fix!

@ksuess thanks for put an eye on this!

PR #3591, opened 2022-08-28
Use pangea dnd

Xhale1:
`react-beautiful-dnd` is no longer maintained and will not support React 18.

This PR migrates to a community-led fork which now supports React 18 and strict mode, [@hello-pangea/dnd](https://github.com/hello-pangea/dnd). It should be a drop-in replacement, requiring only modified imports.

The one caveat is that `@hello-pangea/dnd` upgraded to `react-redux` v8, while Volto uses v7. It might be worth it to investigate upgrading.


sneridagh:
@Xhale1 We finally decided to move soon to @dnd-kit Thanks for your contribution!

PR #3190, opened 2022-03-24
Relative path redo WIP

tiberiuichim:
Based on previous effort from cekk, giulia and all. I'm trying to go at this from a clean state, adding things step by step so that I understand them. I'll provide proper attributions at the end.

PR #3592, opened 2022-08-29
Prefix path router

giuliaghisini:
Enable to publish Volto site under a prefixPath, for example 
www.mymainsite.com/prefixPath


src/helpers/Api/Api.js
```diff
@@ -23,14 +23,22 @@ function formatUrl(path) {
...
     apiPath = settings.internalApiPath;
   } else if (settings.apiPath) {
     apiPath = settings.apiPath;
   }
 
+  let adjustedPath = path[0] !== '/' ? `/${path}` : path;
+  if (
+    config.settings.prefixPath &&
+    adjustedPath.startsWith(config.settings.prefixPath)
+  ) {
```
mamico, line 37: same as above

mamico (changes requested):

src/components/theme/View/FileView.jsx
```diff
@@ -25,7 +25,7 @@ const FileView = ({ content }) => (
       <p className="documentDescription">{content.description}</p>
     )}
     {content.file?.download && (
-      <a href={flattenToAppURL(content.file.download)}>
+      <a href={addPrefixPath(flattenToAppURL(content.file.download))}>
```
giuliaghisini, line 28: do not use addPrefixPath.. use UniversalLink instead <a> tag, it uses router that makes the magic with prefix path!

src/components/theme/View/ImageView.jsx
```diff
@@ -28,7 +28,7 @@ const ImageView = ({ content }) => (
       <p className="documentDescription">{content.description}</p>
     )}
     {content?.image?.download && (
-      <a href={flattenToAppURL(content.image.download)}>
+      <a href={addPrefixPath(flattenToAppURL(content.image.download))}>
```
giuliaghisini, line 31: do not use addPrefixPath.. use UniversalLink instead <a> tag, it uses router that makes the magic with prefix path!

src/components/theme/View/FileView.jsx
pnicolli: Changed

src/components/theme/View/ImageView.jsx
pnicolli: Changed

pnicolli:
Link view now working, Image view still has issues with the download link which is not working ([see this example](https://www.comune.re.it/rosebud/img/chisiamo.jpeg))

mamico (approved):

sneridagh:
During the Volto Team meeting we agreed that we would have to put in place a whole round of Cypress tests pointing to a deployment like this one. I can imagine that in the future will be quite easy to break the whole feature if one does not have in mind it.

I think that although time consuming, could be easy to implement. What do you think?

sneridagh:
@mamico @giuliaghisini could you share the reverse proxy config you have on such deployments?

cekk:
@sneridagh our deployment config is a bit complicated because there are several urls and frontend names, and it's made in varnish and not in nginx/apache.

The conf itself for using this branch is easy. Here is an example for nginx:

    upstream backend {
      server host.docker.internal:8080;
    }
    upstream frontend {
      server host.docker.internal:3000;
    }

    server {
      listen 80  default_server;
      server_name  localhost;

      location ~ /foo/\+\+api\+\+($|/) {
          rewrite ^/foo/\+\+api\+\+($|/.*) /VirtualHostBase/http/${server_name}/Plone/++api++/VirtualHostRoot/_vh_foo$1 break;
          proxy_pass http://backend;
      }


      location ~ /foo($|/) {
        
        # the standard nginx conf
        ...

        proxy_pass http://frontend;
      }
    }

tiberiuichim:
@cekk One other thing is important to document, the prefix path /foo corresponds to the Plone root, not a /foo subfolder in Plone.

cekk:
Yes, _/foo_ points to the root of **Plone** site

sneridagh:
@cekk To make it work I had to launch Volto with:

```
RAZZLE_PREFIX_PATH=/foo RAZZLE_API_PATH=http://localhost/foo yarn start
```

I would have expect that given a RAZZLE_PREFIX_PATH, the other would have adjusted automatically (as seamless mode promises). I am doing something wrong? because given a look at the code, it seems it should, right?

sneridagh:
@cekk Forget the question, I'm still asleep.

sneridagh:
Added tentative tests: #3719 see comments.

packages/volto-slate/src/editor/plugins/StyleMenu/index.js
```diff
@@ -4,16 +4,24 @@ import './style.less';
 
 export default function install(config) {
   const { slate } = config.settings;
+  slate.styleMenu = slate.styleMenu || {};
```
nileshgulia1: Sorry about this...(. Removed.

nileshgulia1:
I would like to take this forward. Locally looks good. I will checkout this branch on one of our projects first and see if I find some issues.

#### State of this PR:
This PR is based on `basename` react-router prop which adds an initial entry to browser "history" stack and allows all links to be prefixed with basename by react-router. The static assets like images need to be explicitly prefixed. I amended an express-middleware which re-directs to app base path on initial load. 

There is another approach in [this](https://github.com/plone/volto/pull/3190) PR, which uses a store enhancer middleware to prefix all the router paths(amending history accordingly) and modified `flattenToAppUrl` method. 

What's left are the cypress tests https://github.com/plone/volto/pull/3719/ which should also account for prefix path in the URLs. I will try to have a look into them.

I personally like the basename approach and let react-router handle the prefixes. However, we need to think about the non-router links and static assets. What do you think @sneridagh @davisagli @tiberiuichim @pnicolli ?

wesleybl:
@giuliaghisini @nileshgulia1 @sneridagh any chance we have this feature in master?

tiberiuichim:
@wesleybl it's due to the changes in razzle.config.js

wesleybl:
@tiberiuichim I fixed this in: https://github.com/plone/volto/pull/3592/commits/f011df6ea0a6e2c86dbb58a5abbb143887d17357

src/helpers/Url/Url.js
```diff
@@ -155,11 +155,21 @@ export const isCmsUi = memoize((currentPathname) => {
...
     ? html
-        .replace(new RegExp(settings.internalApiPath, 'g'), '')
-        .replace(new RegExp(settings.apiPath, 'g'), '')
-    : html.replace(new RegExp(settings.apiPath, 'g'), '');
+        .replace(new RegExp(settings.internalApiPath, 'g'), replacer)
+        .replace(new RegExp(settings.apiPath, 'g'), replacer)
+    : html.replace(new RegExp(settings.apiPath, 'g'), replacer);
+}
+
+export function addPrefixPath(src) {
```
wesleybl, line 166: @sneridagh @nileshgulia1 I think this function is not needed. I put a `return src` on the first line of it, and everything worked fine.

wesleybl: Although it works, without adding the prefix, the srcs and hrefs are missing the prefix in the generated HTML, which can be confusing. So it's better to keep the prefix.

wesleybl:
Documentation and testing was done in #3719 and merged here.

wesleybl (approved):
LGTM

wesleybl:
I did the rebase with the master and did tests. Everything looks ok now. Can you please have a look and merge?

@sneridagh @nileshgulia1 @pnicolli @cekk @tiberiuichim

nileshgulia1:
Let's discuss this in today's @plone/volto-team meeting!

ericof:
@sneridagh, @nileshgulia1, what is still needed to merge this to 18?

src/config/index.js
```diff
@@ -247,6 +249,48 @@ config.settings.apiExpanders = [
   },
 ];
 
+if (prefixPath) {
+  const ControlPanelAsyncPropExtender = {
+    path: `${prefixPath}/controlpanel`,
```
nileshgulia1, line 254: > Note: There's still an issue with asyncConnect Calls. I'm looking into it.

This fixes the issue. The async action dispatcher in https://github.com/plone/volto/blob/7b85e97c5e7c5046f1fce432bbbfd3f6842bd15f/src/components/manage/Controlpanels/Controlpanels.jsx#L311 doesn't seem to work well in ssr when prefixed paths. Do you see any other way of doing this without per route extenders?

wesleybl (approved):
I did some tests and everything worked!

nileshgulia1:
> Note: There's still an issue with asyncConnect Calls. I'm looking into it.

Fixed in [3ad97df](https://github.com/plone/volto/pull/3592/commits/3ad97df8467f54d72a1fa1a5c85ce093bef7e3d3). @ericof I'm ready on my part. It would be great if we have more eyes on it. Thanks @wesleybl for review!

ericof:
@sneridagh, @davisagli Could you take a look at this as well?

teekuningas:
Yes, by creating and applying patch. Otherwise it seems to work as expected but the multiple sites under same domain is a problem. What we want is to lightly create new sites using a single docker image and give each a different prefix.

wesleybl (changes requested):
`addPrefixPath` must also be used in the `Logo` component:

https://github.com/plone/volto/blob/85e3147530f363b1478ca211e80ff90ca758a991/packages/volto/src/components/theme/Logo/Logo.jsx#L51

.github/workflows/acceptance.yml
```diff
@@ -1226,3 +1226,173 @@ jobs:
...
           path: packages/volto/cypress/videos
+
+  prefix:
+    if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name != github.event.pull_request.base.repo.full_name
+    name: Prefixed site
+    runs-on: ubuntu-latest
+    timeout-minutes: 35
+    strategy:
+      matrix:
+        node-version: [18.x]
```
wesleybl, line 1237: This test better be run with Node 20

.github/workflows/acceptance.yml
```diff
@@ -1226,3 +1226,173 @@ jobs:
...
+          path: packages/volto/cypress/videos
+
+  multilingualprefix:
+    if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name != github.event.pull_request.base.repo.full_name
+    name: Multilingual in Prefixed site
+    runs-on: ubuntu-latest
+    timeout-minutes: 35
+    strategy:
+      matrix:
+        node-version: [18.x]
```
wesleybl, line 1322: This test better be run with Node 20

teekuningas:
Also, there are some changes to Image.jsx from 2 months ago that should be taken into account.. this might be enough:
```
diff --git a/packages/volto/src/components/theme/Image/Image.jsx b/packages/volto/src/components/theme/Image/Image.jsx
index db4125554..39ae2013c 100644
--- a/packages/volto/src/components/theme/Image/Image.jsx
+++ b/packages/volto/src/components/theme/Image/Image.jsx
@@ -67,7 +67,7 @@ export default function Image({
       attrs.srcSet = sortedScales
         .map(
           (scale) =>
-            `${flattenToAppURL(basePath)}/${scale.download} ${scale.width}w`,
+            `${addPrefixPath(flattenToAppURL(basePath))}/${scale.download} ${scale.width}w`,
         )
         .join(', ');
     }
```

.github/workflows/acceptance.yml
wesleybl: Done in https://github.com/plone/volto/pull/3592/commits/3938b217555311e266c5f76d142309f12e622f36

.github/workflows/acceptance.yml
wesleybl: Done in https://github.com/plone/volto/pull/3592/commits/3938b217555311e266c5f76d142309f12e622f36

wesleybl:
> addPrefixPath must also be used in the Logo component

Done in https://github.com/plone/volto/pull/3592/commits/61d96d2a48d9cdfc11b841f3050700b5d0facec4

wesleybl:
> Also, there are some changes to Image.jsx from 2 months ago that should be taken into account

@teekuningas done in https://github.com/plone/volto/pull/3592/commits/c7938167f5f28334f37eca55f8b132f70c382b2a

wesleybl (approved):
LGTM

wesleybl:
I updated the branch, fixed some issues and tested. It's all OK. Can anyone else review? Can we merge this PR?

@sneridagh @tiberiuichim @nileshgulia1 @pnicolli @ericof

nileshgulia1:
@wesleybl I was wondering why those 26 tests skipped. As far as I remember we should tests this also with multilingual websites and in seamless mode covering all core cypress tests.

wesleybl:
> As far as I remember we should test this also with multilingual websites and in seamless mode covering all core cypress tests

@nileshgulia1 we have a lot of cypress tests, as can be seen in: https://github.com/plone/volto/tree/main/packages/volto/cypress/tests

Should all these tests be run with prefixed? I'm afraid the tests will take too long.

wesleybl:
@nileshgulia1 hmm, I noticed now that you suggested core tests. Minimal tests are being tested today. I will replace it with core tests.

wesleybl:
@nileshgulia1 @sneridagh some tests expect a URL without the prefix. For example:

https://github.com/plone/volto/blob/81d0369e1628b3072f9e47d14831a2618d2633b6/packages/volto/cypress/tests/core/basic/actions.js#L65

However, when we run the prefixed tests, we have the prefix there and the test fails.

What is the best option in this case? Do a test using the `RAZZLE_PREFIX_PATH` environment variable? Test only the end of the URL?

wesleybl:
@sneridagh I managed to do it with an environment variable. For example:

```javascript
describe('Tests', () => {
  let prefixPath;
  beforeEach(() => {
    prefixPath = Cypress.env('prefixPath') || '';
  });
  it('test', function () {
    const url = `${prefixPath}/my-page-rename/contents`
  });
});
```

nileshgulia1:
> @nileshgulia1 @sneridagh some tests expect a URL without the prefix. For example:
> 
> https://github.com/plone/volto/blob/81d0369e1628b3072f9e47d14831a2618d2633b6/packages/volto/cypress/tests/core/basic/actions.js#L65
> 
> However, when we run the prefixed tests, we have the prefix there and the test fails.
> 
> What is the best option in this case? Do a test using the `RAZZLE_PREFIX_PATH` environment variable? Test only the end of the URL?

We need to [strip](https://github.com/eea/volto-prefixpath/blob/223e4e9de1bac2f9f63e59d21d0bd66bad8408bf/src/customizations/volto/actions/clipboard/clipboard.js#L25) out the prefix paths for some cases like copy/renaming contents or creating aliases. You could write an exception if there is a prefix, then remove it 🤔

nileshgulia1:
@wesleybl Did you checked if the moving/renaming of content works locally in this branch? We are using another solution in https://github.com/eea/volto-prefixpath/tree/master which might not be the case with other prefixed sites.

wesleybl:
> We need to [strip](https://github.com/eea/volto-prefixpath/blob/223e4e9de1bac2f9f63e59d21d0bd66bad8408bf/src/customizations/volto/actions/clipboard/clipboard.js#L25) out the prefix paths for some cases like copy/renaming contents or creating aliases. You could write an exception if there is a prefix, then remove it

@nileshgulia1 I will try to use this: https://github.com/plone/volto/pull/3592#issuecomment-1997827066

> Did you checked if the moving/renaming of content works locally in this branch?

Yes, this works.

wesleybl:
@nileshgulia1 I needed to revert part of the commit: https://github.com/plone/volto/commit/3ad97df8467f54d72a1fa1a5c85ce093bef7e3d3

The problem of not loading properties occurs in more routes, not just in
/controlpanel. We need a more generic solution.

I will check this.

wesleybl:
@nileshgulia1 I fixed the problem by adding prefixed routes. See: https://github.com/plone/volto/pull/3592/commits/149c8998babe318387722f9b7d26240c1c8615e1

Please check if it works for you.

wesleybl:
@sneridagh @nileshgulia1 add core test with prefixed path. Can you take a look please? Can we merge this PR?

docs/source/deploying/prefixed-root.md
```diff
@@ -0,0 +1,58 @@
...
+```shell
+DEBUG_HPM=true DEBUG=superagent RAZZLE_PREFIX_PATH=/my-prefix yarn start
+```
+
+The prefix location will be used regardless of how you start Volto, whether in development or production mode.
+When developing, though, if your backend is something other then `http://localhost:8080`, you'll need to provide your own solution for how to handle things.
+
+For a production setup, when hosting Volto behind a proxy HTTP server, you can configure your rewrite rules to something like the following, in this case for Apache.
+
+```apache
```
sneridagh, line 33: The recommended deployment setup is using Docker. I know that the documentation about this is in a sorry state, but we cannot add more confusion about this by dropping an `Apache` config here. I don't have a good solution for this one :( 

In the upcoming sprints we should work on documenting properly the deployment story.

/cc @ericof @davisagli @stevepiercy @fredvd

docs/source/deploying/prefixed-root.md
```diff
@@ -0,0 +1,58 @@
...
+RewriteRule ^/level-1/level-2(.*) http://volto:3000/level-1/level-2$1 [P,L]
+```
+
+And start Volto with the following.
+
+```shell
+RAZZLE_PREFIX_PATH=/level1/level2 RAZZLE_API_PATH=http://example.com/level1/level2 yarn start
+```
+
+Note, as you'll be integrating Volto with an existing website, you need to configure `config.settings.externalRoutes` so that the router knows which routes it should consider internal.
```
sneridagh, line 52: Need more elaboration and reasoning why this is necessary.

docs/source/deploying/prefixed-root.md
```diff
@@ -0,0 +1,58 @@
...
+
+If you need to debug and understand how the requests are rewritten by the Volto SSR, you can add the following environment variables to the Volto start line:
+
+
+```shell
+DEBUG_HPM=true DEBUG=superagent RAZZLE_PREFIX_PATH=/my-prefix yarn start
+```
+
+The prefix location will be used regardless of how you start Volto, whether in development or production mode.
+When developing, though, if your backend is something other then `http://localhost:8080`, you'll need to provide your own solution for how to handle things.
```
sneridagh, line 29: So in development I don't have to set it up and act as I'm in the root then?

packages/volto/cypress/tests/core/basic/actions.js
```diff
@@ -44,7 +46,7 @@ describe('actions Tests', () => {
       cy.get('a[class="icon-align-name"]').should(
         'have.attr',
         'href',
-        '/my-page-1/contents',
+        `${prefixPath}/my-page-1/contents`,
```
sneridagh, line 49: Can't we do it more elegant by invoking the `baseUrl` from Cypress config?

packages/volto/cypress/support/commands.js
```diff
@@ -837,38 +842,14 @@ Cypress.Commands.add('clickSlateButton', (button, timeout = 1000) => {
   }).click({ force: true }); // force click is needed to ensure the button in visible in view.
 });
 
-// Helper functions
```
sneridagh, line 840: +1 I started this already in a recent PR. A merge might be needed.

packages/volto/razzle.config.js
```diff
@@ -356,6 +356,15 @@ const defaultModify = ({
     };
   }
 
+  const prefixPath = process.env.RAZZLE_PREFIX_PATH || '';
+
+  if (prefixPath) {
+    if (target === 'web' && dev) {
+      config.devServer.devMiddleware.publicPath = prefixPath;
+    }
+    const pp = config.output.publicPath;
```
sneridagh, line 365: No cryptic vars, please.

And regarding the dev server, is it necessary to also use the prefix while developing?

packages/volto/src/components/manage/Controlpanels/Controlpanels.jsx
```diff
@@ -130,64 +136,66 @@ function Controlpanels({
     : [];
   const { filterControlPanels } = config.settings;
 
-  const filteredControlPanels = map(
```
sneridagh, line 133: I don't get the reason for this. Which is the rationale?

packages/volto/src/components/manage/UniversalLink/UniversalLink.jsx
```diff
@@ -8,10 +8,11 @@ import PropTypes from 'prop-types';
 import { HashLink as Link } from 'react-router-hash-link';
 import { useSelector } from 'react-redux';
 import {
+  addPrefixPath,
   flattenToAppURL,
   isInternalURL,
   URLUtils,
-} from '@plone/volto/helpers/Url/Url';
```
sneridagh, line 14: We decided not using barrel indexes in Volto anymore, so going back here is not an option.

packages/volto/src/components/manage/UniversalLink/UniversalLink.jsx
```diff
@@ -109,7 +110,7 @@ const UniversalLink = ({
   } else if (isDownload) {
     tag = (
       <a
-        href={flattenToAppURL(url)}
+        href={addPrefixPath(flattenToAppURL(url))}
```
sneridagh, line 113: So what if instead of modifying any invocation to `flattenToAppURL` (and I'm sure we are missing a lot), we incorporate this to it? I think it's not acceptable to modify every single apparition of it and the future ones.

sneridagh (changes requested):
@plone/volto-team We need to talk about this in our next meeting.

Finally I took a long look at it, and TBH I have lots of doubts. Sorry to be that pessimistic, but it's my sincere opinion.

I understand that this is a long requested feature. However, implement this if our Router does not support it OOTB (which is what it's happening) seems to me a lot. A lot of code to maintain, a lot of pitfalls that lie ahead if we adopt this now. 

Take the example of Remix. They just implemented this, and baked it into latest RR and their Vite build this last month. They did it by using a global `baseName` support: https://remix.run/blog/remix-vite-stable#basename-support 

I can tell you this, complicating Volto codebase will only make harder future refactors of its fundamental components. I don't want this to happen. I'm also convinced that make the current tests pass is not enough. I'm sure we have lots of edge cases uncovered.

This is intrusive enough for it to have its own major release alone, and I won't include it in 18... It has enough changes in there, and it's important enough already, being the targeted Plone 6.1 release, and we are already at the end of the alpha phase.

Let's discuss it in the next meeting, let's see if we can polish rough edges. I'd love that everyone involved and interested be there to talk about it.

wesleybl:
> This is intrusive enough for it to have its own major release alone, and I won't include it in 18...

@sneridagh so would this be for Volto 19? Any predictions on when Volto 19 will start?

stevepiercy:
> Any predictions on when Volto 19 will start?

It already has, by virtue of this pull request. I think this is the second feature designated for Volto 19, after my PLIP for timezone, date, and time support. I think it would be good to create a PLIP for discussion and add it to the [Volto Roadmap](https://github.com/plone/volto/projects/17) GitHub project. I also created a [19.x.x milestone](https://github.com/plone/volto/milestone/28) to start tagging issues and pull requests.

packages/volto/src/components/manage/UniversalLink/UniversalLink.jsx
nileshgulia1: @sneridagh I encountered challenges when attempting to integrate it directly into the "flattenToAppUrl" [function](https://github.com/eea/volto-prefixpath/blob/master/src/customizations/volto/helpers/Url/Url.js#L108), primarily due to the occurrence of duplicate prefixes because we already have a router basename set in this PR. Just to let you know we have recently deployed https://water.europa.eu/freshwater which is using bits and pieces of this PR. We did some customizations which will be removed when volto 17 is integrated into that site. I'm not sure if we applied special deployment tactics maybe @olimpiurob or @claudiaifrim  knows more about this context. 

Just for people who wants to test this feature, currently for development we use
```
RAZZLE_PREFIX_PATH=/freshwater RAZZLE_INTERNAL_API_PATH=http://x.x.x.x:port/Plone yarn start

```
You might also need to set `RAZZLE_API_PATH` if your backend lives in another site than /Plone.

packages/volto/src/components/manage/Controlpanels/Controlpanels.jsx
nileshgulia1: might not need it after [9a6b25f](https://github.com/plone/volto/pull/3592/commits/9a6b25f08b11f463b7d15b3af3a2151aa6b99407) ?
-cc @wesleybl

wesleybl: In https://github.com/plone/volto/commit/3ad97df8467f54d72a1fa1a5c85ce093bef7e3d3, @nileshgulia1 performed a validation to see if there was a controlpanel. I kept that validation. See below:

https://github.com/plone/volto/pull/3592/files#diff-10e8b3997f0087a776219f684dfa7c6440933357dee431c5614d6d4289191c00R139

This prevents an on-screen error if the backend does not return the controlpanels.

packages/volto/razzle.config.js
nileshgulia1: In webpack 5, the public path in dev mode needs to be prefixed as well. `http://localhost:3001/foo/static/js/client.js`

packages/volto/src/components/manage/UniversalLink/UniversalLink.jsx
wesleybl: @nileshgulia1 I didn't understand the difficulty of using `addPrefixPath` inside `flattenToAppURL`. Are there situations where it is not necessary to use `addPrefixPath`?

docs/source/deploying/prefixed-root.md
stevepiercy, line 1: I think that this page needs to have better narrative structure, such as the following:

- Introduction that states the essential and basic requirements for a deployment, regardless of root or non-root. This might already exist somewhere in the [Deploying](https://6.docs.plone.org/volto/deploying/index.html) section. Side note: that entire section needs to be cleaned up, and I think the index page should have this introduction as its introduction, so we can easily refer to it. We can also use an include to avoid copy-paste drift.
- Clarify what these requirements do and why they are necessary.
- Then add to the basic requirements those things that are required for a non-root deployment.
- Again, clarify what these requirements do and why they are necessary.
- Finally show examples of implementation details, whether it is Apache, nginx, or whatever. It is OK to have multiple examples. In fact I would use Sphinx Design with a [tabbed code example](https://sphinx-design.readthedocs.io/en/latest/tabs.html#tabbed-code-examples) interface.

stevepiercy (commented):
I've put this item on the agenda for the next Volto Team Meeting to discuss documentation, implementation details, targeting a Volto release, and possibly other items that come up.

packages/volto/src/components/manage/UniversalLink/UniversalLink.jsx
nileshgulia1: @wesleybl Basically paths that are rendered using `<Link />` or `<UniversalLink/>` need not be prefixed. This also includes hashlinks. In volto-prefixpath we don't set the router basename as prefix, thus we use this approach. 
The image components in our addons were our major source of trouble because in volto 17 we have a single Image component and the rule is to use it everywhere to render images. So, we use this approach until we migrate all addons to volto 17.

packages/volto/cypress/tests/core/basic/actions.js
wesleybl: @sneridagh I started looking into this but that won't be possible. `baseURL` contains the URL with the domain. For example:

`http://localhost/foo`

But links in HTML are relative. For example:

`/foo/my-document`

nileshgulia1:
Minutes from the @plone/volto-team meeting:

1. It was suggested that in dev mode, the prefix should not be a concern. Instead, configurations should be adjusted to exclusively utilize prefix during deployment.
2. Explore the possibility to integrate the functionalities of addprefixPath into flattenToAppUrl covering all the edge cases we might have. react-router v6 has [basename](https://reactrouter.com/en/6.22.3/routers/create-browser-router#basename) support. We need to further investigate how this feature is intended to function, both with and without the React Router upgrade( we are still using v5). The current implementation may compel future developers to rely extensively on this additional helper. Also what about the existing addons? they will have to be migrated as well. 
3. The current cypress testing strategy needs a rehaul according to point 2. It looks unsustainable to incorporate addPrefixPath across all forthcoming tests. Perhaps there is a way to run tests just using a cypress baseUrl? What about the internalUrls those need to be prefixed like Images, links? 
4. In the context of deployment documentation, we should explicitly outline if we need to use apache rules or docker.

packages/volto/src/components/manage/UniversalLink/UniversalLink.jsx
wesleybl: > I encountered challenges when attempting to integrate it directly into the "flattenToAppUrl" [function](https://github.com/eea/volto-prefixpath/blob/master/src/customizations/volto/helpers/Url/Url.js#L108), primarily due to the occurrence of duplicate prefixes because we already have a router basename set in this PR.

@nileshgulia1 did you change this behavior in [this commit](https://github.com/plone/volto/pull/3592/commits/433ad1cb272130ebfdd8b5945c08dedbcc7d68f0)?

Now `flattenToAppURL` takes care of everything?

wesleybl: @nileshgulia1 with its implementation, we have a duplicate prefix in the URL, when we click, for example, on an item in the navigation menu. It would be what you said here:

> I encountered challenges when attempting to integrate it directly into the "flattenToAppUrl" [function](https://github.com/eea/volto-prefixpath/blob/master/src/customizations/volto/helpers/Url/Url.js#L108), primarily due to the occurrence of duplicate prefixes because we already have a router basename set in this PR.

Any suggestions on how to fix this?

packages/volto/src/helpers/Url/Url.js
```diff
@@ -94,14 +100,30 @@ export function getView(url) {
...
       .replace(settings.apiPath, '')
-      .replace(settings.publicURL, '')
-  );
+      .replace(settings.publicURL, '');
+  const internalURL =
+    adjustedUrl &&
+    (adjustedUrl.indexOf(settings.publicURL) !== -1 ||
+      (settings.internalApiPath &&
+        adjustedUrl.indexOf(settings.internalApiPath) !== -1) ||
+      adjustedUrl.indexOf(settings.apiPath) !== -1 ||
```
wesleybl, line 118: It is not necessary to test `settings.publicURL`, `settings.internalApiPath` and `settings.internalApiPath` here, as they have been removed in the `adjustedUrl` assignment.

packages/volto/src/helpers/Url/Url.js
```diff
@@ -94,14 +100,30 @@ export function getView(url) {
...
+      adjustedUrl.charAt(0) === '/' ||
+      adjustedUrl.charAt(0) === '.');
+  //using isInternalUrl method causes infinite loop with special externalRoutes defined in wise-marine #264955
+  if (internalURL && !adjustedUrl.startsWith('#')) {
```
wesleybl, line 122: @nileshgulia1  @sneridagh Can `flattenToAppURL` receive external URLs or full URLs from the portal itself?

`internalURL` will never start with `#`.

nileshgulia1:
> > It was suggested that in dev mode, the prefix should not be a concern. Instead, configurations should be adjusted to exclusively utilize prefix during deployment.
> 
> @nileshgulia1 I think you implemented this in [this commit](https://github.com/plone/volto/pull/3592/commits/433ad1cb272130ebfdd8b5945c08dedbcc7d68f0#diff-70e7c8ca6db238bdc528d00a8db1084788711e492bb9641e473e62e04d855269).
> 
> But what is the motivation for this? When we develop the prefix path, we will have to restart and wait for it to compile. As it was, if the RAZZLE_PREFIX_PATH variable was not informed, it would not influence the "normal" development of the Volto.

Yes, ideally in development, we should use just `yarn start`, the prefix should only be used in production. I didn't want to add the logic inside flattenToAppUrl initially, as we are setting the router basename [here](https://github.com/plone/volto/pull/3592/files#diff-b58a9be9f71dd1872f4e64518ed88f4ebb30163671f03c808a6cbe8329f4e084R20). We need to think about the cases where we need to prefix the path "manually" as most of the cases are covered by router basename as prefix.  Given this, I think many of these conditions in the method can be removed or reconsidered.

wesleybl:
> Yes, ideally in development, we should use just yarn start, the prefix should only be used in production.

@nileshgulia1 But this is bad for development. I'm doing tests here and for each change in the code I need to stop the server, compile, which takes time, and start the server again. This is not productive. Can we go back to the way it was before?

> I didn't want to add the logic inside flattenToAppUrl initially, as we are setting the router basename [here](https://github.com/plone/volto/pull/3592/files#diff-b58a9be9f71dd1872f4e64518ed88f4ebb30163671f03c808a6cbe8329f4e084R20). We need to think about the cases where we need to prefix the path "manually" as most of the cases are covered by router basename as prefix. Given this, I think many of these conditions in the method can be removed or reconsidered.

I'm analyzing it here and I've come to the conclusion that it's a bad idea prefix in `flattenToAppUrl` function. As you said, `Link` and `UniversalLink` components do not need a prefix. And they are mostly used in the Volto. So, only a few `a` tags would need to be prefixed, which are the minority in Volto. I think it's better to have a separate function for `a`. There is no free lunch here. Can we go back to having a separate prefix function?

CC @sneridagh

nileshgulia1:
> But this is bad for development. I'm doing tests here and for each change in the code I need to stop the server, compile, which takes time, and start the server again. This is not productive. Can we go back to the way it was before?

Why do you need to restart the server? What you mentioned here probably because of broken devServer. I will look into it tomorrow.

wesleybl:
> Why do you need to restart the server? What you mentioned here probably because of broken devServer.

@nileshgulia1 after your commit, the prefix no longer works in development mode. That is, with the command:

```
RAZZLE_PREFIX_PATH=/my-prefix pnpm --filter @plone/volto start
```
The prefix doesn't work. It only works in production mode, i.e.:

```
RAZZLE_PREFIX_PATH=/my-prefix pnpm build && pnpm --filter @plone/volto start:prod
```

And in production mode, a change to a file is not effective immediately. We need to compile again.

wesleybl:
@nileshgulia1 @sneridagh Cannot add prefix in `flattenToAppUrl`. To see:

https://github.com/plone/volto/pull/3592#issuecomment-2223666941

So I returned with the `addPrefixPath` function, and called it where necessary. I don't see a problem with using this function this way. Basically it is called in structural components, with `UniversalLink` and `Image`. From what I understand, the use of these components is strongly recommended. So "forgetting to prefix" will be very rare. If we use these components, the prefix will already be done.

I have the time and sponsorship to work with this functionality now. From my point of view, it's working well. Tests are passing now. I need guidance on what I should do to improve this PR.

wesleybl:
> Why do you need to restart the server? What you mentioned here probably because of broken devServer. I will look into it tomorrow.

@nileshgulia1 so can I go back to reading the `RAZZLE PREFIX PATH` variable in development mode?

nileshgulia1:
@wesleybl so the idea is like this, if we remove prefix-path addition in flattenToAppUrl, and since we are using the basename we don't have to prefix manually. All the links with a react-router `<Link/>` should already have a prefix. The basename=/prefix means by-default / points to /prefix. So if you have anywhere in your project, `<Logo href="/" />` you will end up with a trailing slash like so `/prefix/`.  This is a bug in react-router v5. They [fixed](https://github.com/remix-run/react-router/pull/8861) that in latest version, dunno know if it was backported. 

Regarding the development recipe, I think the website itself shouldn't care about the prefix in dev. The prefix should only be used in production. If you want to test it, you can do `RAZZLE_PREFIX_PATH=/prefix yarn build` && `RAZZLE_PREFIX_PATH=/prefix yarn start:prod`, and see if it has the prefix.

nileshgulia1:
>>>So I returned with the addPrefixPath function, and called it where necessary. I don't see a problem with using this function this way. Basically it is called in structural components, with UniversalLink and Image

I think it's worth a try not to depend on addPrefixPath function. I would not force anyone to use it in their new addons or components.

wesleybl:
> If you want to test it, you can do RAZZLE_PREFIX_PATH=/prefix yarn build && RAZZLE_PREFIX_PATH=/prefix yarn start:prod, and see if it has the prefix.

@nileshgulia1 the problem is precisely when we are developing the prefix functionality and we want to see the website with the prefix. The `RAZZLE_PREFIX_PATH=/prefix yarn build` command takes a long time. And for every change we make to a file, we need to compile it again to see the changes. Hot reload does not work in production mode. What is the advantage of not using RAZZLE_PREFIX_PATH in development mode? I think we should allow this, so that it is not necessary to compile for each modification.

nileshgulia1:
>the problem is precisely when we are developing the prefix functionality and we want to see the website with the prefix.

That is the thing, why do we want to develop with prefix differently from a normal case? If you want to develop with prefix, you shouldn't do anything special in development. At least that's what we have discussed in @plone/volto-team meeting. Can't we keep things minimal? So just do RAZZLE_PREFIX_PATH=/preffix yarn build && RAZZLE_PREFIX_PATH=/preffix  arn start:prod. In any case I would like to hear more thoughts. -cc @sneridagh @giuliaghisini @cekk

wesleybl:
> That is the thing, why do we want to develop with prefix differently from a normal case?

@nileshgulia1 when I say "developing the prefix functionality", I'm talking about situations like fixing the test of a prefixed CI job. Or making changes like prefixing inside or outside the `flattenToAppUrl` function. In these cases, I want to use the `RAZZLE_PREFIX_PATH` variable in development mode, so that it is not necessary to compile with each file change.

For "normal development", we can start without the variable, as is normally done today. It would be transparent.

wesleybl:
> I think it's worth a try not to depend on addPrefixPath function. I would not force anyone to use it in their new addons or components.

@nileshgulia1 In order to avoid having an `addPrefixPath` function and prefixing it in the `flattenToAppUrl` function, we need to remove the basename from the route, so as not to duplicate the prefix in the Link and UniversalLinks components, since the `flattenToAppUrl` function is called in these components. But everywhere I've researched about how to use prefixes in React, it's recommended to use basename in the route. So I'm not sure if it would work without the basename. This could make things more complicated than having the `addPrefixPath` function. Even if it works without the basename and the prefix is ​​done in `flattenToAppUrl`, I think we would still have situations in which the action performed by `flattenToAppUrl` would be necessary but would not require prefixing. It won't always be necessary to do both.

I'll try to make a branch removing basename to see how things behave.

wesleybl:
@nileshgulia1 @sneridagh I'm implementing the removal of `basename` from the route and the creation of the prefix in the `flattenToAppURL` function. It works. But we would have to use the `flattenToAppURL` function everywhere that uses the `Link` component. I noticed that in many places where the `Link` component is used, the `flattenToAppURL` function is not used. For example:

https://github.com/plone/volto/blob/6f18663d012b1cdcaeb04d15a79f669973330109/packages/volto/src/components/manage/Controlpanels/Controlpanels.jsx#L209

https://github.com/plone/volto/blob/6f18663d012b1cdcaeb04d15a79f669973330109/packages/volto/src/components/manage/Controlpanels/Controlpanels.jsx#L231

In the end, we would have to change many more files than with the use of `addPrefixPath` separately. So in my opinion, it is better to keep the `addPrefixPath` function and the use of `basename` in the route. What do you think?

sneridagh:
@wesleybl sorry, I can't follow. Tomorrow I can try to take a full read, but at this time of the year I can't devote any effort to make this happen. What do you mean by removing the basename from the router? This is the full base of the feature and we can't remove it. So I'm quite shocked.

We've talked about this feature during today's Volto team meeting. 

1. This feature is NOT trivial.
2. It should be done in a conscious, sustainable way, with enough vision and not jeopardizing the future development of Volto.
3. We can't just hack the full Volto routes and branch the while code only because of that.
4. If we can't do it under these premises, then it won't happen. We have to find the method to do it in the right, elegant way. Not quick and dirty. TBH, I'd love that someone else from the Volto Team steps up and help with this.
4. The base set by RedTurtle in the initial PRs was good but somehow it derived and derailed in some bad decisions.
5. Right now I can't put any energy on this, probably I won't until some sprint happens, and it seems that only @nileshgulia1 and EdW are interested in that, if so, we will need resources when it matters (see below).
6. We need consensus from the Volto Team, so we need a formal PLIP. Do we have already one, well argumented, etc?
7. We need someone who really needs it and have it working and have the feedback from it. This can't happen if it's not battle tested.

So I propose to start over during a sprint, in a joint effort, step by step. Ferrara happens at the end of September. We can start set the bases in there. The Plone Conf is not that far, and the ones that are interested in these feature will attend there.

sneridagh:
@wesleybl ok, @nileshgulia1 created the PLIP: https://github.com/plone/volto/issues/4290

I'd love to talk to you as well in person (I already did with Nilesh this morning in the Volto Team Meeting). I'm open for a talk during this week (not on Thursday) Let me know if you can.

nileshgulia1:
@wesleybl I agree that we should have a basename set, that is the starting point in the PR, it takes care of prefixing for us. We need to identify the parts where the prefix is manually needed. For me, focus is more on trying to avoid affecting the future development of volto. Or if we should really introduce env vars starting with RAZZLE_ now. There are lof of efforts that uses another build system and maybe in future we might shift there. 
 @wesleybl If you plan to attend PloneConf, then let's discuss this in detail there during sprints or in general.

wesleybl:
> We need to identify the parts where the prefix is manually needed.

@nileshgulia1 basically, we need to manually prefix in the HTML tags `a` and `img`, since React Router does not influence these tags. That's why we need the `addPrefixPath` function. In this PR, we have already replaced the use of the `img` tag with the `Image` component, and we use `addPrefixPath` inside the `Image` component. Volto also practically does not use the `a` tag for links. It doesn't even make sense to use the `a` tag in React. The only situation in which the `a` tag is necessary is for file downloads. But file downloads are encapsulated in the `UniversalLink` component and there we use `addPrefixPath`.

In short, if we use the `UniversalLink`, `Link` and `Image` components, everything will be fine.

wesleybl:
> I'd love to talk to you as well in person

@sneridagh I sent you an email.

nileshgulia1:
> > I'd love to talk to you as well in person
> 
> 
> 
> @sneridagh I sent you an email.

I'll be happy to join as well. Just ping me in discord.

wesleybl:
@nileshgulia1 @sneridagh I documented in [PLIP](https://github.com/plone/volto/issues/4290) situations where it is necessary to manually prefix, when using `react-router`'s `basename`. Is basically what is being done in this PR.

teekuningas:
I wonder if this or similar could be included..

https://github.com/plone/volto/pull/3592#issuecomment-1956592764

Anyway, I hope this will merged soon! :- )

wesleybl:
> I wonder if this or similar could be included..
>
> https://github.com/plone/volto/pull/3592#issuecomment-1956592764

@teekuningas In our use case, we want cookies to be shared. But perhaps your suggestion could be a setting.

teekuningas:
> > I wonder if this or similar could be included..
> > [#3592 (comment)](https://github.com/plone/volto/pull/3592#issuecomment-1956592764)
> 
> @teekuningas In our use case, we want cookies to be shared. But perhaps your suggestion could be a setting.

Interesting, I wonder if that would work for us too.. anyway, a setting would be enough, yes.

nileshgulia1:
@wesleybl @teekuningas I talked to @sneridagh during PloneConf, and we agreed that we should rewrite the PR. We should extract the useful parts from it and possibly rewrite the PLIP to ensure it works with the current setup and can be easily upgraded when we move to Seven. What do you think? I agree to provide support maybe we should get in touch soon and target a sprint for it?

wesleybl:
> I talked to @sneridagh during PloneConf, and we agreed that we should rewrite the PR

@nileshgulia1 That's exactly what I started doing with #6754. But instead of doing it all in one big PR, I think it's easier to split it up into multiple PRs.

wesleybl:
> Also I found that after we upload the logo through controlpanel this fails to work as plone_site_logo comes as site root URL and Plone doesn't know about the prefix.

@nileshgulia1 This commit fixes that: https://github.com/plone/volto/pull/3592/commits/5ae6259dc833f5eab06beb73bee2f36a524cb005

packages/volto/src/components/manage/Controlpanels/Controlpanels.jsx
wesleybl: @sneridagh @nileshgulia1 this is no longer needed. I removed it in: https://github.com/plone/volto/pull/3592/commits/ea5db08546380bfc7b0d389d46507b041c4c8e44

wesleybl:
> The basename=/prefix means by-default / points to /prefix. So if you have anywhere in your project, <Logo href="/" /> you will end up with a trailing slash like so /prefix/. This is a bug in react-router v5. They https://github.com/remix-run/react-router/pull/8861 that in latest version, dunno know if it was backported.

@nileshgulia1 do you know of any workaround to fix this? Or would it only be with react-router v5?

I tried to fix this in https://github.com/plone/volto/pull/3592/commits/7f435b0309f11002b1ea85e6d4f980edb7798a7f

But it didn't work. It gives a Not Found in SSR. I had to undo it in: https://github.com/plone/volto/pull/3592/commits/01148eb0f6b1ab3bce24d1997e5c59b416357548

nileshgulia1:
> > The basename=/prefix means by-default / points to /prefix. So if you have anywhere in your project,  you will end up with a trailing slash like so /prefix/. This is a bug in react-router v5. They [remix-run/react-router#8861](https://github.com/remix-run/react-router/pull/8861) that in latest version, dunno know if it was backported.
> 
> @nileshgulia1 do you know of any workaround to fix this? Or would it only be with react-router v5?
> 
> I tried to fix this in [7f435b0](https://github.com/plone/volto/commit/7f435b0309f11002b1ea85e6d4f980edb7798a7f)
> 
> But it didn't work. It gives a Not Found in SSR. I had to undo it in: [01148eb](https://github.com/plone/volto/commit/01148eb0f6b1ab3bce24d1997e5c59b416357548)

I don't think that fix was backported to RRv5. If you use https://github.com/plone/volto/blob/ea5db08546380bfc7b0d389d46507b041c4c8e44/packages/volto/src/routes.js#L333
then, you should not use the basename.

nileshgulia1:
@wesleybl That is the point, if you are using basename, ideally there's isn't any need to prefix routes manually and that's actually the best way. This is what we are using in our project.
But the way how RR handles the root path( with basename set) was a blocker for us. 

IIRC, <Link to href="" /> should point to /prefix, but currently its pointing to / resulting on /prefix/.

Maybe you can write a redux middleware which intercepts '@@router_location_change' and handle trailing slashes more correctly. I tried it but I don't remember if it solved my problem.

nileshgulia1:
@wesleybl I'm in favour of closing this PR and let's do further development in https://github.com/plone/volto/tree/prefix-path-revised ? This branch is new and clean with the minimal setup needed for this feature to work. For the moment, let's try to make it work with RR basename (I'm aware of the trailing slash situation). I'm not using this branch anywhere, just wanted to ask if you (or anyone else) are using this branch anywhere on your projects?

wesleybl:
@nileshgulia1 I agree that this PR should never be merged. I think the best thing to do is to split it into several smaller PRs. But I think it's worth having a place where we can test everything working. I think it should be turned into a draft.

I was using this branch in a PoC. But we ended up creating an addon. But everything I learned in the PoC I put in this PR. So, I don't think the branch should be deleted. It can serve as a reference source.

I think it's a good idea to have a new branch, like yours. But I was thinking of a different strategy. I think the basename should be the last PR to be made. I was thinking first of doing things that could be in Volto, even without the basename. Like centralizations. For example: https://github.com/plone/volto/pull/6754

Anyway, I'll follow your branch. If you need help, let me know. Thanks!

teekuningas:
We are still using it.. but we can migrate as long as there is a working alternative.

nileshgulia1:
Lets continue the work in https://github.com/plone/volto/pull/6897

PR #3614, opened 2022-09-06
Added scroll into view.

robgietema:


davisagli:
@robgietema I think we want the ability to set this for specific instances of the Slate component (i.e. in the sidebar), not a global setting.

PR #3615, opened 2022-09-07
Back to classic workflow transition names, adding i18n to the change Workflow dropdown

sneridagh:
image

image


src/helpers/Workflows/Workflows.js
```diff
@@ -0,0 +1,58 @@
...
+ * @param  {Object} transition
+ *  {
+ *     '@id': 'http://localhost:3000/de/@workflow/publish',
+ *     title: 'Veröffentlichen'
+ *  }
+ * @return {Object} The react-select options object
+ *  { value: 'public', label: 'Published', color: '#007bc1', url: 'http://localhost:3000/de/@workflow/publish' }
+ */
+export function getWorkflowOptions(transition) {
+  const mapping = config.settings.workflowMapping;
```
tiberiuichim, line 44: this should be documented here https://6.dev-docs.plone.org/volto/configuration/settings-reference.html

tiberiuichim (approved):

docs/source/configuration/settings-reference.md
```diff
@@ -225,6 +225,25 @@ controlpanels
...
+      pending: { value: 'pending', color: '#f6a808' },
+      send_back: { value: 'private', color: '#ed4033' },
+      retract: { value: 'private', color: '#ed4033' },
+      reject: { value: 'private', color: '#ed4033' },
+      submit: { value: 'review', color: '#f4e037' },
+    };
+    ```
+
+    It's meant to be extended with your own workflows/transitions.
+    It is recommended that you could assign the same color to the transition as the destination state, so the user can have the visual hint to which state are they transitioning to.
```
stevepiercy, line 246:
```suggestion
    It is recommended to assign the same color to the transition as the destination state, so the user can have the visual hint to which state are they transitioning to.
```

stevepiercy (changes requested):
LGTM with one tiny change.

PR #3634, opened 2022-09-13
Undo `html_static_path` configuration in `plone/documentation`, and restore image and its referenced path in `plone/volto`.

stevepiercy:
I just realized I made a huge goof regarding the `html_static_path` configuration. Apparently you don't need to specify a path to copy images from submodules as long as they have a reference in the Markdown files. I failed to realize that Sphinx automagically copies images from submodules into `_build/_images` instead of `_build/_static` and writes the HTML to the appropriate directory. TIL.

This PR puts it all back to normal.

See https://github.com/plone/documentation/pull/1316


davisagli (approved):
Not sure I follow the overall change here, but it seems to work in the preview.

PR #3639, opened 2022-09-14
Fix workflow and display select in toolbar in case that the option sp…

sneridagh:
…ans several lines

image

image

### DIFF
--- a/src/components/manage/Display/Display.jsx
+++ b/src/components/manage/Display/Display.jsx
@@ -28,9 +28,9 @@ const Option = injectLazyLibs('reactSelect')((props) => {
     <Option {...props}>
       <div>{props.label}</div>
       {props.isFocused && !props.isSelected && (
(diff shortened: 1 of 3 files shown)


src/components/manage/Display/Display.jsx
```diff
@@ -101,6 +101,12 @@ const customSelectStyles = {
     ':active': {
       backgroundColor: null,
     },
+    span: {
+      flex: '0 0 auto',
+    },
+    svg: {
```
davisagli, line 107: @sneridagh Is this actually a CSS property, or a typo?

giuliaghisini (approved):

src/components/manage/Display/Display.jsx
sneridagh: `react-select` uses Emotion for inner styling. In Emotion you can actually do this, create inner selectors from the styling properties.

davisagli (approved):

PR #3654, opened 2022-09-19
Fix backspace selection error.

robgietema:
Since setting the selection has a timeout an error occurs when you press backspace multiple times. This is because the editor tries to set the selection of the first update when the content is already updated and so the selection is out of range. The issue which is raised in the comments is tested and also works without the timeout;

- moving between blocks with up/down arrows
- clicking inside the text of a block should not reset the cursor position to head/end of block text
- split the block with enter in the text

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Bugfix
 
+- Fix selection error when pressing backspace @robgietema
+
 ### Internal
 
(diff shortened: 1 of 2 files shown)


tiberiuichim:
Somewhere along the way the "restore cursor position to join spot, after joining two blocks" has been lost.

tiberiuichim (approved):

Issue #3647, opened 2022-09-16
Press Enter in some blocks does not focus on the text block below


wesleybl:
**Describe the bug**

When we have the focus on some blocks, and we type Enter, the focus does not move to the text block created below.

**To Reproduce**

1. Go to Editing a Document.
2. Add for example a listing block.
3. Press Enter.

**Expected behavior**
The block created below must receive the focus, with the cursor.

**Screenshots**
![Enter in block](https://user-images.githubusercontent.com/3297795/190714327-fb5c3dcb-2ef2-4bc5-a6be-01c7b2e4fac6.gif)

**Software (please complete the following information):**

- OS: Ubuntu 22
- Browser: Chrome
- Volto Version: 16.0.0-alpha.33 
- Plone Version: 6.0.0b2
- Plone REST API Version: The one used in Plone

**Additional context**
This behavior also occurs in Add-on blocks

PR #3659, opened 2022-09-19
Press Enter in some blocks does not focus on the text block below

dobri1408:
Fixes https://github.com/plone/volto/issues/3647.
I saw that all the time when  onAddBlock was executed, also the onSelectBlock was because we wanted to select the added block, but with an exception, that I fixed it now.

### DIFF
--- a/cypress/tests/core/blocks/blocks-autofocus.js
+++ b/cypress/tests/core/blocks/blocks-autofocus.js
@@ -1,0 +1,172 @@
+describe('New Block Auto Focus Tests', () => {
+  beforeEach(() => {
+    // given a logged in editor and a page in edit mode
+    cy.visit('/');
+    cy.autologin();
+    cy.createContent({
+      contentType: 'Document',
+      contentId: 'my-page',
+      contentTitle: 'My Page',
+    });
+    cy.visit('/my-page');
+    cy.waitForResourceToLoad('@navigation');
+    cy.waitForResourceToLoad('@breadcrumbs');
+    cy.waitForResourceToLoad('@actions');
+    cy.waitForResourceToLoad('@types');
+    cy.waitForResourceToLoad('my-page');
+    cy.navigate('/my-page/edit');
+  });
+
+  it('Press Enter on a description block adds new autofocused default block', () => {
+    cy.intercept('GET', '/**/my-page').as('content');
+    cy.intercept('PATCH', '*').as('save');
+    // when I add a text block
+    //add listing block
+    cy.getSlate().click();
+    cy.get('button.block-add-button').click();
+    cy.get('.blocks-chooser .title').contains('Text').click();
+    cy.get('.blocks-chooser .text').contains('Description').click();
+    cy.get('.documentDescription').first().click().type('{enter}');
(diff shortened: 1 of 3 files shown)


wesleybl (approved):
I had seen just that. So LGTM!

tiberiuichim (approved):
It would be nice if this would be tested with cypress.

nileshgulia1 (approved):

dobri1408:
Hi @tiberiuichim, great idea, I am on it!

dobri1408:
I added the cypress tests

tiberiuichim (approved):

cypress/tests/core/blocks/blocks-autofocus.js
```diff
@@ -0,0 +1,172 @@
...
+    cy.navigate('/my-page/edit');
+  });
+
+  it('Press Enter on a description block adds new autofocused default block', () => {
+    cy.intercept('GET', '/**/my-page').as('content');
+    cy.intercept('PATCH', '*').as('save');
+    // when I add a text block
+    //add listing block
+    cy.getSlate().click();
+    cy.get('button.block-add-button').click();
```
wesleybl, line 26: I don't understand cypress, but it seems that this lines can stay in the `beforeEach` function, since they are the same in all tests.

wesleybl (changes requested):

tiberiuichim:
There is a conceptual issue with the HTML block, in that it should handle Enter key on its own, but I think it's a different discussion. For context, see https://community.plone.org/t/how-does-volto-handle-the-enter-key/15663

wesleybl (approved):
LGTM

nileshgulia1 (approved):

sneridagh:
Not sure on this one, I think it could break quite a few blocks out there. Did you tested in all stock blocks? If we are "breaking" the behavior in HTML block, we should fix it as well. How about nested ones?

sneridagh:
Also, another thing I just realised... How do you create a new block below your block, if ENTER now selects the next one?
Food for thought...

dobri1408:
Hi @sneridagh, the initial target of the pr was to solve the issue with the unfocused new blocks. This issue is solved just by adding the function onSelectBlock where was supposed to be. If you consider that the tests are not good, I can remove them, and let just the fix be merged. The tests have the scope to select each block press enter and see if the newly added block is focused. If the tests are a good idea, but there is a problem with the HTML block, I can remove the HTML block.

sneridagh:
@dobri1408 ok! let me test it, I might be wrong in my assumption. Please bare with me!

wesleybl:
> If we are "breaking" the behavior in HTML block, we should fix it as well.

@sneridagh hhe HTML block is broken in master, regardless of the code of this PR. It must be fixed in another PR. It's a different problem.

> How do you create a new block below your block, if ENTER now selects the next one?

Enter is already creating a new block. But this new block is not getting the focus (cursor in text block input). This PR makes the course stay in the text block that is already created.

sneridagh (approved):
LGTM!

PR #3661, opened 2022-09-20
Fix virtualization (windowing) when displaying options with long titles for select widgets

sneridagh:
Turns out that the library I introduced for making the windowing in case that we have lots of options in selects/arrays can't handle variable height options. So I had to change it with one that does have support for this: `react-virtualized`. I gave some love to the select widgets too, adding breaking words in the options.

I left `react-window` library, in case that someone is using it. Both are lazy loaded, so they won't have impact on the build.

Before:
image

After:
image

### DIFF
--- a/src/components/manage/Widgets/useDimensions.js
+++ b/src/components/manage/Widgets/useDimensions.js
@@ -1,0 +1,51 @@
+import { useState, useCallback, useEffect } from 'react';
+
+function getDimensionObject(node) {
+  const rect = node.getBoundingClientRect();
+
+  if (rect.toJSON) {
+    return rect.toJSON();
+  } else {
+    return {
+      width: rect.width,
+      height: rect.height,
+      top: rect.top || rect.y,
+      left: rect.left || rect.x,
+      x: rect.x || rect.left,
+      y: rect.y || rect.top,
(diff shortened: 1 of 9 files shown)


src/components/manage/Widgets/useDimensions.js
```diff
@@ -0,0 +1,51 @@
...
+      left: rect.left || rect.x,
+      x: rect.x || rect.left,
+      y: rect.y || rect.top,
+      right: rect.right,
+      bottom: rect.bottom,
+    };
+  }
+}
+
+function useDimensions() {
```
tiberiuichim, line 22: This is super, but doesn't appear to be used, at least in the PR

sneridagh: I had it in store, I did some trials, finally went for a full library solution. But I'll add it as a helper in another PR. :)

davisagli:
@sneridagh

> I left react-window library, in case that someone is using it.

If volto itself no longer uses it, I would remove it, and mention it in the changelog and/or upgrade docs.

davisagli (approved):

sneridagh:
@davisagli done.

PR #3674, opened 2022-09-24
Add missing `moduleNameMapper`s to default app `package.json` jest co…

sneridagh:
…nfig

### DIFF
--- a/packages/generator-volto/CHANGELOG.md
+++ b/packages/generator-volto/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Bugfix
 
 - Add alias to ESlint config in project generator @sneridagh
+- Add missing `moduleNameMapper`s to default app `package.json` jest config @sneridagh
 
 ### Internal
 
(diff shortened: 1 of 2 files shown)


davisagli (approved):
I'm fine with the specific fix. I wish there was some way to avoid this though; it's a lot of configuration that needs to be copied and maintained in each app.

tiberiuichim:
> I'm fine with the specific fix. I wish there was some way to avoid this though; it's a lot of configuration that needs to be copied and maintained in each app.

Which is why I'm a big fan of treating projects as throw-away and moving all functionality to addons. When there's a new Volto release, just regenerate the project, copy the dependencies part and install all your addons.

sneridagh:
@davisagli yes, it is... in this case, it's because `jest` keeps wanting to compile everything under the sun of the working directory, no matter if it is going to test it (`testMatch`) or not... It's a real pita.

We thought at some point we could write an updater tool that takes care of all these changes, as they happen...

PR #3676, opened 2022-09-24
Last amendments before merging

sneridagh:
Bring back `postcss` to generator add notice of downgrade in docs and changelog.


packages/generator-volto/generators/app/templates/package.json.tpl
```diff
@@ -141,6 +141,7 @@
     "eslint-plugin-prettier": "3.1.3",
     "jest-junit": "8.0.0",
     "mrs-developer": "*",
+    "postcss": "8.4.13",
```
davisagli, line 144: @sneridagh volto's own package.json has postcss 8.3.11; should we keep this in sync, to avoid installing multiple copies?

sneridagh: Done.

PR #3678, opened 2022-09-25
Remove Razzle as dep from `@plone/scripts`

sneridagh:
I checked this out and I don't think it's necessary any longer. On dependency less!

Also, reminder that after this, we might want to remove the Volto dependency on `@plone/scripts` as well, and point `razzle.config.js` reference to i18n module in there for an static reference to `package/scripts/i18n`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,8 @@ See https://6.dev-docs.plone.org/volto/upgrade-guide/index.html for more informa
 
 ### Internal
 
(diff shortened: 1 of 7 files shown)


davisagli (approved):

davisagli:
> after this, we might want to remove the Volto dependency on @plone/scripts as well, and point razzle.config.js reference to i18n module in there for an static reference to package/scripts/i18n.

IIRC I tried to do that, but found it didn't work when volto's razzle.config.js was included from an app's razzle.config.js

PR #3684, opened 2022-09-26
Fix `listing` block in SSR, now it is fully variations aware, and the…

sneridagh:
… config is passed to the SSR querystring action

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 
 ### Bugfix
 
-- Fix `listing` block in SSR, now it's fully variations aware, and the config is passed to the SSR querystring action @sneridagh
+- Fix `listing` block in SSR, now that it is fully variations aware and the configuration is passed to the SSR `querystring` action. @sneridagh
 
 ### Internal
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Bugfix
 
+- Fix `listing` block in SSR, now it's fully variations aware, and the config is passed to the SSR querystring action @sneridagh
+
 ### Internal
 
 ### Documentation
--- a/src/components/manage/Blocks/Listing/getAsyncData.js
+++ b/src/components/manage/Blocks/Listing/getAsyncData.js
@@ -1,13 +1,16 @@
 import { getQueryStringResults } from '@plone/volto/actions';
+import { resolveBlockExtensions } from '@plone/volto/helpers';
 
-export default ({ dispatch, data, path }) => {
+export default ({ dispatch, data, path, blocksConfig }) => {
+  const { resolvedExtensions } = resolveBlockExtensions(data, blocksConfig);
+
   return [
     dispatch(
       getQueryStringResults(
         path,
         {
           ...data.querystring,
-          ...(data.variation?.fullobjects
+          ...(resolvedExtensions?.variation?.fullobjects
             ? { fullobjects: 1 }
             : { metadata_fields: '_all' }),
         },
--- a/src/components/theme/App/App.jsx
+++ b/src/components/theme/App/App.jsx
@@ -223,6 +223,7 @@ export const fetchContent = async ({ store, location }) => {
         location,
         id,
         data,
+        blocksConfig,
       });
       if (!p?.length) {
         throw new Error(
--- a/src/helpers/Extensions/index.js
+++ b/src/helpers/Extensions/index.js
@@ -1,2 +1,5 @@
 export * from './withBlockSchemaEnhancer';
-export withBlockExtensions, { resolveExtension } from './withBlockExtensions';
+export withBlockExtensions, {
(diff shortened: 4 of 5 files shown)


tiberiuichim (approved):

CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
 
 ### Bugfix
 
+- Fix `listing` block in SSR, now it's fully variations aware, and the config is passed to the SSR querystring action @sneridagh
```
stevepiercy, line 11:
```suggestion
- Fix `listing` block in SSR, now that it is fully variations aware and the configuration is passed to the SSR `querystring` action. @sneridagh
```

stevepiercy (changes requested):

stevepiercy (approved):

Rudd-O:
This seems correct in my view, though I have no way to effectively evaluate it.  Tests seem to pass.

The fact that the promise of rendering a site to nonjs visitors / search engines was broken... and this seems to be the direction that the stewards of Plone are going with... breaks my heart.

It is the obvious unspoken conclusion that a client-side rendering strategy was eventually going to get us all to the point that client rendering (= js) turned off will net visitors no (or broken) client rendering, no matter how much we want to render for the client server-side.  We just can't test for that on every individual Web site.  If the thing renders the titles but won't render the contents, and there's the odd idiot like me who browses the web without cookies or JS by default, well who's gonna test that?  RMS is busy reading email from 1989.

This should probably be merged anyway.

Sad.

Rudd-O (commented):
^^^

davisagli:
@Rudd-O Your response to someone fixing a server-side rendering issue is to criticize the team for not caring enough about server-side rendering? Wow, I bet that doesn't help motivate the people that are volunteering time to work on it.

I think there are feasible strategies for testing that server-side rendering isn't broken. i.e. load key pages as an anonymous user with cypress and test that the DOM is the same before and after client-side rendering. If you care about it, it'd be more effective to work on making it better than complain about the current situation not meeting your needs.

I hope this doesn't come across as disrespectful. To some extent, I can understand your frustration. But your comment struck a bit of a nerve because I know the Volto team cares about and has put effort into this.

stevepiercy:
https://2022.ploneconf.org looks great to me!

Screen Shot 2022-09-26 at 6 48 29 PM

davisagli:
@stevepiercy The listing of talks is missing key text though (speaker and abstract). I think that is what this will help with...

tiberiuichim:
@Rudd-O I want to point out that it's not "Volto developers don't care about JS disabled", as @sneridagh and I worked hard to make that happen. It's a bug due to a rather new / uncommon feature that was added to Volto. All this is rather fresh stuff, probably around 6 months. There's a ton of development to do in Volto world, we're trying to make a really attractive platform for developers.

mpeeters (approved):

sneridagh:
@Rudd-O Thanks for report this bug on Discord, good catch! 

First of all, I have to state in case of someone reach this PR at some point, and reads it, that Volto in SSR rendering works well, as the many projects developed by the community that are currently in production since _years_ ago. 

This piece of code is probably one of the more complex in Volto, as usually happens when talking about SSR features. It's one that @tiberiuichim and myself spent quite a few time to figure it out and find the right implementation. Finally we come up with a solution inspired in how NextJS is doing things and we manage to address it. Yes, it had an issue, you identified it and pointed it out in the Discord channel and it got fixed within the day... This says a lot about this community, right?

It's true that this community still haven't found time to test and assess these edge cases (since plain page SSR works since years ago), but we will eventually do. Maybe you could take a look and help with it?

I wish we could do something to cheer you up. Volto is not the default UI for Plone 6 because "the stewards of Plone" decided so. It's because the Plone community decided so. Plone 6 Volto is a proven solution that powers dozens of websites nowadays and renew its prowess every day. Are you attending the conference? I'd like to be able to address and convince you of these points and dissipate any concern you could have about the future of Plone.

Rudd-O:
> @Rudd-O Your response to someone fixing a server-side rendering issue is to criticize the team for not caring enough about server-side rendering? Wow, I bet that doesn't help motivate the people that are volunteering time to work on it.
> 
> I think there are feasible strategies for testing that server-side rendering isn't broken. i.e. load key pages as an anonymous user with cypress and test that the DOM is the same before and after client-side rendering. If you care about it, it'd be more effective to work on making it better than complain about the current situation not meeting your needs.
> 
> I hope this doesn't come across as disrespectful. To some extent, I can understand your frustration. But your comment struck a bit of a nerve because I know the Volto team cares about and has put effort into this.

No disrespect taken.  I am frustrated, but I recognize that the Volto team puts effort in ensuring that both alternatives work fine.

Totally stoked, however, at how quickly the issue was remedied.  Yes, that does say a lot about the community, and it's only positive things.  Let's figure out a way to never have this issue crop up in the future ever again.

Issue #3649, opened 2022-09-17
Crash in multilingual website after saving


tiberiuichim:
On a multilingual site, when saving, I get a crash

```
Cannot read properties of undefined (reading 'translations')

    at Toolbar (http://localhost:3001/static/js/client.chunk.js:70824:5)
```

Logging the `this.props.content`, I can see that at one point the content becomes:

```
lock: {locked: false, stealable: true}
```

pnicolli:
Yeah I discussed it at the Volto Team Meeting two weeks ago. At first I thought it was about overriding the content, but it's actually because the content is reset on route change, then the unlock_content reply comes in and you get the situation when you only have the lock info in the content, then the get_content info comes in from the backend and overrides the lock info. If the page is rendered before you receive the get_content reply, the error is thrown.

Fixed in #3685

davisagli:
This was fixed a couple years ago in https://github.com/plone/volto/pull/4377

PR #3685, opened 2022-09-27
Fix race condition in rendering the toolbar in multilingual sites after saving a page

pnicolli:
This happens very rarely but we saw it happen in production. The PATCH to the server goes through and the saving happens, then the user is rerouted to the view, the content reducer is reset and if the rendering happens at that time, before the GET response comes, the view breaks.

Also, in this PR is a minor fix for the content reducer that currently loses locking info when the GET_CONTENT_SUCCESS action comes.


tiberiuichim (approved):
I've noticed the problem too

The worse part, as far as I'm concerned, is that at no point should the store.content be completely replaced. https://github.com/plone/volto/issues/3649

It implies a problem with the lock API design. If possible, the full content should be refresheed together with the lock (if no other solution is found).

sneridagh:
I tried to fix that also, but it turns out the lock mechanism is broken. Maybe we should revisit it and rethink it again.

avoinea (approved):

erral:
We have faced this too.

pnicolli:
> I tried to fix that also, but it turns out the lock mechanism is broken (see tests). Maybe we should revisit it and rethink it again.

@sneridagh Yes it could probably benefit from a rework, but I believe that a very small fix like this could be released in the meantime. This is still breaking multilingual sites from time to time, when I forget to backport this customization to new clients.

pnicolli:
Closing because fixed in #4377

PR #3699, opened 2022-09-29
Support Node 18

sneridagh:
It will be LTS on 2022-10-25, better get prepared.

It seems to me that it's a bit cumbersome that time of the year that we support up to 3 Node versions (Oct-Apr) when 14 will be deprecated... it's CI time nonetheless...

Maybe something we can talk during the conference. /cc @plone/volto-team


nzambello:
I'm using Volto on node 18 for a while, didn't notice any issue

nzambello:
Tests failing ref: https://github.com/webpack/webpack/issues/14532

sneridagh:
This is pitiful: https://github.com/webpack/webpack/issues/14532#issuecomment-947525539

I investigated it a bit. It seems that they won't fix Webpack 4. Either we go full length and upgrade to 5, or we cope with `NODE_OPTIONS=--openssl-legacy-provider` solution... :( 

Great, and it seems that the option is invalid for NodeJS < 18 ... 

Node 18 is LTS in 2022-10-25 :/ 

/cc @plone/volto-team

davisagli:
@sneridagh I guess we need both: for the short term, add `NODE_OPTIONS=--openssl-legacy-provider` for Node 18 only, and in the long term, we do need to update to a modern webpack version sometime. It should be easier now that we have Razzle 4, which supports webpack 5. But it'll take a bit of updating our webpack plugins.

tisto (approved):

davisagli:
We added support for Node 18 in #4086 instead.

PR #3703, opened 2022-09-30
fix: concatenate multilingualRoutes and externalRoutes (if available) to defaultRoutes

### LINKED ISSUE #3653: Default routes provided by the generator don't include everything
https://github.com/plone/volto/blob/05162223a7c08000f3cf86b8ebdd784cbef04fe7/packages/generator-volto/generators/app/templates/src/routes.js#L7

As you see, in the `routes.js` the generator uses the `default_routes`. But the multilanguage routes are not included in the default routes, which is a problem. 

All these:

https://github.com/plone/volto/blob/05162223a7c08000f3cf86b8ebdd784cbef04fe7/src/routes.js#L272

and https://github.com/plone/volto/blob/05162223a7c08000f3cf86b8ebdd784cbef04fe7/src/routes.js#L266-L269

should be moved in the default routes.

erral:
Fixes #3653


src/routes.js
```diff
@@ -85,10 +85,8 @@ export const multilingualRoutes = [
   },
 ];
 
-export const defaultRoutes = (
-  (config.settings?.isMultilingual && multilingualRoutes) ||
-  []
-).concat([
+export const defaultRoutes = [
+  ...((config.settings?.isMultilingual && multilingualRoutes) || []),
```
erral, line 89: :upside_down_face:

ionlizarazu (approved):

tiberiuichim:
May I suggest to move also the external routes fix? Otherwise it's only a partial fix. I know that CodeSyntax worked on the multilingual, but I don't think there are any boundaries when it comes to Volto codebase.

erral:
No problem with that, I moved them to the defaultRoutes

sneridagh (approved):

PR #3704, opened 2022-09-30
Extend Id widget validation rules to accept a dot "."

reebalazs:


reebalazs:
@sneridagh WIP because somewhat incomplete. The tests must be run and the snapshot generated and added to the repo.

Also all languages need an update in the translation at some later point.

sneridagh:
@reebalazs LGTM, but I have a question. Could be that two dots in the name it does fail? Could you please check? 
/cc @davisagli

stevepiercy:
You could have an ID of `......`.

https://regex101.com/r/Xzc7oa/1

Do you want to allow that?

sneridagh:
@stevepiercy nope, sorry I'm not a regex-guru :) I tend to trust them all.

I was asking @davisagli because this has to match with the current id validator algorithm that is present in Plone.

stevepiercy:
Unfortunately I am not familiar enough with the Plone code base to find this, but I know regex well enough to be dangerous.

What is the [IdWidget](https://6.dev-docs.plone.org/storybook/?path=/story/edit-widgets-id--text&globals=measureEnabled:false) used for in Plone?

Better yet, if you point me at the Plone code in question, I could find the regex in use and convert it from Python to JS syntax.

davisagli (approved):
@stevepiercy @sneridagh The IdWidget is used for the `id` field, which is called "Short Name" in the UI. It can be used to rename the object (i.e. the name that becomes part of the URL) from its edit form.

id validation on the backend is complicated, there are parts in Plone (https://github.com/plone/plone.base/blob/main/src/plone/base/utils.py#L361) and in Zope (https://github.com/zopefoundation/Zope/blob/master/src/OFS/ObjectManager.py#L84). So it's quite possible there are additional fixes needed, but this particular change looks good to me. If the entered id passes validation on the frontend but not on the backend, the user will still see a validation error, so it's not a huge problem.

src/components/manage/Widgets/__snapshots__/IdWidget.test.jsx.snap
```diff
@@ -181,7 +226,7 @@ exports[`IdWidget renders an id widget with invalid characters 1`] = `
           <div
             class="ui red pointing basic label"
           >
-            Only lowercase letters (a-z) without accents, numbers (0-9), and the two characters "-" and "_" are allowed.
+            Only lowercase letters (a-z) without accents, numbers (0-9), and the two characters "-" and "_", and dot "." are allowed.
```
davisagli, line 229: @reebalazs This is confusing now, it says two characters and then lists 3 characters.
```suggestion
            Only lowercase letters (a-z) without accents, numbers (0-9), and the characters "-", "_", and "." are allowed.
```

davisagli (changes requested):
I just realized the help text needs to be adjusted.

stevepiercy:
@davisagli I agree it is not a huge problem. But if the validation fails on the frontend, does it send a request to the backend? If so, that would be a nice thing to have.

I also want to ensure that we do not prevent valid IDs from getting to the backend, which this PR fixes by allowing `.` in the ID (but not only `.` and `..` as the entire ID, per https://github.com/zopefoundation/Zope/blob/d1814d0a6bddb615629b552de10e9aa5ad30a6da/src/OFS/ObjectManager.py#L98-L100). Are there other valid characters that we should let through, like `☃` or em-dash? I did not find those as invalid characters in the backend, unless I missed something.

davisagli:
> But if the validation fails on the frontend, does it send a request to the backend? If so, that would be a nice thing to have.

Generally no, but in the case of this widget, yes. It was already possible to set an id that included a period, if you simply ignored the frontend message telling you you couldn't!

> I also want to ensure that we do not prevent valid IDs from getting to the backend, which this PR fixes by allowing . in the ID (but not only . and .. as the entire ID, per https://github.com/zopefoundation/Zope/blob/d1814d0a6bddb615629b552de10e9aa5ad30a6da/src/OFS/ObjectManager.py#L98-L100). Are there other valid characters that we should let through, like ☃ or em-dash? I did not find those as invalid characters in the backend, unless I missed something.

It looks like Zope allows any character above ASCII codepoint `7F` except for `<`, `>`, `&`, and `/`. An underscore is also disallowed at the beginning of the id. At least at first inspection Zope doesn't disallow characters above the ASCII range, but Plone does, by using an ASCIILine field for this field in the schema (https://github.com/plone/plone.app.dexterity/blob/master/plone/app/dexterity/behaviors/id.py#L25)

stevepiercy:
Let me try to summarize this discussion and suggest next steps.

1. Make the suggested changes to this PR, then merge. It fixes an issue now.
2. Create a new issue that defines what we want the frontend to do based on what the backend allows, specifically the `and`ing of all these:
    - No new lines (https://zopeschema.readthedocs.io/en/latest/fields.html#asciiline)
    - Allow only 7-bit bytes characters (https://zopeschema.readthedocs.io/en/latest/fields.html#ascii)
    - No characters that are illegal in URLs: `\x00-\x1F`, `\x7F`, `<`, `>`, and `&`. (https://github.com/zopefoundation/Zope/blob/d1814d0a6bddb615629b552de10e9aa5ad30a6da/src/OFS/ObjectManager.py#L81-L94)
    - Not the empty string `""`. (https://github.com/zopefoundation/Zope/blob/d1814d0a6bddb615629b552de10e9aa5ad30a6da/src/OFS/ObjectManager.py#L90)
    - Is not `.` or `..` (can be traversed). (https://github.com/zopefoundation/Zope/blob/d1814d0a6bddb615629b552de10e9aa5ad30a6da/src/OFS/ObjectManager.py#L98)
    - Cannot start with `_`. (https://github.com/zopefoundation/Zope/blob/d1814d0a6bddb615629b552de10e9aa5ad30a6da/src/OFS/ObjectManager.py#L101)
    - Cannot start with `aq_`. (https://github.com/zopefoundation/Zope/blob/d1814d0a6bddb615629b552de10e9aa5ad30a6da/src/OFS/ObjectManager.py#L105)
    - Cannot end with `__`. (https://github.com/zopefoundation/Zope/blob/d1814d0a6bddb615629b552de10e9aa5ad30a6da/src/OFS/ObjectManager.py#L108)
    - Cannot start with `@@` or `++`. (https://github.com/zopefoundation/Zope/blob/d1814d0a6bddb615629b552de10e9aa5ad30a6da/src/OFS/ObjectManager.py#L112)
    - Cannot be `REQUEST`. (https://github.com/zopefoundation/Zope/blob/d1814d0a6bddb615629b552de10e9aa5ad30a6da/src/OFS/ObjectManager.py#L131)
    - Cannot contain `/`. (https://github.com/zopefoundation/Zope/blob/d1814d0a6bddb615629b552de10e9aa5ad30a6da/src/OFS/ObjectManager.py#L133)

@davisagli how does that sound? Is item 2 complete (or complete enough)?

davisagli:
@stevepiercy Looks good. Thanks, that's a helpful summary.

stevepiercy:
> 2\. Create a new issue that defines what we want the frontend to do based on what the backend allows, specifically the `and`ing of all these:

Created new issue https://github.com/plone/volto/issues/3716

src/components/manage/Widgets/__snapshots__/IdWidget.test.jsx.snap
reebalazs: Done.

reebalazs:
If I understood correctly, the validation remains as it is (multiple dots are currently allowed, and even a trailing dot would pass), and the complex rules go in #3716.

So the only update I did was updating the validation message.

PR #3710, opened 2022-10-02
Add user form password reset request

JeffersonBledsoe:
This PR updates the 'Add user' form to include the 'Send a confirmation mail with a link to set the password' option and including. The existing reset mechanism all seems to work out the box, with an email being sent like below:

```
Welcome,
Your user account has been created. Your username is jeff. 
Please activate it by visiting http://localhost:3000/passwordreset/f2f3718cd7684f0e9d14899fde0de4e5?userid=jeff Please activate your account before 09 Oct, 2022 11:09 AM
```

The given token all works with the existing `passwordreset` form.

Screenshot 2022-10-02 at 12 25 42 pm

I tried to write a cypress test for this workflow using techniques found in <https://www.cypress.io/blog/2021/05/11/testing-html-emails-using-cypress>, but couldn't find a way to simply disable the robot server MockMailHost. I'm happy to finish this test off if anybody knows of a way to disable it in the Volto test setup without too much technical debt!


sneridagh (approved):

sneridagh:
@JeffersonBledsoe what do you mean by disabling the MockMailHost? If you don't have it, the mail will fail to send, right?
What's the problem in Cypress? Do you want to catch something?

JeffersonBledsoe:
@sneridagh I was setting up a test mailserver in Cypress so I could pull the URL out and visit it in the cypress test to check the `passwordreset` page works with the URL that is requested, but I couldn't get Plone to send the mail. Code below is a snippet of how it should work, but mail 

cypress.config.js
```js
    setupNodeEvents(on, config) {
      const port = 25;
      const mailServer = ms.init(port);
      console.log(`mail server started at port ${port}`);
      let lastEmail = {};

      mailServer.bind((addr, id, email) => {
        lastEmail[email.headers.to] = email.html || email.body;
      });

      on('task', {
        resetEmails(email) {
          if (email) {
            delete lastEmail[email];
          } else {
            lastEmail = {};
          }
          return null;
        },
        getLastEmail(email) {
          return lastEmail[email] || null;
        },
      });
    },

```

user-control-panel.js
```js
// In the cypress test:
   cy.setRegistry('plone.email_from_address', 'no-reply@plone.org');
   cy.setRegistry('plone.email_from_name', 'Plone');
// Fill in the form, request a user password and wait a bit for the mail to send

   cy.task('getLastEmail', email).then((email) => {
      cy.log('Getting last email');
      cy.log(email);
      if (email) {
// We should have the password reset email here
        debugger;
      }
    });
```

sneridagh:
@JeffersonBledsoe did you try with a Cypress interceptor?

sneridagh:
@JeffersonBledsoe Maybe I'm missing something, but do you want to test if the mail was sent as expected? Since this is part of the backend, I'll assume it's done well. There are tests there to ensure that, if not you can add it in `plone.restapi`. From my point of view, we don't need that we test this from the frontend in an acceptance test.

sneridagh:
@JeffersonBledsoe I'm in Discord now if you want to talk :)

JeffersonBledsoe:
@sneridagh I wanted to test the mail was sent and the token that was received still worked with the frontend, but if you're happy with the assumption that the token will be fine then I certainly am! I'm around in Discord now if you wanted to carry this on :)

PR #3721, opened 2022-10-04
Add proper webserver with reverse proxy with seamless mode

sneridagh:


sneridagh:
hehe, it will pass now.

PR #3737, opened 2022-10-06
Sort control panels alphabetically by title within each group

JeffersonBledsoe:
The control panels within each group are sorted alphabetically


stevepiercy:
Will sorting work for all languages?

Will the sort be within groups of control panels?

erral:
AFAIK if you are using the title they will be sorted by title, because p.restapi takes care of exposing the translated title in the relevant controlpanel. But I can provide some screenshot in some languages to check it.

erral:
But we already fixed it some time ago in classic to have it sorted by title:

- https://github.com/plone/Products.CMFPlone/pull/3093
- https://github.com/plone/Products.CMFPlone/pull/3094

We should have it both in classic and volto in the same way.  I don't know which one (I prefer have it sorted) but the same.

sneridagh:
I do not know now the state in classic, but I agree that we should follow the same approach.

JeffersonBledsoe:
I updated the PR title to reflect that it's sorted within each group.

> if you are using the title they will be sorted by title

@erral That was my understanding too, but I haven't verified this yet.

> People preferred to have the icon in the same place across languages than have them alphabetically ordered.

@sneridagh My experience with multi-lingual sites is very limited, but I imagine most users would use a single language most of the time? I agree that it's not ideal having them move in cases when they do switch, but I can't think of another way right now for them to be logically sorted the other 80% of the time. I'll finish this PR off, can always revisit it :)

erral:
Here are the screenshots using the branch in this PR with Plone 6b3

controlpanels-fr

controlpanel-es

controlpanels-eu

davisagli:
Hmm, why are there different numbers of control panels shown in the different languages?

erral:
Good question :thinking: I think that's because I didn't change the Volto language when testing this, I only modified Plone language, but I will check it again

erral:
## French
controlpanels-fr-v2

## Basque

controlpanels-eu-v2


## Spanish
controlpanel-es-v2


Here we have the screenshots with frontend + backend both configured correctly with the same language.

The controlpanel differences in French are because of some controlpanels are directly coded in Volto and are not comming from Plone, so their category indicator must be translated into French in the PO/JSON file and it is not. If you check it carefully we have a category called `Général` (coming from Plone) and the one below called `General` which comes from Volto and it is not translated yet.

erral:
@JeffersonBledsoe can you add a changelog entry so that the build is green?

JeffersonBledsoe:
@erral  I'll sort that shortly, but I've still some tests failing locally

tiberiuichim (approved):

JeffersonBledsoe:
@tiberiuichim Update the snapshot and did some more testing, all finished :)

JeffersonBledsoe:
@sneridagh Thoughts on including this in the Plone6 release?

erral (approved):
:+1:

sneridagh:
@JeffersonBledsoe upss I merged your other PR related to the SSR control panel and now there is a conflict that must be resolved.

tkimnguyen:
I have reached my (very low) limit of comprehension about why Controlpanels.test.jsx is problematic :)

sneridagh:
Doh, clear example of buried PR... should we resurect it?
BTW, I took a quick look, so we are sorting by (English - not translated GS property) title? Also in classic?

sneridagh:
Ok I just checked and the endpoint return the translations. @davisagli ok! let's do it.

sneridagh (approved):

PR #3748, opened 2022-10-11
Update the Python minimum requirement to align with README.md and fix…

stevepiercy:
… English grammar


README.md
```diff
@@ -54,7 +54,7 @@ First get all the requirements installed on your system.
 ### Prerequisites
 
 - [Node.js LTS (16.x)](https://nodejs.org/)
-- [Python 3.8.x](https://python.org/) or
+- [Python 3.8.x or later](https://python.org/)
```
davisagli, line 57: @stevepiercy We probably need to be more specific.

The supported Python versions depend on which version of the Plone backend is used. Plone 5.2 supports Python 3.6-3.8. Plone 6 supports Python 3.8-3.10. No Plone release officially supports (pre-release) Python 3.11 yet.

stevepiercy: Agreed, but I am unsure of how all the versions align. Is this correct? Volto alignment with Python and Plone is less clear to me.

| Plone | Python | Volto |
|---|---|---|
| 5.2 | 2.7, 3.6-3.8 | ?-15.0 |
| 6.0 | 3.8-3.10 | 15.0 |

stevepiercy: Are there earlier versions of Volto that work with Plone 5.2? If so, should we mention them?

Volto 16 is still alpha, and Plone 6 is beta. Should we mention those details, too?

README.md
```diff
@@ -54,9 +54,18 @@ First get all the requirements installed on your system.
...
 - [Node.js LTS (16.x)](https://nodejs.org/)
-- [Python 3.8.x](https://python.org/) or
+- [Python](https://python.org/) - See below for specific versions.
 - [Docker](https://www.docker.com/get-started) (if using the Plone docker images)
 
+The versions of Python that are supported in Volto depend on the version of Plone that you use.
+
+| Plone | Python | Volto |
+|---|---|---|
+| 5.2 | 2.7, 3.6-3.8 | 14.0(? or earlier?)-15.0 |
```
davisagli, line 64: Probably easiest to just recommend Volto 15. I am not sure about earlier versions, since I just came back in June. @sneridagh?

davisagli (approved):
Looks accurate. I would just leave out the mention of Volto versions earlier than 15.

README.md
stevepiercy: Done.

davisagli (approved):

PR #3755, opened 2022-10-12
(fix): You may need an additional loader to handle the result of these loaders. (NULLISH OPERATOR

dobri1408:
The fix for the cypress tests that were falling because of nullish operator.


sneridagh:
@dobri1408 can you point me to one of this failing tests? Also, it seems that something is wrong in this PR. Also, I think that the plugins you are adding are already added by Razzle...

PR #3782, opened 2022-10-17
TextLine Edit component

sneridagh:


src/components/manage/TextLineEdit/TextLineEdit.jsx
```diff
@@ -0,0 +1,214 @@
...
+import cx from 'classnames';
+
+const messages = defineMessages({
+  title: {
+    id: 'Type the heading…',
+    defaultMessage: 'Type the heading…',
+  },
+});
+
+function usePrevious(value) {
```
tiberiuichim, line 23: We should put this in the helpers.

src/components/manage/TextLineEdit/TextLineEdit.jsx
```diff
@@ -55,14 +40,22 @@ export const TextLineEdit = (props) => {
...
 
   const [editor] = useState(withReact(createEditor()));
   const [initialValue] = useState([
     {
       type: P,
-      children: [{ text: metadata?.['title'] || properties?.['title'] || '' }],
+      children: [
+        {
+          text:
+            data?.[fieldDataName] ||
```
tiberiuichim, line 52: I think the operator to be used here is `??`

tiberiuichim: Though I don't think it would make a difference for this use case.

Issue #3783, opened 2022-10-17
DateWidget Year Selector


arky:
**Is your feature request related to a problem? Please describe.**
The volto date widget (singledatepicker) has usability issues. It is not possible to quickly switch years without resorting to swiping multiple times. AFAIK there isn't a shortcut for switching years with keyboard.

**Describe the solution you'd like**
In Plone Classic interface, you have separate year selector to select a target year.
![210723c965eaf4ebe05c2bd27506497078fdb97b](https://user-images.githubusercontent.com/808895/196225720-b87e16ea-8081-40b8-bff8-37eeb046f9db.png)

**Describe alternatives you've considered**
react-dates has introduced 'renderMonthElement' that allows you to create month and year dropdown selectors similar to plone classic interface.

yashnerkar:
I would like to work on it

yashnerkar:
can I work on it?

stevepiercy:
`<input type='date'>` and all of its attributes, including `pattern` for graceful degradation, `min`, and `max`, would be my preference as a default date widget. If someone needs a more robust date widget, there are plenty of them out there. I usually regret adopting JavaScript date and time pickers in my projects, and recently have taken great pleasure in ripping them out to use the HTML version.

arky:
+1 for @tiberiuichim suggestion on using input HTML element.

PR #3784, opened 2022-10-17
Add Month, Year Selector

arky:
Provide month and Year drop-download selector using renderMonthElement. This provides a fix for https://github.com/plone/volto/issues/3783

### DIFF
--- a/src/components/manage/Widgets/DatetimeWidget.jsx
+++ b/src/components/manage/Widgets/DatetimeWidget.jsx
@@ -122,6 +122,48 @@ export class DatetimeWidgetComponent extends Component {
     return this.props.dateOnly || this.props.widget === 'date';
   }
 
+  renderMonthElement = ({ month, onMonthSelect, onYearSelect }) => {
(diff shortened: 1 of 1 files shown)


src/components/manage/Widgets/DatetimeWidget.jsx
```diff
@@ -122,6 +122,48 @@ export class DatetimeWidgetComponent extends Component {
     return this.props.dateOnly || this.props.widget === 'date';
   }
 
+  renderMonthElement = ({ month, onMonthSelect, onYearSelect }) => {
+    let i;
+    let years = [];
+    for (i = this.moment().year(); i >= this.moment().year() - 100; i--) {
```
davisagli, line 128: I don't think we should assume the widget is being used to select a date in the past. Maybe default to showing 100 years in the past + 20 in the future, but also provide widget options to customize the range?

davisagli:
Closing because there is a specific widget we'd like to use for this; see #5662

PR #3788, opened 2022-10-21
Create and use express middleware adapter for @portrait endpoint

instification:
Currently portraits are fetched using `flattenToBackendURL`, which is used in the `<img src` tag.

This is problematic when using `internalApiPath` as it then asks the browser to fetch the image from a non-accessible url. For places where the avatar is rendered directly from the SSR code it also doesn't get updated (e.g. comments that display from the ssr and don't trigger a backend call).

This PR creates a simple middleware adapter, the same as for images, which proxies (with auth) to the backend and returns the portrait.

It then uses `flattenToAppUrl` where portraits are used, namely personal tools and comments.


tiberiuichim (approved):

src/express-middleware/portrait.js
```diff
@@ -0,0 +1,28 @@
...
+      // Just forward the headers that we need
+      HEADERS.forEach((header) => {
+        if (resource.headers[header]) {
+          res.set(header, resource.headers[header]);
+        }
+      });
+      res.send(resource.body);
+    })
+    .catch(next);
+}
```
davisagli, line 17: @instification Would it make sense to reuse imageMiddleware here instead of duplicating the code?

sneridagh (approved):
LGTM, waiting for resolving David's question.

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 ### Feature
 
+- Added link integrity potential breakage warning message when deleting a referenced page @danielamormocea
+- Added new components & interfaces for content-rules `Rules` control in Volto. Rules management in both controlpanel and object view. @andreiggr
+- Added portrait middleware adapter. @instification
```
davisagli, line 11: @instification Careful, it looks like you picked up some unrelated changelog entries while merging/rebasing

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 ### Feature
 
+- Added link integrity potential breakage warning message when deleting a referenced page @danielamormocea
+- Added new components & interfaces for content-rules `Rules` control in Volto. Rules management in both controlpanel and object view. @andreiggr
```
davisagli, line 10:
```suggestion
```

CHANGELOG.md
davisagli: I fixed it.

davisagli (approved):

PR #3804, opened 2022-10-26
Allow dumping addon dependency graph to a .dot file

tiberiuichim:
out

### DIFF
--- a/addon-registry.js
+++ b/addon-registry.js
@@ -472,6 +472,37 @@ class AddonConfigurationRegistry {
   getAddonDependencies() {
     return getAddonsLoaderChain(this.dependencyGraph);
   }
+
+  getDotDependencyGraph() {
+    const seen = [];
+    let out = `digraph {
+  layout="fdp"
+  beautify=true
+  sep="+10"
+  splines=curved
+
+  "@package" [color = red fillcolor=yellow style=filled]
+`;
+    let queue = ['@package'];
+    let name;
+
+    while (queue.length > 0) {
+      name = queue.pop();
+
+      const deps = this.dependencyGraph.directDependenciesOf(name);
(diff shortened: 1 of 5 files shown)


addon-registry.js
```diff
@@ -472,6 +472,37 @@ class AddonConfigurationRegistry {
   getAddonDependencies() {
     return getAddonsLoaderChain(this.dependencyGraph);
   }
+
+  getDotDependencyGraph() {
+    const seen = [];
```
davisagli, line 477: Probably doesn't matter at all for performance at this scale, but it might make sense to use a Set here. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set

create-addons-loader.js
```diff
@@ -1,4 +1,5 @@
 const fs = require('fs');
+const path = require('path');
```
davisagli, line 2: Unused import?

docs/source/configuration/environmentvariables.md
```diff
@@ -93,6 +93,14 @@ also
 DEBUG=volto:* yarn start
 ```
 
+#### `DEBUG_ADDONS_LOADER`
+
+Set `DEBUG_ADDONS_LOADER=true` to have Volto generate a file, `addon-dependency-graph.dot` which contains a graph of all the loaded addons. You can convert this file to an image with:
```
davisagli, line 98: (should mention that it requires having graphviz installed)

tiberiuichim:
@davisagli Thanks for the review! I've tweaked the code based on your suggestions

davisagli (approved):
I assume the failing netlify build is unrelated; it looks like it tried to use node 18

avoinea (approved):

sneridagh (approved):

PR #3808, opened 2022-10-27
Refactor alignwidget

sneridagh:


src/components/manage/Widgets/AlignWidget.jsx
```diff
@@ -63,7 +63,10 @@ const AlignWidget = (props) => {
     }
   });
 
-  const icons = { ...ICON_MAP, ...actions_icon_map };
+  const actions_info = {
+    ...default_actions_info({ intl }),
+    ...actions_icon_map,
```
ichim-david, line 68: @sneridagh since the mapping is no longer only about the icon map but also about the title of the
action, how about naming it **actions_options**? 
It makes more sense to have such a naming as there I can configure the id, the icon to use and the title of the action

sneridagh: Fixed.

src/components/manage/Widgets/AlignWidget.jsx
```diff
@@ -1,31 +1,94 @@
...
+
+  React.useEffect(() => {
+    if (!props.value && props.default) {
+      props.onChange(props.id, props.default);
+    }
+  });
+
+  const actions_info = {
+    ...default_actions_info({ intl }),
+    ...actions_info_map,
```
davisagli, line 68: @sneridagh Shouldn't these be camel case?

sneridagh: Fixed!

tiberiuichim (commented):
If you ask me, I would not have an AlignWidget, just a Choice widget that renders as buttons.

sneridagh:
@ichim-david @tiberiuichim ok, how about an align widget (so this is the "old" one, behaving as we expect), then a `ButtonsWidget` that behave completely arbitrarily?

ichim-david:
@sneridagh as I've mentioned in our comments, I still would of enhanced the current work to allow the modification of **defaultActionsInfo**  and the passing of an **actions** parameter that overrides the default actions from AlignWidget so that I can reuse the widget to construct other widgets as you can see in the volto-block-style repo
https://github.com/eea/volto-block-style/blob/develop/src/Widgets/Align.jsx#L67
https://github.com/eea/volto-block-style/blob/develop/src/Widgets/TextAlign.jsx#L37

I would love for the Align widget from Volto to have this behavior of well so that I can just use these widgets as fallbacks when using Volto < 16 but it's up to you to consider if you want to make it even more extensible or move this extensibility to another time in another widget as you've mentioned above.

sneridagh:
@ichim-david a bit more closer, tell me what do you think now.

src/components/manage/Widgets/AlignWidget.jsx
```diff
@@ -50,46 +48,14 @@ const AlignWidget = (props) => {
...
-    if (!props.value && props.default) {
-      props.onChange(props.id, props.default);
-    }
-  });
-
-  const actionsInfo = {
-    ...defaultActionsInfo({ intl }),
-    ...actionsInfoMap,
-  };
+  const actionsInfo = actionsInfoMap || defaultActionsInfo({ intl });
```
ichim-david, line 55: @sneridagh  I think you should bring back 
  const actionsInfo = {
    ...defaultActionsInfo({ intl }),
    ...actionsInfoMap,
  };
otherwise the actionsInfo behaves as a total replacement and it might not be what is needed
Think about this code
 <AlignWidget
          id="align"
          title="Alignment"
          fieldSet="default"
          onChange={() => {}}
          actions={['left', 'right', 'additional']}
          actionsInfoMap={{
            additional: [imageFullSVG, 'Additional action title'],
          }}

I don't want to have to pass left and right since it's already in the defaults
but because of the condition, it's either actionsInfoMap or defaultActionsInfo

sneridagh: but then I don't know how to do what you requested the other day...

sneridagh: Or then you'll use the `ButtonsWidget` instead if you need it?

src/components/manage/Widgets/ButtonsWidget.jsx
```diff
@@ -0,0 +1,41 @@
...
+    <FormFieldWrapper {...props} className="align-widget">
+      <div className="align-buttons">
+        {actions.map((action) => (
+          <Button.Group key={action}>
+            <Button
+              icon
+              basic
+              aria-label={actionsInfoMap[action][1]}
+              onClick={() => onChange(id, action)}
+              active={(action === 'center' && !value) || value === action}
```
tiberiuichim, line 26: hardcoded action?

sneridagh: Fixed.

sneridagh:
@ichim-david @tiberiuichim ok to merge it? What do you think?

tiberiuichim (approved):
I'm ok with the current shape. @ichim-david ?

ichim-david (approved):
@sneridagh I'm ok as well with these changes.

sneridagh:
@ichim-david @tiberiuichim ok thanks!

CHANGELOG.md
```diff
@@ -59,6 +59,7 @@ See https://6.dev-docs.plone.org/volto/upgrade-guide/index.html for more informa
 - Use `View comments` and `Reply to item` permissions in `Comments` component. @razvanMiu
 - Added portrait middleware adapter. @instification
 - Allow dumping the addon dependency graph to a .dot file. Start Volto with `DEBUG_ADDONS_LOADER=true yarn start`, `addon-dependency-graph.dot` will be created in your project folder. @tiberiuichim
+- Improve the `AlignWidget`, add `narrow` fix default support @sneridagh
```
davisagli, line 62: @sneridagh This got added to an old release. One way to avoid this is to wait and add the changelog entry right before merging.

Issue #3791, opened 2022-10-24
No button to test mailhost in Mail control panel


tiberiuichim:
There is a difference in functionality between the Plone Classic and Volto versions

MdSahil-oss:
@tiberiuichim Would you like to elaborate the issue, I wanna work on this.

tiberiuichim:
![image](https://user-images.githubusercontent.com/141568/198306259-034151e2-7e64-4374-bab7-9e50f5c958a3.png)

https://6-classic.demo.plone.org/@@mail-controlpanel

MdSahil-oss:
@tiberiuichim Thanks! understood the issue. Soon will be raising a PR for this enhancement.

MdSahil-oss:
@tiberiuichim does this look good ?

![Screenshot from 2022-10-31 18-50-55](https://user-images.githubusercontent.com/85174511/199023184-027b5813-e353-40aa-ac94-33798a41e10a.png)

tiberiuichim:
@MdSahil-oss Yes, as far as I can say. You'll need to look at the plone.restapi implementation for the "save and send test email". Also, don't forget about the error messages.

MdSahil-oss:
@tiberiuichim I spent some time on plone.restapi repository to find implementation of `Save and send test email` but couldn't find anything relevant, So thinking to implement a new method in plone.restapi repository which will "send a test email first and then trigger saving operation". what's your opinion on this.

tiberiuichim:
I know that @avoinea was recently looking into the "mail sending options" of plone.restapi. Any pointers, on this, Alin?

MdSahil-oss:
@avoinea Do you have any idea of it ?

avoinea:
@MdSahil-oss 

1. There is an endpoint about sending arbitrary emails https://plonerestapi.readthedocs.io/en/latest/endpoints/email-send.html
2. And also for e-mail notifications https://plonerestapi.readthedocs.io/en/latest/endpoints/email-notification.html

Thus, I think we can use the second option for this feature request.

MdSahil-oss:
@avoinea Are you sure ? because the logic that is implemented in the email-notification file is like:
```
data = json_body(self.request)

        sender_from_address = data.get("from", None)
        message = data.get("message", None)
        sender_fullname = data.get("name", "")
        subject = data.get("subject", "")

        if not sender_from_address or not message:
            raise BadRequest("Missing from or message parameters")

[...]

PR #3814, opened 2022-10-31
WIP: Save and send test email button insertion in mail controlpanel

MdSahil-oss:
Save and send test email button insertion in mail controlpanel.
This PR is related to the issue #3791


davisagli:
Old PR that was never completed

Issue #3811, opened 2022-10-29
Block Shortcuts style shouts with UPPERCASE ALL THE THINGS


stevepiercy:
# Describe the bug

When a block is active, and the user clicks `Block` on the right-hand side, the content is mostly uppercase, which is rude and shouty.

## Expected behavior

It would be nice to have sentence case friendly text. Also some text needs to be updated and added.

### ANGRY BLOCK TEXT

![Google Chrome014](https://user-images.githubusercontent.com/102112/198844083-954e368d-95be-4f07-9847-0f0fe3d20065.png)

### Friendly block text

![Google Chrome015](https://user-images.githubusercontent.com/102112/198844097-b4fc1c1c-c4d7-46ee-8b50-b650dcbecc09.png)

## Proposal

This could be resolved by removing this transformation. I don't know how that would affect other parts of the UI.

```css
.ui.secondary.attached.segment {
    text-transform: uppercase;
}
```

Also I would suggest other text updates:

Change block type: `/` to bring up a menu of block types. Use up/down arrow keys and `Enter`, or tap or click, to select a block type.

Strong/Bold: `mod+b`
Emphasis/Italic: `mod+i`
Underline: `mod+u`
Strikethrough: `mod+s`

The modifier key (`mod`) on Windows is `Ctrl`, and on macOS it is `⌘` or `command`.
```

## Software (please complete the following information):

https://6.demo.plone.org/edit

davisagli:
Probably we could detect whether the current OS is Mac or not, so that we can show the correct modifier key rather than making the reader put the pieces together.

tiberiuichim:
The sidebar for the Slate text is mostly an experiment that was never finished.

When we've started to use Volto, the feedback from our users was that it was confusing to have the sidebar automatically switch between the Document and the Block tab (when focusing a Title or Text block, it switches to Document, otherwise it sticks to the Block panel). With draft, when you activated the Block tab panel, there was an empty panel with a big "forbidden" icon. So with volto-slate we've tried to provide "something" for the sidebar. We had the idea that we want volto-slate extensions, and we need to provide a space for them to render, so we've decided just to show the shortcuts, as a sort of placeholder for some future, more relevant content. So the whole "shortcuts + markdown help" can be considered an experiment that can be removed or improved.

ksuess:
[...]

Issue #1867, opened 2020-10-02
Proposal for cut/copy/paste of Volto blocks


tiberiuichim:
I'd like to implement a cut/copy/paste of Volto blocks, I'm looking for feedback on my ideas. I can try to make this as an addon, but I'm missing some plug in points into Volto to make this painless.

So: we add either add new buttons in the Toolbar or in the Blocks/Edit.jsx component. The buttons would be:

- copy the current block
- cut the current block
- (if a clipboard exists) paste the current block

We could also think about a toggle to enter a multi-selection state, where we can checkbox-select multiple blocks (maybe even applying an overlay on top of each of the blocks, and just click-select them).

The paste would only be enabled for empty text blocks, so it's easy to understand what happens. 

This is a sensitive area, I'd want to be involved in the implementation of this idea because I know (our) other form block frameworks. Should we pursue this before or after the Form.jsx refactoring?

tisto:
@tiberiuichim this is something that I would consider to be a core-feature of Volto. Clients always ask us for this feature. Maybe we can ask @albertcasado where we should put the buttons to copy/cut/paste the current block (maybe below the trash can?). Pasting might be harder to figure out (where to put the button).

tiberiuichim:
Below the trashcan is not a good place, the block can be limited to a single line of text. The current blocks editing wrappers are a bit limited, there's not a lot of room for expansion. If we go with something in the block editing area, I'd put it left of the drag/choose block button. Facebook implements their editor like this.

iFlameing:
@tiberiuichim @tisto Here is the pr which I implemented a few months back https://github.com/plone/volto/pull/1567/files

tiberiuichim:
@iFlameing So we put the paste button in the block chooser and the copy button in the block sidebar. I like it. There could be some pitfalls around this, but the initial pain is small. Could you add some screenshots maybe?

tisto:
@tiberiuichim  I am just checking the copy/paste functionality and I was wondering if it is possible to copy/paste a set of blocks from one page to another. I tried to copy a few blocks "SHIFT+COMMAND" on an existing page, then I created a new page and tried to paste with "SHIFT+COMMAND+P" or "COMMAND+P". I guess this is not yet supported, right?

tisto:
@tiberiuichim I found the paste button in the toolbar. Sorry for the noise. :)

tiberiuichim:
[...]

Issue #1851, opened 2020-09-26
Empty placeholders text blocks should not render themselves as break


tiberiuichim:
aka "Volto's "new block" placeholder story"

I see that the empty block placeholder has started to render itself as a `<br/>`, which is fine in a way. The editor wants a new empty space, just like it wants it in a real word processor.

But making the empty blocks significant means that an editor might delete one now, because he doesn't want it, and then get stuck because there's no obvious way to create a new block after another one without a text input.

In my view the empty placeholder block should not be created when inserting non-text blocks. Instead, a plus icon should be shown (centered) below the current block, alowing the editor to insert a new block as desired.

If we continue to make empty paragraph blocks significant then we need to "own" this by not inserting them randomly.

nileshgulia1:
> But making the empty blocks significant means that an editor might delete one now, because he doesn't want it, and then get stuck because there's no obvious way to create a new block after another one without a text input.

In my sense, if we press enter( on a non-text block )and add another block we get a text block which is default and then we can add/replace another blocks with it. By using your opinion we can save user a click when adding non-text block. But again we have to re-think when we decided to make text block as default as a placeholder to render when there's nothing.

tiberiuichim:
I think the bigger issue is that now that we've started to use Volto for real applications we go beyond the territory of what's inside Volto by default and we hit borders that we didn't know exist. One of these borders is the lack of comprehensive guidelines on Volto UX. At this moment we're just winging it: we, as developers, can get to an agreement on code issues, but not having an agreement on UX just leads to frustration and non-productivity.

I don't know if Volto was inspired by Wordpress' Gutenberg and it seems that the scope of that editor is a lot bigger (it wants to be usable on mobiles, which is not a specific goal for Volto, I think), but I'll be using their guidelines as far as possible: https://developer.wordpress.org/block-editor/designers/block-design/

tisto:
@tiberiuichim Gutenberg and Medium.com were certainly our role models. We want something as simple as Medium with some features from Gutenberg (but not as many as Gutenberg to ensure better usability).

I haven't made up my mind about the placeholder text blocks TBH. Though, I think a new text block should be created when the user clicks in the area below the last block. This is how medium does it and with this maybe we would be able to remove the default block.

In any case. Those are things that you have to prototype and try out. Hard to judge without an example.

PR #3815, opened 2022-11-01
Proposal for a more intuitive add block button

### LINKED ISSUE #682: UX: Make "+" more obvious (was: Volto editor first time use, adding images is difficult)
I tested the Volto editor last week as a complete newbie: I didn't look at any screencasts the last few months or documentation so I was as 'blank as possible'.  I first played around on the Volto demo site and later installed Volto with locally because I really thought at first something was wrong with the online demo site. 

This is a 'curse of knowledge' issue: once you know how it works, the problem is gone. And you want to forget as soon as possible that you didn't understand something so obvious and simple from the start....

Inserting an image is not logical:  I couldn't figure out 'when' the (+) sign appears next to text. At first I tried to to modify an existing page (the homepage) and couldn't insert any image, but you have to know you first have to go into the last text block at the end and then press enter to create a new empty text block and only on an empty block there's the option to convert it to other media. duh.

From just visually inspecting the page, checking menu's or typing text in existing blocks you get no visual feedback at all this is "the way".

add_image

One easy way to fix this would be to always keep an empty block with the (+) symbol once you are in editing mode. 

If you create a new page first instead of modifying the homepage,  the (+) symbol is there, but even then I succeeded in wandering around Volto for 10 more minutes because my first response was to fill the element with text and feeling silly wondering where the (+) button went. 

Having an insert something option at the bottom of the editor would also have made the next task easier: Once I add an image block, I couldn't start a new text block below it, except for:

*  first creating an empty text block above the image, 
* then first type some text in it because otherwise there's no drag handle
* and then dragging it below the image
* now delete or replace the text.
 
which feels like a work around.

Add_text_below_image

davisagli:
Here is a prototype for a more intuitive add block button, which fixes #682. Here's a screencast showing how it compares to the status quo (sound on):

https://user-images.githubusercontent.com/49014/199147930-dd0964bb-ca3b-42b8-a349-10efd539efef.mov

Details:
- The add button is now shown below any selected block, instead of only at the left of empty text blocks. This provides a more obvious way to add a block between existing blocks, even if the user doesn't know that the Enter key does this.
- The button is also always shown below the last block, even if it isn't selected, to make it extra obvious how to add a block at the bottom of the page.
- The button is centered horizontally. This avoids the problem with it falling out of the viewport at some widths, and also keeps it well clear of the delete button.
- Adding text blocks with the Enter key still works as before.
- Clicking the add button on an empty text block will replace that block instead of adding below it, to help avoid leaving empty text blocks around accidentally.
- In terms of implementation, the button is added in the EditBlockWrapper, so it should work regardless of block type. It's also positioned relative to the EditBlockWrapper. I tested with a project that has customized the block padding and margins, and it still looked fine.

I'm interested in feedback on whether this is desirable and whether there are edge cases I overlooked. If people are generally in favor, I'm happy to continue the work to update the tests and such.

(I'm aware that the Quanta design also includes a new block add button. This is my attempt to borrow the best parts of that idea without tying it to the rest of Quanta, and also to think through some details that aren't covered in that design like how to make the button work in between multiple blocks without messing up the spacing.)


djay:
@davisagli looks good and solves the problems highlighted in #682. 
vs the notion way of doing it I suggested in #682 there is not that much pros and cons but I will add these two points
But they aren't important compared to getting this into plone 6 before release IMO.
- having the + on left does make it less jarring that add on an empty block changes the current block vs insert it below at other times. But on the other hand the + below makes it more obvious when inserting below an existing block. 
- the notion "click below the last block to create a new text block" as opposed to permanently showing a + does look a little cleaner and still works even then the page is completely blank. But you could also argue the plus makes it obvious for people who are looking rather than clicking. I personally think they will click below as the first thing they will try so going without the permanent + is fine IMO.
notion create new block

stevepiercy:
This is a vast improvement. I like the appearance of only the active block's `+` and the last block's `+`, or if they happen to be one and the same, then just the one.

I also noticed that when a user hover overs a given block, its outline appears, but not its `+`. Adding a `+` on hover would save a click. Instead of click to activate, then click to add block, one could hover and click the `+` to add a block. It might be tricky to implement, though, and I think this improvement is good enough as is. I would not extend this one-fewer-click thing to drag or delete, as I think those actions should be more deliberate.

davisagli:
> Adding a + on hover would save a click. Instead of click to activate, then click to add block, one could hover and click the + to add a block.

I'm open to trying that, but was trying to avoid hover effects since they don't work well on touch devices.

djay:
> > Adding a + on hover would save a click. Instead of click to activate, then click to add block, one could hover and click the + to add a block.
> 
> I'm open to trying that, but was trying to avoid hover effects since they don't work well on touch devices.

I think it's fine since its a progressive enhancement as mobile users would just click.

I also don't see a problem with being able to drag or delete from hover. There is undo and I can't see it being done by accident that easily.

davisagli:
I don't feel strongly for or against the idea of showing the buttons on hover, but it'd be significantly more work and I don't want to add it to the scope of this PR.

stevepiercy:
> There is undo and I can't see it being done by accident that easily.

OMG! Why is this _not_ in the docs?

https://6.dev-docs.plone.org/search.html?q=undo&doc_section=volto

I guess because there is no end-user Editor documentation for Volto? I have an idea here: https://github.com/plone/volto/issues/3811#issuecomment-1298030528

packages/volto-slate/src/blocks/Text/DefaultTextBlockEditor.jsx
```diff
@@ -246,22 +235,6 @@ export const DefaultTextBlockEditor = (props) => {
           }}
         </Dropzone>
 
-        {selected && !data.plaintext?.trim() && !disableNewBlocks && (
```
tiberiuichim, line 249: I think it's great news that the Text block no longer has to take care of rendering the Add block button and it's instead moved to the Wrapper, as it should be.

tiberiuichim:
I'm +1 on merging it ASAP.

erral:
Yes, I would definitely add it in 16

djay:
@sneridagh I'd say it has to go into 6.0. The first time user experience is very important IMO.

@davisagli before this is merged I think there needs to be a change to take care of the case where no blocks are left on the page. The assumption that the title block will always be there only holds for the out of the box experience Page type and the first thing anyone does is create a landing page type that removes the title so you can use something like a hero block instead.
As you can see in the example below I can delete the top block and therefore every block
image

my 2c opinion on how to fix this would would still be use the text cursor and clicking in the blank space below the last block (or no blocks) to create an empty text block. This is something that can be added in addition to the + at the bottom of a block and solves the zero block problem.
I have less concerns that discoverability is an issue than @sneridagh since users tend to click on things when they want something to happen in that space. its natural.
But I guess a permanent + is ok too to solve the zero block problem.

djay (changes requested):
need to solve the zero block/empty page problem

sneridagh:
I'd say that if we can solve it, let's find it out (Quanta has the final + always in place).

However, we had this situation since day one in Volto and we had no complaints reporting this as feedback from our clients. Although certainly a possible use case, it's something that could happen quite unlikely. There are in place also worksronds for it (undo).

But as I said let's think about it. We have to not make tthis a blocker either, once we have an acceptable solution for the rest.

fredvd:
@davisagli Wow. thank you so much. This looks a lot more intuitive for first time or occasional users. 

The only inconsistency / edge case I still see  is with current Volto behavior. The original (+) button was never an add button but a change/replace button: it changed the fresh empty text block that was added through hitting 'ENTER' in the previous block.  Now that the + is a real 'add' in this proposal, it would still be useful to have  'change' button or some other mechanism where empty inserted text blocks (or placeholders) can changed into another type of block.  

For existing users that have the Enter for a new block in their muscle memory, you could open the block selector automatically when giving a second enter.   And mouse users clicking the center (+) already operate the mouse and naturally continue selecting the block type. (Is the block selector keyboard navigable? ) 

But there is still the edge case of an empty (previously) created text block which the editor wants to change, and now there is only the (+) that correctly adds a new block below, but the change button is gone. 

My gut feeling on the (+) being already there on hover to save a click is to hide it andkeep the UX cleaner and simple to remember. first click always selects a block, second click is an action (edit, add).  But as Steve piercy writes it does save an extra click in mouse operation. Undecided.

davisagli:
@djay I agree we should look for a good solution for when there are no blocks. I'll take a look at that.

@fredvd Actually the existing (+) button is not a change/replace button. I just tried it on 6.demo.plone.org and it inserts the new block above the empty text block, without removing the empty text block. I didn't show it in the video but my implementation works like this:
1. If the user clicked the add button on an empty text block, the new block replaces the empty text block.
2. If the user clicked the add button on any other block, the new block is added below the existing block.
So, it is either an add or a replace depending on the context.

I agree the block chooser should be accessible by the keyboard, but I'm not sure what is the best shortcut to use.

tiberiuichim:
Another hidden volto feature :) On an empty text block, type a slash (/), you'll get a dropdown with available blocks, which you can navigate via keyboard

davisagli:
@tiberiuichim I was a bit confused why the `/` dropdown is different from the main block chooser. I guess that's intentional to optimize for keyboard navigation?

djay:
Notion makes this feature less hidden by having a empty block hint of "/ for commands" which is a good idea.

They also make it a shortcut you can use anytime by allowing you to use it anywhere in a text block. At the start is insert before. Anywhere else is insert after. And empty block it replaces the current block. 

It don't see why we couldn't copy all of that behaviour.

tiberiuichim:
@davisagli Yes, it's optimized for keyboard navigation. Actually, I've omitted the part that it also acts as a filter. So if you type "/im" you'll get a dropdown with only two entries, "Image" and "Images grid" (on 6.demo.plone.org)

tisto:
In general, I like the idea. Though, I'd like to show this to our UX experts (@albertcasado et al) to get feedback. That is quite a fundamental change in how Volto works and it definitely needs careful consideration and a major release. We have to take into consideration that our current users will be confused by the change and that our clients will need to amend all their documentation and screenshots.

davisagli:
@tisto Can we get that feedback quickly? I agree it's a change with some impact, so all the more reason to try to get it done before 16 is released rather than wait a long time to fix this usability problem, in my opinion.

tisto:
@davisagli I showed the new implementation to one of our UX experts (in our current main project :)) and his feedback is that overall he thinks this is an improvement. He had two things that we could improve:

- the plus symbol shows up in the block that is currently selected and in the last block, this might confuse people. We used to show the last plus to indicate to editors they can add something. Though, maybe we just want to show the + on the last block if no other block is selected
- the drag handle currently does not show up on an empty text block. This is because we used to show the + icon there. We should amend this as well

Do we have the new implementation shipped somewhere where we can point people to toy around with it?

tiberiuichim:
>  Though, maybe we just want to show the + on the last block if no other block is selected

We already handle this case, we have code that ensures that there's always an empty placeholder block created at the bottom. To test: 6.demo.plone.org, create new Page, on the text placeholder block, click + to open Block Chooser, select Image block. A new placeholder block will be created at the bottom.

davisagli:
> We already handle this case, we have code that ensures that there's always an empty placeholder block created at the bottom. To test: 6.demo.plone.org, create new Page, on the text placeholder block, click + to open Block Chooser, select Image block. A new placeholder block will be created at the bottom.

@tiberiuichim We've found this to be a bit of a problem, because editors forget to remove the empty block at the end, and it messes up the spacing before the footer. In my implementation, I removed it because now the add button is available on any block, so the empty placeholder block doesn't seem necessary.

> the plus symbol shows up in the block that is currently selected and in the last block, this might confuse people. We used to show the last plus to indicate to editors they can add something. Though, maybe we just want to show the + on the last block if no other block is selected

I'll try this.

> the drag handle currently does not show up on an empty text block. This is because we used to show the + icon there. We should amend this as well

Good point.

tiberiuichim:
@davisagli throw-back to an old ticket, https://github.com/plone/volto/issues/1851

davisagli:
Here is a demo site where you can play around with the new add button: https://plone-frontend.onrender.com/

This is running on render.com's free tier, so the instance shuts down after 15 minutes of inactivity. It will probably time out when you first try to visit it, but should work if you try again after a minute or two.

The demo includes several of the suggested improvements from this discussion:
- The add button is no longer always shown on the last block, unless it is selected.
- I added a feature to detect a click in the empty area below the last block, and add a new text block (same effect as hitting Enter while the last block is selected). This also solves the problem of how to add a new block if the page is completely empty.
- The drag handle now appears on empty text blocks.

tiberiuichim:
image

There's a zindex problem with the block chooser.

tiberiuichim:
I think the + on the left side in current Volto is "stranger" then the proposal here, we're just used to it being there. I think the this proposal is good, it solves UI problems for volto. Let's hear more feedback.

davisagli:
@tiberiuichim Interesting, I can reproduce the z-index problem but don't understand it. It happens when I open the chooser between two image blocks, but not when I open it above both of them. The problem seems to go away if I disable this CSS rule: `.ui.drag.block.image { z-index: 2; }`

tiberiuichim:
CSS shenanigans

sneridagh:
@davisagli take a look at the pr that brought that in. I don't remember why it's there, but I can imagine it's because of something.

sneridagh:
I'll also try how the nedted blocks addons behave with this change

davisagli:
@sneridagh note the demo site is deployed from my fork which has a few changes that aren't in this PR yet: https://github.com/davisagli/volto/tree/add-block

stevepiercy:
It's amazing how a small change makes a huge difference. This is very useful. Thank you!

sneridagh:
:( https://plone-frontend.onrender.com/ does not respond now

sneridagh:
It looks good to me, one thing though

> * I added a feature to detect a click in the empty area below the last block, and add a new text block (same effect as hitting Enter while the last block is selected). This also solves the problem of how to add a new block if the page is completely empty.

The text cursor appears all around the editor area, if you click on a side it does nothing, but if you click above the title, it also creates the empty block at the bottom.

I'd like to discuss this next week in the Volto Team Meeting. We could reschedule it to a time that @davisagli could attend too. 
/cc @plone/volto-team

davisagli:
> The text cursor appears all around the editor area, if you click on a side it does nothing, but if you click above the title, it also creates the empty block at the bottom.

@sneridagh Fixed; now it should only add a block when the click is actually below the bottom of the existing blocks. But I haven't thought of a good way to only show the text cursor there. That seems like a pretty minor issue to me.

src/components/manage/Blocks/Block/BlocksForm.jsx
```diff
@@ -161,6 +173,12 @@ const BlocksForm = (props) => {
     }
   }
 
+  useEvent('voltoClickBelowContent', () => {
```
tiberiuichim, line 176: There can be multiple BlocksForm on a screen, at one moment. The BlocksForm is also used by the "blocks container" addons, the accordion, column, tabs, etc.

I propose we move this to the main Edit / Add form, if feasable.

tiberiuichim (changes requested):

src/components/manage/Blocks/Block/BlocksForm.jsx
davisagli: @tiberiuichim Thanks, that is a helpful hint. I think handling the event needs to be here in BlocksForm, so that we have access to `onAddBlock` and `onSelectBlock`. But it looks like I can use the `isMainForm` prop to make sure that only the top-level BlocksForm listens for the event. Right?

src/components/manage/Blocks/Block/BlocksForm.jsx
```diff
@@ -18,6 +18,18 @@ import { useDispatch } from 'react-redux';
...
+  useEffect(() => {
+    // initiate the event handler
+    document.addEventListener(event, handler);
+
+    // this will clean up the event every time the component is re-rendered
+    return function cleanup() {
+      document.removeEventListener(event, handler);
+    };
+  });
+};
```
davisagli, line 31: Note to self: move this to helpers

src/components/manage/BlockChooser/BlockChooserButton.jsx
```diff
@@ -75,7 +74,7 @@ const BlockChooserButton = (props) => {
 
   return (
     <>
-      {!disableNewBlocks && !blockHasValue(data) && (
+      {!disableNewBlocks && (
```
davisagli, line 77: Note to self: make sure disableNewBlocks still works

theme/themes/pastanaga/extras/blocks.less
```diff
@@ -470,18 +469,19 @@ body.has-toolbar.has-sidebar-collapsed .ui.wrapper > .ui.inner.block.full {
...
   z-index: 2;
-  // margin-top: -21px;
-  top: -2px;
-  padding: 0;
-  border: none;
+  bottom: -22px;
+  left: 50%;
+  padding: 0 !important;
+  border: none !important;
+  border-radius: 50% !important;
```
davisagli, line 478: @sneridagh I don't like needing to add all these `!important` rules. They are there to override some of the styles inherited from `.ui.basic.button`. Can I remove some or all of those classes since this is not formatted like a normal UI button?

src/components/manage/Blocks/Text/Edit.jsx
```diff
@@ -329,20 +328,6 @@ export class EditComponent extends Component {
...
-            onInsertBlock={(id, value) => {
-              this.props.onSelectBlock(this.props.onInsertBlock(id, value));
-            }}
-            allowedBlocks={this.props.allowedBlocks}
-            blocksConfig={this.props.blocksConfig}
-            size="24px"
-            className="block-add-button"
-            properties={this.props.properties}
-          />
-        )}
```
davisagli, line 345: To do: test I didn't break draftjs blocks

src/components/manage/Blocks/Block/BlocksForm.jsx
tiberiuichim: Right. I forgot that that flag exists.

src/components/manage/BlockChooser/BlockChooserButton.jsx
tiberiuichim: @davisagli I'm not sure how to test this in vanila. Maybe @avoinea knows?

davisagli: It's available as an option if I enable editable blocks for a content type and edit the layout. It's currently not working, but I think it should be easy to fix.

davisagli: Actually it's working fine here (add button does not show for the block with this setting turned on) but we are missing checks for disableNewBlocks if the user creates a new block using Enter or clicking below all blocks.

src/components/manage/Blocks/Block/BlocksForm.jsx
davisagli: @tiberiuichim https://github.com/plone/volto/commit/a39b81a50b0065ca1b8c8901d795429ca9cb7ec8

ksuess:
First of all: This is a great enhancement!

I think it is important, and you do it this way, to show add buttons only on the selected block. Otherwise add buttons on every block would be a distraction.

I assume it is out of scope of this PR: For blocks like `@eeacms/volto-columns-block` i would like to have the same behavior: show "+" button on bottom border if a sub block is selected.

tiberiuichim:
@ksuess volto-columns-block has its own EditWrapper, similar to what David added here, so we need to add the button in there.

theme/themes/pastanaga/extras/blocks.less
sneridagh: In semantic, since it's so especific from the default CSS itself, sometimes there's no other way... we gave up on that long time ago. The component is using the semantic button, you'll have to remove it and use a normal button, then define all the CSS from scratch to make that work.

src/components/manage/Blocks/Block/BlocksForm.jsx
sneridagh: WP always have a virtual "placeholder" after the last block. It's invisible until you hover over it. In mobile, we could show it always.

sneridagh:
For the sake of completion, this was the initial mockup from PastanagaUI:

image

theme/themes/pastanaga/extras/blocks.less
```diff
@@ -470,18 +469,19 @@ body.has-toolbar.has-sidebar-collapsed .ui.wrapper > .ui.inner.block.full {
...
-  // .ui.basic.button.block-delete-button {
   position: absolute;
   z-index: 2;
-  // margin-top: -21px;
-  top: -2px;
-  padding: 0;
-  border: none;
+  bottom: -22px;
+  left: 50%;
+  padding: 0 !important;
```
avoinea, line 476: IMHO, Changing the padding here could improve the UI a little:

```
0 1rem !important
```

plus-button

avoinea:
@davisagli There is an issue with TTW DX Layout editor. When you press Enter all restrictions on that block are wiped out. See:

[plus-button-in-layout.webm.webm](https://user-images.githubusercontent.com/323385/200550067-4754a5ee-1e0b-47b4-b1cc-061aa2ffce88.webm)

src/components/manage/Blocks/Block/EditBlockWrapper.jsx
```diff
@@ -71,6 +87,25 @@ const EditBlockWrapper = (props) => {
...
+              data={data}
+              block={block}
+              onInsertBlock={(id, value) => {
+                if (blockHasValue(data)) {
+                  onSelectBlock(onInsertBlock(id, value));
+                } else {
+                  onChangeBlock(id, value);
+                }
+              }}
+              onMutateBlock={onMutateBlock}
```
sneridagh, line 101: Just realised of this after a closer test. We are mutating the current block if the block is empty. This differs of the current implementation. Maybe that's the reason why the layout editor is "broken"? /cc @avoinea 

This seems quite innocuous change, but it's important.

davisagli:
> @davisagli There is an issue with TTW DX Layout editor. When you press Enter all restrictions on that block are wiped out. See:

@avoinea I think this is a separate issue. I can reproduce it also on https://6.demo.plone.org/, and my work here doesn't make any changes to what happens when Enter is pressed.

avoinea:
@davisagli I see

djay (approved):

davisagli:
Just to give a bit of an update here... Sorry for the radio silence; I've been focused on another topic at work and was on vacation over the weekend, so I haven't had time to get back to this PR yet.

In general it sounds like there is a lot of interest and I intend to keep working on it. I am also planning to take a look at making the new add button controlled by a (temporary) setting, so that there is a bit more flexibility in how it gets rolled out for existing projects. #3896 explains the reasoning for this more.

theme/themes/pastanaga/extras/blocks.less
davisagli: @avoinea I think that looks fine, but it looks odd if the page has a different background color, because of the way I'm using border-radius to create a circle around the icon. I'm not sure how to create this space while matching the background color, so I think I'll leave it as is for now.

src/components/manage/Blocks/Block/EditBlockWrapper.jsx
davisagli: @sneridagh This is an intentional change, to avoid leaving empty text blocks lying around. Let me know if you want to discuss further. The bug with being able to add blocks with the Enter key even if that was disabled in the layout editor appears to be unrelated to this PR's changes.

davisagli:
I think I've addressed all the feedback (or responded why I think it's out of scope) and it's ready for final review.

djay:
@davisagli 

While I agree experimental feature flags is a good idea and does provide a way to move things forward for some UX features my 2c is that this shouldn't apply in this case. I believe this change should be turned on by default for everyone and perhaps have a way to disable it if required.

Why?

1) its a bug fix not a feature. 
- it's a 4yo bug (https://github.com/plone/volto/issues/682.)
- if you ask anyone to try volto the first time they will invariably get confused on how to insert. this is going to affect a lot of people. Plone 6 audience is one that should not require training. We should be aiming to release an intuitive UX.
- yes it should have been fixed a long time ago. I have ideas on how to fix the process to prevent this in the future but this is a different discussion.

2) demos, downloads, plone 6 docs are all going to be released with a UX that is confusing to a first time user. first impressions count. 6.1 may not come any time soon. Releasing with a marketing message that the new editor experience is easy would be a plus. 

3) its a non breaking fix. it doesn't change of the existing way of inserting works. you can still press enter.

4) it would be a releasing a UX that can't be used on mobile which seems a shame in 2022. (there are other UX preventing this too which should also be fixed before plone 6 IMO). Again Plone 6 first impressions and marketing would benefit from mobile editing label if the fixes are relatively simple.

Maybe we can have 2 flags? an experimental feature flag and a pre-6.0 UX flag?

djay (changes requested):
I think it should be on by default for plone 6.0 and have the ability to turn it off if desired for the sake of not upsetting existing users

davisagli:
> I think it should be on by default for plone 6.0 and have the ability to turn it off if desired for the sake of not upsetting existing users

@djay This is exactly how I've already implemented it, so I'm confused why you've requested changes.

djay (approved):
@davisagli sorry about that. That's great news. 
it wasn't clear from your comments since https://github.com/plone/volto/pull/3815 only talked about a single experimental feature flag that would be turned off by default?

davisagli:
@djay The proposal is one flag per feature, not a single flag that turns on all the experimental features.

davisagli:
@tiberiuichim github shows that you requested changes a while ago; could you update your review?

@sneridagh Can we merge this for Volto 16?

tiberiuichim (approved):

davisagli:
I discussed this PR this morning with @tisto, @ericof, and @sneridagh. Timo is still concerned about merging it with the flag turned on when it hasn't been tested in any actual projects.

So, we plan to merge it with the flag turned off, to make sure it's included in the release candidates. We intend to switch it to be on by default by the time of a final release of Volto 16 as long as we hear positive feedback from a real project.

@djay I know this isn't quite your ideal, and I agree with your arguments for why it should be on by default. But I think it's the path we need to follow to actually get it into Volto 16 while keeping everyone happy. It should work out the way you want unless there is some unanticipated surprise.

djay:
@davisagli I totally understand the need for caution. It's not an ideal situation to be introducing bug fixes so late.
However if the intention is to have it on by default I don't understand the idea of turning it off now? Only a few will know to turn it on and test. Surely the only way to know if there are problems is to have as many people use it as possible? Otherwise we will get to the end of the RC and still be as uncertain about turning it on as we are now?

sneridagh:
My 5c, with the release manager hat on:

When we started Volto 16, we knew that it was going to be Plone 6 final. So we envisioned this release as "the one with slate" and bugfixing release with these goals in mind:

- It will be breaking but without any hard compromise (except the slate one, which in turn, is a big one).
- It will not have any big barrier that could prevent any early adopters sites built in previous versions of Volto to go to Plone 6.
- It will not present any substantial visual change (as in theming) that could break existing themes.

When David came with these proposal, I had to admit that I was +1 to include it in 16 without any further concerns. However, reviewing and writing down again the main goals we had for 16, I admit that it's a hard one to swallow. Since it's a substantial change on UX and how editor users interacts with content, and merging this at the last minute before RC happens without proper real-life and battle testing it, seems wrong to me.

So David came up also with the experimental flags, which allows us to have an scape hatch with this kind of situations. To be honest, I like it a lot. Other frameworks like Next.js are doing it (even Razzle did it in the past). The good thing is that by using them, we could continue releasing all the improvements under 16 without having to move to canary for testing it.

So my proposal is that we don't adopt this change lightly and we take our time to test it properly. Then after this cautionary time passes then we promote it to default, in Volto 17 (since it's a breaking) and then release it along with Plone 6.1 (which will happen much earlier than we are used to). This will give us time to adjust and improve it as long as we receive feedback from our users and projects. 

I know that this could upset people, but we have to make the cut at some point, and 16 is already due since a long time ago and Plone 6 needs to fly now.

davisagli:
@sneridagh I am fine with whatever you decide as the release manager. Did I understand correctly that you are planning to merge it for 16 (with the setting turned off)? Because until it is merged and released, I'm not sure how further testing can happen in the context of real projects.

If I understood that right, then I think it is ready.

tiberiuichim (approved):

djay:
@sneridagh what usability testing has plone 6 undergone?

djay:
@sneridagh @davisagli @tiberiuichim @ericof  I don't think anyone is upset. and that is the wrong way to make these decisions in anycase.
It personally doesn't affect me. I can tell me clients to avoid these bugs and we patch make other fixes in our builds anyway.

UX is not a design you get done once but a continuous process of actively looking for feedback from the users you want rather than the users you have. The users we want aren't going to say they are upset in a github ticket. They just move on.
 
It's suprising to me that given kitconcept have produced this amazing software, put so much money into it, pushed hard to get it as the default plone 6 experience and then when it's 95% of the way to being perfect, let a few small problems get in the way of a great first user plone 6 experience.  
Obviously this wasn't the intention to have an editing experience that requires explanation and training several years ago but that is what plone 6 will be released with IMO.
image

Other than adding new blocks being confusing if not taught how, the mobile experience is unusable, #3856, cut, copy paste is undiscoverable and unusable on mobile - #1867, inability to close the settings preventing editing on mobile, #3910. 

@sneridagh While is great to be able to put one hat on and ignore others, unfortunately there is a UX/product owner hat missing to balance the technical risk with the end user risk. Plone very much misses this role and I'm beginning to think it should be a paid position like release manager. Someone everyone respects like @albertcasado should incentivised to be able to do the thankless job of usability testing and providing quarterly reports on what the biggest UX issues are with Plone. 

I suspect now that one reason UX bugs don't get fixed is it's just a horrible experience. Everyone chiming in with an opinion and often work getting rejected because there is no consensus and no hard data and strong "meritocratic" voices incentivised to push for no changes to the UI. I'd certainly be more willing to put more developer resources into fixing these issues if I knew a UX person blessed it first and would fight for the fix going forward. Maybe this is the reason why #682, a 4yo bug that everyone who first tried volto knew about, got left unfixed for 4 years?

davisagli:
@djay Would you be willing to post that as a new thread on community.plone.org instead of here? I think you brought up some good topics for discussion, but it goes beyond the focus of my PR.

djay:
@davisagli done in https://community.plone.org/t/should-plone-have-a-ux-manager-role/15891 but I zero expectation anything good will come of it. open source and UX just don't mix.

PR #3820, opened 2022-11-01
Show utc time in history view

iRohitSingh:


iRohitSingh:
Screenshot 1944-08-10 at 21 46 04

erral:
I mean to have something like `toLocalizedTime` that we have in Plone that handles the showing internally

erral:
Why do we want to show the UTC time as a tooltip? It would help to show the date in full format, but I don't think that using UTC will help here.

sneridagh:
@iRohitSingh we need to show the client's browser local time.

iRohitSingh:
@sneridagh When we see the history version locally it shows the local time of user but when we deployed the site on server it is only providing the UTC time. We are using the same component everywhere at it works perfectly fine. In folder content view also the time is localized even in server but in history version view it is not showing the local time.

I investigated and found out that there is slight difference between the response of restapi. In folder content there is additional information which is there which is missing in history '+' information which is missing in history view response.

history restapi response.
Screenshot 2022-10-28 at 6 50 24 PM

foldercontent item response.
Screenshot 2022-10-28 at 6 49 40 PM

src/components/theme/FormattedDate/FormattedDate.jsx
```diff
@@ -13,17 +13,30 @@ const FormattedDate = ({
   relative,
   className,
   locale,
+  isHistory,
```
sneridagh, line 16: This is plain wrong, do not change the default behavior of generic components please.

If the attribute `title` has a flaw, let's fix it.

src/components/manage/History/__snapshots__/History.test.jsx.snap
```diff
@@ -241,7 +241,7 @@ exports[`History renders a history component 1`] = `
           >
             <time
               dateTime="2017-04-19T14:09:34+02:00"
-              title="Wednesday, April 19, 2017 at 12:09 PM"
+              title="4/19/2017, 12:09:34 PM UTC"
```
sneridagh, line 244: We don't want UTC, we want local time shown.

sneridagh: Take into account, that if not noted in the TZ properties of the date string, it's UTC always.

src/components/manage/History/History.jsx
```diff
@@ -229,7 +229,7 @@ class History extends Component {
                   </Table.Cell>
                   <Table.Cell>{entry.actor.fullname}</Table.Cell>
                   <Table.Cell>
-                    <FormattedDate date={entry.time} />
+                    <FormattedDate isHistory={true} date={entry.time} />
```
sneridagh, line 232: Please don't modify a component that has a generic purpose like this.

sneridagh:
Closing this, the source of the problem is the history endpoint, who is not returning a TZ'd date.

PR #3821, opened 2022-11-01
Use TextLineEdit instead of DraftJS

sneridagh:


tiberiuichim:
Awesome! Next step, @plone-collective/volto-draftjs

PR #3822, opened 2022-11-01
Add composeSchema helper

tiberiuichim:


davisagli:
Looks useful. I guess https://6.dev-docs.plone.org/volto/blocks/extensions.html would be the place to mention this in docs.

docs/source/blocks/extensions.md
```diff
@@ -160,6 +160,17 @@ The `schemaEnhancer` is a generic extension mechanism provided by
 it for the `variation` extension.
 ```
 
+Volto provides a helper to combine multiple schemaEnhancers into a single
```
stevepiercy, line 163: It is confusing to have a mix of "schemaEnhancers" and "schema enhancers". In narrative text, I think the latter is preferred, whereas in code the former should be an inline literal.

```suggestion
Volto provides a helper to combine multiple schema enhancers into a single
```

docs/source/blocks/extensions.md
```diff
@@ -160,6 +160,17 @@ The `schemaEnhancer` is a generic extension mechanism provided by
...
+function, so that multiple schema enhancers can be used. For example:
+
+```
```
stevepiercy, line 166: I'm not sure whether the syntax is js or jsx.

````suggestion
```js
````

**Review by stevepiercy — CHANGES_REQUESTED** [review]
Minor tweak.

File: docs/source/blocks/extensions.md [hunk 3]
```diff
@@ -160,16 +160,16 @@ The `schemaEnhancer` is a generic extension mechanism provided by
 it for the `variation` extension.
 ```
 
-Volto provides a helper to combine multiple schemaEnhancers into a single
-function, so that multiple schema enhancers can be used. For example:
+Volto provides a helper to combine multiple `schemaEnhancer` into a single
```
stevepiercy, line 163: "objects" might not be the correct word, perhaps "extensions" or "extension mechanisms", following the pattern from the previous note? Or we could get technical and call it "thingies". 😜 

```suggestion
Volto provides a helper to combine multiple `schemaEnhancer` objects into a single
```

docs/source/blocks/extensions.md
```diff
@@ -160,16 +160,16 @@ The `schemaEnhancer` is a generic extension mechanism provided by
...
+function. This allows creating clean, single purpose reusable schema enhancers:
```
stevepiercy, line 164:
```suggestion
function. This allows creating clean, single purpose, reusable schema enhancers:
```

stevepiercy (changes requested):
I am not sure of the right technical JavaScript term to use. Feel free to replace it with something more appropriate.

stevepiercy (approved):

sneridagh (approved):

PR #3830, opened 2022-11-03
Fix undefined error when there is no block present in a page

iFlameing:


davisagli:
@iFlameing You came up with almost the same fix that @reebalazs did in #3831! I would lean toward using @reebalazs' PR, because it includes a test.

sneridagh:
+1, closing this.

PR #3831, opened 2022-11-03
Be more robust towards invalid block configuration

reebalazs:
Be more robust towards invalid block configuration

Do not fail when the layout contains a block that does not exists.
This happened recently in a project where a bug caused faulty block
configuration in the history.

- Show missing blocks in the layout as "Invalid block ..."
- Remove such blocks when saving the page next

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Bugfix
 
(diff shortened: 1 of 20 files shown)


davisagli (approved):
LGTM

sneridagh (approved):

PR #3837, opened 2022-11-04
Upgrade slate versions

tiberiuichim:
This is just to clearly separate the slate packages upgrade

There's actually a bug somewhere, triggered by a change in slate-react 0.82.2 https://github.com/ianstormtaylor/slate/blob/main/packages/slate-react/CHANGELOG.md#0822

When clicking at the end of a slate block, there's a nasty error raised.

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 
 ### Internal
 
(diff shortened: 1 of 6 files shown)


tiberiuichim:
It looks like there's a new attribute being passed, `aria-multiline={true}`, snapshots will need to be updated.

PR #3838, opened 2022-11-04
restrict css selector for error message (volto-slate)

mamico:
The css selector, in `packages/volto-slate`, for the error message is not specific enough.

As a result, for example, after a validation error in a `form`, the class `error` is added to the `form`  and elements within it unexpectedly turn red.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 ### Breaking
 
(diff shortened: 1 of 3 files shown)


tiberiuichim (approved):

sneridagh (approved):
Changelog needed.

PR #3861, opened 2022-11-09
Rearranging value of contextURL in ObjectBroswer so that currentPath …

iFlameing:
…take precedence over other value.

we are doing this because currentPath is the props that is passed by the developer to set the context URL, putting at the end of or operator is sometimes is buggy.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,7 +20,7 @@ See https://6.dev-docs.plone.org/volto/upgrade-guide/index.html for more informa
 - Improve the algorithm that calculates the position of the Slate Toolbar @tiberiuichim
 - The `_unwrapElement` of the volto-slate `ElementEditor` will return an updated range (selection) of the unwrapped element.
 - Replace the main client entry point in `start-client.jsx` anonymous function for a named one. @sneridagh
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -20,6 +20,7 @@ See https://6.dev-docs.plone.org/volto/upgrade-guide/index.html for more informa
 - Improve the algorithm that calculates the position of the Slate Toolbar @tiberiuichim
 - The `_unwrapElement` of the volto-slate `ElementEditor` will return an updated range (selection) of the unwrapped element.
 - Replace the main client entry point in `start-client.jsx` anonymous function for a named one. @sneridagh
+- Rearranging value of contextURL in ObjectBroswer so that currentPath take precedence over other value. @iFlameing
```
davisagli, line 23: I'd suggest rewording this to focus on the impact of the change rather than the implementation details:

```suggestion
- Fix `currentPath` option for `openObjectBrowser`. @iFlameing
```

davisagli (approved):
I tested and this works when currentPath is passed in to openObjectBrowser, and also the fallback to use the current path works when it is not passed in.

I just have a small suggestion to tweak the changelog entry.

sneridagh (approved):

Issue #3474, opened 2022-07-13
When used as folder contents, withQuerystringResults triggers an unneeded network call


tiberiuichim:
https://github.com/plone/volto/blob/92af5ac14730815d925e5d74f3c2d003b1210535/src/components/manage/Blocks/Listing/withQuerystringResults.jsx#L104

Unless the page is different then 0, there's no need to trigger a network fetch, the children are already in the content.

PR #3866, opened 2022-11-09
Use default query in SSR rendering and client fetch of the listing block when we don't have a search criteria

### LINKED ISSUE #3839: "Exception: No query supplied" in the listing block
**Describe the bug**
When we load a page that has a listing block, without a search criteria, we get the error below in Plone:

```bash
2022-11-04 16:51:47 ERROR [Zope.SiteErrorLog:35][waitress-3] Exception: http://localhost:3000/@querystring-search
Traceback (innermost last):
  Module ZPublisher.WSGIPublisher, line 167, in transaction_pubevents
  Module ZPublisher.WSGIPublisher, line 376, in publish_module
  Module ZPublisher.WSGIPublisher, line 271, in publish
  Module ZPublisher.mapply, line 85, in mapply
  Module ZPublisher.WSGIPublisher, line 68, in call_object
  Module plone.rest.service, line 22, in __call__
  Module plone.restapi.services, line 19, in render
  Module plone.restapi.services.querystringsearch.get, line 30, in reply
Exception: No query supplied
```

Who makes this call is [getAsyncData.js](https://github.com/plone/volto/blob/master/src/components/manage/Blocks/Listing/getAsyncData.js)

It sends the `query` parameter as an empty list.

It looks like it's an attempt to render the block server-side, but I don't quite understand it. The fact is that the block works without getAsyncData. So I'm not quite sure what the idea is here.

**To Reproduce**

1. Create a listing block at home, without any search criteria.
2. Press F5 on home.
3. See the Plone log.

**Expected behavior**
No error in Plone.

- OS: Ubuntu 22
- Browser: Chrome
- Volto Version: 16.0.0-alpha.48
- Plone Version 6.0.0b3
- Plone REST API Version: 8.30.0

wesleybl:
Now we use `getQueryStringResults` on `withQuerystringResults` even when we have no search criteria.

The query used will have the same result as the items key of the content json.

Fix #3839

This also makes #3474 obsolete.


wesleybl:
Can someone please take a look here?

@tiberiuichim @sneridagh @ksuess

sneridagh:
@wesleybl sorry for not answering before, since I find it valuable, I think we should discuss it in the Volto Team meeting, since it changes completely the default behavior and this could break feature expectations in existing sites.

Next meeting is next Tuesday. 

/cc @plone/volto-team Feedback?

sneridagh:
@wesleybl also, it's a breaking change (and a bug fix) since it modifies default behavior. Also it differs from the default collections in Plone classic. Could you please note it accordingly in the changelog? Thanks!

wesleybl:
> since it changes completely the default behavior and this could break feature expectations in existing sites.

@sneridagh from the end user's point of view, I tried to do it in a way that doesn't change. The default query that I put, returns the same items that are returned today, when it has no criteria.

> it's a breaking change (and a bug fix) since it modifies default behavior. Also it differs from the default collections in Plone classic. Could you please note it accordingly in the changelog?

I put the entry in Breaking.

wesleybl:
@sneridagh has this been discussed?

tisto:
@sneridagh @wesleybl I checked out the PR locally and I could not spot any difference in the editor UI:

Edit-Page (1)

The fix is definitely beneficial. If there is no user-facing change, we might get away without a breaking change. Do I understand correctly that we send a default query in case the user did not fill out anything to avoid the REST API error on the backend?

Sorry, I did not look into the code yet. Just did a quick superficial review.

wesleybl:
> Do I understand correctly that we send a default query in case the user did not fill out anything to avoid the REST API error on the backend?

Yes. We sent a default query to `getAsyncData.js`, avoiding the backend error. This query is the same used in `withQuerystringResults.jsx` when the user does not select a search criterion. And this default query returns the same result that the code without this PR returns today. So I don't think this will change the result of blocks that don't have a search criteria.

wesleybl:
This PR fix yet another problem that occurs today. Today, if you create a listing block without filling in the search criteria and save, it will return the contents of the container. If you edit it again by selecting the search criteria, it will show the criteria result. But from then on, if you remove the search criteria, it no longer returns the contents of the container. It will return an empty list.

tiberiuichim:
So, let's summarize what this PR is about:

With current volto, the listing block behaves like:

|            | SSR                                     | clientside                                                       |
|------------|-----------------------------------------|------------------------------------------------------------------|
| no query   | incorrectly calls getQuerystringResults | needesly calls getContent()                                      |
| with query | calls getQuerystringResults             | flashes folder contents before showing querystringsearch results |

With this PR, will always call getQuerystringResults. We lose the cool feature we already get the content items for free (which is a source of problems if they're not serialized in the same way).

In principle I would be ok with this change. It's just a matter of being clear about expected behavior.

@sneridagh , @plone/volto-team let's move forward with this PR. What do you guys think?

wesleybl:
> the folder content items now receive more brain data, there's an extra request on all folder contents.

@tiberiuichim current templates work fine with basic metadata. Then we can change it so that the queries don't fetch all the metadata.

An extra request already occurs today, but for `/++api++/` instead of `@querystring-search`

tiberiuichim:
@wesleybl I think the plone.restapi serializer should be enhanced for the items, so that the children are summarized with the same serializer as catalog batch results. This would make sure that the result of content.items is the same as the querystringsearch.

I know that in default Volto this doesn't make a difference, but in projects we have the case where we use "enhanced result items", for example to show the tags. It confusing for the editors when they add a new listing block, they first set the variation and are surprised that they have results, but the information is not there.

wesleybl:
I saw a problem with this implementation. When we are going to add new content, the [initialPath](https://github.com/plone/volto/blob/0f459fed33d72826696c1aae5bb40f220485acb4/src/components/manage/Blocks/Listing/withQuerystringResults.jsx#L33) will be the current folder, since the content we are adding has not yet been created. So, at that moment, the listed contents will be those of the current folder. But I would expect no content to be listed since the content has not yet been created. After saving, the content is created and initialPath will be the path of the new content. At this time, no content will be listed, as there are no created contents within the created content yet.

I tried to do a test to check if current URL ends with `/add` and not call the actions that do the search in that situation. But lint complained that the hooks calls were conditional, and that's not good.

Another option I see is to search for a URL that does not exist, for example for `/undefined`

pnicolli:
@wesleybl is this PR still relevant? I see that the original issue #3839 is closed, maybe this was fixed somewhere else?

wesleybl:
@pnicolli It seems the issue has been fixed, but not for all situations. See @mliebischer' comment:

https://github.com/plone/volto/issues/3839#issuecomment-2637410384

I think the way it is here would fix those problems. So I still think it's relevant.

stevepiercy:
@wesleybl would you please resolve the conflicts and push? Thank you!

I also requested that @mliebischer test out this branch to see if it resolves his report.

wesleybl:
> @wesleybl would you please resolve the conflicts and push? Thank you!

@stevepiercy I fixed the conflicts and updated the branch.

I also fixed https://github.com/plone/volto/pull/3866#issuecomment-1428193405 by testing inside `useDeepCompareEffect`.

wesleybl:
@stevepiercy @sneridagh @pnicolli I updated the branch and fixed the tests. It's ready for review.

stevepiercy (changes requested):
All breaking changes must have an entry in the [Upgrade Guide](https://6.docs.plone.org/volto/upgrade-guide/index.html), per https://6.docs.plone.org/volto/contributing/index.html#document-breaking-changes

packages/volto/news/3839.breaking
```diff
@@ -0,0 +1 @@
+Use default query in SSR rendering and client fetch of the listing block when we don't have a search criteria. Fix https://github.com/plone/volto/issues/3839 @wesleybl
```
stevepiercy, line 1: FYI, by virtue of the name of this file, when this gets included in a release, the process will automatically create an item under the "Breaking" heading with a link to the issue #3839. See for example https://6.docs.plone.org/volto/release-notes/#breaking

```suggestion
Use default query in SSR rendering and client fetch of the listing block when we don't have a search criteria. @wesleybl
```

stevepiercy (changes requested):
All breaking changes must have an entry in the [Upgrade Guide](https://6.docs.plone.org/volto/upgrade-guide/index.html), per https://6.docs.plone.org/volto/contributing/index.html#document-breaking-changes

sneridagh (commented):
@plone/volto-team Every time the defaults in the listing/search block are touched, I get shivers... LGTM, but please, can I get a review from someone else? Thanks!

ksuess:
I see the urge to not break the current behaviour: Show contained items, if no criteria given.

IMHO a listing without criteria should show nothing.

stevepiercy:
If it's not breaking, then the news file needs to be renamed to something more appropriate to its type.

wesleybl:
See the behavior with this PR. I will change it to bugfix.

[recording (1).webm](https://github.com/user-attachments/assets/32fa8bbc-6b96-4eaa-867c-b683e5d5006f)

wesleybl:
@sneridagh @tiberiuichim @pnicolli Can I change these changes to bugfix instead of breacking? The way to do the search has changed but the end result is the same.

sneridagh:
@wesleybl the video looks good, I'd like a review from someone else of the Volto Team.

@ksuess during 19 alphas, we can consider changing the default behavior.

wesleybl:
@sneridagh @sneridagh I changed it to bugfix. Can you please take a look?

stevepiercy (approved):
Change log looks good, but this still needs a technical review from @plone/volto-team that is not @sneridagh.

stevepiercy:
I'm fixing broken links in https://github.com/plone/volto/pull/7010. As soon as that is green, I'll merge it, then update this PR so that the readme linkcheck passes.

pnicolli:
The new implementation looks fine, there's one thing still bothering me that we should find more time to test, which is the cypress test changing the queryCounter from 1 to 3. Do we need to have more than 1 query? This needs to at least be documented in the PR if we decide we need it.

wesleybl:
> The new implementation looks fine, there's one thing still bothering me that we should find more time to test, which is the cypress test changing the queryCounter from 1 to 3. Do we need to have more than 1 query? This needs to at least be documented in the PR if we decide we need it.

@pnicolli This PR changes to use `getQueryStringResults` even when we have no criteria. So the number of requests to `@querystring-search` will increase in this situation.

PR #3871, opened 2022-11-10
Add support for loading core add-ons from the `packages` folder defined in Volto's `package.json`

sneridagh:
So, all add-ons defined there will get into registered add-ons.


CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ See https://6.dev-docs.plone.org/volto/upgrade-guide/index.html for more informa
 
 - Japanese translation updated @terapyon
 - Improve the `AlignWidget`, add `narrow` fix default support @sneridagh
+- Standardize the existence of core `packages` folder. You can define a number of core add-ons present in this folder and they get configured automatically @sneridagh
```
davisagli, line 16: @sneridagh what does "configured automatically" mean? I see below that you have to put packagesFolderAddons in package.json, which doesn't look automatic.

And I'm not sure I understand the point of making it configurable in volto's package.json instead of just hardcoding the list of packages to load.

tiberiuichim (approved):

CHANGELOG.md
sneridagh: Yeah, I had this doubt as well... I had a quick talk to @tiberiuichim and it seemed to us that could be interesting at some point that the list is more "accessible" in terms to be able to configure it, if required.

Background: I want to start developing quanta as a core package, located in the `packages` folder. Same for other potential experimental features that could be implemented as add-ons, but located also in core. These add-ons could be published as separated add-ons so they could be installed in projects separately.

sneridagh: Using this approach, separated from core, is also less prone to have conflicts and you can locate it in long life branches.

sneridagh: I will reword the Changelog.

davisagli: @sneridagh It makes sense to me to have the ability to put optional addons in the volto repository. I'm just confused because you're describing something that is optional per package, but the PR description and the changelog both say that all packages will be loaded automatically. Maybe they are just out of date?

tiberiuichim: @davisagli 

For context and "background": our [packages](https://github.com/plone/volto/tree/master/packages) folder right now has a bunch of things that are not Volto addons, so we can't really use that location for an automatic "all these are volto addons". volto-slate was the first thing that was added there, and to make it work, we include a special case in the addons-registry so that the webpack config to point to that location. Victor's PR fixes the hack by making it a pattern. So it's really an internal setting for Volto itself, nothing that would concern projects, so it doesn't have to be pretty.

tiberiuichim (changes requested):
I think the documentation part belongs to Developer Guidelines. It's an internal feature that belongs to Volto, it should only concern Volto Core developers. 

https://6.dev-docs.plone.org/volto/developer-guidelines/index.html

stevepiercy:
> I think the documentation part belongs to Developer Guidelines. It's an internal feature that belongs to Volto, it should only concern Volto Core developers.

I agree. This is not configuration, but core to Volto.

docs/source/configuration/core-addons.md
```diff
@@ -0,0 +1,24 @@
+# Volto Core add-ons
```
stevepiercy, line 1: This filename should align with its title. Please rename and move to `developer-guidelines/core-add-ons.md`, and update the indexes accordingly.

docs/source/configuration/core-addons.md
```diff
@@ -0,0 +1,24 @@
...
+
+Volto has the concept of "core add-ons".
+They are add-ons that are always installed in Volto.
+They are located separately from the main code in the `packages` folder.
+These add-ons can be released as standalone add-ons, and used in projects that are using versions of Volto that does not have them by default.
```
stevepiercy, line 6:
```suggestion
These add-ons can be released as standalone add-ons, and used in projects that are using versions of Volto that do not have them by default.
```

docs/source/configuration/core-addons.md
```diff
@@ -0,0 +1,24 @@
...
+Another useful use case is the introduction of new or experimental features that need to be developed side-by-side with core Volto, but not enabled by default.
+The idea is that, if needed, you can use these packages released as standalone add-ons installed in your old Volto projects to test them or to have a bleeding edge feature without having to upgrade to the latest Volto version.
```
stevepiercy, line 8:
```suggestion
You can use these packages released as standalone add-ons installed in your old Volto projects to test them or to have a bleeding edge feature without having to upgrade to the latest Volto version.
```

docs/source/configuration/core-addons.md
```diff
@@ -0,0 +1,24 @@
...
+
+```json
+  "packagesFolderAddons": {
+    "volto-slate": {
+      "package": "@plone/volto-slate"
+    }
+  },
+```
+
+The structure is similar to `mrs-developer.json` one, so the key is the name of the folder, then an object with at least the key `package` with the name of the package that is going to be used.
```
stevepiercy, line 24:
```suggestion
The structure is similar to `mrs-developer.json`.
The key is the name of the folder, then an object with at least the key `package` with its value as the name of the package that is going to be used.
```

stevepiercy (changes requested):
Minor tweaks

docs/source/configuration/core-addons.md
sneridagh: Done.

PR #3873, opened 2022-11-11
WIP Fix drag handler alignment for screen width < largeMonitorBreakpoint

reebalazs:
Below 1224px, Both the drag handler and the delete button were pushed inside the block editing border. This fix places them back outside like they should be in all resolutions. Making sure that alignment is the same both with normal blocks and the title block.

Note that the main part of this fix is fixing an override that was not
doing the override (so an explicit bug), but when I was there I also adjusted both icons in
all resolutions correctly.


sneridagh:
@reebalazs I tried it out and the handlers are not visible :(

image

reebalazs:
It looks like the patches are project specific, I was under the impression that this is a Volto issue. Closing.

PR #3875, opened 2022-11-11
Storybook component update

danalvrz:
In progress, not ready to merge.


tiberiuichim:
@danalvrz I think this PR is very interesting, even in its current shape. 

I think it's a bad idea to place the mock image in the src/static, maybe you could find its place in another location, more "test-related" ? Or even use an external hosted image, that would work as well.

danalvrz:
@tiberiuichim thanks for the comments! I actually noticed the folder's bad placement, but forgot to move it. Its new location should be better.

src/components/manage/Widgets/ColorPickerWidget.stories.jsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import ColorPickerWidget from './ColorPickerWidget';
+import WidgetStory from './story';
+
+export const Default = WidgetStory.bind({
+  widget: ColorPickerWidget,
+});
+
+Default.args = {
+  id: 'localhost:3000/something',
```
davisagli, line 10: @danalvrz The `id` used by this widget is a field id, not the full path for the page whose data includes that field. I'd suggest using `favoriteColor` as the example id here, and "Favorite Color" as the title.

(We should also use a more realistic field id and title in the stories for the other widgets.)

src/components/theme/PreviewImage/PreviewImage.stories.jsx
```diff
@@ -0,0 +1,31 @@
...
+    <Wrapper location={{ pathname: '/folder2/folder21/doc212' }}>
+      <div className="ui segment form attached" style={{ width: '400px' }}>
+        <PreviewImage item={item} alt={alt} size={size} {...rest} />
+      </div>
+    </Wrapper>
+  );
+};
+
+const item = {
+  '@id': '/static/media/.storybook/static/preview.png',
```
davisagli, line 17: @danalvrz Is the nebula image you added in this PR in the public domain? If not, we should replace it with an image that we are allowed to use.

src/components/theme/Widgets/Login.stories.jsx
```diff
@@ -0,0 +1,43 @@
...
+  );
+};
+
+export const Default = StoryComponent.bind({});
+Default.args = {
+  login: () => null,
+};
+
+export default {
+  title: 'Internal Components/Login',
```
davisagli, line 33: @danalvrz Let's put both this and the PreviewImage story in the "Public Components" section instead of "Internal Components", since they are things that can be seen by users who have not logged in yet, and not only by editors.

davisagli (approved):

davisagli:
@sneridagh I think this is ready. Leaving it for you to merge in case there's anything you want us to do differently.

tiberiuichim (approved):

sneridagh (approved):

PR #3876, opened 2022-11-12
Load core add-ons configuration as any other add-on.

sneridagh:


tiberiuichim (approved):

sneridagh:
@tiberiuichim there were some errors in tests, and when I dig in, I've found a couple of inconsistencies in the addons list in the registry. I've found that it's better to add them to the list in the dependency graph rather than in the place they were before.

See: https://github.com/plone/volto/pull/3876/commits/557bd4c6ae281d94b4093b8d6c39d506362a1d8a

Let me know what do you think.

PR #3877, opened 2022-11-12
[WIP] Quanta

sneridagh:
Initial implementation of Quanta as a core add-on.

# Tasks

- [x] Port all existing code from the volto-quanta repo
- [x] Add sass loading via razzle plugin
- [x] Enabled by default


davisagli:
@sneridagh Is there any reason we should keep this open?

PR #3892, opened 2022-11-15
Fix keyboard accessibility issue of Clear button in Folder content view

iFlameing:


src/components/theme/Icon/Icon.jsx
```diff
@@ -32,12 +32,13 @@ const defaultSize = '36px';
...
  * for further reference see {@link https://kitconcept.com/blog/pastanaga-icon-system/ | here}
  */
-const Icon = ({ name, size, color, className, title, onClick }) => (
+const Icon = ({ name, size, color, className, title, onClick, tabIndex }) => (
   <svg
     xmlns={name.attributes && name.attributes.xmlns}
     viewBox={name.attributes && name.attributes.viewBox}
     style={{ height: size, width: 'auto', fill: color || 'currentColor' }}
     className={className ? `icon ${className}` : 'icon'}
+    tabIndex={tabIndex}
```
davisagli, line 41: @iFlameing I think we should make tabIndex default to 0 if it is undefined AND onClick is not undefined. That way icons with a click handler will always be reachable by the keyboard, even if the developer using the Icon element does not remember to add the tabindex. Does that make sense?

sneridagh, line 41: Not sure that this belong to the `Icon` component... since it's not a form element at all. Could be that without the proper aria elements the screen readers get confused. I'll add it always to the wrapping element in case that it has an active part of an action or handler. 

In summary, I won't have handlers that are directly mapped to standalone icons. I know there is an "onClick" in there, but I think that it's wrong, and we should deprecate and remove it at some point.

sneridagh (changes requested):

src/components/theme/Icon/Icon.jsx
sneridagh: yes, it would belong in the Contents component, I'd say. One example that we have are the buttons in toolbars and widgets, that contain a single icon. I guess here should be the same.

sneridagh: I'm quite looking these days at `react-aria` from Adobe:

https://react-spectrum.adobe.com/react-aria/useSearchField.html#styling

iFlameing: Should I proceed with the button component?

davisagli: @iFlameing Please try it, but stop and tell us if it is difficult for some reason.

sneridagh (approved):

davisagli (approved):
LGTM

PR #3900, opened 2022-11-16
Ignore /.tool-versions

JeffersonBledsoe:
I use [asdf](https://asdf-vm.com/) for managing my python and node versions rather than pyenv and nvm. While it can read from the `.python-version` and `.nvmrc` files, it's sometimes easier to use it's own config too which is a `.tool-versions` file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -80,3 +80,4 @@ public/critical.css
 # Sphinx and MyST
 docs/_build/
 /.python-version
+/.tool-versions
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 
 ### Internal
 
+- Ignore `.tool-versions` file
+
 ### Documentation
 
 ## 16.0.0-alpha.50 (2022-11-15)


tiberiuichim (approved):

JeffersonBledsoe:
Do we have a 'useful tools' section of the docs outside of the [Effective volto training's development tools chapter
(https://training.plone.org/5/effective-volto/development.html)? I agree, would be nice to have it documented somewhere alongside all the other possible tools we could use with Volto.

@stevepiercy Thoughts on including a 'development tools' section in the Plone docs?

stevepiercy:
There is an effort to standardize some tools for the entire Plone GitHub organization, such as CI tools. I think @mauritsvanrees has a vision about that. I cannot remember who was leading that effort or where it is.

We don't have such a section in Plone 6 Documentation. I don't think it belongs in the Plone 6 Docs, unless it gets standardized for the organization. If a specific project, such as Volto, wants to document it, then that's fine, we'll just pull in their docs.

For the overall Plone 6 Docs, I would much prefer that effort be spent on [filling in the existing blanks in the docs](https://github.com/plone/documentation/issues) over taking on new tasks.

mauritsvanrees:
> There is an effort to standardize some tools for the entire Plone GitHub organization, such as CI tools. I think @mauritsvanrees has a vision about that. I cannot remember who was leading that effort or where it is.

That would mostly be @ericof. Tools:

- Docker image to run code-quality checks: https://github.com/plone/code-quality
- GitHub Action that uses the docker image: https://github.com/plone/code-analysis-action

Issue #3894, opened 2022-11-15
Remove control panel settings that are not relevant in volto


davisagli:
There are some control panel settings that are shown in Volto but only relevant to the Classic UI. We should remove or hide them.

Language control panel:
- [x] `Show language flags`
- [x] `Always show language selector`

Search control panel:
- [x] `Enable LiveSearch`
- Some of the others might also be irrelevant, need to check more closely

Site control panel:
- [x] I'm not sure if `Display publication date` is used in volto
- [x] `Icon visibility`
- [x] All of the settings related to thumbnails
- [x] `Toolbar position`
- [x] `Relative URL for the toolbar logo`
- [x] Default page IDs

Editing control panel:
- [x] `Available editors`
- [x] `Default editor`
- [x] `Enable External Editor feature`

Image Handling control panel
- [x] `High pixel density mode`
- [x] `Image quality at 2x`
- [x] `Image quallity at 3x`
- [x] `Picture variants`
- [x] `Enable image captioning`

Markup control panel:
- [x] Remove the entire control panel (it controls the content types available in tinymce, not volto)

PR #3904, opened 2022-11-16
Filter settings in control panels

davisagli:
Refs #3894 
- [x] Issue ticket to-do items
- [x] Cypress testing


src/components/manage/Controlpanels/Controlpanel.jsx
```diff
@@ -141,6 +141,35 @@ class Controlpanel extends Component {
   }
   form = React.createRef();
 
+  // Goal would be to replace the Form schema prop in line 182, with this function
+  filterSchema = (controlpanel) => {
```
davisagli, line 145: @danalvrz I would move this function outside of the Controlpanel class, since it doesn't use other things from the class.

src/components/manage/Controlpanels/Controlpanel.jsx
```diff
@@ -141,6 +141,35 @@ class Controlpanel extends Component {
...
+    const panelType = controlpanel['@id'].replace('/@controlpanels/', '');
```
davisagli, line 146: @danalvrz The `@id` may be a complete URL, not just a path. I checked and on the demo site it is e.g. `https://6.demo.plone.org/@controlpanels/date-and-time`. So I would try to get the text after the last slash, rather than replacing `/@controlpanels/`

src/components/manage/Controlpanels/Controlpanel.jsx
```diff
@@ -141,6 +141,35 @@ class Controlpanel extends Component {
...
+    }
+
+    // Creates modified version of properties
+    const newPropertiesObj = Object.keys(controlpanel.schema.properties)
+      .filter((key) => !unwantedSettings.includes(key))
+      .reduce((obj, key) => {
+        return Object.assign(obj, {
+          [key]: controlpanel.schema.properties[key],
+        });
+      }, {});
```
davisagli, line 165: @danalvrz I usually try to avoid using reduce, because I always have to stop and think for a minute to remember how it works. Here is another way to do it that I think is easier to read:
```
const newPropertiesObj = Object.fromEntries(
   Object.entries(controlpanel.schema.properties).filter(
      ([key, val]) => !unwantedSettings.includes(key)
   )
);```

File: src/components/manage/Controlpanels/Controlpanel.jsx [hunk 4]
```diff
@@ -141,6 +141,35 @@ class Controlpanel extends Component {
...
+
+    //Should return clone of props.controlpanel.schema, with updated properties
+    return { ...controlpanel.schema, properties: newPropertiesObj };
+    //console.log(controlpanel.schema); //original obj
+    //console.log({ ...controlpanel.schema, properties: newPropertiesObj });
```
davisagli, line 170: Remove the commented debugging code before submitting the final PR

src/components/manage/Controlpanels/Controlpanel.jsx
```diff
@@ -141,6 +141,35 @@ class Controlpanel extends Component {
...
+    let unwantedSettings = [];
+
+    // Allows to specify fields to remove per controlpanel
+    switch (panelType) {
+      case 'language':
+        unwantedSettings = ['display_flags', 'always_show_selector'];
+        break;
+      default:
+        unwantedSettings = ['none!'];
+    }
```
davisagli, line 156: This is fine. But the way I would usually expect to see this written in Javascript is to build a nested object:
```
const unwantedSettings = {
  language: ['display_flags', 'always_show_selector'],
};
```

And then use `unwantedSettings[panelType] || []` below

src/components/manage/Controlpanels/Controlpanel.jsx
```diff
@@ -19,6 +19,7 @@ import { updateControlpanel, getControlpanel } from '@plone/volto/actions';
 
 import saveSVG from '@plone/volto/icons/save.svg';
 import clearSVG from '@plone/volto/icons/clear.svg';
+import { rest } from 'lodash';
```
davisagli, line 22: @danalvrz This is no longer used, so should be removed

src/components/manage/Controlpanels/Controlpanel.jsx
```diff
@@ -43,6 +44,41 @@ const messages = defineMessages({
   },
 });
 
+// Filters props.controlpanel.schema to only valid/relevant fields
+const filterSchema = (controlpanel) => {
+  const panelType = controlpanel['@id'].substring(
+    controlpanel['@id'].lastIndexOf('/') + 1,
+  );
```
davisagli, line 51: @danalvrz You could do
```suggestion
  const panelType = controlpanel['@id'].split('/').pop();
```

davisagli (commented):
@danalvrz I think this looks pretty good. Please continue excluding fields from the other control panels, and work on a cypress test.

We should also try it "for real" in the browser, and make sure it's actually possible to save each control panel. I'm guessing it doesn't work if we exclude a field that is required on the backend...

src/components/manage/Controlpanels/Controlpanel.jsx
davisagli: (If you look at the checks at the bottom of the PR, you'll see that the eslint check failed because of this)

cypress/tests/core/basic/language-controlpanel.js
```diff
@@ -0,0 +1,27 @@
+describe('Language control-panel', () => {
+  beforeEach(() => {
+    cy.visit('/');
+    cy.autologin();
+    cy.waitForResourceToLoad('@navigation');
+    cy.waitForResourceToLoad('@breadcrumbs');
+    cy.waitForResourceToLoad('@actions');
+    //cy.waitForResourceToLoad('@types'); // Uncommenting creates error: > Timeout resolving resource: @types (type undefined)
```
davisagli, line 8: @danalvrz That makes sense. The `@types` endpoint is used to get the schema for a content item when it is being viewed or edited, but since this is not a content view, it is not used. You can remove this line with the comment.

src/components/manage/Controlpanels/Controlpanel.jsx
```diff
@@ -43,6 +44,66 @@ const messages = defineMessages({
...
+    //   'default_page',
+    // ],
+    // editing: ['available_editors', 'default_editor', 'ext_editor'],
+    // imaging: [
+    //   'highpixeldensity_scales',
+    //   'quality_2x',
+    //   'quality_3x',
+    //   'picture_variants',
+    //   'image_captioning',
+    // ],
```
davisagli, line 79: @danalvrz Let's go ahead and uncomment these, if it is working correctly when you test locally.

sneridagh:
@danalvrz looks good! I've discussed with @tiberiuichim and we thought that the filter function should live in the settings, so one can override it and enhance it if needed. That's why I created #3914 and I merged into this one.

Thanks a lot!

tiberiuichim (approved):

wesleybl:
@davisagli couldn't the [filterControlPanelsSchema](https://github.com/plone/volto/blob/6d5ac5fa9eee70d2315aa4bba7e28ab3b0a0d2c4/src/config/ControlPanels.js#L46) function be in a utils, and only have the `unwantedSettings` variable in the Volto configuration? This would make it easier for a portal to add fields. Today I need to customize the entire function to do this.

davisagli:
@wesleybl That would indeed be a better way to do it. I probably won't spend more time on it but if you'd like to submit a PR, I'll review it.

PR #3908, opened 2022-11-17
Allow passing arbitrary style and props to an Icon's SVG

JeffersonBledsoe:
There's many circumstances where we may want to customise the SVG render by an `Icon`, such as passing `aria-hidden` to use it as an icon button, or passing in a style to change the SVG to a block element.

### DIFF
--- a/src/components/theme/Icon/Icon.jsx
+++ b/src/components/theme/Icon/Icon.jsx
@@ -32,13 +32,28 @@ const defaultSize = '36px';
  *
  * for further reference see {@link https://kitconcept.com/blog/pastanaga-icon-system/ | here}
  */
-const Icon = ({ name, size, color, className, title, onClick }) => (
+const Icon = ({
+  name,
+  size,
+  color,
+  className,
+  title,
+  onClick,
+  style = {},
+  ...props
+}) => (
   <svg
     xmlns={name.attributes && name.attributes.xmlns}
     viewBox={name.attributes && name.attributes.viewBox}
-    style={{ height: size, width: 'auto', fill: color || 'currentColor' }}
+    style={{
+      height: size,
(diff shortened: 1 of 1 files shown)


sneridagh:
@JeffersonBledsoe I had this discussion yesterday and I think we have to leave the icons as a decorative thing, and use the proper (form) elements to have these attributes (aria, tabIndex, etc).

I have a close eye on `react-aria` lately, and it advocates for this: 
https://react-spectrum.adobe.com/react-aria/useSearchField.html#styling

As you could see, we introduced an `onClick` on it, but I think it's utterly wrong, the action should go into a form element. At some moment, we should remove it from the Icon component.

Regarding the styles, I'd be ok, but you can already pass a `className` to it. However, I'd be -1 for passing all props to the svg.

JeffersonBledsoe:
> @JeffersonBledsoe I had this discussion yesterday and I think we have to leave the icons as a decorative thing, and use the proper (form) elements to have these attributes (aria, tabIndex, etc).
@sneridagh Perhaps this PR opens up too many issues then, but I think we should at least be able to set `aria-hidden` on an icon to remove it from the accessibility tree. For example, all of the current Toolbar's Icon's get read as a group in VoiceOver currently because they have an accessible name and an image in, even though the image is purely decorative from this perspective. There maybe be more properties that could apply in the case where the SVG is decorative which is why I reached for allowing any prop, but I suppose we could always adjust the Icon component as these needs come up!

JeffersonBledsoe:
@sneridagh I've updated this PR to only allow `id`, `styles` and `aria-hidden` as discussed

sneridagh (approved):

tiberiuichim (approved):

PR #3911, opened 2022-11-17
Objectwidget defaultvalues

tiberiuichim:


sneridagh (approved):

tiberiuichim:
@sneridagh It occured to me that we shouldn't be handling default values at the widget level, we should be handling them at "form" level, just like ObjectWidget is doing.

So, in my changeset, this is wrong:

image

while this is correct and the way to do it (and extend it to the Form or Field implementations as well):

image

The reason is one of simplifying the API that the Widgets need to implement. Let's ask them to deal only with value + onChange, just like a React controlled input needs, and avoid the confusion of `default` vs `defaultValue`. We handle it at form level.

src/components/manage/Form/__snapshots__/InlineForm.test.jsx.snap
```diff
@@ -66,7 +66,7 @@ exports[`Form renders a form component with defaults in the schema - Checkboxes
            - 
           No description
            -
-          No value
+          false
```
tiberiuichim, line 69: This is a bug fix. The test was faulty. It explicitely says "Form renders a form component with defaults in the schema - Checkboxes", passes "false" as value to the checkbox, so this is correct.

sneridagh (approved):

PR #3912, opened 2022-11-17
Add a cog icon into the toolbar

### LINKED ISSUE #3910: can't edit on mobile as hard to close and reopen settings
# Problems.

Currently editing on mobile is not really possible partly due to how settings sidebar works

While editing on mobile might not be a goal for 6.0 release it seems a shame to let something small like this to get in the way doing basic edits.

## Blocks settings
- block settings open by default on mobile taking up most of the screen
- Blocks setting hide is not obvious. very hard to discover (mobile and desktop)
- settings reopen is also hard to find (mobile and desktop)
- settings reopen is close to the edge of the screen so hard to click on some mobiles

# Describe the solution you'd like

## toggle settings button
Implement the quanta solution of toggle in the toolbar now.
Note: this is from quanta design. Quanta toolbar is needed for this, and we'd still keep the tabs in the settings

image

image

   - cog in bottom of toobar which toggles the open and close state of the settings sidebar
       - con: bit weird to have toggle on left to open something on the right?
       - should be at the bottom of toolbar since settings is not changing the page state.
   - close button top left in settings
   - don't change how choice of block vs page settings shown for now
   - keep the current "hardly visible" open close toggle for now
   - make an experimental feature - https://github.com/plone/volto/issues/3896
      - turned on my default?
   - if there is a quanta block menu then there could also be a "settings" item there that specifically opens block settings
   - why:
       - fixes: not obvious how to close settings
       - fixes: not obvious how reopen settings 
       - fixes: hard to reopen settings on mobile
 
What to do about undo, cut and paste?
 - for the moment put them above the settings button since least amount of change
    con: they don't below together and undo buttons should probably move

# Alternatives considered

- settings in toolbar always opens page settings

## Placement of button and undo

- put settings next to save and cancel
  - con: its confusing. it's not a page level action
 - propose putting undo into a context menu in the toolbar
    - it doesn't make sense to put settings with undo since they are different kind of actions
    - context menu would be above save so it's in the right place on horizontal mobile view
    - pro: can put other functions later like find and replace
 - propose removing cut and paste from toolbar and using the [material design select pattern instead](https://github.com/plone/volto/issues/1867)

- gestures only for undo/redo
   - undo should not be visible says [apple](https://developer.apple.com/design/human-interface-guidelines/patterns/undo-and-redo/). we should use a [gesture](https://stackoverflow.com/questions/58547067/where-are-the-gesture-recognizers-for-the-ios-13-undo-redo-gestures), 
   - android doesn't have a standard gesture for undo
- perhaps have [snackbar](https://m2.material.io/components/snackbars#anatomy) to allow for UI to undo
- have an edit mode
   - if an edit is made then replace save cancel bar with undo bar with ```settings          undo redo accept```
   - tick/accept returns you to save and cancel bar
   - this is what google docs mobile does (but then they don't have a "save")

# Additional context

full [quanta design](https://www.figma.com/file/YpaRSKjFcFaAkVYnTSVGuz/Volto-UI-%E2%80%93-Quanta?node-id=1786%3A286) which combines block menu and slate into one. It switched between block settings and page settings based on if a block was selected or not?

JeffersonBledsoe:
This PR adds a settings icon into the toolbar to open and close the sidebar, and a 'Close' icon in the sidebar to make it clear that it can be hidden and how. The current handle at the edge of the settings draw isn't particularly clear and gets missed or be difficult to press by some users, particularly on mobile where screens are smaller and may have curved glass preventing clicks. This change is very similar to the Quanta design

https://user-images.githubusercontent.com/30210785/202449569-af21fa6d-0b2e-4d2e-82e3-453ee34e602a.mov

Screenshot 2022-11-17 at 12 47 23 pm

## Todo:

- Move icon below the 'undo/ redo' on desktop

Closes #3910


tiberiuichim (approved):

JeffersonBledsoe:
@tiberiuichim I'm running into Pluggable problems again. I need the plug dependency so that the component is correctly re-rendered when the sidebar is opened/ closed, but it seems that this causes the pluggable to re-order the plugs and so the button gets moved to above the undo/ redo. Is it possible to force the Plug to be in a certain position, or is this entirely up to the Pluggable?

tiberiuichim:
@JeffersonBledsoe you can pass an order to the Plug

djay:
I think the only difference to the quanta design is in quanta this how you open the page settings only since the blocks vs page tabs are gone. 
To make it closer you could make this button switch to the page settings

djay:
I think I was wrong before when I said the toolbar settings icon should open the page settings only. The quanta design is a bit hard to work out the intention.

Throughout the design in figma the toolbar settings is highlighted when the block settings or the page settings is open. So I think @albertcasado did mean for it relate to the sidebar in general rather than the page settings only.

image
image
image


I really like the design with no tabs. It's cleaner and more mobile friendly saving space.

But with no tabs, what action shows the page settings when a block settings are shown?

The ways I can think of that are consistent with the figma design are
 1. clicking the toolbar settings toggles between block and page settings rather than open/close. Close icon on either settings is the only way to hide the sidebar.
    - if the sidebar is closed then the toolbar settings should open the sidebar and switch to page settings. 
 2. OR clicking the toolbar settings always opens page settings and never toggles. but also happens to indicate if the sidebar is open as well? 
 3. OR - The toolbar settings is a toggle to open/close the sidebar and you have to do something like unselect a block to show the page settings?

but how these all work in with ARIA I'm not sure.
Of those options I think 1 is the more logical choice and it's not incompatible with keeping the tabs for now and it's not incompatible with clicking off a block to switch to page settings either.

JeffersonBledsoe:
CC @Wagner3UB I know you're working on some sidebar stuff so just thought you might want to take a look over this incase it changes your work

Wagner3UB:
@JeffersonBledsoe I have some personal opinions about this new behavior:

- First of all, I don't agree with the "always start close" behavior. We are entering the edit section of the site, and I expected to see some "tools" to change and customize things, which is a well-known behavior and has always been there. Why do I need to take another action to do something that was already available? In my opinion, we shouldn't "clean up" this view but rather show how it works right away. Adding another step creates the feeling that we're not fully inside the "edit page" yet.

- I'm not a fan of placing the action button on the left toolbar. I would suggest that a labeled icon attached to the right sidebar would work better. When clicked, the icon disappears, and by clicking the "X" inside the sidebar, the icon reappears. However, I am aware that this could create some accessibility issues, such as "how do I navigate there using the keyboard?"

It will be something similar to this:

Schermata 2024-09-25 alle 12 37 17

JeffersonBledsoe:
@Wagner3UB Agreed on both points! I think there is some differing opinion on these still though across the community. As this is behind a feature flag to start with, what do you think about having configuration options for both of the points you raised? E.g:

```js
config.experimental.sidebarButton = {
  position: 'toolbar' // Could also be 'floating'
  defaultOpenState: 'open' // Could also be 'closed'
```

djay:
I think open by default and having some new floating settings button can be added later if people really find they need it.
One of the main motivations behind this change is that it's currently unusable on mobile.
You can't have the sidebar open by default on mobile and you really don't want a floating element blocking editing on mobile either. 
These are how quanta looks on mobile. It makes sense where it is.

image
image

I personally don't think it's going to be a problem in practice to have a button that looks like a settings button result in opening the settings even if where it opens is further away. It's still obvious. and it's way more obvious than whats there now.
As for opening it by default I think the same holds.
But it might be a good idea to have the sidebar open to the state it was on the last edit. ie set a cookie.

JeffersonBledsoe:
This PR now MOSTLY is feature-flag only. It does refactor the sidebar to be controlled by a reducer now rather than it controlling itself, however this should make it easier to control other ways in the future should we need it (e.g. open/ close via the keyboard) I've introduced the following feature flags:

```
    sidebarCloseButton: false,
    sidebarToggleButton: {
      position: false, // Can also be 'toolbar' or 'floating'
    },
```

The sidebarCloseButton flag has been introduced so integrators can experiment with improving the mobile experience without introducing the open/ close toggle button as it was discussed at the Salamina sprint that the close button could be a useful feature on it's own.
The sidebar toggle button has been moved behind the`config.experimental.sidebarToggleButton?.position` flag. The This currently only supports 'toolbar' as shown in this PR, however the designers and developers I talked to at the sprint mostly preferred having a button next to the sidebar as they felt it was more logical that the thing it was next to was the thing it controlled.

PR #3916, opened 2022-11-18
Fix spelling in translation

iFlameing:


iFlameing:
@giuliaghisini Do you remember why you added the `(object list)` in translation id?

sneridagh (changes requested):
@iFlameing if you change the id, then all the existing translations have to be updated, and to start the translation is lost for all languages. I'd rather not change the id, leave it as it is, and change the translation you want to change.

iFlameing:
@sneridagh we found the problem. I am creating a new one with only spelling fix.

PR #3921, opened 2022-11-18
Introduce settings.styleClassNameConverters

tiberiuichim:


sneridagh (approved):

src/helpers/Blocks/Blocks.test.js
```diff
@@ -639,5 +663,48 @@ describe('Blocks', () => {
...
+        color: 'red',
+        'inverted:bool': false,
+      };
+      expect(buildStyleClassNamesFromData(styles)).toEqual([
+        'has--color--red',
+        '',
+      ]);
+    });
+
+    it('Ugly edge cases', () => {
```
tiberiuichim, line 700: I don't know if fixing this edge case would be breaking. @sneridagh ?

sneridagh: I'd say it's a bugfix, not in the spec of the feature.

ichim-david (commented):
@sneridagh @tiberiuichim naming is so hard in programming, I like the flexibility of this pull request I have just a small reservation for the noprefix option as I interpret when I read it that has-- will therefore not be added.
I suggested perhaps :omit 
"Theme:omit": "primary" so i think I will only get primary. Maybe you guys have another naming idea that work even better, I figure that perhaps this comment will spark an even better naming suggestion

src/helpers/Blocks/Blocks.test.js
tiberiuichim: Fixed

tiberiuichim:
@ichim-david In the styling you have two parts, you have to take both of them into account: the key and the value. We're stashing extra information in the field name (the :sufix part), but we can't make it confusing to what the result will be, as it doesn't take both parts of the information into account. You'll say "I don't care about the value, I'm only interested in getting rid of those damn classname labels", that's false, because you also want the case where "inverted:bool" = true.

So, "noprefix" is a processing instruction for the end result. And you have a configuration setting in the registry where you can define your own custom converters. So, if you think "omit" is better, register it yourself.

PR #3925, opened 2022-11-18
Fix initial block settings data defaults, improve `applySchemaDefaults` and `ObjectWidget` schema support

tiberiuichim:
I'll try to describe what this PR fixes:

We always had a problem with block default values extracted from schemas. The solution we had in place in InlineForm (based on repeat triggering onChangeField) is buggy because there wasn't a good way to avoid overriding the block value with old values. The reason is that `onChangeField` triggers a React state change. React batches multiple state changes, so in effect we're getting simultaneous calls like `[setState({...oldState, newValue}), setState({...oldState, newValue})]`, and as you can see, the state is non-deterministic, it will overwrite itself with old values stored in a closure.

I've recently introduced a bug in the handling of default values, by making all fields like `<Field value={props.value ?? props.default} />`. This supperficially looks ok, because the inputs show up the correct default value. The problem is that this value is not propagated to the parent form state.

So the solution is to always make sure that, when we create a new "anything data" state based on a schema, we always try to apply the schema defaults on that data, _before_ it's being passed to a callback handler and stored in a state. We do this in the BlocksForm on the `onAddBlock` and `onInsertBlock` (please look at this, maybe we need to do it in other block operations as well) and in the ObjectListWidget (which instantiates new objects based on schemas, when user clicks Add Item).

The ObjectWidget default value is another problem. If I look at a schema that has an `object` widget, I say "it's a regular field, you need to define the `default` prop for it, and that's it". But let's be more friendly, also support the existing code that uses `default` in objectwidget schemas. So we detect the object widget schema in `applySchemaDefaults`, and compute a value based on that. But then we have another issue: the ObjectWidget field value is a JS object (`{}`), but the `ObjectListWidget` also defines a `schema` prop, but its value is an Array (`[]`). So the ObjectListWidget schema prop should be renamed to `itemSchema`. For now I've added the condition in `applySchemaDefaults`.

To make sure we clean our act in InlineForm, we use `onChangeFormData`, which can mutate the whole form state, rather then one field at a time. In `BlockDataform`, we create the `onChangeFormData` by reusing the passed down `onChangeBlock`. So we get to the final thing, if you want properly working default values in your block, you should pass down `onChangeBlock` to your `BlockDataForm`

TODO:

- [x] console.warn if BlockDataForm is not passed down onChangeBlock
- [x] check if intl is passed to applySchemaDefaults, for BBB
- [x] more tests for applySchemaDefaults


tiberiuichim:
TODO: check if `intl` is passed to applySchemaDefaults, if it's not passed, don't try to compute schema defaults.

Also console.warn if BlockDataForm is not passed onChangeBlock.

tiberiuichim:
>  So the ObjectListWidget schema prop should be renamed to itemSchema

There's two proposal on that name:

- itemSchema
- objectSchema

My feeling, and the reason I think itemSchema is better, is because if we name it objectSchema, but the ObjectWidget keeps the `schema` prop, but it's still an "object schema", then why is it not named `objectSchema` as well? So, `itemSchema` to be different from "object schema", because the ObjectListWidget needs a different name for the schema.

sneridagh (approved):

PR #3937, opened 2022-11-20
More block defaults edge cases

sneridagh:
@tiberiuichim cherry picked your last two commits. Let's use the "normal" flow too, then backport to 16.

Supersedes: #3931


tiberiuichim:
@sneridagh Except for the changelog, I think all the code is here.

PR #3962, opened 2022-11-23
Fix user search by full name in users control panel

reebalazs:
Follow up on https://github.com/plone/plone.restapi/commit/6760d55d5fd5ce10adae86b8c79612d5b5545910

This is actually already implemented in the actions but the calling place still missed the change

In short, we want containment search in full name (`search` parameter) and not in the id (`query` parameter, the old way kept for compatibility)


sneridagh (approved):

ksuess:
Please change the CHANGELOG and add my handle. You are using my impementation for the user group panel, offered but not used in user panel.

Issue #3995, opened 2022-11-28
Docs: Settings reference guide is broken


stevepiercy:
**Describe the bug**

[Settings reference guide](https://6.dev-docs.plone.org/volto/configuration/settings-reference.html#) is broken.

PR incoming.

PR #3996, opened 2022-11-28
Close the open glossary list

stevepiercy:
Fixes https://github.com/plone/volto/issues/3995

### DIFF
--- a/docs/source/configuration/settings-reference.md
+++ b/docs/source/configuration/settings-reference.md
@@ -232,6 +232,8 @@ styleClassNameConverters
     data to actual class names. You can customize the generated classname by
     registering fieldnames with names such as `<fieldname>:<converterName>`,
     where the converter is registered here.
+```
+
 
 ## Views settings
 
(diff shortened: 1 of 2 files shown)


davisagli (approved):

docs/source/configuration/settings-reference.md
```diff
@@ -232,6 +232,8 @@ styleClassNameConverters
     data to actual class names. You can customize the generated classname by
     registering fieldnames with names such as `<fieldname>:<converterName>`,
     where the converter is registered here.
+```
```
ksuess, line 235: The glossary has nested directives. I wonder that the build does not complain and asks for increased number of ticks.

ksuess (approved):

docs/source/configuration/settings-reference.md
stevepiercy: The nested directives are indented, thus their context starts new, and are not really "nested". If they were at the same level of indentation, then the outer directives would need to have extra backticks for their nested directives.

BTW, I had the same question, and my testing confirmed this.

PR #3997, opened 2022-11-28
Update razzle and react-dev-utils

davisagli:


tiberiuichim (approved):

sneridagh:
@davisagli @tiberiuichim this is a breaking change, since we are forced to have the dependency set in all projects, so we have to tell everybody to update, if not, the build gets crazy if it has two different Razzle (with their own deps) in there.

We could do it, but maybe it's not worth to push for this minor change now?

davisagli:
@sneridagh I don't think it will work if we can never make bugfix updates to our dependencies. We need to do that to fix security vulnerabilities.

But the main point here is the update to react-dev-utils (because that unblocks updating some transitive dependencies that are vulnerable). I just updated razzle in passing because I noticed there was a new bugfix release. So I could try taking out the razzle update.

sneridagh:
@davisagli yes, I know... this sucks... I just wanted to bring it to the table. Maybe we could find a solution for it... but as long as we have to maintain the direct dependencies in the projects, we will have to deal with this somehow. Even if we make that super easy to manage for the integrators, it would imply an action to be taken, thus the need for mark it as breaking. 😞

davisagli:
@sneridagh It might not be necessary for projects to specify a razzle version directly now that we removed the razzle dependency from `@plone/scripts`...would need to experiment more to be sure.

davisagli:
@sneridagh btw, maybe we should have a test that runs using a project that was generated with volto 16.0.0 rather than the latest generator, so that we can catch if we change something that requires project updates?

davisagli:
@sneridagh Anyway, we can still merge it for master and just not backport it to 16.x.x, right?

pnicolli:
> @davisagli yes, I know... this sucks... I just wanted to bring it to the table. Maybe we could find a solution for it... but as long as we have to maintain the direct dependencies in the projects, we will have to deal with this somehow.

iirc, the npm way of doing this is peerDependencies, razzle should be one of them. I would really really love to have an overhaul of the volto dependency system to be more standard, and yarn3 already pushed us somewhat in this direction, but it's quite a biggie and definitely breaking, but also very beneficial in the long run imho.

...in my dreams volto should become just the framework to be used as an installed dependency, not a razzle project itself. 😬  I meant to discuss this point sometimes in the future, but this thread called for a spoiler :)

davisagli:
@sneridagh We should revisit this one for 17

PR #4003, opened 2022-11-28
Fix select widget loosing focus when the value has changed

reebalazs:
The select widget is doing unnecessary re-renders which will result in rendering a new input field. Consequently it looses focus. To overcome the problem we use the autoFocus property on this widget.


reebalazs:
A deeper explanation. The problem is probably in react-select, whenever the value is changed, it destroys the input element and creates a new one, in effect it looses the focus. It should not happen this way.

For the fix to work, there are different components:

- When the form field value is changed, I'm setting a state variable `inFocus` saying `This element should come to focus now`.
- I am utilizing the already existing `focus` property. This is previously used to bring the first field into focus when the form mounts. This behaviour is kept intact.
- In addition fixing the select field, which, unlike other fields, does not handle the focus property. I am using `autoFocus` to bring the component into focus.

I am not sure from the commit history who is the author of `focus` but in any case please comment if you think anything should be changed or improved here!

tiberiuichim (commented):
Is there a plan to also set focus false by bluring the input?

reebalazs:
> Is there a plan to also set focus false by bluring the input?

Could be done, but manual testing shows that it does not matter. The user can leave the field and the focus does not return. 

However to be prudent, yes it could be done, but only if really necessary, because it would cause an extra render to the entire form.

If you believe we should go for it, I can add this.

tiberiuichim:
@reebalazs No, I just wanted to understand the problem better.

sneridagh:
@reebalazs I tried out locally, and it seems that it does not work. I tried in a Vanilla Volto, Language control panel. Also, we might want it for other select widgets?

reebalazs:
@sneridagh I see the problem with the language control panel, but that's not the issue that I've fixed. My fix includes specifically the case of ArrayWidget, which implements a multiple select. (While, the language selector is a single select, not using an ArrayWidget)

This particular problem is caused by a broken widget, meaning it's not Volto that is broken but the 3rd party component. Any widget should not ever destroy and recreate the input which currently has the focus. I can see now that besides the ArrayWidget, other widget types are also broken in this respect. Unfortunately, I think there is no "generic fix" for the problem: each widget type will require an individual fix or workaround to ensure they work correctly.

I believe it would make sense that I amend this PR and cover the focus fix for the single value select widget (the language selector case). Following that, any new widget type that we also discover broken, could be fixed in follow up PRs.

For the reference, I include the video to demonstrate the issue that I fixed with the ArrayWidget, with the Creator field as an example.

https://user-images.githubusercontent.com/490635/205888640-f9653999-d907-4c71-accd-7bed94e744b0.mp4

reebalazs:
Upon investigation the focus problems with single select (for example language), this is what I found out:

The single select does _not_ suffer from the same problem as the `ArrayWidget`, as it's not destroying the component if the value is updated. So the current solution does not apply to the single select case.

Further investigation shows that the single select widget actually _does_ keep the focus correctly, which you can notice only if you continue typing. So the focus is not lost, only the cursor disappears.

It turns out that there is an open ticket about this in react-select: https://github.com/JedWatson/react-select/issues/3832

Since this is an unrelated issue, I suggest that the current PR is merged and a separate ticket is created about the single select widget. Once the above issue is fixed in the 3rd party component, the problem should be solved. Also, a workaround may be possible but the direct application of the workaround suggestion in the ticket did not work for me. Some modification of the customSelectStyles will likely solve the problem, if there is an interest in pursuing this further.

@sneridagh can you please confirm that the Language field, does in fact keep the focus, only the cursor becomes invisible, but you can continue typing, proving the focus is correct? Or do you see any other issue that we could address here?

CHANGELOG.md
```diff
@@ -54,6 +54,7 @@
 - Improve collapsing of whitespace when pasting to slate text block @tiberiuichim
 - Avoid warning for missing value in NumberWidget @tiberiuichim
 - Fix crash in Slate link editing in a dexterity field @tiberiuichim
+- Fix select widget loosing focus when the value has changed @reebalazs
```
davisagli, line 57: @reebalazs For future PRs, please note that the changelog now uses towncrier. That will help avoid putting the release note under the wrong release, like here.

reebalazs: Sorry for that - old habits leave on :)

PR #4012, opened 2022-12-01
Fix subscript and supscript active at same time.

iFlameing:
@tiberiuichim I don't know is it correct but it works. 

screencast

https://user-images.githubusercontent.com/33936987/205070416-40aa58ba-f06b-485f-bfd9-f615c074eaba.mov

### DIFF
--- a/news/4011.bugfix
+++ b/news/4011.bugfix
@@ -1,0 +1,1 @@
+Fix subscript and supscript active at same time. @iFlameing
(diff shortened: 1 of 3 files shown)


tiberiuichim:
@iFlameing The fix is good, but let's generalize this behavior (I don't like special cases). We could have a setting for "exclusiveElements", which could looks like: 

```
config.settings.slate.exclusiveElements = [
    ['sup', 'sub']
];
```

Then you could make the checks by iterating over that list.

iFlameing:
@tiberiuichim Done from my side and checked for other values also. Can you please review this once again?

tiberiuichim (approved):

sneridagh (approved):

PR #4023, opened 2022-12-03
Pnpm second try

tiberiuichim:


tiberiuichim:
To completely switch, we'd have to change the `.husky/*` scripts.

I know that we don't want to switch, but I think having pnpm as an option is a healthy thing, overall.

tiberiuichim:
Another mention, the pnpm plugin for yarn is incomplete, it doesn't install the scripts of  the packages, so yarn can't fully replace pnpm.

tiberiuichim:
Interesting, with pnpm node_modules we get smaller bundles:

## bundles (`pnpm build`) with pnpm 

image

## bundles with `yarn build`

After nuking node_modules and running `yarn`

image

sneridagh:
> Another mention, the pnpm plugin for yarn is incomplete, it doesn't install the scripts of the packages, so yarn can't fully replace pnpm.

@tiberiuichim what do you mean? What's the culprit?

sneridagh:
@tiberiuichim also, we'll need some kind of deps tracking/sync if we want it in projects too, right?

tiberiuichim:
> > Another mention, the pnpm plugin for yarn is incomplete, it doesn't install the scripts of the packages, so yarn can't fully replace pnpm.
> 
> @tiberiuichim what do you mean? What's the culprit?

yarn + pnpm plugin (setting `nodeLinker: pnpm` in `.yarnrc.yml` doesn't create any `node_modules/.bin/` scripts. You can test this by making the change in yarnrc.yml, nuke the node_modules and run yarn. The problem is that all of our CI scripts depend on having the bin scripts installed. I would consider it a bug in the yarn pnpm plugin... I don't know. Maybe there's a way to fix it.

tiberiuichim:
@sneridagh I think it would be fine, no dependency tracking needed, except to add more of the dev tools (eslint, etc): 

https://github.com/plone/volto/blob/7abef0a3f3f1fd507a04551635ef644cd09c3b87/packages/generator-volto/generators/app/templates/package.json.tpl#L132-L147

tiberiuichim:
@sneridagh (this is just my reasoning, I haven't actually tested anything) I think with the pnpm node_modules layout volto and Volto projects, in general, are less sensitive to depending on a good set of .lock files, as the hoisting issues are completely avoided.

PR #4070, opened 2022-12-08
Don't crash on unknown blocks

tiberiuichim:


nileshgulia1 (approved):
LGTM. I too faced this issue.

davisagli (commented):
I feel like I've seen similar fixes before. It would be good to have a test that unknown blocks don't break rendering the page.

sneridagh (approved):

davisagli (approved):
Thanks!

sneridagh (approved):

PR #4075, opened 2022-12-08
Generator --canary option always gets latest Volto, including any alpha or RC

tiberiuichim:
This is a bugfix, but might be considered "breaking", as the version that will be fetched will no longer necessarily an RC. We use `semver` to sort the released versions, so from a semantic versioniong, this mechanism is better. So it's a bug fix.


avoinea (approved):

davisagli (commented):
Makes sense to me. If I ask for canary, I want the latest, no matter whether it's an alpha or final.

I think the logic here is correct, but I don't see a test demonstrating that, for example, 16.0.0 sorts before 16.0.0-rc1. You could certainly argue that it's not our job to verify that semver does its job correctly. But it is our job to make sure we're using it correctly, and it's hard to tell just by looking at the code. It would be easy to write this test if you move the sorting into a separate function from fetching the dist-tags.

tiberiuichim:
@davisagli I mocked the http get, this makes the test safe as we're now testing for a fixed version number.

davisagli (approved):

sneridagh (approved):

PR #4085, opened 2022-12-11
Fix typo in english translation and add missing french translation

mpeeters:


erral (commented):
Shouldn't we change the original string instead of modifying the English translation? Otherwise we will only have these 2 strings translated.

sneridagh:
@mpeeters I agree with @erral here. Could you please take a look? I'd like to merge it asap, today is the day!

jimbiscuit:
@sneridagh I fix the defaultMessage in the component and remove the translation on po file

ericof (approved):

erral (approved):

sneridagh (approved):

PR #4086, opened 2022-12-12
Node 18 and webpack 5 support

davisagli:
Still lots to test and fix, but here's a start that makes it possible to start volto without errors with webpack 5.

(Targeting volto 17)


src/express-middleware/sitemap.js
```diff
@@ -1,5 +1,5 @@
 import express from 'express';
-import { generateSitemap } from '@plone/volto/helpers';
+import { generateSitemap } from '@plone/volto/helpers/Sitemap/Sitemap';
```
davisagli, line 2: I removed this helper module from helpers/index.js because it's for the server side only and uses the node `zlib` module.

src/components/manage/Controlpanels/VersionOverview.jsx
```diff
@@ -7,7 +7,7 @@ import React from 'react';
 import { FormattedMessage } from 'react-intl';
 import { isEmpty } from 'lodash';
 
-import { version as voltoVersion } from '../../../../package.json';
+import packageJson from '../../../../package.json';
```
davisagli, line 10: Importing specific keys from a JSON file is deprecated.

webpack-plugins/webpack-svg-plugin.js
```diff
@@ -16,10 +16,17 @@ module.exports = {
...
+                name: 'preset-default',
+                params: {
+                  overrides: {
+                    convertPathData: false,
+                    removeViewBox: false,
+                  },
+                },
+              },
+              'removeTitle',
+              'removeUselessStrokeAndFill',
```
davisagli, line 29: Changed config format for newer version of `svgo`

webpack-plugins/webpack-relative-resolver.js
```diff
@@ -33,24 +35,36 @@ class RelativeResolverPlugin {
...
+              nextRequest,
+              null,
+              resolveContext,
+              callback,
+            );
+          } else {
+            callback();
+          }
+        },
+      );
```
davisagli, line 67: Changed API for resolver plugins

tiberiuichim:
https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#webpack4-support-discontinued one more reason to push for this

sneridagh:
### Feedback

I tested it out with Node 18.14.0:
- no problems in Vanilla Volto
- yalc'd the PR in our latest project, all good

The idea is to release it with Volto 17, I'd merge it and start an alpha stage.

@plone/volto-team it would be great if you try it out also in your projects. Here are the detailed instructions.

```
nvm use 18
npm -g i yalc
```

in your local Volto checkout:
```
cd volto
git co webpack-5
yalc publish
```

in your project:
```
yalc add @plone/volto --no-pure
yarn
yarn start
```

after testing: 

```
yalc remove --all
```

sneridagh:
@tiberiuichim since you wrote the webpack plugins it would be great if you could take a second look at them, to make sure that they are migrated correctly.

sneridagh:
@davisagli waiting for your last polishing and review so we can mark it as "ready for review"

davisagli:
@sneridagh I think this is ready for review. Note that in order to merge it, we need to update which checks are required for branch protection, since the checks now run with node 16 and 18 instead of 14 and 16.

sneridagh (approved):
LGTM, this is huge! Thanks @davisagli for working on it!

src/components/manage/Controlpanels/VersionOverview.jsx
```diff
@@ -7,14 +7,8 @@ import React from 'react';
...
 
-import {
-  version as voltoVersion,
-  name as voltoName,
-} from '../../../../package.json';
-import {
-  version as projectVersion,
-  name as projectName,
-} from '@root/../package.json';
+import voltoPackageJson from '../../../../package.json';
```
tiberiuichim, line 10:
```suggestion
import voltoPackageJson from '@plone/volto/../package.json';
```

docs/source/upgrade-guide/index.md
```diff
@@ -24,6 +24,29 @@ The generator will also "update" your project with the latest changes, and propo
...
+### Ending support for NodeJS 14
+
+Long Term Support for NodeJS 14 by the NodeJS community ends in April 2023.
+Volto 17 no longer supports NodeJS 14.
+Please update your projects to a NodeJS LTS version, where either 16 or 18 is supported at the moment of this writing.
+Version 18 is recommended.
+
+#### localhost now resolves to an IPv6 address
+
+NodeJS 18 prefers to resolve `localhost` to an IPv6 address instead of IPv4.
```
tiberiuichim, line 40: I was thinking that `RAZZLE_API_PATH` is kind of "deprecated" due to seamless mode. We should push the `RAZZLE_DEVPROXY_API_PATH` more to the front to make people understand they don't need RAZZLE_API_PATH normally, only if you do development on a non-standard Plone location

Would this imply that we'd have to do development on 127.0.0.1:3000 ?

tiberiuichim (approved):

docs/source/upgrade-guide/index.md
wesleybl: > I was thinking that RAZZLE_API_PATH is kind of "deprecated" due to seamless mode.

I was only able to get Volto working in production with the `RAZZLE_API_PATH`. What am I supposed to do then?

tiberiuichim: @wesleybl Check that your seamless mode is properly installed. You should have nginx redirect the `++api++` paths directly to the plone backend.

erral: @wesleybl we have Volto with https with seamless mode without using any environment variable. [Here is](https://github.com/codesyntax/bobtemplates.cs/blob/master/bobtemplates/cs/cs_plone_volto_buildout/templates/volto.tpl#L84-L91) our nginx configuration.

sneridagh: @wesleybl It's me to blame that the documentation on seamless mode is not more clarifying or understandable, I apologize for that :( You are not the first with doubts about it. Probably the evolution of how things were and how they are now through the years are to blame, in the meanwhile please bear with us.

It's on my list to clarify the whole seamless mode documentation and explanations. I might also create a video about it. The idea is that seamless mode IS the standard de facto way to go for _any_ deployment, and _deprecate_ RAZZLE_API_PATH and do not point people in the documentation to this, unless _really_ necessary.

Thanks @erral for the pointers, @wesleybl the main idea is that Volto (as part of seamless) is capable to detect the HOST header and X-FORWARDED-PROTO or the request protocol, and with this info, it _infers_ the `RAZZLE_API_PATH` from them. So ideally one does not need to set the `RAZZLE_API_PATH` if the webserver (Nginx, Traefik, etc) provides this headers.

I hope this helps you! Also knowing your struggle with the seamless docs would help to build a better ones!

ericof: @wesleybl We had a similar problem in a setup with Traefik + Varnish (due to a misconfiguration of Traefik). The questions here would be:
* How is your set up?
*

wesleybl: @erral @sneridagh @ericof thanks for the clarifications! It wasn't clear to me in the documentation that the protocol came from `X-FORWARDED-PROTO`. I'm using `nginx`. And I have the following configuration in all locations:

```nginx
  proxy_set_header X-Forwarded-Proto $scheme;
  proxy_set_header Host $http_host;
```

Still, it's not working for me.

I'm not using Varnish.

davisagli: @sneridagh Where is the logic for that (inferring the api path from the incoming requests)? What I am seeing is that in production if there is no RAZZLE_API_PATH, then config.apiPath remains empty: https://github.com/plone/volto/blob/master/src/config/index.js#L43

@wesleybl It might help to know more context on where the `++api++` URLs with http are coming from. If they are generated in the Plone backend, then the problem could be that nginx is not using https in the virtual hosting path passed to Zope. We have Traefik proxying `/++api++` directly to Zope with this rewrite: `/VirtualHostBase/https/[hostname]/Plone/++api++/VirtualHostRoot/$1`

davisagli: @tiberiuichim 

> Would this imply that we'd have to do development on 127.0.0.1:3000?

You can still access localhost:3000 in the browser. I guess if the hostname resolves to both an IPv4 and IPv6 address, then either the browser is trying both of them or it is preferring the IPv4 one.

The problem is when Node 17+ is the client. It is no longer preferring IPv4, but is taking the resolved IP addresses in the order returned by the OS. If IPv6 is first, then it will only try that one.

wesleybl: I debugged the request, and `X-FORWARDED-PROTO` was coming as `http`, even when the request was `https`. So I changed the nginx config to this:

```nginx
  map $http_x_forwarded_proto $thescheme {
    default $scheme;
    https https;
  }

  ...

  proxy_set_header X-Forwarded-Proto $thescheme;
```

Accord with: https://serverfault.com/questions/515957/how-to-have-nginx-forward-the-http-x-forwarded-proto-header

I also have a haproxy in front of nginx.

So it worked!

Thank you all!

PR #4115, opened 2022-12-13
Remove clean-css from the CSS minimizing step

sneridagh:
This removes CleanCSS from the minimiser build step in production mode. The project is unmaintained: 
https://github.com/clean-css/clean-css/discussions/1209

There was a problem when using container queries (@container) since it was unrecognising it, they were wiped out.

Razzle was using it (no clue why) instead of using cssnano (the default in css-minimizer-webpack-plugin). This change relies on it for the minimising step instead.

Also the build now seems to be improved by 600B 😱

```
 114.61 KB (-691 B)    build/public/static/css/client.2a553559.chunk.css
```

So good riddance...

### DIFF
--- a/razzle.config.js
+++ b/razzle.config.js
@@ -93,6 +93,9 @@ const defaultModify = ({
       ? 'static/js/[name].js'
       : 'static/js/[name].[chunkhash:8].js';
 
+    const TerserPlugin = require('terser-webpack-plugin');
+    const CssMinimizerPlugin = require('css-minimizer-webpack-plugin');
+
     config.optimization = Object.assign({}, config.optimization, {
       runtimeChunk: true,
       splitChunks: {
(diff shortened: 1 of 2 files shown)


razzle.config.js
```diff
@@ -93,6 +93,9 @@ const defaultModify = ({
       ? 'static/js/[name].js'
       : 'static/js/[name].[chunkhash:8].js';
 
+    const TerserPlugin = require('terser-webpack-plugin');
+    const CssMinimizerPlugin = require('css-minimizer-webpack-plugin');
```
pnicolli, line 97: These are probably better if moved to the top of the file for code quality and readability

tiberiuichim (approved):

razzle.config.js
sneridagh: ups, true!

sneridagh: done.

Issue #3960, opened 2022-11-23
Slate Table block doesn't match /table as block input shortcut


tiberiuichim:
![image](https://user-images.githubusercontent.com/141568/203514832-782935ed-05f7-4024-b497-9c4cf75d93f9.png)

Volto 16.0.0

iRohitSingh:
@tiberiuichim because id of table block is slate "slateTable" and in code we are matching first letter. That`s why we are unable to see table block. 
What should be my next step?

<img width="518" alt="Screenshot 1944-09-02 at 21 14 02" src="https://user-images.githubusercontent.com/61353484/203588713-b354f45f-9f2b-4572-97a3-17a7f5c027c6.png">

tiberiuichim:
@iRohitSingh one quick solution would be that the quicksearch should do a match over the entire text (I'd go as far as block title, but that's discussable. @sneridagh thoughts on this?)

davisagli:
@tiberiuichim @iRohitSingh I'm not sure whether we should look for a match that is not at the start of the block id. But I do think we should also match on the block title (case insensitively). The user is more likely to know the title than the internal id, which might not even be in their language.

PR #4116, opened 2022-12-13
Change SlashMenu to include block title in shortcut search

danalvrz:
Fixes https://github.com/plone/volto/issues/3960

### DIFF
--- a/cypress/tests/core/volto-slate/09-block-slate-slashmenu.js
+++ b/cypress/tests/core/volto-slate/09-block-slate-slashmenu.js
@@ -1,0 +1,18 @@
+import { slateBeforeEach } from '../../../support/e2e';
+
+describe('SlashMenu Test: Shortcuts', () => {
+  beforeEach(slateBeforeEach);
+
+  it('As editor I can crate a Description block using the SlashMenu shortcut', function () {
+    // Use SlashMenu shorcut to create a Description block and type some text in it
+    cy.getSlateEditorAndType('/D').type('{enter}');
+    cy.get('.documentDescription').type('This is a description.');
+    cy.get('.documentDescription').type('{enter}');
+
+    // Save
+    cy.toolbarSave();
+
+    // then the page view should contain a Description with the right text in it
+    cy.get('.documentDescription').contains('This is a description.');
+  });
+});
--- a/cypress/tests/core/volto-slate/09-block-slate-slashmenu.js
+++ b/cypress/tests/core/volto-slate/09-block-slate-slashmenu.js
@@ -3,9 +3,9 @@ import { slateBeforeEach } from '../../../support/e2e';
 describe('SlashMenu Test: Shortcuts', () => {
   beforeEach(slateBeforeEach);
 
-  it('As editor I can crate a Description block using the SlashMenu shortcut', function () {
-    // Use SlashMenu shorcut to create a Description block and type some text in it
-    cy.getSlateEditorAndType('/D').type('{enter}');
+  it('As editor I can create a Description block using the SlashMenu shortcut', function () {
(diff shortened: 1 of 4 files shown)


packages/volto-slate/src/blocks/Text/SlashMenu.jsx
```diff
@@ -110,7 +114,10 @@ const PersistentSlashMenu = ({ editor }) => {
       )
         .filter(
           // TODO: make it work with intl?
-          (block) => slashCommand && block.id.indexOf(slashCommand[1]) === 0,
+          (block) =>
+            slashCommand &&
+            (block.id.indexOf(slashCommand[1]) === 0 ||
+              block.title.toLowerCase().indexOf(slashCommand[1]) === 0),
```
davisagli, line 120: @danalvrz Try typing upper and lower case forms of both ids and titles. I think you need to lower case both the values from the block data and the text that the user typed, before comparing.

cypress/tests/core/volto-slate/09-block-slate-slashmenu.js
```diff
@@ -0,0 +1,18 @@
+import { slateBeforeEach } from '../../../support/e2e';
+
+describe('SlashMenu Test: Shortcuts', () => {
+  beforeEach(slateBeforeEach);
+
+  it('As editor I can crate a Description block using the SlashMenu shortcut', function () {
```
davisagli, line 6:
```suggestion
  it('As editor I can create a Description block using the SlashMenu shortcut', function () {
```

cypress/tests/core/volto-slate/09-block-slate-slashmenu.js
```diff
@@ -0,0 +1,18 @@
...
+    // Use SlashMenu shorcut to create a Description block and type some text in it
```
davisagli, line 7:
```suggestion
    // Use SlashMenu shortcut to create a Description block and type some text in it
```

cypress/tests/core/volto-slate/09-block-slate-slashmenu.js
```diff
@@ -0,0 +1,18 @@
...
+    cy.getSlateEditorAndType('/D').type('{enter}');
```
davisagli, line 8: I'd suggest adding more of the block title, just to make the test more robust in case other blocks with the letter D are added in the future and it's no longer the first option.

```suggestion
    cy.getSlateEditorAndType('/Desc').type('{enter}');
```

packages/volto-slate/src/blocks/Text/SlashMenu.jsx
```diff
@@ -110,7 +117,10 @@ const PersistentSlashMenu = ({ editor }) => {
       )
         .filter(
           // TODO: make it work with intl?
```
davisagli, line 119: Matching the title should take care of making it work in other languages, so we can remove the comment.

```suggestion
```

packages/volto-slate/src/blocks/Text/keyboard/slashMenu.js
```diff
@@ -1,5 +1,6 @@
 export const slashMenu = ({ editor, event }) => {
   if (!editor.showSlashMenu) return;
+  event.preventDefault();
```
davisagli, line 3: @danalvrz This may be called for any key press, so we should only prevent the default action if one of the handlers is used here!

So I would replace line 14 with this:
```
if (handler) {
    event.preventDefault();
    handler();
}
```

davisagli (changes requested):

davisagli (approved):

tiberiuichim (approved):

PR #4141, opened 2022-12-16
Use popperjs in BlockChooser, move the markup to the bottom of the bo…

sneridagh:
…dy tag.

This PR makes `BlockChooser` component to work with `popperjs`. It keeps the visibility of the popped element (as in a tooltip/popup/etc). Also, moved the inserted markup to the bottom of the BODY tag, so it does not interfere with any other element in the flow (so no z-index needed, etc). Turns out that the interactions with elements inside a container-sized elements are not well resolved (or it's on purpose for some reason), and the popped element does not retain z-index between them.

Overall, I think this is how it was supposed to work.

It's arguable that this is a breaking change. It doesn't seem to me so, since no HTML is touched, and the basic functionality is intact.


tiberiuichim (approved):

cypress/support/commands.js
```diff
@@ -762,6 +762,12 @@ function createHtmlPasteEvent(htmlContent) {
   );
 }
 
+Cypress.Commands.add('addNewBlock', (blockName, createNewSlate = false) => {
+  let block;
+  block = cy.getSlate(createNewSlate).type(`/${blockName}{enter}`);
+  return block;
+});
```
davisagli, line 769: I like the addition of this command, but it means most of the tests are now implicitly testing the slash menu instead of the block chooser. Do we need to add an end-to-end test for the block chooser specifically, or is there still one that covers it?

davisagli:
@sneridagh If I have a text block positioned at the top of the window and I click the button to add a block, it scrolls the page up, which feels surprising.

Before click:
Screen Shot 2022-12-19 at 12 52 44 PM

After click:
Screen Shot 2022-12-19 at 12 53 03 PM

davisagli:
@sneridagh And this is what I see if I click the button for a block at the bottom of the page. The menu opens mostly outside of the visible viewport.

Screen Shot 2022-12-19 at 12 56 53 PM

This is in Firefox.

cypress/support/commands.js
sneridagh: Yes, we need to test the blockchooser in addition, although I'd generalise the new command for normal block creation.

davisagli: I agree on that, it's a good idea.

sneridagh:
@davisagli https://github.com/plone/volto/pull/4141/commits/eb76172806e61dc32146b649380349df670b7a6b fixes the "jump".

sneridagh:
Volto Team Meeting decision:

- Better slate it for 17.
- Test it in other blocks-in-block add-ons (EEA, RedTurtle)

PR #4142, opened 2022-12-16
Add ImageWidget with upload/drop/external and inline/widget capabilities

sneridagh:
So it can be used outside the Image Block. This is intended to be used in blocks that upload/selects an image. It can also be used as standalone image widget.

## Inline in the block UI:

image

## As a widget in the sidebar:

image

**Filled**

image

## Usage

Configurable via the `inline` prop, in the block UI would be:
```
        <div>
          {props.isEditMode && (
            <ImageWidget
              {...props}
              inline
              // Since we are using a component that has a widget interface
              // we need to adapt its props to it
              id="url"
              onChange={(id, value) => {
                onChangeBlock(block, {
                  ...data,
                  [id]: value,
                });
              }}
            />
          )}
        </div>
```

or as widget:

```
      an-image: {
        title: 'image',
        widget: 'image',
      },
```


tiberiuichim (approved):

src/components/manage/Widgets/ImageWidget.jsx
```diff
@@ -0,0 +1,324 @@
...
+  const dispatch = useDispatch();
+
+  const request = useSelector((state) => state.content.subrequests[block]);
+  const content = request?.data;
+  const urlUploaded = content ? content['@id'] : null;
+  const requestLoaded = request ? request.loaded : null;
+
+  useEffect(() => {
+    if (loading) {
+      if (loading && requestLoaded && uploading) {
```
pnicolli, line 65: Duplicate `loading` check. Removing the first one should be just fine.

pnicolli (changes requested):

src/components/manage/Widgets/ImageWidget.jsx
sneridagh: done.

pnicolli (approved):

JeffersonBledsoe:
@sneridagh Possible to add a couple of options to configure whether to show the icon and customise the message? This could be used with e.g. video block and custom file blocks

sneridagh:
@JeffersonBledsoe yes, we could do that, also the icon. 

@bipoza I think that https://github.com/plone/volto/pull/4179 should use this instead of moving all the logic in the video Edit component. We can customise it further as @JeffersonBledsoe suggested.

sneridagh:
@dobri1408 I've continued the work in here in:

https://github.com/kitconcept/volto-highlight-block/blob/main/src/components/ImageWidget/ImageWidget.jsx

I polished it quite a bit, please use it as reference if you want to continue working on it.

it's not an easy one since in the widget you don't have access to things that you need for upload then deal with the duality upload/reference because the shape of the object is different just uploaded/catalog. Ping me if you have any question.

/cc @ichim-david

dobri1408:
I have moved the pr here https://github.com/plone/volto/pull/5607, beacuse of the conflicts, and the latest changes.

davisagli:
@dobri1408 Thanks for the updated PR. I'll close this one.

PR #4149, opened 2022-12-19
Change issue_format to link to pull requests instead of issues

stevepiercy:
Change the output template to generate links to pull requests instead of issues, as every change must have a pull request, but not necessarily have an issue. This will also reduce the number of redirects of issues to pull requests.

### DIFF
--- a/news/4149.internal
+++ b/news/4149.internal
@@ -1,0 +1,1 @@
+Change the issue_format to generate links to pull requests instead of issues. @stevepiercy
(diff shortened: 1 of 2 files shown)


mauritsvanrees:
This relies on the majority of existing news snippets to already point to a PR. Can you see if this is the case, and that the link check errors are actually reduced? Otherwise there is not much point I guess.

I think it is still better to use an issue number, if this exists.
Scenario:

- You create issue 1.
- You solve this issue, create PR 2 for master, and create `news/2.bugfix`.
- You backport the PR to Volto 16 and have PR 3.

What do you do then? Do you rename the snippet to `news/3.bugfix`? That is a bit of a hassle. But if you do not do this, the generated changelog for Volto 16 will point to the master PR.

stevepiercy:
I see your point.

I think you've made a good argument for developers to always create an issue before submitting a PR, just as this PR demonstrates. 😉 

The only exception would be for backports.

I can filter out the redirects from linkcheck so it does not fail. Closing.

PR #4151, opened 2022-12-19
Don't crash when dealing with locales that don't have po/json files

tiberiuichim:


src/middleware/api.js
```diff
@@ -203,9 +203,11 @@ const apiMiddlewareFactory = (api) => ({ dispatch, getState }) => (next) => (
             config.settings.supportedLanguages.includes(lang)
           ) {
             const langFileName = normalizeLanguageName(lang);
-            import('~/../locales/' + langFileName + '.json').then((locale) => {
-              dispatch(changeLanguage(lang, locale.default));
-            });
+            import('~/../locales/' + langFileName + '.json')
```
sneridagh, line 206: Still tilde here?

sneridagh (commented):
Shouldn't we return English (which has all the default messages) instead of an empty object?

sneridagh:
@tiberiuichim What do you think about the above comment? I think we should return English (as it has all the default messages in there) if no lang file found.

tiberiuichim:
@sneridagh I haven't had a chance to test it yet.

sneridagh:
@tiberiuichim I think we should not abandon this, let's try to push it forward.

davisagli:
Outdated, this was done in https://github.com/plone/volto/pull/7955

Issue #3868, opened 2022-11-10
Pagination component does not keep state when navigating away from it


erral:
**Describe the bug**

When the user navigates away from a listing that has pagination, the listing block doesn't keep the state on which was rendered, and if the user goes back to the listing block, the pagination state is recalculated and rendered again from the 1st page.

**To Reproduce**

1. Add a Listing or a Search block and configure it in such a way that it has more than 1 page in the results.
2. Browse the page with the block you have just created and go to the 2nd page
3. Click in a result item and go to its page
4. Use the back button in the browser
5. The listing block results are rendered again but they show the 1st page

**Expected behavior**
After going back I expect to be in the same page that I was before I left the page

**Additional context**
There is no querystring like ?page=1, or b_start=20 or something like that that could help on getting the actual page the user was on.

tiberiuichim:
there needs to be something like what SearchBlock is using: https://github.com/plone/volto/blob/master/src/components/manage/Blocks/Search/hocs/withSearch.jsx#L200

This is a complicated problem space: it's easy to overdo it and end up with performance issues, or long/ugly URLs

davisagli:
@tiberiuichim What kind of performance issues are you thinking about? You mean from clients trying to change the batch size to load too many items at once, or search engines following the links to index all of the pages without rate limiting?

erral:
We've been hit by search engines trying all kind of fancy request
parameters in classic Plone UI, so this could be an issue here too

tiberiuichim:
@davisagli I think I used the wrong term when I wrote "performance". I was mainly thinking of how to stash all the listing block states in URLs, considering that there can be multiple listing blocks on a page.

davisagli:
@tiberiuichim I see. Yeah, that's tricky. To keep things simpler, could we decide to only support this for the first listing block on a page?

Or maybe we could stash the pagination state in sessionStorage keyed by block id and not try to sync it with the URL.

tiberiuichim:
We could use the [persistentReducers](https://6.dev-docs.plone.org/volto/configuration/settings-reference.html?highlight=persistent%20reducers#:~:text=A%20list%20of%20reducer%20names%20that%20should%20use%20the%20browser%27s%20localstorage%20to%20persist%20their%20data.) to keep the state synced to localStorage.

I wish, though, that the current implementation of our persistentReducers would use session storage instead of local storage.

erral:
[...]

PR #4159, opened 2022-12-21
Add current page parameter to the route in the listing and search block pagination

bipoza:
Added the current page parameter to the route in the listing and search block pagination.

Fixes: #3868 

## Demo with listing block
https://user-images.githubusercontent.com/26112509/208886147-6c51f379-3be2-4e83-9743-d3d9abb823e4.mov

## Demo with search block
https://user-images.githubusercontent.com/26112509/208886239-67a762a6-8170-49ac-8ae6-4858c0afac11.mov

## Demo combining multiple lists
https://user-images.githubusercontent.com/26112509/208963510-32668c24-76f0-4d72-be96-16766aa2e3c0.mov

## QueryString compatibility:
To be able to combine the searchBlock with the pagination parameters, we had to change the parameters of the searchBlock from hash to search (# -> ?).

We are contemplating the possibility of maintaining both types of routes so as not to generate a breaking change.

In general the compatibility is good except for a small error in the design of the SearchInput. With the old hash url, the clear icon does not render well, although the functionality is correct.

https://user-images.githubusercontent.com/26112509/208971981-2bc180ac-9c3a-4228-988f-cf239fe8081f.mov


ionlizarazu:
The bug you mention exists in the current (16.4.0) Volto version

erral (approved):

sneridagh:
@bipoza @erral wow! thanks for working in this. I know is specially difficult to accomplish. How are we maintaining the sync with the state-URL? I know this is a major pain point. Could we have unit tests for this feature? How about acceptance tests?

What we are trying to accomplish here is quite complex and I'd like to discuss this in the next Volto Team meeting with the team, and specially @tiberiuichim who tried to implement this in the past. 

I'd like to get this well, specially agree on the format and serialization of the URL query parameters, because if we have to change them in the future for whatever reason, it will force a breaking change (people having saved URLs as bookmarks, etc)

sneridagh:
@bipoza the towncrier file extension is wrong, could you please amend it? I think it's ready to merge since long time ago. I have doubts if it's a bugfix or a feature...

erral:
We have seen that we have a bug in the search block with this PR.

We are preparing a fix and some tests to improve the PR.

src/helpers/Utils/usePagination.js
```diff
@@ -24,26 +27,29 @@ const useCreatePageQueryStringKey = (id) => {
  * A pagination helper that tracks the query and resets pagination in case the
  * query changes.
  */
-export const usePagination = (id = null, defaultPage = 1) => {
+export const usePagination = (blockId = null, defaultPage = '1') => {
```
tiberiuichim, line 30: I'd think we should stick to a generic `id` rather then `blockId` and a number for the defaultPage.

ionlizarazu: You mean the "blockId" naming or something else?

ionlizarazu:
We have written some tests for the usePagination.js file

erral:
@sneridagh we think this is ready

PR #4161, opened 2022-12-21
fix: use Grid instead of Table in DiffField

### LINKED ISSUE #4160: Use Grid instead of Table in Diff View
**Describe the bug**

The Diff view for a block based content is too wide and it goes over the screen width, see the attached image.

**To Reproduce**

1. Go to 'News
2. Add a News Item, enter some blocks as text. Save
3. Modify the News Item and change some text
4. Go to the Edit bar -> three dots -> History
5. Click on the three dots of the latest revision and click 'View Changes'

**Expected behavior**
I expect to see the content in the the content area

**Screenshots**
Pantaila-argazkia 2022-12-21 13-36-46

The solution is to modify the `DiffField` component to use `Grid` instead of `Table`.

erral:
Fixes #4160 

Before the fixes:
Pantaila-argazkia 2022-12-21 13-36-46

After the fixes:
Pantaila-argazkia 2022-12-22 08-17-25


sneridagh (commented):
@plone/volto-team 😩 breaking or bugfix?

We really have to write this down and decide the definition of breaking, specially in this edge cases.

/cc @tisto

erral:
Yes... the line between the two is very narrow...

pnicolli:
Imho we should file this as bugfix. There is no way I can imagine that this breaks a site, unless maaaaybe somebody really did go as deep as adding styles to this view, which I believe is very very unlikely.
Am i missing something here?

I still agree this is debatable, of course, and should maybe be discussed and written down as a policy.

tiberiuichim:
It's a bug fix.

image

tiberiuichim (approved):

tiberiuichim:
@sneridagh Let's get this merged, it's painful to have this bug.

PR #4163, opened 2022-12-21
Fix provide a method to disable the searching capabilities of SortOn …

iFlameing:
…component of Search Block


tiberiuichim:
@iFlameing I think we can just set it to false directly, right? I mean, I don't imagine that you'd want a searchable "sort on" control. It's just an oversight, if you ask me.

sneridagh:
@tiberiuichim do you think so? We had this doubt... @iFlameing please set it to false then!

iFlameing:
@tiberiuichim @sneridagh done from my side.

sneridagh (approved):

Issue #4164, opened 2022-12-21
Router Links do not support anchors


jackahl:
**Describe the bug**
When you want use the normal React-router link component to navigate to an anchor id on another page, you will still end up at the top of the other page. This seems to be the default behavior with react router though...

**To Reproduce**
1. Create some Component in Volto, that links to an anchor on another page
2. Click that link
-> >You end up at the top of the other page

**Expected behavior**
You would want to end up scrolled to the anchors position.

**Software (please complete the following information):**
any

**Additional context**

yashnerkar:
Can I work on this?

tiberiuichim:
@yashnerkar there's been recent work on this issue, I think, one of the PRs.

Take issues from the "Polished Volto" Classic Project, those are usually more introductory. 

For example this issue: https://github.com/plone/volto/issues/3424

stevepiercy:
@yashnerkar you can filter issues by label to find things to work on. Filter by [33 needs: docs](https://github.com/plone/volto/labels/33%20needs%3A%20docs) or ["41 lvl: easy"](https://github.com/plone/volto/labels/41%20lvl%3A%20easy).

We especially need help with writing the user manual for Editors.

- https://github.com/plone/volto/issues/3827
- https://6.docs.plone.org/volto/user-manual/blocks.html#facets

PR #4165, opened 2022-12-21
React router hash link

jackahl:
potential fix for #4164


tiberiuichim (approved):

ksuess (approved):
Even scrolling with offset works fine.
Just on pages with an accordion, the scrolling does not touch exactly the offset (default or with scroll function).

+1 from me.

ksuess:
The ScrollToTop still interferes. 
I would like to suggest to modify the default "scroll to top" behavior by adding a condition "only if no hash".
https://github.com/plone/volto/blob/abe52c2881603f6387b0a5c9de19c876ae52d8ea/src/helpers/ScrollToTop/ScrollToTop.jsx#L29-L33

```
  componentDidUpdate(prevProps) {
    if (
      this.props.location?.pathname !== prevProps.location?.pathname &&
      !this.props.location?.hash
    ) {
      window.scrollTo(0, 0);
    }
  }
```

sneridagh:
During the Volto Team meeting we decided to:
- adopt the library into Volto itself
- take advantage of the above and expose as Volto settings some of the library settings
- Fine tuning the whole setup (ScrollToTop)

@jackahl , @ksuess can you coordinate the efforts? Thanks!

jackahl:
> During the Volto Team meeting we decided to:
> 
>     * adopt the library into Volto itself
> 
>     * take advantage of the above and expose as Volto settings some of the library settings
> 
>     * Fine tuning the whole setup (ScrollToTop)
> 
> 
> @jackahl , @ksuess can you coordinate the efforts? Thanks!

sure. Pushed the code for @ksuess proposal. 
@sneridagh what features of the library do you want to expose via the registry? The smoothscroll?

sneridagh:
@jackahl yes, and she mentioned another one I think.

jackahl:
@sneridagh in the libraries documentation (https://github.com/rafgraph/react-router-hash-link#readme) I could not see any other setting, that would make sense to put into the registry, except for the smooth scroll. Did I miss something?

sneridagh:
Let's wait for @ksuess to weigh in. I'll look into the current shape! Thanks!

ksuess:
Originally I thought to export a scrollWithOffset function. But now I think the use case is too rare.

What about providing an example in the documentation:

````
Scrolling with offset:

```js
const scrollWithOffset = (el, offset = null) => {
  let yCoordinate = el.getBoundingClientRect().top + window.pageYOffset;
  const yOffset = offset ? -offset : 0;
  window.scrollTo(0, yCoordinate + yOffset, { behavior: 'smooth' });
};
```

```html
<ConditionalLink
  to={`${el.pageurl}#${el.id}`}
  condition={!isEditMode}
  scroll={(el) => scrollWithOffset(el, 25)}
>
  {el.text}
</ConditionalLink>
```
````

**Comment by jackahl** [comment]
@ksuess I agree with it probably being a rather rare usecase. If we include a registry option for every edgecase we might bloat the registry unnecessarily and make it harder to use. And I just had the same thought about our documentation. If we include too many edgecases there it might become harder to use as well. Especially if its documented somewhere else (the packages documentation) anyways. Maybe a reference to the packages documentation somewhere in our docs might be enough? Though I am not sure how much of an 'edgecase' offset scrolling is. Thats something that more experienced devs than me are probably better in determining.

**Comment by ksuess** [comment]
> @ksuess I agree with it probably being a rather rare usecase. If we include a registry option for every edgecase we might bloat the registry unnecessarily and make it harder to use. And I just had the same thought about our documentation. If we include too many edgecases there it might become harder to use as well. Especially if its documented somewhere else (the packages documentation) anyways. Maybe a reference to the packages documentation somewhere in our docs might be enough? Though I am not sure how much of an 'edgecase' offset scrolling is. Thats something that more experienced devs than me are probably better in determining.

Yes, sure, I'm fine with not including the offset example.
If you integrate the code of the package in Volto, you will mention the package. Who has the need for an offset can check its README and find the link to the tip.

**Review by ksuess — APPROVED** [review]

**Comment by tiberiuichim** [comment]
@sneridagh I'm fine with the merge now.

**Comment by ksuess** [comment]
I'm fine with merging now with the dependency.

**Review by sneridagh — APPROVED** [review]

File: src/helpers/ScrollToTop/ScrollToTop.jsx [hunk 1]
```diff
@@ -27,7 +27,10 @@ class ScrollToTop extends React.Component {
    * @memberof ScrollToTop
    */
   componentDidUpdate(prevProps) {
-    if (this.props.location?.pathname !== prevProps.location?.pathname) {
+    if (
+      !this.props.location?.pathname.hash &&
```
ksuess, line 31: Typo. This should read
!this.props.location?.hash &&

ksuess (commented):
@jackahl, can you please fix the condition on hash.

src/helpers/ScrollToTop/ScrollToTop.jsx
nileshgulia1: I took care of this in https://github.com/plone/volto/pull/4287/files#diff-1712f4ab0dc421cacb610886108bb1e5cd7978212604901774cf341a5fcf9ed1R30

nileshgulia1: Also, we need to account for "copy link to highlight" feature on browsers. Right now it causes problems.

PR #4176, opened 2022-12-23
Show project name and version in control panel

sneridagh:
We find this useful in our projects.

image

### DIFF
--- a/news/4176.feature
+++ b/news/4176.feature
@@ -1,0 +1,1 @@
+Show project name and version in control panel @sneridagh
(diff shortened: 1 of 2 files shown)


pnicolli (approved):
We find it useful as well from time to time. I am inclined to say "yeah go for it" therefore the approval, the only thing picking my brain is the fact that I would really like to start suggesting to build Volto projects only in add-ons instead of putting stuff in the src folder. As I said, this is something to be considered for the future, so for now I would say "yes, merge this"

sneridagh:
@pnicolli Agreed, we went the "all add-ons" strategy since long time ago, and go with it must be the recommended and documented way to go. 

However, the overall version of the project (as the sum of all add-ons, with their versions and glue), is still relevant and it's tied to the project versioning and often to the changelog generation on release.

PR #4178, opened 2022-12-23
Open all accordion'd content in InlineForm by default, allow arbitrarily close any number of them

sneridagh:
**UPDATED 09/03/2023:**

Issue: The state (collapsed/open) of the fieldset accordions in the `InlineForm`/`BlockDataForm` is not being preserved. 
This is a UX problem when the user is editing content and finds out that the state of the settings UI is not the same as he just left before focusing in another block. 

Another "problem" is that the accordions start closed by default. This could be problematic because to my experience some users oversee that the fieldset could be "opened" when they first access the block settings.

I quote "problem" because could be that someone could prefer it to be otherwise. 

This PR is fixing both issues by using [`jotai`](https://jotai.org/) library to save the state for each block. We are adding a new library here, but I think given the power that it provides, it's worth to test it. It only adds 2k (!) to the build.

This is a breaking change, but we have the ability to put it under a configuration setting to control the default behavior, then we could backport it to 16.

I leave the Volto Team meeting the final decision.


sneridagh:
@plone/volto-team okay, could be that some Cypress tests break... I'd be ok if we decide it to mark it as breaking and go for 17.

tiberiuichim (approved):

sneridagh:
Volto Team decission:
- Make the behavior configurable, as opt-in.

PR #4187, opened 2022-12-24
Improve matching in keyboard slash menu

### LINKED ISSUE #4173: Block types menu does not show some blocks when filtering by first letter
**Describe the bug**

The block types menu does not show some blocks, and shows some unexpected blocks, when filtering by the first letter.

**To Reproduce**

1. Go to https://6.demo.plone.org/edit
2. Insert a new block.
3. Type `/t` to bring up the block types menu.
4. Table does not appear in the filtered results, but three other block types that start with "T" appear.

I tested additional filters for blocks that are in the full list and found more surprising results.

- `g`: Grid does not appear.
- `s`: Returns Search and System as expected, but also Table and Text.
- `t`: Returns only Table of Contents, Teaser, and Teaser Grid, but not Tablet and Text.

**Expected behavior**

The appropriate blocks should appear when its first letter is typed.

**Screenshots**

Screen Shot 2022-12-22 at 11 23 48 PM

davisagli:
Fixes #4173 

This adds matching on any substring (not just at the start of the block title & id), but scores the results so that initial matches are still preferred.

### DIFF
--- a/cypress/tests/core/volto-slate/09-block-slate-slashmenu.js
+++ b/cypress/tests/core/volto-slate/09-block-slate-slashmenu.js
@@ -3,16 +3,12 @@ import { slateBeforeEach } from '../../../support/e2e';
 describe('SlashMenu Test: Shortcuts', () => {
   beforeEach(slateBeforeEach);
 
-  it('As editor I can create a Description block using the SlashMenu shortcut', function () {
-    // Use SlashMenu shortcut to create a Description block and type some text in it
-    cy.getSlateEditorAndType('/Desc').type('{enter}');
-    cy.get('.documentDescription').type('This is a description.');
-    cy.get('.documentDescription').type('{enter}');
-
-    // Save
+  it('As editor I can create a Table block using the SlashMenu shortcut', function () {
+    // Use SlashMenu shortcut to create a table block
+    cy.getSlateEditorAndType('/t').type('{enter}');
     cy.toolbarSave();
 
-    // then the page view should contain a Description with the right text in it
-    cy.get('.documentDescription').contains('This is a description.');
+    // then the page view should contain a table
+    cy.get('#page-document table').should('be.visible');
   });
 });
--- a/news/4187.bugfix
+++ b/news/4187.bugfix
@@ -1,0 +1,1 @@
+Improve matching in keyboard slash menu. [davisagli]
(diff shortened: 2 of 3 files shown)


tiberiuichim:
My opinnion on this is that we should avoid matching with blockIds. They're hidden information, and as you know, confusing (why is "s" matching "Text" ?). So, only match titles.

davisagli:
@tiberiuichim I agree. I removed the matching on block id, and also made sure that it filters and sorts using the localized block titles.

cypress/tests/core/volto-slate/09-block-slate-slashmenu.js
```diff
@@ -3,16 +3,12 @@ import { slateBeforeEach } from '../../../support/e2e';
...
-    // Save
+  it('As editor I can create a Table block using the SlashMenu shortcut', function () {
+    // Use SlashMenu shortcut to create a table block
+    cy.getSlateEditorAndType('/t').type('{enter}');
     cy.toolbarSave();
 
-    // then the page view should contain a Description with the right text in it
-    cy.get('.documentDescription').contains('This is a description.');
+    // then the page view should contain a table
+    cy.get('#page-document table').should('be.visible');
```
stevepiercy, line 12: Should this be a new test, instead of modifying an existing test?

stevepiercy (approved):
Thank you!

cypress/tests/core/volto-slate/09-block-slate-slashmenu.js
davisagli: @stevepiercy Cypress tests are slow. I decided it made more sense to modify the existing test to cover more functionality (it's still verifying that the blocks were filtered correctly, but now also verifies that it does a case-insensitive match anywhere in the block title, not only a case-sensitive match at the beginning) than add another very similar test.

If someone is motivated to add unit tests that cover more specific cases, that would be fine, but I won't be able to spend more time on this now.

tiberiuichim (approved):

cypress/tests/core/volto-slate/09-block-slate-slashmenu.js
stevepiercy: That's good thinking.

danalvrz (approved):
That was a great update!

PR #4189, opened 2022-12-27
Fixed maxLength validation typo

pnicolli:
I am unsure if this needs to be a breaking change or just a bugfix. Theoretically, if this `maxLength` feature was working at all without this fix, this should be a breaking change even if it is just a typo fix, but I honestly doubt that it was working since the dx controlpanel sends `maxLength` as a key for this validation option, while the validation function was checking for `maxLengthj`. @avoinea was it working for you? You added this kind of validation so I assume you have tried to use it somewhere.


src/helpers/FormValidation/FormValidation.js
```diff
@@ -4,37 +4,39 @@ import { messages } from '../MessageLabels/MessageLabels';
...
  * @param {string | number} value can compare '47' < 50
  * @param {string | number} valueToCompare can compare '47' < 50
  * @param {string} minCriterion
  * @param {Function} intlFunc
  */
-const isMinPropertyValid = (value, valueToCompare, maxCriterion, intlFunc) => {
+const isMinPropertyValid = (value, valueToCompare, minCriterion, intlFunc) => {
   const isValid = valueToCompare !== undefined ? value >= valueToCompare : true;
-  return validationMessage(isValid, maxCriterion, valueToCompare, intlFunc);
+  return validationMessage(isValid, minCriterion, valueToCompare, intlFunc);
```
pnicolli, line 39: All these changes are just code style and consistency, for easier readability, they are not really needed for the PR.

sneridagh (approved):
I'd say we go with a bugfix, but let's see what @avoinea has to say about it.

avoinea (approved):

PR #4190, opened 2022-12-27
bugfix: wrong conditional proprieties on ObjectBrowser

Wagner3UB:
**Describe the bug**
When we are selecting items to be correlated with has a limit set, using the maximumSelectionSize variable, this limit is not taken and can be overcome. Example, if the variable was set to be 10, when 10 items are selected the sidebar behavior works and closes the sidebar, but it's possible to re-open the select field / sidebar and keep selecting items one by one updating the final number, exceeding the maximum number (12 of 10).

**To Reproduce**
Steps to reproduce the behavior: (see the video below)

1. on an select box with uses objectBrowser, selecting multiples items with uses the maximunSelectionSize variable ( !== null) . 
2. Select all possible items, in my case, 10.
3. the sidebar will close
4. re-open the sidebar clicking on the select box once again
5. pic another item. The sidebar will close again, but the behavior continues and one more item will be added, showing 11 of 10.

**Expected behavior**
The selector must stop when the limit was reached. 10 of 10 max.  

**Screenshots**

https://user-images.githubusercontent.com/60133113/209677612-7ab5c7f5-9d3e-4f91-a571-e60b5ef4a69b.mov

- OS: iOS
- Browser chrome
- Volto Version 16.5.0
- Plone Version [e.g. 5.2.2]
- Plone REST API Version [e.g. 7.0.1]


sneridagh (commented):
Given the amount of tests broken, I'd say there was a reason for that condition to be like this, and the fix should happen in another point.

deodorhunter:
@pnicolli  @sneridagh tried a different approach to tackle this one. All tests are green now, we can discuss if this has the expected behaviour.

news/4190.bugfix
```diff
@@ -0,0 +1 @@
+"Fixed wrong conditional proprieties on ObjectBrowser for multiple selection - @deodorhunter @Wagner3UB"
```
stevepiercy, line 1:
```suggestion
Fixed wrong conditional proprieties on `ObjectBrowser` for multiple selection. @deodorhunter @Wagner3UB
```

stevepiercy (changes requested):
Fix formatting of news item.

pnicolli (approved):

sneridagh (approved):

PR #4197, opened 2022-12-29
change hotkey for strikethrough in Slate editor

### LINKED ISSUE #4196: In Slate editor "ctrl+s" hotkey causes conflicts with it's default behaviour
**Describe the bug**
In Volto Slate editor `MOD+S` (`ctrl+s` for me, a windows user)  is used for adding a Strikethrough style but when focus is not on block, pressing `ctrl+s` opens page save dialog (as this is its default behavior in chrome ). which is quite common behavior.

Also, we may use `ctrl+s` for saving content. See [#4168](https://github.com/plone/volto/issues/4168). Feedback Welcome!

**To Reproduce**

1. In Edit mode choose plain text Block
2. Write something and click outside the block.
3. Now press ctrl+s

.
- OS: [Windows]
- Browser [ chrome, ]
- Volto Version [16]
- Plone Version [6]
- Plone REST API Version [e.g. 7.0.1]

MAX-786:
fix #4196 
Change hotkey from `ctrl+s` to `ctrl+d`

### DIFF
--- a/packages/volto-slate/src/editor/config.jsx
+++ b/packages/volto-slate/src/editor/config.jsx
@@ -207,7 +207,7 @@ export const hotkeys = {
   'mod+b': { format: 'strong', type: 'inline' },
   'mod+i': { format: 'em', type: 'inline' },
   'mod+u': { format: 'u', type: 'inline' },
-  'mod+s': { format: 'del', type: 'inline' },
+  'mod+d': { format: 'del', type: 'inline' },
   // 'mod+`': { format: 'code', type: 'inline' },
   // TODO: more hotkeys, including from plugins!
 };
--- a/news/4196.bugfix
+++ b/news/4196.bugfix
@@ -1,0 +1,1 @@
+fix Slate editor "ctrl+s" hotkey causes conflicts with it's default behaviour. @MAX-786


tiberiuichim (approved):

davisagli:
Why ctrl+d? Is there another editor that uses this key combination for strikethrough? It would probably make sense to use something that might be familiar to someone. Google Docs uses cmd+shift+x (I'm on a Mac, probably ctrl+shift+x elsewhere)

MAX-786:
i started a discussion on community to know most common key combo for strikethrough, ctrl+d looked good to me (ms word user) .Thanks for suggestion!
One thing is that adding 'shift' for one command.... wouldn't it be confusing tho?  Thanks!

JeffersonBledsoe:
Word online uses `CMD+SHIFT+X` too (And I'd guess that's ctrl on non-mac). CMD+D is the browser bookmark shortcut normally, might help to not conflict with this too?

packages/volto-slate/src/editor/config.jsx
```diff
@@ -207,7 +207,7 @@ export const hotkeys = {
   'mod+b': { format: 'strong', type: 'inline' },
   'mod+i': { format: 'em', type: 'inline' },
   'mod+u': { format: 'u', type: 'inline' },
-  'mod+s': { format: 'del', type: 'inline' },
+  'mod+d': { format: 'del', type: 'inline' },
```
nileshgulia1, line 210: I would also not recommend using just ctrl+d as it's a bookmark shortcut and interferes with the browser. Gutenberg uses opt+ctrl+d in osx and shfit+alt+d in windows. Maybe you could take inspiration from that or else.

MAX-786: Thanks both of you for your insights :). I forgot about bookmark shortcut.
As JeffersonBledsoe and others suggested, `mod+shift+x` would be best then.

MAX-786:
Done!

nileshgulia1 (approved):

Issue #4195, opened 2022-12-29
Docs: Remove buildout from install backend


stevepiercy:
In [Install a Plone backend locally without Docker](https://6.docs.plone.org/volto/configuration/backend.html#install-a-plone-backend-locally-without-docker) the Volto docs still recommend to use buildout. The new method is now in https://6.docs.plone.org/install/install-from-packages.html.

I would suggest that we remove buildout references and instead direct users to the new way.

@tiberiuichim @sneridagh opinions?

tiberiuichim:
Of course

stevepiercy:
On second thought, I think the entire page should be removed, and the Configuration section should have a new introduction to direct developers to install Plone via containers or packages using the new methods.

But this ties back into several open issues about reorganizing the Volto install docs within the Plone 6 docs. https://github.com/plone/volto/issues/3998#issuecomment-1328914246

Because there are so many different parts to install Volto, I created a new mini-project, [Install Docs](https://github.com/orgs/plone/projects/16/views/1).

Once we take inventory of all the open issues related to Install Docs, then we can come up with a plan for how to reorganize things.

SaiRev0:
Is the issue still open, I would like to work on this issue

stevepiercy:
@SaiRev0 no. See comment in https://github.com/plone/volto/pull/4212#issuecomment-1369706496.

If you want an easy issue to work on, please search for the label ["41 lvl: easy"](https://github.com/plone/volto/labels/41%20lvl%3A%20easy). Thank you!

SaiRev0:
@stevepiercy Thanks for the info

davisagli:
Fixed in https://github.com/plone/volto/pull/5287

stevepiercy:
@davisagli partially fixed, I think, unless I misunderstand. We need to remove the section, correct?

davisagli:
@stevepiercy The info I left on the page is still relevant.

stevepiercy:
@davisagli I mean, if you say it is fixed, then it should be closed, right? What else needs to be done for this issue?

davisagli:
I think it is done, I just missed closing it.

stevepiercy:
Ah, yeah, and needing to update submodule tips, plus cache, it took a while to get removed. We're good!

PR #4212, opened 2023-01-03
 #4195 issue solved 

KartikBudania:
Doc is updated, let me know if we need some changes


stevepiercy:
@KartikBudania read all the comments in #4195. We don't even have a plan for how to proceed yet.

Aside from that, before any contribution can be accepted in the Plone organization for documentation, start by reading and following: https://6.docs.plone.org/contributing/index.html

KartikBudania:
@stevepiercy thanks, for letting me know about the process of contribution.
My bad for #4195, I will take care of it from now onwards.

PR #4213, opened 2023-01-03
Search block: Allow site viewers to select 'Relevance' as a sort option

### LINKED ISSUE #4000: Default 'Relevance' sort order missing from search block's `Sort on options`
**Describe the bug**
If a `sort_on` key is not specified, Plone will order a querystring search by 'relevance', which is often the desired behaviour. A search block enables site viewer sorting options and will default to the `relevance` sorting if one is not selected. Not only is it not possible to return to the `relevance` sorting once an option has been set, it isn't clear that Plone is even performing a `relevance` sort initially as it isn't mentioned anywhere. It would nice for site editors to be able to add a `Relevance` option to the search block's sort on options.

**To Reproduce**

1. Create a search block
2. Enable the `Show sorting` option in the `Controls` fieldset
3. Try to add an option to allow sorting by `relevance`

**Expected behavior**
An option for `relevance` is shown in the dropdown. This option would simply remove the `sort_on` key from the querystring, allowing Plone to return to it's default sort order.

JeffersonBledsoe:
Changes the 'No selection' option in the 'Sort on' for the search block to instead be a selectable 'Relevance' option. If no text has been searched for, this is displayed as 'Unsorted'. Also includes the following minor changes to make the 'Relevance' option make more sense:

- The 'Sort direction' buttons will be hidden if we're using 'Relevance/ unsorted'
- The sorter won't be displayed if no options are set in 'Sort on options'

https://user-images.githubusercontent.com/30210785/210405052-f7569308-2ab7-48b7-9b33-224b864eaf5c.mov

Closes #4000

### DIFF
--- a/src/components/manage/Blocks/Search/components/SortOn.jsx
+++ b/src/components/manage/Blocks/Search/components/SortOn.jsx
@@ -62,18 +62,17 @@ const SortOn = (props) => {
 
   const activeSortOn = sortOn || data?.query?.sort_on || '';
 
-  const { sortOnOptions = [] } = data;
+  const noValueLabel = searchedText
+    ? intl.formatMessage(messages.relevance)
+    : intl.formatMessage(messages.unsorted);
+
   const value = {
     value: activeSortOn,
     label:
       activeSortOn && sortable_indexes
         ? sortable_indexes[activeSortOn]?.title
-        : activeSortOn || searchedText
-        ? intl.formatMessage(messages.relevance)
-        : intl.formatMessage(messages.unsorted),
+        : noValueLabel,
   };
-
-  debugger;
 
   return (
     <div className="search-sort-wrapper">
@@ -91,10 +90,10 @@ const SortOn = (props) => {
           theme={selectTheme}
           components={{ DropdownIndicator, Option }}
           options={[
-            { value: '', label: 'Relevance' },
+            { value: '', label: noValueLabel },
             ...sortOnOptions.map((k) => ({
               value: k,
-              label: sortable_indexes[k]?.title || k,
+              label: k === '' ? noValueLabel : sortable_indexes[k]?.title || k,
             })),
           ]}
           value={value}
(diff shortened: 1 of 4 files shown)


sneridagh (changes requested):
@JeffersonBledsoe LGTM, but instead of showing/hide the sort buttons I'd disable them (the Button component allows it using `disabled`), to avoid the UI "jump".

What do you think?

/cc @tiberiuichim

tiberiuichim:
@sneridagh Your proposal sounds good to me.

JeffersonBledsoe:
@sneridagh Updated to have the buttons disabled rather than hidden :)

avoinea (approved):

sneridagh (commented):
Taking a closer look now, and since we are updating the Left/Right/Top components and passing down the new prop. Wouldn't it be breaking? As it requires this update to existing shadows, if the SortOn component is changed and expects it.

@JeffersonBledsoe what happens if the prop are not passed down?

JeffersonBledsoe:
@sneridagh It should still safely fall back to the existing behaviour. I will do some more testing to verify this behaviour though!
But yes, to get this new behaviour it would require an update to shadows unfortunately.

pnicolli:
@JeffersonBledsoe any news on this?

pnicolli:
Since we are now working on Volto 19 this could definitely be something that we do even if it's breaking.
@JeffersonBledsoe is this still relevant? Should we move this forward? If this is the case, let's try and recollect ideas on what's missing.

JeffersonBledsoe:
@pnicolli Thanks for the ping. I'm hoping to look at this this week actually! There's a slight disconnect between this PR and the discussion in https://github.com/plone/volto/issues/4000, I'll ping the Volto team once I get this back up to date and tidy up a few things with it.

djay:
I've updated https://github.com/plone/volto/issues/4000 with what we think the correct UX should be for the behaviour of sort_on dropdown and relevance search is being auto-switched to.
and in the next week or so we will revise this PR to implement this. Any comments on the proposed UX please add to https://github.com/plone/volto/issues/4000

Issue #4131, opened 2022-12-15
Right side "sidebar handle" button is not noticeable 


MAX-786:
**Describe the bug**
The right side `sidebar handle`, people had issues and didn't know how to open and close it. What's worse (for new Volto editors), once you close the sidebar, it stays closed and the whole thing is lacking contrast and you won't know how to open it. It's a major UX problem.

**To Reproduce**

1. Open Volto Editor, and click on any block.
2. On the right side, the `sidebar handle` button is not visible clearly.
3. If you click on it and close the sidebar, it will remain closed.

**Expected behavior**
The sidebar handle button should be more opaque so it can be easily noticeable. And also it should open again when you click on any block.

**Screenshots**

https://user-images.githubusercontent.com/99530996/207945357-aff82593-48cd-47de-9087-52e2d92de1b0.mp4

**Software (please complete the following information):**

- Volto Version [e.g. 8.5.0]
- Plone Version [e.g. 5.2.2]
- Plone REST API Version [e.g. 7.0.1]

JeffersonBledsoe:
Related issue #3910 with a WIP PR to resolve it #3912 . More than happy for somebody to finish the PR off if they wanted :)

prithvi009:
hey! @MAX-786 can I work on this issue?

stevepiercy:
@prithvi009 in the Plone organization, you may pick up any open issue or PR, assign yourself to it (Assignees in the upper right corner of GitHub), and work on it. There is no need to ask, unless someone else is actively working on it.

See https://6.docs.plone.org/volto/developer-guidelines/index.html for details.

And welcome to Plone! We're a friendly bunch.

MAX-786:
@prithvi009 Sure! as Steve already said, but u may have noticed this that we already have this related issue [#3910](https://github.com/plone/volto/issues/3910) and WIP PR [#3912](https://github.com/plone/volto/pull/3912). Feel free to look into it and work on it if u like :) .

PR #4214, opened 2023-01-04
fixed bug #4131

prithvi009:
Right Side "sidebar handle" made visible


sneridagh:
@prithvi009 could you please open another PR using a local branch in the Volto repo instead of using a fork? Then the checks will run correctly. Also, link the issue to the PR. Thanks!

PR #4223, opened 2023-01-05
(Fix) Select Widgets scrolls the page when the options are not visible

dobri1408:
BUG: 
You can see that when you press a dropdown, and the options content doesn't fit in the view, the page is getting scrolled, but when the content is visible, the page is not getting scrolled.
This bug occurs for all the dropdown fields.
Be aware that this bug occurs only when the page is big enough so you can scroll it. 
Example: 

https://user-images.githubusercontent.com/50819975/210770088-33e9f746-d754-4f95-8e60-35bac3d71442.mp4

FIX:

The problem is not coming from the volto itself but from a library that volto uses, named react-select

Other reports complain about this problem; see here: 
https://stackoverflow.com/questions/61999859/react-select-causing-page-to-scroll-down
So the fix is to force the react-select to not scroll into view by using prop: menuShouldScrollIntoView={false}


avoinea (approved):

avoinea:
@sneridagh @tiberiuichim Please add your thoughts here.

tiberiuichim:
@avoinea @dobri1408 Does it fix https://github.com/plone/volto/issues/4230 ?

dobri1408:
@tiberiuichim 

Yes, I tested it, and it does fix it.

dobri1408:
The bug is visible on demo.plone.org, but you must check that the page is long enough to scroll it.

tiberiuichim:
I see the problem is relevant only in the sidebar, so the fix is not generic

tiberiuichim (approved):

sneridagh (approved):

Issue #4103, opened 2022-12-12
Schemas should be injected to the EditBlock component


tiberiuichim:
Right now, to have proper "default values" support in a block, you should pass have the schema in the block config, because it is computed in `applyBlockDefaults` in the `Blocks/Block/Edit.jsx` component that [bootstraps a block's edit component](https://github.com/plone/volto/blob/c96c911b7b05e3a0c2ea33f2de7d987a3d41e742/src/components/manage/Blocks/Block/Edit.jsx#L173).

`applyBlockDefaults` reads the [block's config and looks for the schema](https://github.com/plone/volto/blob/c96c911b7b05e3a0c2ea33f2de7d987a3d41e742/src/helpers/Blocks/Blocks.js#L431-L433).

In the block edit component we have to create the schema again:

https://github.com/plone/volto/blob/24c04c7fde707f2dbb32fa8722e8f725874a81b0/src/components/manage/Blocks/ToC/Edit.jsx#L11

This is creates the opportunity for confusion. The applyBlockDefaults will magically create a schema, it's being applied even before the real TocBlockEdit component is created, but then the TocBlockEdit needs to recreate the schema on its own. I know the history of why it's like this, we did baby steps in the direction of where we're now. (I use the TOC block as an example, this isssue is not about TOC).

My proposal: create the schema in the EditBlockWrapper, pass that schema to applyBlockDefaults, pass it to the block edit components.

PR #4231, opened 2023-01-05
Read listing block schema from configuration registry

pnicolli:
This is to make the schema easier to customize. It becomes as easy as overriding the `blockSchema` key of the listing block with something imported from a file like this:

```javascript
import ListingBlockSchema from '@plone/volto/components/manage/Blocks/Listing/schema';

export const schemaListing = (props) => {
  const baseSchema = ListingBlockSchema(props);

  // customize schema

  return baseSchema;
}
```

I'm not sure if `schemaEnhancer` is an option for customization here, it is listed in the docs as the go-to way for customizing schema for variations, but not for the entire block.


tiberiuichim:
100% this.

https://github.com/plone/volto/blob/a7e3e08a9d8fa4a829cad04c13566f117004ce30/src/helpers/Blocks/Blocks.js#L431-L432

We should set `blockSchema` in all components, it's setting an example and the only way form defaults are properly computed. What's more, my impression is that the schema should be injected into the block components (and fully under the control of the schemaEnhancers), rather then instantiated from the block edit components. See https://github.com/plone/volto/issues/4103

There's another use case that I hit recently: the data that we present to the block view should be based on the schema. Let's assume that I have a variation where I add a new field. I use the variation, set a value for the field, but then change the variation. The field widget is no longer presented, but the value is recorded in the block data, which leads to bugs.

tiberiuichim:
@pnicolli It's possible to set a schemaEnhancer for the block, also for the variations. The block's schemaEnhancer is applied last.

So its: `blockSchema -> variation schemaEnhancer -> block schemaEnhancer`

sneridagh:
+1 we need to make all the blocks (stock and add-ons) to use this approach. 

I'd say it's not breaking... what about you?

pnicolli:
Yeah I don't think this is breaking anything.

pnicolli:
I'm not sure I understood if something is missing here in your opinion, or if this is just a little forgotten, I forgot it myself in the last week. @sneridagh @tiberiuichim

tiberiuichim (approved):

tiberiuichim:
@pnicolli My views on this PR are like this:

In the ideal scenario the listing block schema should not need to be computed/acquired by any listing block components, it should be automatically injected as a prop from the Volto's Edit block machinery. 

We could go with the changes here, but with the understanding that it's a temporary fix.

sneridagh (approved):

Issue #4234, opened 2023-01-07
Multiple listing blocks: Rendering fails for copy-pasted listing blocks


ksuess:
**To Reproduce**
- Create listing block.
- copy paste listing block and modify criteria
- both blocks are rendering the same despite different criteria

**Software (please complete the following information):**
Current Volto

ksuess:
Problem is, that for the rendering, the block id included in content.blocks[blockid] is used. This is the same for original and copy. :-(

ksuess:
Question here: Why is the block id included in content.blocks[blockid] at all? It's redundant and raises an error when duplicating listing blocks.

ksuess:
I would like to modify the copy change action instead of eliminating the block id. Eliminating the redundant block id would mean a breaking change.

Example for page with duplicated and modified listing blocks:

```
{
  "content": {
    "@components": "{actions: {…}, aliases: {…}, breadcrumbs: {…}, cata…}",
    "@id": "http://igib.example.com/api/dokumentation/manual-navigations-seiten/uebersicht-komponente-lena",
    "@type": "Module Navigation Page",
    "UID": "7a28fc0e7c3147669a14413ad8f44488",
    "allow_discussion": false,
    "blocks": {
      "3343eead-7d86-4eb7-9393-0ab4a7d1aaa6": {
        "@type": "listing",
        "block": "8d726b48-f73b-4fcf-b851-7590bdc3e35e",
        "headline": "Anleitungen",
        "headlineTag": "h2",
        "query": "[]",
        "querystring": "{query: Array(2), sort_order: \"ascending\"}",
        "variation": "default"
      },
      "8d726b48-f73b-4fcf-b851-7590bdc3e35e": {
        "@type": "listing",
        "block": "8d726b48-f73b-4fcf-b851-7590bdc3e35e",
        "headline": "Anleitungen",
        "headlineTag": "h2",
        "query": "[]",
        "querystring": "{query: Array(2), sort_order: \"ascending\"}",
        "variation": "default"
      },
      "eaef6ccb-2fa5-4d7c-aeae-555cae09a122": {
        "@type": "title"
      }
    },
    "blocks_layout": "{items: Array(3)}",
    "contributors": "[]",
    "created": "2023-01-09T07:49:42+00:00",
    "creators": "[\"admin\"]",
    "description": "",
    "effective": null,
    "exclude_from_nav": false,
    "expires": null,
    "id": "uebersicht-komponente-lena",
    "is_folderish": false,
[...]

PR #4239, opened 2023-01-09
Synchronize redundant block id in listing block on pasting: blocks[bl…

ksuess:
…ockid].block
fix https://github.com/plone/volto/issues/4234

### DIFF
--- a/news/4234.bugfix
+++ b/news/4234.bugfix
@@ -1,0 +1,1 @@
+Synchronize redundant block id in listing block on pasting: blocks[blockid].block @ksuess
--- a/src/components/manage/Form/BlocksToolbar.jsx
+++ b/src/components/manage/Form/BlocksToolbar.jsx
@@ -100,7 +100,15 @@ export class BlocksToolbarComponent extends React.Component {
             : [uuid(), blockData]
           : [uuid(), blockData]; // if cut/pasting blocks, we don't clone
       })
-      .filter((info) => !!info); // some blocks may refuse to be copied
+      .filter((info) => !!info) // some blocks may refuse to be copied
+      .map(([id, data]) => {
+        // sync redundant block id in listing block: blocks[blockid].block
+        if (data.block) {
+          let new_data = Object.assign({}, data);
+          new_data.block = id;
+          return [id, new_data];
+        } else return [id, data];
+      });
     const blocksFieldname = getBlocksFieldname(formData);
     const blocksLayoutFieldname = getBlocksLayoutFieldname(formData);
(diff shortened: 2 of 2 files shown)


ksuess:
Maybe this bug fix is a breaking one: If somebody relies in `getAsyncData` on block uid in data, they have to rewrite their function. See https://github.com/plone/volto/blob/05f09ce261f78dfddd1bb4bcd3ed1c965b004337/docs/source/blocks/ssr.md

tiberiuichim (approved):
The signature of `getQueryStringResults` hasn't changed. The documentation change is only a "fantasy" example, it would not represent a breaking change, IMHO.

image

sneridagh:
@tiberiuichim @ksuess we are not persisting `block` anymore in the block data... I'd say it's breaking :/ I know it shouldn't be there, but still... maybe someone is relying on it (eg. shadows).

Let's discuss it in the next meeting.

sneridagh:
Volto Team Meeting:
We will try to make the PR not breaking. We could also have a cleaning (breaking) PR for 17.

sneridagh:
@ksuess I'd merge this right away, as a breaking change. Then afterwards work it out if it make sense for 16 too. Are you ok?

sneridagh:
@ksuess Would you change the type of change? and possibly write something brief in the upgrade guide? Thanks!

ksuess:
I thought about it back and forth. I do not see anymore a breaking change.

What is fixed here, is that the listing block does not save the block id in blocks data anymore.
If somebody still does this, no problem occures, as this fix is just taking the id from blocks props, which is always available.
Do you agree?

src/components/manage/Blocks/Search/SearchBlockView.jsx
```diff
@@ -89,6 +89,7 @@ const SearchBlockView = (props) => {
         setSelectedView={setSelectedView}
       >
         <ListingBody
+          id={id}
```
davisagli, line 92: @ksuess For the most part I agree with your analysis that the change is not breaking, but I think there is a possibility for it to break here if a project has customized the SearchBlockView. Then the ListingBody would not receive the `id` unless the customization is updated.

This could be handled in withQuerystringResults by making it use the `id` if present, otherwise fall back to the old approach (`data.block`).

ksuess: Updated.
https://github.com/plone/volto/pull/4239/commits/fec8fea12cbec0a65b99d90e23a662c79b0ee2ef

PR #4282, opened 2023-01-18
Add displayName when registering a component in the component registry

sneridagh:
This fixes this pattern in React dev tools was showing `Anonymous` as the name of the component.

```
  const Image = config.getComponent('Image').component || DefaultImage;
```

Then in JSX

```
<Image .../>
```

### DIFF
--- a/news/4282.feature
+++ b/news/4282.feature
@@ -1,0 +1,1 @@
+Add displayName when registering a component @sneridagh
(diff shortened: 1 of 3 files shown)


tiberiuichim (commented):
Some components may already have displayName set.

sneridagh:
@tiberiuichim mmmm true.

tiberiuichim (approved):

PR #4284, opened 2023-01-19
Allow passing a custom block to focus to TextLineEdit

JeffersonBledsoe:
The TextLineEdit block makes the assumption that the block whose data it is editing is the block that should be focused when it is clicked. While this is a valid assumption most of the time, it does not work in scenarios where there may not be a block on the page to select, or where the block is a child block of another block and the child block may not want to be selectable

### DIFF
--- a/src/components/manage/TextLineEdit/TextLineEdit.jsx
+++ b/src/components/manage/TextLineEdit/TextLineEdit.jsx
@@ -158,7 +158,7 @@ export const TextLineEdit = (props) => {
     if (blockToFocus) {
       onSelectBlock(blockToFocus);
     }
-  }, [block, onSelectBlock]);
+  }, [block, blockToFocus, onSelectBlock]);
 
   const RenderTag = renderTag || 'h1';
 


JeffersonBledsoe:
Good to know! It would be nice to make building WYSIWYG blocks easier and those changes look like a nice step in that direction :) @sneridagh Happy for this change?

tiberiuichim:
@JeffersonBledsoe how do you inject the `blockToFocus`? I think, in light of the future changes, we may want to make this feature more generic, by passing an onFocus to the TextLineEdit, so you could write a wrapper for it (or just integrate that directly.

```
const CustomFocusLineEdit = ({blockToFocus, ...props}) => <TextLineEdit {...props} onFocus={() => props.onSelectBlock(blockToFocus)} />

...

<CustomFocusLineEdit {...props} blockToFocus={X} />
```

sneridagh:
@JeffersonBledsoe , @tiberiuichim showed me yesterday a PoC on the field block aware component. I think he's working on it for the new Hero block.

PR #4285, opened 2023-01-19
Read from api extenders in reducers

sneridagh:


src/config/index.js
```diff
@@ -75,7 +75,13 @@ let config = {
...
     publicURL,
     apiPath,
-    apiExpanders: [],
+    apiExpanders: [
+      // Uncomment or add the following expanders for only issuing a single request
+      // {
+      //   match: '',
+      //   GET_CONTENT: ['breadcrumbs', 'navigation', 'actions', 'types'],
+      // },
+    ],
```
pnicolli, line 84: If I understand this correctly, this new feature is opt-in; if these lines are left commented it will be like it's now, meaning the expanders are not used, is it correct?
If it's correct, would it be a breaking change to have this new feature on by default? At first glance, this new feature only seems beneficial, am I missing something?

tiberiuichim: Only benefits, no drawbacks, as far as I can see, nothing breaks.

The only breaking change would be if you have already declared your own expanders, now you have to combine the lists.

sneridagh: @pnicolli I only left it as documentation, we can move it to a proper section in the docs.

This was long due, in fact, if you wanted to accomplish this, you had to shadow your own reducers/components and make the changes included in this PR. If you don't do anything, all will continue working (even if you've defined custom shadows taking care of this. As a matter of a fact the existing shadows related will now render redundant, and you may want to remove them. Also, review what you include in `apiExpanders` will be also a good idea, after this is merged, but is not a breaking change.

tiberiuichim (approved):

src/config/index.js
pnicolli: @sneridagh  Yeah, this is actually awesome :)
My final goal with the comment was: if it's not a breaking change, and I believe it is not, imho we should enable this by default instead of only suggesting it with a comment and documentation. I wouldn't want developers to have to remember to enable this feature themselves, if it has no downsides.

sneridagh: @pnicolli doing that is a breaking, I'd say :) My proposal: Let's leave it opt-in for now, then in 17 as opt-out. What do you think?

sneridagh: (It is breaking because people that already does it, or have any customisation, would have to adapt/merge configs).

pnicolli: > Let's leave it opt-in for now, then in 17 as opt-out. What do you think?

Ok makes sense then, let's go with this approach

pnicolli (approved):

PR #4299, opened 2023-01-23
[WIP] Undo block paste will now reload clipboard to unpasted stuff

### LINKED ISSUE #4063: Undoing a block paste operation should reload the clipboard with the un-pasted stuff
It's not uncommon to paste stuff in the wrong place. The first inclination, and most expedient thing to do, is to undo the paste and then paste again in the right place. Currently the clipboard is not reloaded when you undo a paste. It probably should be.

MAX-786:
fixes #4063 

## Approach I used

1. In `Form.jsx` I used blocksClipboard global state from redux store and passed it to `<UndoToolbar />` component's "state" property.
2. Inside `useUndoManager.js` I checked if state.blocksClipboard is empty or not when user click undo button (i.e. inside doUndo) and if clipboard is not empty then 
3. dispatches an action `setBlocksClipboard` using store which is injected from `start-client.js` to the `useUndoManager.js`.

### DIFF
--- a/src/helpers/UndoManager/useUndoManager.js
+++ b/src/helpers/UndoManager/useUndoManager.js
@@ -1,8 +1,15 @@
 import React from 'react';
 import Undoo from 'undoo';
 
+import { setBlocksClipboard } from '@plone/volto/actions';
+
 // Code based on Apache-2.0 License
 // https://github.com/reaviz/reaflow/blob/78d60aa04f514947a17097c906efdbbd6bae5080/src/helpers/useUndo.ts
+
+let store;
+export const injectStore = (_store) => {
+  store = _store
+}
 
 const useUndoManager = (
   state,
@@ -25,6 +32,15 @@ const useUndoManager = (
       const nextRedo = manager.current.canRedo();
       setCanUndo(nextUndo);
       setCanRedo(nextRedo);
+      
+      console.log(state);
+      if (Object.keys(state.blocksClipboard).length !== 0) {
+        console.log("Undo Clipboard!");
+        const actionType = Object.keys(state.blocksClipboard)[0];
+
+        const blocksData = state.blocksClipboard?.copy ? state.blocksClipboard.copy : state.blocksClipboard.cut;
+        store.dispatch(setBlocksClipboard({ [actionType]: blocksData }));
+      }
 
       callbackRef.current({
         state,
--- a/src/components/manage/Form/Form.jsx
+++ b/src/components/manage/Form/Form.jsx
@@ -565,6 +565,7 @@ class Form extends Component {
               formData: this.state.formData,
               selected: this.state.selected,
               multiSelected: this.state.multiSelected,
+              blocksClipboard: this.props.blocksClipboard,
             }}
             enableHotKeys
             onUndoRedo={({ state }) => this.setState(state)}
@@ -790,6 +791,10 @@ class Form extends Component {
 
 const FormIntl = injectIntl(Form, { forwardRef: true });
 
+const mapBlocksClipboardToProps = state => ({
+  blocksClipboard: state.blocksClipboard,
+})
+
 export default compose(
-  connect(null, { setSidebarTab }, null, { forwardRef: true }),
+  connect(mapBlocksClipboardToProps, { setSidebarTab }, null, { forwardRef: true }),
 )(FormIntl);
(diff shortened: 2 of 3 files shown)


MAX-786:
@tiberiuichim , can you please check the approach I used? Thanks! :)

I have checked that it may not break other code and not introduce any new bugs but need others feedback.

tiberiuichim:
@MAX-786 

Looks like a good start!

There's no need for the store middleware, get the dispatch with `useDispatch()`. Check volto's src code, there's many examples with that.

MAX-786:
> There's no need for the store middleware, get the dispatch with useDispatch(). Check volto's src code, there's many examples with that.

Thanks for the insight! can you please review it if possible?

MAX-786:
Done! all green :)

src/helpers/UndoManager/useUndoManager.js
```diff
@@ -46,6 +46,7 @@ const useUndoManager = (
         canRedo: nextRedo,
       });
     });
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
tiberiuichim, line 49: If possible, solve this so that the disable is not required

MAX-786: Sure!... adding dispatch to the dependency list will do ig.

src/helpers/UndoManager/useUndoManager.js
```diff
@@ -26,14 +30,23 @@ const useUndoManager = (
       setCanUndo(nextUndo);
       setCanRedo(nextRedo);
 
+      if (Object.keys(state.blocksClipboard).length !== 0) {
+        const actionType = Object.keys(state.blocksClipboard)[0];
+
+        const blocksData = state.blocksClipboard?.copy
```
tiberiuichim, line 36: hmm... I can think of a scenario where the clipboard would be poluted. One thing to know about this "blocks clipboard", it uses localStorage, so it can be used from multiple browser tabs.

Suppose this scenario:

- I select and copy & paste some blocks
- I go on another page on the same site, I make some modifications. 
- If I undo the modifications, is there a chance that the clipboard from the other page would be brought back into action?

MAX-786: > If I undo the modifications, is there a chance that the clipboard from the other page would be brought back into action?

AFAIK, it will, since both share the same undo stack and undoo manager saves the state with blocksClipboard and if that object is not empty and doUndo is called irrelevant to the page site it will dispatch an action to setBlocksClipboard.

But I may not be able to understand that do we want this to happen?

tiberiuichim: @MAX-786 Yes, I think mixing the blocks copied from another page with the undo stack of the current page is not desirable. I'm not sure how to solve this issue, in principle we want to know where those blocks come from.

It would be possible (as an idea) if, instead of reading the copied blocks from the store, we would push "block copy actions" into the undo stack. Just an idea, if nothing comes out of it, we may as well leave the current PR as it is. It's not perfect, but it would be fine. @kenmanheimer what do you think of this?

MAX-786: Sorry for the delayed response, stuck in clg work. 
> It would be possible (as an idea) if, instead of reading the copied blocks from the store, we would push "block copy actions" into the undo stack.

Yes, indeed this will be the better way of doing it but I wasn't successfully able to accomplish it and I took this simple method to resolve this issue for now and maybe in another issue we can improve it.

MAX-786:
Oh sorry, I missed that this PR was still open😅.

> It would be possible (as an idea) if, instead of reading the copied blocks from the store, we would push "block copy actions" into the undo stack. 

Agreed and tried to implement this and I am having an issue here, I spent a significant amount of time on it but got nowhere if possible @tiberiuichim can you please help me out here:

### My Approach: 

- I exported a function `pushToUndoStack()` from `undoManager.js`  which saves a state object in undo stack with dispatch function.
- When you paste blocks, this `pushToUndoStack()` is called.

### Issue:

- edit some blocks
- then select some blocks and copy/cut these blocks.
- now when you paste them and undo, it'll be done.
- but when you further go back it doesn't works.

Please checkout [`additon-undo-manager`](https://github.com/plone/volto/tree/addition-undo-manager) branch for the source code and if you have any approach in mind please let me know.

pnicolli:
From Volto Team Meeting 2025-03-25: we discussed this and think this is still relevant. We should also consider the option to just avoid clearing the clipboard when we paste things. It opens up the possibility to paste more than once, fixes the original issue; on the other hand, we should understand when or how to clear the clipboard.

PR #4311, opened 2023-01-25
Update volto-slate to check for slateSettings before falling back to config

danalvrz:
- Updated volto-slate to check for slateSettings before falling back to config, so that other blocks can have custom slate settings.


davisagli:
@tiberiuichim Our use case here is that we want to disable some slate toolbar buttons in a particular block. What do you think?

sneridagh (approved):

tiberiuichim (approved):
Nice! I always wanted this.

PR #4315, opened 2023-01-26
Allow the site root to be passed directly to UniversalLink

JeffersonBledsoe:
Currently, the UniversalLink produces an `Invalid item passed to UniversalLink` warning when you pass the site root to it due to the implicit type cast to a bool [here](https://github.com/plone/volto/blob/410b1e9ba577115a4eb39b27938975c9d9812a67/src/components/manage/UniversalLink/UniversalLink.jsx#L33). Rather than forcing developers to convert `''` to `'/'` each time they want to use the link, it would be nice if it were handled internally.

A question:

Should the URL be `settings.isMultilingual ? `/${lang}` : '/'` rather than only `/` to support multi-lingual sites?


davisagli (approved):
Seems fine to me.

> Should the URL be settings.isMultilingual ? /${lang} : '/' rather than only / to support multi-lingual sites?

I think it'll end up there anyway, via a redirect.

Please add a towncrier changelog entry.

davisagli:
@JeffersonBledsoe I just saw this was opened from your fork, which means the CI does not run. Could you please close it and open a new PR using a branch in the plone/volto repository?

pnicolli:
CI in forks is disabled by default but can be enabled, iirc. Correct me if I'm wrong.

stevepiercy:
> CI in forks is disabled by default but can be enabled, iirc. Correct me if I'm wrong.

We can configure workflows to run on any event. Volto is configured like so:

https://github.com/plone/volto/blob/master/.github/workflows/unit.yml#L2

```yml
on: [push]
```

This means that a job runs only when a push is made to a _branch in the repository_. If you open a PR from your own fork, then CI would run only on your repo when you push to it.

We could open it up to the outside world. From the [GitHub workflow docs](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request):

```yml
on:
  pull_request:
    types: [opened, reopened]
```

Details: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request

sneridagh:
@stevepiercy we could give it a try. @davisagli @ericof what do you think?

davisagli:
@sneridagh @stevepiercy I think it would fail, because CI running on a fork would not have access to some of the secrets used in the CI

stevepiercy:
@davisagli definitely true for a `push` event, but I don't know about a `pull_request` event made to the volto repo. I know in Pyramid `pull_request` events from remotes will build. See https://github.com/Pylons/pyramid/blob/main/.github/workflows/ci-tests.yml

stevepiercy:
Here's one PR that came from an external fork to Pyramid, to confirm.

https://github.com/Pylons/pyramid/pull/3707

sneridagh:
@JeffersonBledsoe there are some CI failures, do you mind take a look?

JeffersonBledsoe:
Superseded by https://github.com/plone/volto/pull/5263

Issue #3735, opened 2022-10-06
Issues when generating translations po/json files


erral:
We have faced the following issue:

We added volto-form-block in one of our projects and found that some of the screen messages were left untranslated. It was a bit strange because they were basic select widgets, so we started digging into this.

volto-form-block includes some widgets in its code, and it happens to reuse the same FormattedMessage IDs as in Volto.

When the frontend i18n script collects Volto's translations and then collects volto-form-block's, the second ones override the first ones, and if the translations of the second one are empty, the original translations are cleared. 

The problem lies that here we do not have the concept of translation domains that we have in the backend, so everything is compiled together and the last one wins.

I suggest 2 approaches that can be taken here:

- Stablish a convention to namespace the message IDs in Volto and addons. Something like 'volto.select', this way we use project specific msgids and will not clash with others coming from addons.

- Rework the i18n script, so that if it finds an empty translation for a given msgid, does not update the translation po/json, provided that it previously exists.

One and the other are not mutually exclusive, we can adopt both.

The ideal scenario would be to have something like translation domains ñ, but I don't know whether that's possible in JS world. 

Opinions?

erral:
CC @ionlizarazu

PR #4316, opened 2023-01-26
Add filter to remove entries with empty msgstr

danalvrz:
Regarding: [Issue 3735](https://github.com/plone/volto/issues/3735)
- Added filter to function `poToJson()` to remove entries with empty `msgstr`


news/4316.bugfix
```diff
@@ -0,0 +1 @@
+Added filter to function poToJson() to remove entries with empty msgstr @danalvrz
```
davisagli, line 1:
```suggestion
Fixed i18n script to avoid overwriting translations with an empty msgstr @danalvrz
```

packages/scripts/i18n.cjs
```diff
@@ -174,6 +174,7 @@ function poToJson({ registry, addonMode }) {
 
     // Write the corresponding language JSON, cover the special EN use case for including
     // defaults if not present
+    items = [...items].filter((item) => item.msgstr[0] !== ''); // Remove entries with msgstr = “”
```
davisagli, line 177:
```suggestion
    items = items.filter((item) => item.msgstr[0] !== ''); // Remove entries with msgstr = “”
```

davisagli (changes requested):

davisagli (approved):

erral:
I have some doubts here.

In some languages some msgstr-s are left untranslated on purpose or with an empty space for instance, because the language doesn't need that translation.

For instance there are several msgstr-s left empy in the calendar recurrence widget left untranslated in basque, because the langue structure doesn't need them.

What would happen with such translations?

See: https://github.com/plone/volto/blob/master/locales/eu/LC_MESSAGES/volto.po#L4288

davisagli:
@erral Well, this doesn't change anything about how a msgstr with a space is handled.

But I think you're right that we need to be a bit more careful so that we don't lose the defaultMessage if all msgstrs are empty.

tiberiuichim:
@erral My first guess is that the basque translation issue is a smell, probably the intl message in that place is improperly formed and should use dynamic var places to incorporate the values, rather then using some connecting words as separate entities.

sneridagh:
@davisagli @danalvrz 

If I understood the use case, is that:
- We have a translation for a msgid in place
- An add-on that reuses the msgid for some reason does not have that translation
- Then, as the add-ons are processed by the `addons` list order, could be that one translation gets overwritten.

Here could be a legit question: Do we want that add-ons overwrite the core translations? or other add-ons translations? I'd say we want this to happen, and let the add-on order to determine who wins.

If we want to fix the issue, I think that the placement of the fix is wrong. We have to fine tune the merging of the .po-s instead of the final output to the .json file.

erral:
> @erral My first guess is that the basque translation issue is a smell, probably the intl message in that place is improperly formed and should use dynamic var places to incorporate the values, rather then using some connecting words as separate entities.

the recurrence widget is indeed the hell for translators. There are several widgets with prepositions to translate. Something like this:

Every 2nd Tuesday

Which is handled like

Every: msgid
2: dropdown
nd: msgid
Tuesday: dropdown

Which in Basque, but I am pretty sure in some other language aswell is:

2. Astearte guztietan

There is no translation for Every, 2nd is a dot and not a word, and we need an additional word after Tuesday.

It is a corner case indeed, but we need to support those empty strings.

tiberiuichim:
@erral bah. Maybe we should refactor that widget, to be easier to translate.

davisagli:
@erral @sneridagh We updated the implementation so that it allows empty translations to be overwritten, but does not lose them if they are empty everywhere. Does it look okay now?

sneridagh (approved):
LGTM, @erral ?

Issue #4341, opened 2023-01-31
The ListingBody component should be exported unconnected


tiberiuichim:
It's hard to insert custom behavior between the withQuerystringResults and the ListingBody anonymous function component, because of the way the code is structured. 

My proposal is to declare the ListingBody (and export it) separately from the withQuerystringResults call, and export default the `withQuerystringResults(ListingBodyComponent)`

ajmalwali7:
I have opened a PR on this issue

ajmalwali7:
#4342

wesleybl:
@ajmalwali7  @prithvi009 I think it's good for you to talk. You are working on the same issue:

#4344
#4343

Abro0058T:
@ajmalwali7  & @prithvi009 please let me know if anyone of you are working on this issue . I am planing to work on this .

PR #4343, opened 2023-02-01
ListingBody (and export it) separated from the withQuerystringResults…

prithvi009:
#4341 ListingBody (and export it) separately from the withQuerystringResults call and export default the withQuerystringResults(ListingBodyComponent)

### DIFF
--- a/src/components/manage/Blocks/Listing/ListingBody.jsx
+++ b/src/components/manage/Blocks/Listing/ListingBody.jsx
@@ -8,7 +8,7 @@ import withQuerystringResults from './withQuerystringResults';
 import paginationLeftSVG from '@plone/volto/icons/left-key.svg';
 import paginationRightSVG from '@plone/volto/icons/right-key.svg';
 
-const ListingBody = withQuerystringResults((props) => {
+const ListingBody = (props) => {
   const {
     data = {},
     isEditMode,
@@ -111,6 +111,6 @@ const ListingBody = withQuerystringResults((props) => {
       </Dimmer>
     </div>
   );
-});
+};
 
-export default injectIntl(ListingBody);
+export default injectIntl(withQuerystringResults(ListingBody));
--- a/news/4341.enhancement
+++ b/news/4341.enhancement
@@ -1,0 +1,1 @@
+ListingBody (and export it) separated from the withQuerystringResults call @prithvi009


src/components/manage/Blocks/Listing/ListingBody.jsx
```diff
@@ -8,7 +8,7 @@ import withQuerystringResults from './withQuerystringResults';
 import paginationLeftSVG from '@plone/volto/icons/left-key.svg';
 import paginationRightSVG from '@plone/volto/icons/right-key.svg';
 
-const ListingBody = withQuerystringResults((props) => {
+const ListingBody = (props) => {
```
tiberiuichim, line 11: rename to `ListingBodyComponent`, and also export it

src/components/manage/Blocks/Listing/ListingBody.jsx
```diff
@@ -8,7 +8,7 @@ import withQuerystringResults from './withQuerystringResults';
 import paginationLeftSVG from '@plone/volto/icons/left-key.svg';
 import paginationRightSVG from '@plone/volto/icons/right-key.svg';
 
-const ListingBody = (props) => {
+const ListingBodyComponent = (props) => {
```
akshatgarg12, line 11: I guess you still have to export this

tiberiuichim: yes

tiberiuichim (changes requested):
The requirement for this ticket is like this:

- the basic component should be **exported** as a named export. That mean `export const ListingBodyComponent = (props) => ...`.
- the default export should be the one that's wrapped with `withQuerystringResults` and injectIntl. 
- there should be 2 exports in this module!

Akshat2Jain:
> The requirement for this ticket is like this:
> 
> * the basic component should be **exported** as a named export. That mean `export const ListingBodyComponent = (props) => ...`.
> * the default export should be the one that's wrapped with `withQuerystringResults` and injectIntl.
> * there should be 2 exports in this module!


 @tiberiuichim  Are you saying something like this 

```
export const ListingBodyComponent = (props) => {
  // Component implementation
};
```
```
const ListingBody = withQuerystringResults(injectIntl(ListingBodyComponent));

export default ListingBody;
```

Issue #4130, opened 2022-12-15
In SlateToolbar, when you click outside LinkEditor gets closed


MAX-786:
**Describe the bug**
While using `SlateToolbar` to create a LInk, the first time `Add Link` Button works fine but if you click outside, `LinkEditor` will get close and also if you click on `Add Link` Button again to open `LinkEditor`, it won't respond.

**To Reproduce**

1. Create any plain text Block in Volto.
2. Select some text so SlateToolbar will show.
3. On the SlateToolbar click on Add Link `🔗` Button.
4. Click anywhere outside the LinkEditor.
5. Now try to open LinkEditor again by clicking on `🔗`.

**Expected behavior**
Add Link `🔗` Button should work again after when you click outside and click on it again.

**Screenshots**

https://user-images.githubusercontent.com/99530996/207936798-88d1985b-66f4-48e8-ae05-3c85ea070897.mp4

**Software (please complete the following information):**

- OS: Windows 11
- Browser: chrome
- Volto Version [e.g. 8.5.0]
- Plone Version [e.g. 5.2.2]
- Plone REST API Version [e.g. 7.0.1]

CannedShroud:
Adding to the details of this issue, I noticed that closing the SlateBar and opening it again, resets this behaviour and we can click on the button to access the Add Link menu

PR #4348, opened 2023-02-02
fix : on clicking Add Link Button again to open LinkEditor, it won't respond.

suman9893:
fix : #4130

Don't reset savedPosition for onOverrideContent will successfully fix it .


tiberiuichim:
@suman9893 the behavior of "LinkEditor closes when you click outside" is intentional. From the original bug report, I would flag only 

> "and also if you click on Add Link Button again to open LinkEditor, it won't respond."

suman9893:
@tiberiuichim  I have understood what you have stated. Thanks for guiding. May you please check again ?

davisagli:
Duplicate of https://github.com/plone/volto/pull/4146

Issue #4230, opened 2023-01-05
Select widget steals focus


tiberiuichim:
Tested on demo.plone.org, 

Volto 16.5.0
Plone 6.0.0
plone.restapi 8.32.6
CMF 2.6.0
Zope 5.7.3
Python 3.11.1 (main, Dec 8 2022, 00:19:30) [GCC 10.2.1 20210110]
PIL 9.3.0 (Pillow)
Add-ons
@plone/volto-slate 16.5.0
@kitconcept/volto-blocks-grid 5.2.0

[vokoscreenNG-2023-01-05_14-08-26.webm](https://user-images.githubusercontent.com/141568/210777044-338b6577-f744-4009-aadc-6b48d510300c.webm)

sneridagh:
@reebalazs It seems that this is still happening under some instantiations of the select widgets. 

@avoinea as we talked during the Volto Team Meeting, could you please coordinate the efforts to fix this?

PR #4350, opened 2023-02-02
fix: #4230 select widget steals focus

tedw87:
The issue was that the focus was always set on the first field, [regardless of what was selected](https://github.com/plone/volto/pull/4350/files#diff-d959bdcdd6c623daf4101c3f276e852e02f0da86ae0b30738305cba0c03f3be0L684) so I've used the the `inFocus` state that is set when the [field is changing its value](https://github.com/plone/volto/blob/4230-select-widget/src/components/manage/Form/Form.jsx#L321-L342).
I've also updated the `inFocus` state on tab change so that the first field of the tab is in [focus by default](https://github.com/plone/volto/pull/4350/files#diff-d959bdcdd6c623daf4101c3f276e852e02f0da86ae0b30738305cba0c03f3be0R256-R262).


tiberiuichim:
@tedw87 Would be nice to get a description of the underlying problem, so that we can understand the fix

tiberiuichim (approved):

avoinea (approved):

avoinea:
@sneridagh This is starting to bite us. Can you please prioritize it?

sneridagh:
@avoinea on it!

PR #4353, opened 2023-02-02
Conditionally render all delete items in confirm widget

### LINKED ISSUE #4336: Delete confirmation screen should not list all the items, in case there are more then X threshold
image

I don't think it's user friendly if the delete confirmation popup message lists all the objects. It should say only "Delete 3 items?" or something like that.

akshatgarg12:
fixes #4336

### DIFF
--- a/news/4336.bugfix
+++ b/news/4336.bugfix
@@ -1,0 +1,1 @@
+bugfix: conditionally render all delete items in confirm widget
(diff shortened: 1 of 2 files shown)


tiberiuichim (approved):

sneridagh:
@akshatgarg12 could you please include screenshots of the final UX for this?

PR #4371, opened 2023-02-06
@plone/client - Agnostic Plone 6 RESTAPI client

sneridagh:
This package is an agnostic client written in TypeScript and built on the top of [@tanstack/query](https://tanstack.com/query) to query Plone 6 backend.

It should be possible to use it in React/Vue/Solid/Svelte projects.


sneridagh:
Moving it to: https://github.com/plone/plone.restapi-client

PR #4379, opened 2023-02-07
Add open external link in a new tab config option.

robgietema:


erral:
There is an advanced link entering menu for Slate that can be enabled and offers an advanced link option entering panel.

This option just makes the open in new tab global, which I think is like a patch... I think what we need is to think better the link entering panel and perhaps document how to enable the advanced one or improve it a little bit.

We have enabled the advanced one in volto-slate in a Volto 15, and it works, but perhaps it needs some love.

The advanced one is documented in the volto-slate repo: https://github.com/eea/volto-slate#3-working-with-links-internal-external-email

tiberiuichim (approved):

robgietema:
> There is an advanced link entering menu for Slate that can be enabled and offers an advanced link option entering panel.
> 
> This option just makes the open in new tab global which I think is like a patch... I think what we need is to think better the link entering panel and perhaps document how to enable the advanced one or improve it a little bit.
> 
> We have enabled the advanced one in volto-slate in a Volto 15, and it works, but perhaps it needs some love.
> 
> The advanced one is documented in the volto-slate repo: https://github.com/eea/volto-slate#3-working-with-links-internal-external-email

I am aware of the advanced menu and that is a great option to select the target on a per link basis. This PR sets the default target of external links which is a slightly different usecase. Both can coexist next to each other.

PR #4381, opened 2023-02-08
Show related items

### LINKED ISSUE #3740: related items don't show by default
on https://6.demo.plone.org/ if I edit the front page and set a related item, it does not appear in the normal view

wesleybl:
I need help with the layout:

related items

Can someone help me?

fixes #3740


ksuess:
I would say
- rewrite Tags.jsx to Segment.Group instead of  Container,
- include Tags.jsx and Related.jsx in a new MetaData.jsx with a Semantic UI Container.

But this would mean a breaking change, I guess.

wesleybl:
@ksuess I don't think it would be a good idea to do a breaking change for this. Someone who customized the Tags might not like it.

I tried to follow the HTML structure of the Semantic UI lists described here:

https://semantic-ui.com/elements/list.html

Of the React components, I used only the Container to wrap everything.

More suggestions?

tiberiuichim:
@sneridagh @tisto any insight on a visual for this?

tisto:
@tiberiuichim I can ask our designer to come up with something. Though, that might take a few days...

ichim-david:
@wesleybl The related functionality is still not present in Volto to the point that it was added as a tutorial block from World Plone Day https://github.com/collective/volto-relateditems-block
I wonder at this point if something will be agreed for core...

wesleybl:
@wesleybl I think related content shouldn't be in blocks. They should appear without explicit user action, as occurs in classic Plone. I will try to update this PR.

packages/volto/src/config/index.js
```diff
@@ -149,6 +149,7 @@ let config = {
     serverConfig,
     storeExtenders: [],
     showTags: true,
+    showRelatedItems: true,
```
ichim-david, line 152: There is still no suggestion for making it prettier, I suggest this feature is off by default making it a non-breaking feature that people can opt to enable and style as needed.

ichim-david (changes requested):
@wesleybl see my review suggestions

packages/volto/src/components/theme/RelatedItems/RelatedItems.jsx
```diff
@@ -0,0 +1,74 @@
...
+ * Related Items component.
+ * @function RelatedItems
+ * @param {array} tags Array of related items.
+ * @returns {string} Markup of the component.
+ */
+
+const RelatedItems = ({ relatedItems, intl }) =>
+  relatedItems && relatedItems.length > 0 ? (
+    <Container>
+      <h4 className="ui header">{intl.formatMessage(messages.relatedItems)}</h4>
```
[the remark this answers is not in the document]
wesleybl: I also changed the HTML to look more like `volto-relateditems-block`. Now the Related Items will look like this:

download (5)

I didn't include the CSS because it didn't make a difference.

packages/volto/src/config/index.js
wesleybl: I changed the setting, but I think it should be true by default, just like it is in the classic UI. This makes it even harder to test.

ichim-david:
@wesleybl don't have time to spend on open-source anymore for today, I already do so for too long :) :( but I plan on checking this tomorrow.

packages/volto/src/components/theme/RelatedItems/RelatedItems.jsx
```diff
@@ -0,0 +1,71 @@
...
+ * @param {array} relatedItems Array of related items.
+ * @returns {JSX.Element} Markup of the component.
+ */
+const RelatedItems = ({ relatedItems, intl }) => {
+  if (!relatedItems || relatedItems.length === 0) {
+    return null;
+  }
+
+  return (
+    <Container style={{ 'margin-top': '20px' }}>
```
ichim-david, line 30: @wesleybl please don't use inline objects to set css, put a class on it and apply the style from css. If someone needs more space they will need to add an important rule to override the inline style.

wesleybl: Done.

packages/volto/src/components/theme/RelatedItems/RelatedItems.jsx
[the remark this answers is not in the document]
ichim-david: > Done. I also changed the HTML to look more like `volto-relateditems-block`. Now the Related Items will look like this:
> 
> download (5)
> 
> I didn't include the CSS because it didn't make a difference.

@wesleybl if this is the final render update the screenshot from the ticket description with this or the latest render.

ichim-david:
@erral @pnicolli @davisagli @ksuess We need to decide on how we want to display the related items in Volto.
@wesleybl has worked on adding the display of the related items as seen in the ticket description.
There are several questions that need answering such as:
1. Are we ok with adding the component after the tags directly in the View template or should we add it through Slots?
2. Are we ok with this minimal styles or do we need something more?
3. This work adds the ability to see the related items in Volto core but during World Plone Day this functionality was added as a block in collective https://github.com/collective/volto-relateditems-block. Do we want to have it as a component added in the template or as a block?
4. What kind of options do we want for this work, do we want to show it in View.jsx or would there be a need to insert it based on certain conditions?
5. Should we have this feature disabled or enabled by default? If we opt for the minimal style and as a non-breaking feature we could go for the opt-in solution where you enable and further style it as needed if you want it.

Unless we all chip in on this pull request, if we have any feedback I don't know how we should treat this work because I don't know what is your expectations for the related items viewing especially given that we haven't added it to core until now.

wesleybl:
@ichim-david when I made this PR, slots didn't exist in Volto. But I think it makes sense now for Related Items to be in slots. I think it would be interesting for them to be in the `belowContent` slot. But for that, Tags would have to be in that slot too, above the related items. See:

https://github.com/plone/volto/blob/8915be36ffb8a896e16ea3bccdae8ceb38a0d937/packages/volto/src/components/theme/View/View.jsx#L258-L263

One detail is that Volto doesn't register anything in the slot. But I don't think there would be a problem registering it, right?

Can I make the change?

ichim-david:
> @ichim-david when I made this PR, slots didn't exist in Volto. But I think it makes sense now for Related Items to be in slots. I think it would be interesting for them to be in the `belowContent` slot. But for that, Tags would have to be in that slot too, above the related items. See:
> 
> https://github.com/plone/volto/blob/8915be36ffb8a896e16ea3bccdae8ceb38a0d937/packages/volto/src/components/theme/View/View.jsx#L258-L263
> 
> One detail is that Volto doesn't register anything in the slot. But I don't think there would be a problem registering it, right?
> 
> Can I make the change?

@wesleybl  I am more in favor of having this functionality as a registration in the below content slot rather than adding more things directly to the template especially since you can even condition the slot addition.
However I don't know what is the sentiment of this pull request and despite my effort to @ people as you have seen so far we have gotten no replies. I think it also has todo with the time of the year, people are either busy preparing for holidays or are on holidays as we speak.

Unless we get feedback from the other I don't know in what direction to point you as I don't know what the other core members think about related items.

As such I would wait for more feedback before any more development as to not have to make too many changes on and on and on.

davisagli:
@wesleybl @ichim-david I like the idea of using slots for this, because then it's pretty easy for an add-on to replace the component if it wants to do something different (maybe even different components for different content types). In that case we might not need the `showRelatedItems` setting (because it's possible to override the slot with an empty component)

erral:
Yes, I also like the idea of having it in the said slot. If I am not wrong, in Classic UI there is a viewlet to render the related items which is rendered in a viewlet manager. And slots are similar to viewlet manager, si it makes total sense.

I also agree with the point that @davisagli raises regarding the need of the `showRelatedItems` setting.

Regarding changing the Tags component and adding it to a Slot, it makes total sense, but I think that's out of the scope of this PR.

wesleybl:
I put the RelatedItems in the slot.

I still think the `showRelatedItems` setting should be `true` by default. Can I change this?

packages/volto/news/3740.feature
```diff
@@ -0,0 +1,2 @@
+- Show related items.
+  @wesleybl
```
stevepiercy, line 2:
```suggestion
```

packages/volto/news/3740.feature
```diff
@@ -0,0 +1,2 @@
+- Show related items.
```
stevepiercy, line 1:
```suggestion
Show related items. @wesleybl
```

stevepiercy (changes requested):
Minor suggestion to news item. This PR needs a technical review.

packages/volto/news/3740.feature
wesleybl: Done

davisagli (approved):
Looks good to me.

> I still think the showRelatedItems setting should be true by default. Can I change this?

@wesleybl You can do this for Volto 18. If you want to backport this to Volto 16 or 17, it should be false by default there.

wesleybl:
> You can do this for Volto 18. If you want to backport this to Volto 16 or 17, it should be false by default there.

Done to Volto 18

fredvd:
I only found this issue this week. Yes, putting the related items in the slot is the most flexible and others a good default experience.

Please consider setups where you do want authors to control the related items in a block the site could be set up to hide th default slot and add a block for the related items.  Both are possible this way.

davisagli:
@ichim-david Could you update your review please?

@plone/volto-team Can we merge this for Volto 18?

ichim-david (approved):

ichim-david:
@davisagli I haven't tested the latest changes but everything that is added came out of our comments on what we think it's best. I trust your judgement on how it works with the slot implementation. At this moment I am getting ready for the trip to Ferrara so I am accepting on principle and not due to further testing.

davisagli:
We discussed in the Volto Team meeting. We missed merging this for Volto 18 so we want to merge it with the setting off by default. We can consider turning it on by default in the future in Volto 19.

wesleybl:
> We discussed in the Volto Team meeting. We missed merging this for Volto 18 so we want to merge it with the setting off by default. We can consider turning it on by default in the future in Volto 19.

@davisagli I changed the default to false.

packages/volto/news/3740.feature
```diff
@@ -0,0 +1 @@
+Show related items. @wesleybl
```
davisagli, line 1:
```suggestion
Add component to show related items, which can be enabled using the `showRelatedItems` setting. @wesleybl
```

packages/volto/src/components/index.js
```diff
@@ -22,6 +22,7 @@ export { default as Title } from '@plone/volto/components/theme/Title/Title';
 export { default as DefaultView } from '@plone/volto/components/theme/View/DefaultView';
 export { default as Pagination } from '@plone/volto/components/theme/Pagination/Pagination';
 export { default as Tags } from '@plone/volto/components/theme/Tags/Tags';
+export { default as RelatedItems } from '@plone/volto/components/theme/RelatedItems/RelatedItems';
```
davisagli, line 25: Let's not add new components to the barrel files, since we want to remove them (#6507)
```suggestion
```

davisagli (approved):

davisagli:
@wesleybl Obrigado!

packages/volto/news/3740.feature
```diff
@@ -0,0 +1 @@
+Add component to show related items, which can be enabled using the `showRelatedItems` setting. @wesleybl
```
tisto, line 1: @sneridagh this could be considered a breaking change. Imagine a client project where the editors "try out" this new feature and the visual design is broken. For instance, we need designs for this for VLT, DLR, FZJ, etc...

Technically it is a new feature but it has the potential to break client expectations. We either have to mark those issues always as breaking or we have to change our internal policy and very closely monitor feature releases internally.

davisagli: @tisto Yeah, I'm confused by your comment. This is why we added the setting and made it disabled by default, so that it is not a breaking change. It's a setting in the Volto config for developers, not for editors.

PR #4413, opened 2023-02-19
Export App and its connectorAppComponent as standalone

tiberiuichim:
With this it's possible to wrap the App in additional context providers. To do this, you need to:

- override the default app component registration:

```
config.registerComponent({name: 'App', component: MyApp})
```

- write your own component:

```
import { App, connectAppComponent} from '@plone/volto/components/theme/App/App';

const MyAppComponent = (props) => <MyContextProvider><App {...props} /><MyContextProvider>;
export default connectAppComponent(MyAppComponent)
```

The benefit of doing it like this is that the `MyApp` component is wrapped in asyncConnect so it has access to the store.

### DIFF
--- a/src/components/theme/App/App.jsx
+++ b/src/components/theme/App/App.jsx
@@ -57,7 +57,7 @@ import LockingToastsFactory from '@plone/volto/components/manage/LockingToastsFa
  * @class App
  * @extends {Component}
  */
-class App extends Component {
+export class App extends Component {
   /**
    * Property types.
    * @property {Object} propTypes Property types.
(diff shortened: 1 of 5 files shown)


tiberiuichim:
I think an entry in the Upgrade Guide is needed, as we're making a change in the default generated route.js

src/config/Components.jsx
```diff
@@ -1,5 +1,6 @@
-import { PreviewImage } from '@plone/volto/components';
+import { App, PreviewImage } from '@plone/volto/components';
```
sneridagh, line 1: I'd start using full paths as default, instead of using the indexes... for code splitting, right?

src/components/theme/App/App.jsx
```diff
@@ -259,72 +259,76 @@ export const fetchContent = async ({ store, location }) => {
...
+        content: state.content.data,
+        apiError: state.apierror.error,
+        connectionRefused: state.apierror.connectionRefused,
+      }),
+      null,
+    ),
+  )(AppComponent);
+}
+
+export default connectAppComponent(App);
```
sneridagh, line 334: So default export is 'connected' then named export is bare component with bare name, right?

Trying to establish conventions (and forward thinking in `@plone/components` as well).

How about naming `connectedAppComponent` instead?

sneridagh (approved):

PR #4429, opened 2023-02-23
Add OK route

### LINKED ISSUE #4375: OK page for Volto
In Plone we currently have the `@@ok` view where Plone returns a simple `OK` message to signal that it is alive and thus this URL can be used in load-balancers or keep-alive services to check that the website is running.

We do not have such a feature in Volto and in some cases if the back-end is not exposed the only way to check if the website is alive is to hit the front-page or some other pages.

I would like to have a non content route called `ok` in Volto, that renders a component that simply returns `OK` like the Plone one does. This way this URL can be added to any service checking the status of the site.

erral:
Fixes #4375 

I think that something like this will work. 

Anyway, I don't know if I am using useEffect here correctly... when running these changes in my laptop, the route runs just once and makes a query to the backend.

I  have hard-coded a back-end query to /, this way we can guarantee that the Plone backend is working.

But, as explained in the issue, this renders the full page HTML and not just the `OK` message that Plone does.

What do you think?


tiberiuichim (commented):
Do we really want to hit the content here? 

If we are to hit the Plone backend, then I would argue that we should hit the same thing, which is Plone's ok view.

avoinea (approved):

erral:
I think that we need to hit the content.

Then we should need a ok view response in JSON, shouldn't we?

erral:
The alternative express-way is in this other commit of another branch:

https://github.com/plone/volto/commit/fd446b83cf868412cd1475e67a4edac9e76f0a6d

tiberiuichim:
https://github.com/plone/Products.CMFPlone/blob/981aadcb4dad75497160a6405e1f01b3759bc345/Products/CMFPlone/browser/okay.py#L4-L26

Here's my proposal:

- let's implement this as an expressjs middleware
- in that middleware we should not do anything more then the equivalent of the Plone view. We should not connect to the backend.

Why? If the ok route is supposed to be used internally by the docker healthcheck, then it should be an atomic operation that correctly expresses the state of the running service. Why would the Volto container be restarted if the Plone container is down? This just makes things worse.

tiberiuichim:
The other branch seems fine except maybe for the caching headers. There should be a PR for it.

erral:
> Why? If the ok route is supposed to be used internally by the docker healthcheck, then it should be an atomic operation that correctly expresses the state of the running service. Why would the Volto container be restarted if the Plone container is down? This just makes things worse.


I had not docker internal health-check in mind, but to know the availability of the full website (front + back). But I get your point.

avoinea:
@erral Is this obsolete in favor of #4432?

erral:
> @erral Is this obsolete in favor of #4432?

yes.

PR #4432, opened 2023-02-24
Add /ok route as an express middleware

### LINKED ISSUE #4375: OK page for Volto

erral:
Fixes #4375 adding an Express middleware

This supersedes #4429


tiberiuichim (approved):

tiberiuichim (changes requested):
In Plone the ok view is namespaced by the `@@` symbols, so there's no chance to overlap something else. Let's do something similar here. Maybe `/_ok` ? 

Also, I don't think there's a need for a setting for this, if we solve the namespacing problem.

davisagli:
How about something like `.well-known/volto/ok` (based on https://en.wikipedia.org/wiki/Well-known_URI)?

Whatever it is, it should be mentioned in the docs.

avoinea:
> In Plone the ok view is namespaced by the `@@` symbols, so there's no chance to overlap something else. Let's do something similar here. Maybe `/_ok` ?
> 
> Also, I don't think there's a need for a setting for this, if we solve the namespacing problem.
> 
> @plone/volto-team thoughts on this?

Well, it also works with `/ok` in Plone Classic and we're doing all health-checks to this path `/ok`. Thus, I would stick to `/ok`. Or use the `.well-known` approach as @davisagli suggested.

sneridagh:
We need a consensus here. I'm ok with using both `/ok` and `.well-known/volto/ok` the reasoning behind both seems ok (no pun intended) for me. Everybody is ok? 😂

tiberiuichim:
I'm +1 on adopting the industry standard.

erral:
I don't see a ok URL under the well-known URLs right? I mean I don't mind using the .well-known approach, but right know there's no such standard for ok right?

We can start having the ok like we have in classic and evolve to whatever is stablished later on.

davisagli:
I was the one who brought up the `.well-known` idea, but I don't see any big problem with `/ok`: it is unlikely to conflict with content, and if it does there is the setting. So I would go ahead with this the way it is. The one thing I am missing in the PR is documentation of the new setting.

stevepiercy:
> The one thing I am missing in the PR is documentation of the new setting.

Would that go here?

https://6.docs.plone.org/volto/configuration/settings-reference.html#server-specific-serverconfig

erral:
I have realized that I was missing the documentation part on this PR as requested by @stevepiercy :facepalm: so I have added it :wink:

docs/source/configuration/settings-reference.md
```diff
@@ -374,8 +374,17 @@ additionalToolbarComponents
       }}
     </Plug>
     ```
-```
+````
```
stevepiercy: This is the closing fence for the opening glossary fence at the start of the page. It should be moved after your new glossary entry.

```suggestion
```

docs/source/configuration/settings-reference.md
```diff
@@ -374,8 +374,17 @@ additionalToolbarComponents
...
+
+okView
+Volto provides a `/ok` URL where it respondes with a text/plain ok response (with an `HTTP 200` status code) to signal third-party health-check services that the Volto process is running correctly.
```
stevepiercy: Indent glossary definitions.

```suggestion
    Volto provides an `/ok` URL where it responds with a `text/plain ok` response, with an `HTTP 200` status code, to signal third party health check services that the Volto process is running correctly.
```

docs/source/configuration/settings-reference.md
```diff
@@ -374,8 +374,17 @@ additionalToolbarComponents
...
+Using this setting one can modify such an URL and configure to respond in some other URL.
```
stevepiercy:
```suggestion
    Using this setting, one can modify such an URL and configure it to respond with another URL.
```

docs/source/configuration/settings-reference.md
```diff
@@ -374,8 +374,17 @@ additionalToolbarComponents
...
+The provided default matches the existing Plone Classic URL.
```
stevepiercy:
```suggestion
    The provided default matches the existing Plone Classic UI URL.
```

docs/source/configuration/settings-reference.md
```diff
@@ -374,8 +374,17 @@ additionalToolbarComponents
...
 
+    ```jsx
+      config.settings.okView = '/site-is-ok'
+
```
stevepiercy, line 394:
```suggestion
```

docs/source/configuration/settings-reference.md
```diff
@@ -374,8 +374,17 @@ additionalToolbarComponents
...
+    ```
```
stevepiercy, line 395: Close glossary fence.

````suggestion
    ```
```
````

stevepiercy (changes requested):
Thanks for the docs. Mostly MyST and English syntax and grammar.

erral:
> Thanks for the docs. Mostly MyST and English syntax and grammar.

thank you as always!

stevepiercy (approved):
Looks good! https://deploy-preview-4432--volto.netlify.app/configuration/settings-reference.html#term-okRoute

stevepiercy:
It is OK to ignore the failing docs job. He Who Shall Not Be Named has once again pulled more feathers from the bird site, now requiring authentication to view a user's twerks.

sneridagh (approved):

PR #4434, opened 2023-02-26
Improvements to dev proxy

davisagli:
1. Prefer RAZZLE_INTERNAL_API_PATH over RAZZLE_API_PATH for devProxyToApiPath. This makes more sense (the proxy target is accessed from the Volto server side) and it makes the dev proxy work without additional configuration when running Volto in a Docker container RAZZLE_API_INTERNAL_API_PATH is already set correctly.
2. Keep the URL protocol from the API path in the virtual hosting path passed to the backend, instead of hardcoding http. This lets the backend generate correct URLs when the backend is served over https.
3. Use a new environment variable, RAZZLE_DEV_PROXY_INSECURE, to control whether the proxy checks certificates on the backend.
4. Always log where API paths are proxied on startup, not only in development mode.
5. In the APIResourceWithAuth helper, always use devProxyToApiPath, not only in development mode. It's a potentially confusing inconsistency.

With these changes, it's possible to run Volto with a remote backend served over https. For example:

    RAZZLE_INTERNAL_API_PATH=https://demo.plone.org RAZZLE_PROXY_REWRITE_TARGET=/++api++ RAZZLE_DEV_PROXY_INSECURE=1 yarn start


davisagli:
Note: I still need to check if there are any implications for the docs.

src/express-middleware/devproxy.js
```diff
@@ -75,12 +75,14 @@ export default function () {
...
         config.settings.proxyRewriteTarget ||
-        `/VirtualHostBase/http/${apiPathURL.hostname}:${apiPathURL.port}${instancePath}/++api++/VirtualHostRoot`;
+        `/VirtualHostBase/${apiPathURL.protocol.slice(0, -1)}/${
+          apiPathURL.hostname
+        }:${apiPathURL.port}${instancePath}/++api++/VirtualHostRoot`;
 
       return `${target}${path.replace('/++api++', '')}`;
     },
     logLevel: process.env.DEBUG_HPM ? 'debug' : 'silent',
-    ...(config.settings?.proxyRewriteTarget?.startsWith('https') && {
```
davisagli, line 83: The old code here was never used, because the settings aren't loaded yet when the proxy middleware is initialized.

src/helpers/Api/APIResourceWithAuth.js
```diff
@@ -21,7 +21,7 @@ export const getAPIResourceWithAuth = (req) =>
     let apiPath = '';
     if (settings.internalApiPath && __SERVER__) {
       apiPath = settings.internalApiPath;
-    } else if (__DEVELOPMENT__ && settings.devProxyToApiPath) {
```
davisagli, line 24: Looks like this one may have broken a test in seamless mode; I need to take a closer look.

src/express-middleware/devproxy.js
sneridagh: Uh, is that so? :( 

Ok!

src/start-server.js
```diff
@@ -19,9 +19,11 @@ export default () => {
       } else {
         console.log(`API server (API_PATH) is set to: ${app.apiPath}`);
       }
-      if (__DEVELOPMENT__ && app.devProxyToApiPath)
+      if (app.devProxyToApiPath)
         console.log(
-          `Using internal proxy: ${app.publicURL} -> ${app.devProxyToApiPath}`,
+          `Proxying API requests from ${app.publicURL}/++api++ to ${
```
sneridagh, line 24: We might want to deprecate the legacy mode at some point too (since this might be false, if used).

sneridagh (approved):
LGTM to me, we should revisit the whole docs on Seamless/Deployment.

sneridagh:
@davisagli seamless traefik docker config:

```http:
  routers:
    frontend:
      rule: "Host(`localhost`)"
      service: frontend
    backend:
      rule: "Host(`localhost`) && PathPrefix(`/++api++`)"
      service: backend
      middlewares:
        - backend

  middlewares:
    backend:
      replacePathRegex:
        regex: "^/\\+\\+api\\+\\+($|/.*)"
        replacement: "/VirtualHostBase/http/localhost/plone/++api++/VirtualHostRoot$1"

  services:
    frontend:
      loadBalancer:
        servers:
          - url: "http://host.docker.internal:3000"
    backend:
      loadBalancer:
        servers:
          - url: "http://host.docker.internal:55001"
```

has `localhost` on it, could be the problem?

src/start-server.js
davisagli: @sneridagh If legacyTraverse is true then the client does not add `/++api++` to URLs it fetches, but the devproxy is still running and would proxy a `/++api++` request to the same location. Right?

src/helpers/Api/APIResourceWithAuth.js
davisagli: @sneridagh Without the check for `__DEVELOPMENT__`, this ends up trying to use the default devProxyToApiPath (http://localhost:8080/Plone) in seamless mode in production, instead of leaving the apiPath blank.

I think there is probably a way to make this less confusing by using a default for internalApiPath instead of devProxyToApiPath, but I'm going to take this change out of this PR for now.

sneridagh:
works like a charm... We should document it properly.

PR #4436, opened 2023-02-27
3762-integrate listing-block headline with with TOC

tedw87:
Configured the `listing-block` by adding the `tocEntry` function used by `block-toc` to generate the table of content

https://user-images.githubusercontent.com/104510089/221514697-77061332-7552-43c9-b220-1a5e7798c402.mov

### DIFF
--- a/src/components/manage/Blocks/ToC/View.jsx
+++ b/src/components/manage/Blocks/ToC/View.jsx
@@ -6,6 +6,9 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import { injectIntl } from 'react-intl';
+import { useSelector } from 'react-redux';
+import isArray from 'lodash/isArray';
+import isObject from 'lodash/isObject';
 import cx from 'classnames';
 import { Message } from 'semantic-ui-react';
 import config from '@plone/volto/registry';
@@ -33,6 +36,9 @@ const View = (props) => {
         : [1, 2, 3, 4, 5, 6],
     [data],
   );
+
+  const querystringsearch = useSelector((state) => state.querystringsearch);
+
   const tocEntries = React.useMemo(() => {
     let rootLevel = Infinity;
     let entries = [];
@@ -49,8 +55,11 @@ const View = (props) => {
       const entry = config.blocks.blocksConfig[block['@type']]?.tocEntry(
         block,
         data,
+        {
+          querystringsearch,
+        },
       );
-      if (entry) {
+      if (isArray(entry)) {
         const level = entry[0];
         const title = entry[1];
         const items = [];
@@ -60,6 +69,25 @@ const View = (props) => {
         if (level < rootLevel) {
           rootLevel = level;
         }
+      }
+      if (isObject(entry) && isArray(entry.entries)) {
+        entry.entries.forEach((e, index) => {
+          const i = `${id}-${index}`;
+          const level = e[0];
+          const title = e[1];
+          const items = [];
+          if (!level || !levels.includes(level)) return;
+          tocEntriesLayout.push(i);
+          tocEntries[i] = {
+            level,
+            title: title || block.plaintext,
+            items,
+            id: i,
+          };
+          if (level < rootLevel) {
+            rootLevel = level;
+          }
+        });
       }
     });
 
@@ -91,7 +119,14 @@ const View = (props) => {
     });
 
     return entries;
-  }, [data, levels, properties, blocksFieldname, blocksLayoutFieldname]);
+  }, [
+    data,
+    levels,
+    properties,
+    blocksFieldname,
+    blocksLayoutFieldname,
+    querystringsearch,
+  ]);
 
   const Renderer = variation?.view;
   return (
--- a/src/config/Blocks.jsx
+++ b/src/config/Blocks.jsx
@@ -285,6 +285,20 @@ const blocksConfig = {
       },
     ],
     getAsyncData: getListingBlockAsyncData,
+    tocEntry: (blockData = {}, _, store) => {
+      const search = store.querystringsearch.subrequests[blockData.block] || {};
+      const levels = {
+        h2: 2,
+        h3: 3,
+      };
+      if (!search.items) return false;
+      return {
+        entries: search.items.map((item) => [
+          levels[blockData.headlineTag],
+          item.Title,
+        ]),
+      };
+    },
   },
   video: {
     id: 'video',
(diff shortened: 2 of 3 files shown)


tiberiuichim:
ah... I'm not sure why we're going this route. We (at least I assume it's a general "we") don't want to include the actual listed items in the Toc, just the block headline.

image

razvanMiu (approved):

src/config/Blocks.jsx
```diff
@@ -289,6 +289,14 @@ const blocksConfig = {
       },
     ],
     getAsyncData: getListingBlockAsyncData,
+    tocEntry: (blockData = {}, _) => {
```
tiberiuichim, line 292: Please document the `tocEntry` setting here: https://6.dev-docs.plone.org/volto/blocks/settings.html

docs/source/blocks/settings.md
```diff
@@ -1,10 +1,10 @@
...
 myst:
   html_meta:
-    "description": "How to configure custom blocks"
-    "property=og:description": "How to configure custom blocks"
-    "property=og:title": "Blocks settings"
-    "keywords": "Volto, Plone, frontend, React, Block settings"
+    'description': 'How to configure custom blocks'
+    'property=og:description': 'How to configure custom blocks'
+    'property=og:title': 'Blocks settings'
+    'keywords': 'Volto, Plone, frontend, React, Block settings'
```
stevepiercy, line 7: Please revert all changes of `"` to `'`, and configure your editor not to make these changes.

docs/source/blocks/settings.md
```diff
@@ -182,13 +196,16 @@ const initialBlocks = {
 `allowed_headline_tags`
 : Allows you to customize the choices of the "Headline Tag" types shown in the block settings by default. It has the following syntax (a list of lists, where a list item consists of `['token', 'display_name']`):
 
-  ```js
-  allowed_headline_tags: [['h2', 'h2'], ['h3', 'h3']]
-  ```
+```js
```
stevepiercy, line 199: Revert all dedents. Indentation is required for MyST syntax.

stevepiercy (changes requested):

stevepiercy (approved):
Docs look good to me now.

src/components/manage/Blocks/ToC/View.jsx
```diff
@@ -61,6 +64,25 @@ const View = (props) => {
...
+            level,
+            title: title || block.plaintext,
+            items,
+            id: i,
+          };
+          if (level < rootLevel) {
+            rootLevel = level;
+          }
+        });
+      }
```
davisagli, line 85: I don't think the changes in this file are needed. The new `tocEntry` function is always returning an array, not an object.

davisagli (changes requested):

docs/source/blocks/settings.md
```diff
@@ -102,7 +102,22 @@ const customBlocks = {
           }
         ]
       }
-    }
+    },
+     // A block can have elements included in ToC (Table of Contents block)
+    tocEntry: (blockData, data) => {
+    // Arguments of the function will be the data of the block
+    // which will interact with ToC and the data of ToC
+      // If a singe entry is needed, the function will return an array with the
```
davisagli, line 110:
```suggestion
      // If a single entry is needed, the function will return an array with the
```

docs/source/blocks/settings.md
```diff
@@ -102,7 +102,22 @@ const customBlocks = {
...
+      // level and the element to be placed in ToC
+      return [level, title]
+      // If there are more entries to be displayed in Toc, the function will
+      // return an object like below:
+      return {
+        // Each nested array represents an elememnt at a specific level
+        // to be displayed in the ToC
+        entries: [[level, title], [level, title]]
+      }
+    },
```
davisagli, line 120: I would remove the part about multiple entries, and the related changes in View.jsx. It's not a bad idea, but it's not needed for this block, and I think it's best to wait to add it until there is a concrete use case, so that we can make sure it's needed.

stevepiercy (changes requested):
Minor grammar fixes.

sneridagh:
@tedw87 could you please take a look at @davisagli suggestions? Thanks!

tiberiuichim:
I haven't checked the code lately, but as far as I understand, the conceptual problem is this: do we want a block to be able to return multiple TOC entries? To which I'd say: yes, of course!

davisagli:
@tiberiuichim I agree with that conceptually, but I would not add it without applying it to a concrete use case to make sure we're designing the right interface for configuring it, and we don't have that use case in the listing block, so I would not add it in this PR. Maybe I'm being too careful.

The particular concern I have with the implementation here is that the caller of tocEntry has to account for the fact that it can return 2 different types. It would be better to make the return type consistent (i.e. always a list of entries). But that's not backwards compatible. So maybe the path forward is to make a new function called tocEntries and deprecate tocEntry.

razvanMiu:
@davisagli @sneridagh @tiberiuichim 
I added a new function that can be defined in blocks configuration (tocEntries) that returns an array of entries. The old one (tocEntry) can still be used and is also required for backward compatibility. [check here](https://github.com/plone/volto/pull/4436/files#diff-bddad260db9ed7beb597b5396f2950460307b8c3d8445ad5013640c628a0a194R124-R131)
I also added a new helper function [getBlocksTocEntries](https://github.com/plone/volto/pull/4436/files#diff-bddad260db9ed7beb597b5396f2950460307b8c3d8445ad5013640c628a0a194R97-R158) that can be used to get the entries of a BlocksForm.

Example of usage:

```
config.blocks.blocksConfig.columnsBlock.tocEntries = (blockData, data) => {
  const columns = blockData.data.blocks_layout.items || [];
  let entries = [];

  columns.forEach((column) => {
    const { blockDataFormEntries } = getBlocksTocEntries(
      {
        ...blockData.data.blocks[column],
      },
      data,
    );
    entries = [...entries, ...blockDataFormEntries];
  });

  return entries;
};
```

docs/source/blocks/settings.md
```diff
@@ -101,7 +101,17 @@ const customBlocks = {
           }
         ]
       }
-    }
+    },
+    // A block can have elements included in ToC (Table of Contents block)
+    // Arguments of the function will be the data of the block and the data of ToC
+    tocEntries: (blockData, data) => {
+      // The function returns an array of entries each entry defined by
```
stevepiercy, line 108:
```suggestion
      // The function returns an array of entries, with each entry defined by
```

davisagli:
Outdated. https://github.com/plone/volto/pull/5146 added support for configuring a tocEntries function in the blocksConfig. It still needs to be added for blocks that have a headline.

PR #4443, opened 2023-03-01
fix: newsitem view blocks wrapper className

nzambello:


davisagli (commented):
Searching the codebase shows `viewwrapper` also in EventView. Does it need to be fixed there too?

Please add a changelog entry in the `news` folder.

davisagli (approved):

PR #4452, opened 2023-03-02
Adds "missing" support for form defaults

tiberiuichim:


docs/source/blocks/editcomponent.md
```diff
@@ -127,6 +127,37 @@ import { Icon } from '@plone/volto/components';
 </SidebarPortal>;
 ```
 
+### Default values support for block schema forms
+
+In the block schema definition you can use the two additional properties for
```
stevepiercy, line 132:
```suggestion
In the block schema definition, you can use the two additional properties for
```

docs/source/blocks/editcomponent.md
```diff
@@ -127,6 +127,37 @@ import { Icon } from '@plone/volto/components';
...
+each field: `default` and `missing`. The `default` value is used to populate
+the data as "initial value". The `missing` value should be used when the field
```
stevepiercy, line 134:
```suggestion
the field with an initial value. The `missing` value should be used when the field
```

docs/source/blocks/editcomponent.md
```diff
@@ -127,6 +127,43 @@ import { Icon } from '@plone/volto/components';
...
+
+- `default` and `missing` are declared: the block form is initialized with
+  the `default` value. When the user clears the field, the field is
+  reinitialized with the `missing` value. This would imply a "required" field.
+- only `default` is declared: the block form is initialized with the `default`
+  value. When the user clears the field, the field remains empty. This would
+  imply an "optional" field.
+- only `missing` is declared: the block form is initialized with the `missing`
+  value. When the user clears the field, the field is reinitialized with the
+  `missing` value.
```
stevepiercy, line 159: I'm curious. Does this follow the same pattern as Colander, which shares its roots with Zope? https://docs.pylonsproject.org/projects/colander/en/latest/api.html#schema-related

We need to add the case when neither `missing` nor `default` is defined.

The scenarios also need to be clarified about when reinitialization occurs (when field is cleared or form submitted) and where (client or server side).

stevepiercy (changes requested):
Hit me up for documenting this one. These concepts are not easy to document, even for native English speakers. Assuming the concepts are similar to the Pylons Project's Colander, I might be able to wordsmith the docs. See https://docs.pylonsproject.org/projects/colander/en/latest/api.html#schema-related

docs/source/blocks/editcomponent.md
tiberiuichim: I think I'm more used to the zope.schema concept, rather then colander.

https://github.com/zopefoundation/zope.schema/blob/89f09bb970bc14f81b724e81174aaaedb8853c3a/src/zope/schema/interfaces.py#L251-L261C9

The major difference to the zope.schema is that the UI is automatically updated based on the schema and user interactions, so it's more a "user interaction" issue rather then data validation

tiberiuichim:
@stevepiercy What can we do about the docs?

docs/source/blocks/editcomponent.md
```diff
@@ -127,6 +127,43 @@ import { Icon } from '@plone/volto/components';
 </SidebarPortal>;
 ```
 
+### Default values support for block schema forms
+
+In the block schema definition, you can use the two additional properties for
+each field: `default` and `missing`. The `default` value is used to populate
+the field with an initial value. The `missing` value should be used when the field
+should always have a value (as required).
+
```
stevepiercy, line 136: Off topic rant: offering suggestions for narrative documentation with arbitrary line breaks sucks. This is one good reason to use one sentence per line.

Anyway, here's my suggestion:

```
In the block schema definition, you can use the two additional properties for each field: `default` and `missing`.

The `default` value is used to set an initial value for the field.
The user may override this value or clear it.

The `missing` value must be a valid value which becomes the field's value to be submitted when the user clears the field.
With `missing`, the field is optional.
Without `missing`, the field is required.
```

docs/source/blocks/editcomponent.md
stevepiercy: OK, that makes a little more sense, I guess kinda like HTML inputs getting their attributes for `value` and `required` set and unset?

I would suggest the following, and I have further questions to help me understand and clarify the intent.

Let's make this section a definition list for ease of legibility.

```
`default` and `missing` are declared
:   The block form is initialized with the `default` value.
    When the user clears the field, the field is reinitialized with the `missing` value.
    This would imply a required field.

`default` only is declared
:   The block form is initialized with the `default` value.
    When the user clears the field, the field remains empty.
    This would imply an optional field.

`missing` only is declared
:   The block form is initialized with the `missing` value.
    When the user clears the field, the field is reinitialized with the `missing` value.

`default` and `missing` are not declared
:   The block form has no initialized value.
    The field is required.
```

Are "block form" and "field" the same thing or different? In my brain, a form is a collection of one or more fields (a schema) with additional attributes.

Also I am struggling with the term "reinitialized". Does that mean the value is shown to the user in the field and/or its HTML attribute `value` gets set to the `missing` value? If so, that seems weird to me that when a user clears a field, then its value is reinitialized. Maybe the value is what gets processed outside the client browser? Please help me understand.

stevepiercy (commented):
It might be easier to explain with a picture or video, too. Sometimes words just don't make it clear, even for developers, but especially for the end user.

tiberiuichim:
@stevepiercy many thanks for the suggestions, they're all really good. I'll try to incorporate them.

stevepiercy:
Can `missing` be an empty string, `None`, or `Null`?

I assume that "form block" and "field" are the same, so I would replace the former with the latter. I also would replace "re/initialized" with "set".

```
`default` and `missing` are declared
:   The field's value is set to the `default` value.
    When the user clears the field, the field is set to the `missing` value.
    This would make the field required.
```

I think using a code example of what actually happens might help, too. For example, if I understand correctly:

````
Let's say that `default=https://default.com` and `missing=https://missing.com`.
When the field is first displayed to the user, and if the field is an HTML text input, its HTML source would render as follows.

```html
<input name="url" value="https://default.com">
```

This would be displayed to the user as follows.

<input name="url" value="https://default.com">

When the user edits the field, but does not clear it, then the value entered by the user is used for the field.
If the user clears the field, then the field's value will be set to the `missing` value.

```html
<input name="url" value="https://missing.com">
```

This would be displayed to the user as follows.

<input name="url" value="https://missing.com">
````

This would render in the docs as shown in the screenshot.

Screen Shot 2023-03-02 at 11 50 26 PM

I am still not sure about how `missing` would actually work, but that's my interpretation at this moment. Please correct me if I am mistaken.

**Comment by tiberiuichim** [comment]
@stevepiercy You got it correct. That's the behavior. I'll update the docs with your suggestions. Many thanks!

**Comment by sneridagh** [comment]
@tiberiuichim is this ready?

File: docs/source/blocks/editcomponent.md [hunk 5]
```diff
@@ -127,6 +127,43 @@ import { Icon } from '@plone/volto/components';
...
+      default: 'https://example.com',
+      missing: 'https://nothing.com',
+    },
+  }
+  // ...
+```
+
+In effect, the following situation matrix will occur:
+
+- `default` and `missing` are declared: the block form is initialized with
```
nileshgulia1, line 151: Nice. I just gave it a try on local vanilla volto. When both the values "default" and "missing" are used, it somehow re-initializes the form field with default value. Here's the schema snippet I created for demo field:
```
demo: {
        title: 'demo',
        widget: 'url',
        default: 'www.default.com',
        missing: 'www.missing.com',
      },

 ```
 
  In addition to this, idk if its a glitch or the way it works, but the values let it be "missing" or "default" started to re-appear after clearing them and editing the form second time🤔

Comment (tiberiuichim) [reply docs/source/blocks/editcomponent.md:151]: @nileshgulia1 that's the effect of setting `missing`. If you don't want that effect, don't set the `missing`.

Btw, this is also the default with current Volto. Try it with this widget:

```
{
        title: 'demo',
        widget: 'url',
        default: 'www.default.com',
      },
```

nileshgulia1: This is how the default behaviour works:

https://user-images.githubusercontent.com/22280901/226318877-30578d6a-51c9-4ba9-b172-4a671d2c9898.mp4

sneridagh:
@tiberiuichim Do you think this is ready?

@davisagli could you please give it a try and review it?

Marked for review during next week sprint.

PR #4471, opened 2023-03-06
Fix history view dropdown for first entry, showing 'Revert to this version option' always

sneridagh:


src/components/manage/History/History.jsx
```diff
@@ -116,22 +116,46 @@ class History extends Component {
...
 
-  /**
-   * Render method.
-   * @method render
-   * @returns {string} Markup for the component.
-   */
-  render() {
+  processHistoryEntries = () => {
+    // Getting the history entries from the props
+    // No clue why the reverse(concat()) is necessary
```
davisagli, line 121: I guess someone just didn't want to figure out how to write a for loop to go through the entries in reverse order?

src/components/manage/History/History.jsx
```diff
@@ -116,22 +116,46 @@ class History extends Component {
...
+        return item;
+      }
+
+      if (item.type === 'versioning' && mark) {
+        return { ...item, show_revert_button: true };
+      }
+
+      return item;
+    });
+  };
```
davisagli, line 146: You could use lodash.find to add `is_current: true` to the first item that has `type: 'versioning'` instead of adding `show_revert_true` to all the others. And then check for `is_current` below.

But, this is good enough.

davisagli (approved):

src/components/manage/History/History.jsx
sneridagh: @davisagli refactored using `find`, I thought it would return all the matching elements. Thanks for the suggestion!

PR #4472, opened 2023-03-06
Fix order of row of long table  in edit and view mode.

### LINKED ISSUE #4473: Fix order of row of long table in edit and view mode
When you have a very large table block in Volto we are getting different rows compared to edit mode.

iFlameing:
If we have a very large table in Volto. In view mode when we are iterating using an object.Keys it is giving us the wrong order of table rows. Because object. keys don't give us keys in order of properties but it is different, which causes different rows in the edit as compared to the view.

Fixes #4473


tiberiuichim (approved):

davisagli:
@iFlameing When you open a PR that is related to an issue, please mention the issue like this in the PR description: "Fixes #4473". Then GitHub will automatically close the issue when the PR is merged.

davisagli (approved):

PR #4497, opened 2023-03-10
Make Drag and Drop list work with container-type inline-size.

robgietema:


sneridagh (approved):
LGTM, merging for now.

Issue #3553, opened 2022-08-05
Datetime value is not correctly rendered in DefaultView


erral:
**Describe the bug**

**To Reproduce**

1. User time zone: "Europe/Madrid"
2. We create a new content-type called "Tender" with a Datetime field 
3. We add a "tender" in the site, we set the "datetime" field to be "2022-08-05 10:00 am"
4. The default view shows "2022-05-05 08:00 am":
5. Example: https://6.demo.plone.org/tender-1

**Expected behavior**
We expect the shown datetime to be "2022-05-05 10:00 am"

It looks like the "datetime view widget" does not handle the timezone of the user in the same way that the edit widget does:
- View Widget: https://github.com/plone/volto/blob/master/src/components/theme/Widgets/DatetimeWidget.jsx
- Edit Widget: https://github.com/plone/volto/blob/master/src/components/manage/Widgets/DatetimeWidget.jsx

PR #4516, opened 2023-03-14
3553 datetime value

tedw87:
https://github.com/plone/volto/issues/3553
Formated the UTC timezone to local time in `DatetimeWidget` view component.


tiberiuichim (changes requested):
moment should be injected as a lazy library

sneridagh:
@tedw87 can't be done with vanilla JS?

https://dockyard.com/blog/2020/02/14/you-probably-don-t-need-moment-js-anymore

erral:
I haven't tested if the implementation fulfills this but either we go with moment or vanilla JS, please be aware that user's browser locale and application's locale do not always match.

I can have my browser in Spanish but browse an English page and expect dates to be in English locale.

avoinea:
I think this is obsolete. Closing.

avoinea:
Fixed by #8236

PR #4517, opened 2023-03-14
Fix devproxy vh path when the public URL port is null

davisagli:
If volto is being accessed without an explicit port (i.e. default port 80 for http or 443 for https) then apiPathURL.port is null, and we should not include it in the virtual host path sent to the backend.

### DIFF
--- a/src/express-middleware/devproxy.js
+++ b/src/express-middleware/devproxy.js
@@ -73,11 +73,12 @@ export default function () {
     },
     pathRewrite: (path, req) => {
       const { apiPathURL, instancePath } = getEnv();
+      const targetPort = apiPathURL.port ? `:${apiPathURL.port}` : '';
       const target =
         config.settings.proxyRewriteTarget ||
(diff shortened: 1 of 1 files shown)


sneridagh:
LGTM

However, I think that we should  have a way to know that the internal proxy is being used in production, when it shouldn't... mark the response with a header?

davisagli:
@sneridagh Or just make sure the way to explicitly disable it works?

sneridagh:
Yeah, maybe even better. An especific env var for that?

davisagli:
@sneridagh I'd say we should make sure that explicitly setting RAZZLE_DEV_PROXY_API_PATH works to disable it like it says in this comment (instead of adding yet another variable): https://github.com/plone/volto/blob/b5bafd5b2be339539f1a384dbe7137235b2506ca/src/config/index.js#L94

Currently the comment is wrong.

davisagli:
This was fixed in https://github.com/plone/volto/pull/7967

PR #4520, opened 2023-03-15
Combined dependency updates

davisagli:
@sneridagh Here's a combined version of all the pending dependency updates submitted by dependabot.


sneridagh (approved):
Thanks a lot!

sneridagh:
@davisagli I guess it also makes sense to prepare one for 16, right? without the specifics (webpack)

PR #4524, opened 2023-03-15
Apply suggestion from browser for password field

### LINKED ISSUE #3990: Apply suggestion from browser for password field
Volto 16.0.0,  6.demo.plone.org

image

lord2anil:
closes #3990

### DIFF
--- a/news/3990.bugfix
+++ b/news/3990.bugfix
@@ -1,0 +1,1 @@
+Apply suggestion from browser for password field @lord2anil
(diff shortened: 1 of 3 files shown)


davisagli:
@stevepiercy No, I don't think it does. It's a hint to the password manager (including one which may be built in to the browser) for which field would be the right one to fill in the current password. The policy decision about whether to allow that to happen would be done elsewhere (in configuration for the browser or password manager), not at the level of a specific site.

PR #4541, opened 2023-03-15
March dependabot dependencies update for 16

sneridagh:


package.json
```diff
@@ -417,6 +416,7 @@
     "identity-obj-proxy": "3.0.0",
     "jest": "26.6.3",
     "jest-environment-jsdom": "^26",
+    "jsonwebtoken": "9.0.0",
```
davisagli, line 419: @sneridagh A major semver update for jsonwebtoken and moving it to devDependencies are not backwards compatible, if someone was using it downstream without putting it in package.json explicitly. I would leave this one out of the backport.

packages/generator-volto/package.json
```diff
@@ -108,7 +108,6 @@
     "fs-extra": "3.0.0",
     "gitly": "2.0.3",
     "mkdirp-then": "1.2.0",
-    "ms": "1.0.0",
```
davisagli, line 111: @sneridagh Also technically not backwards compatible, but it's very unlikely that someone is extending the generator so I wouldn't worry about it.

davisagli (changes requested):
@sneridagh Thanks for preparing this. I have a couple doubts about what to include in the backport.

sneridagh:
@davisagli ok, let's wait for this then.

davisagli:
@sneridagh I think it's important to do the updates which are just patch upgrades though, because there are some security fixes.

davisagli (approved):
@sneridagh I think this is ready now

PR #4545, opened 2023-03-15
removed: is-url dependency

### LINKED ISSUE #4543: Remove `is-url` dependency
**Remove _is-url_ dependency**

A dependency can be removed by adding some regex expressions.
This library is checking for the passed URL is a correct URL or not!

My approach is to remove this dependency and add a file with same functionality by writing some regex expressions.

sudhanshu1309:
A dependency named is-url is removed by adding some regex expression in file isUrl.js
closes #4543

### DIFF
--- a/news/4543.breaking
+++ b/news/4543.breaking
@@ -1,0 +1,1 @@
+Added new file for the checking url. Removed `is-url` dependency. @sudhanshu1309
--- a/package.json
+++ b/package.json
@@ -299,7 +299,6 @@
     "image-extensions": "1.1.0",
     "immutable": "3",
     "is-hotkey": "0.2.0",
-    "is-url": "1.2.4",
     "jest-file": "1.0.0",
     "jwt-decode": "2.2.0",
     "less": "3.11.1",
--- a/packages/volto-slate/src/blocks/Text/extensions/isUrl.js
+++ b/packages/volto-slate/src/blocks/Text/extensions/isUrl.js
@@ -1,0 +1,48 @@
+/**
+ * Expose `isUrl`.
+ */
+
+module.exports = isUrl;
+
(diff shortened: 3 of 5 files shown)


davisagli (commented):
Note: This code is copied from `is-url`. Why do you propose moving it here instead of including `is-url` as a dependency? In general things like parsing URLs is tricky and I would prefer to use a separate library that is more likely to get bugfixes reported from a broader user base than just the Plone community. Although, it does look like there have been no releases of `is-url` recently, so maybe that point doesn't matter much.

davisagli:
@sudhanshu1309 Also, I don't appreciate that you submitted code from another source as if it were your own, without attributing the source. This is not allowed by the Plone contributor agreement. Any code you submit to Plone needs to either be code to which you own the copyright (you wrote it) or code with a license that allows it to be copied elsewhere. This code has such a license, but it specifies that the license must be included with copies of the code, which you did not do.

sudhanshu1309:
@davisagli  I'm quite new to open source, so I wasn't aware of licenses or how they work. I'll keep this in mind from this point onward. My objective was to reduce the dependency on legacy code and ensure that this is maintained and updated in accordance with Plone's requirements. I'll be sure to not repeat this.

sudhanshu1309:
@davisagli should I close this PR or do I need to add a license?

davisagli:
@sudhanshu1309 Okay, thanks for trying to contribute to Plone!

A dependency on code that hasn't been updated in some time is not _necessarily_ a problem, unless that code has issues which aren't being addressed and are causing problems in Plone. For that reason, I will close this PR but we can consider this in the future if there is a specific problem that needs to be addressed in the `is-url` code.

PR #4546, opened 2023-03-15
Update simple-git

davisagli:
And remove it as a direct dependency of `@plone/scripts`, because it is not used directly.

### DIFF
--- a/news/4546.breaking
+++ b/news/4546.breaking
@@ -1,2 +1,0 @@
-Remove dependency of `@plone/scripts` on `simple-git`. It is used by
-`mrs-developer` but not directly. @davisagli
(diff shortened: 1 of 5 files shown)


sneridagh (approved):

sneridagh:
@davisagli Since this is a change in `@plone/scripts` we use to place the changelog in the package.

davisagli:
@sneridagh Oops, I hadn't noticed there is a separate one. Makes sense, I moved it.

davisagli:
@sneridagh Why does it fail the towncrier check now?

sneridagh:
@davisagli my bad... I didn't anticipate this for inner packages changelogs.

PR #4547, opened 2023-03-15
Update parse-url

davisagli:
(and in order to do that, make sure all packages are using the same up to date version of release-it)


sneridagh (approved):
@davisagli somehow this got lost, can I merge it?

davisagli:
@sneridagh Yes. Last time we talked about it we punted it because it is unclear whether it should be a bugfix or breaking, and we ran out of time. It includes a major update to `release-it`, but that is a dev dependency, so maybe it's still fine for volto 16?

davisagli:
This happened in the meantime

PR #4558, opened 2023-03-16
(fix):Object.normaliseMail: Cannot read properties of null

dobri1408:
I found out that if somehow it happens that a link has a null value, it can be caused by a faulty copy-paste, this error can happen. This is due to the fact that the "email"/"telephone" is null or "res.url" is null.
Captură de ecran din 2023-03-16 la 14 54 56

### DIFF
--- a/src/helpers/Url/Url.js
+++ b/src/helpers/Url/Url.js
@@ -264,14 +264,14 @@ export function isTelephone(text) {
 }
 
 export function normaliseMail(email) {
-  if (email.toLowerCase().startsWith('mailto:')) {
+  if (email?.toLowerCase()?.startsWith('mailto:')) {
     return email;
   }
   return `mailto:${email}`;
 }
 
 export function normalizeTelephone(tel) {
-  if (tel.toLowerCase().startsWith('tel:')) {
+  if (tel?.toLowerCase()?.startsWith('tel:')) {
     return tel;
   }
   return `tel:${tel}`;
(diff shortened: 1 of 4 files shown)


tiberiuichim (approved):
Would it be possible to add this test case in the jest test?

dobri1408:
Hello @tiberiuichim, sure, thanks!

dobri1408:
Hi, I added the tests. Please, let me know if anything else should be changed or added.

src/helpers/Url/Url.js
```diff
@@ -300,6 +300,7 @@ export function checkAndNormalizeUrl(url) {
         res.isValid = false;
       }
     }
+    if (res.url == undefined || res.url == null) res.isValid = false;
```
tiberiuichim:
```suggestion
    if (typeof res.url === "undefined" || res.url === null) res.isValid = false;
```

tiberiuichim: Although I see that with modern JS it's possible to compare something to `undefined`.

tiberiuichim:
@dobri1408 please use triple equality operator

tiberiuichim:
I've had to make this customization to make it work: https://github.com/eea/volto-cca-policy/blob/d92a80db622999357310ec9196ee30734789c3c0/src/customizations/volto/components/manage/AnchorPlugin/components/LinkButton/AddLinkForm.jsx

dobri1408:
I fixed the error that caused the internal links to not work anymore.

sneridagh (approved):

Issue #4550, opened 2023-03-16
Should use location.assign rather then location.replace


tiberiuichim:
https://github.com/plone/volto/blob/15b83f26e21b87acaf058fca0ac6016aa3b06f22/src/middleware/blacklistRoutes.js#L21

https://stackoverflow.com/questions/4505798/difference-between-window-location-assign-and-window-location-replace

saadkhan011:
Hi tiberiuichim,
     I want to work on this issue can you assign me this issue so I can start.
Thanks

Krishpanchal:
Hi tiberiuichim, I can work on this issue. You can assign this issue to me.
Thanks

Darkcoder011:
i think You Don't Do that Because 
replace :- will replace the current document and replace the current History with that URL
assign:- will just cause a new document to load

@tiberiuichim

Coder-aadarsh:
Generally location.assign is typically used to navigate to a new URL, while location.replace is used to replace the current URL with a new one.

If the goal is to allow the user to navigate to a new page while preserving the current page in their browser history, then location.assign may be more appropriate. If the goal is to replace the current page in the browser history with a new one, then location.replace may be more appropriate i guess

tiberiuichim:
honestly, I don't understand the difference. I'm the one that wrote the original code, `window.replace()`.

What is the impact for the browser history? What's the practical difference? As far as we can say, window.replace worked fine till now.

Darkcoder011:
the  difference is assign is used to navigate new url but the difference replace the current URL with a new one

tiberiuichim:
@Darkcoder011 but from a practical point of view, what happens?

Because even with window.replace, the back button works in the browser. So, it it a case that the last entry from the "list of urls I've visited" is removed and replaced with the new one? And if I use assign, that doesn't happen?

Darkcoder011:
@tiberiuichim  yes both redirect to new page .
they both do the same job but in difference manner and both have a difference 

when we use assign they can load a new page  

but when we use replace they directly replace that page 

you can see below 
for assign:-
document.location.href = 'http://frontbackend.com'; // go to  homepage
document.location.assign('http://frontbackend.com/Frontend'); // loads a new page

for replace :- 
document.location.href = 'http://frontbackend.com'; // go to  homepage
document.location.replace('http://frontbackend.com/Frontend');  // replace to another webpage

Darkcoder011:
[...]

PR #4561, opened 2023-03-16
fix: Should use location.assign rather then location.replace

suman9893:
fix: #4550

### DIFF
--- a/news/4550.internal
+++ b/news/4550.internal
@@ -1,0 +1,1 @@
+fix: Should use location.assign rather then location.replace . @suman9893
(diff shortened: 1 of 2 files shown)


tiberiuichim:
I wonder if you could also add a cypress test for this? It would be awesome to get this functionality tested. Add the relevant configuration in the core-sandbox package, it's loaded with the tests: https://github.com/plone/volto/blob/48d7d62e8c5f122f6b6cae2860650e3547187ac8/packages/coresandbox/src/index.js#L123-L131

davisagli:
Old, outdated PR. The existing code was correct.

PR #4569, opened 2023-03-18
Fix broken links at ReactJS

stevepiercy:


davisagli (approved):
Annoying as it might be to have to make updates like this, it's pretty cool that we can catch and fix them so quickly. React just launched their new docs within the past few days.

Issue #4018, opened 2022-12-02
Use the Plone schema to build the user add form


wesleybl:
Today, Volto build the form without using the Plone schema:

https://github.com/plone/volto/blob/7f5bf0248ce60a3a23c57f179e0f1c8df1611b91/src/components/manage/Controlpanels/Users/UsersControlpanel.jsx#L412

It would be good to use the Plone schema, used to assemble the Classic Plone form:

https://github.com/plone/plone.app.users/blob/bc0699f70edf8beadebb1b612354327359e2d548/plone/app/users/schema.py#L123

PR #4572, opened 2023-03-18
User Control Panel improvements

### LINKED ISSUE #4551: User control panel improvements
## Abstract

We need some changes in the current Volto user controlpanel to be able to handle users efficiently or at least like it can be done in ClassicUI.

## Motivation

The current user control panel in Volto requires some work to make it more user friendly. For instance it doesn't show any indication when one is searching for users, there is no way to access user information (click and see all user properties) or there is no indication on which is the user login name

## Assumptions

We need to unify the user handling UI and UX to the one existing in ClassicUI.

## Proposal & Implementation

- When "many users" is enabled the User control panel should not show the table headings, but just the search form like ClassicUI does.
- The user controlpanel should show a spinner when the search is being made and no results are present yet.
- The user listing should show not only the user fullname but also the user login name like ClassicUI does.
- The user controlpanel should link each user to a page where all user data should be shown and be modifiable if the current logged in users has the required permissions.
- When any change is done in the User/Role matrix, an indicator should mark that changes are not saved yet (as opposed to the user-group controlpanel were changes are applied immediately when they are made)

## Deliverables

New components will be prepared to handle the user data form, and changes will be made to existing components to match the requirements.

## Risks

I don't see any risks here.

erral:
fixes #4551 and #4018

This PR contains the following functionalities identified in #4551:

- Show the username next to the fullname in the user controlpanel
- Don't show the user table headings when many_users setting is true and no search have been made.
- Show "no results found" when doing a search in the user controlpanel and there are no results.
- Don't show the group table headings when many_groups setting is true and no search have been made.
- Show "no results found" when doing a search in the group controlpanel and there are no results.

And the following identifed in #4018
- Add a button to Edit the user data in a ModalForm following the userschema configured in the back-end


erral:
> 



> > Show "no results found" when doing a search in the user controlpanel and there are no results.
> > Show "no results found" when doing a search in the group controlpanel and there are no results.
> 
> I see the string "undefined" instead of "no results found"

Ups, I will check it.

> 
> I think this PR still doesn't close #4018. I would expect all field names and descriptions to come from the backend. But this can be done in another PR.

The `username`, `password`, `roles` and `groups` fields of the edit form are not part of the userschema, so we need something else in the REST API to handle them in the back-end.

wesleybl:
In classic Plone, when adding a user, fields like Home page, Biography, Location and Portrait are not shown. Here it is being shown. This is a problem?

erral:
> In classic Plone, when adding a user, fields like Home page, Biography, Location and Portrait are not shown. Here it is being shown. This is a problem?

There is a setting in Classic Plone which sets where a field should be available in the user add form and/or in the personal information.

I would say Volto should obey that in the Add and Personal Information form, and show all fields in this specific edit form which will be handled by a Manager anyway, right?

wesleybl:
The problem is that I think these fields don't make sense when adding users. These are very personal user fields, the user is added by another.

> There is a setting in Classic Plone which sets where a field should be available in the user add form and/or in the personal information.

So, I think the Volto should also respect this configuration.

erral:
I have fixed the `undefined`.

The different fieldset for registration and user information form is not supported right now by the  `@userschema` endpoint. This should be handled differently.

wesleybl:
> The different fieldset for registration and user information form is not supported right now by the @userschema endpoint.

It would be nice to open an issue for this.

src/components/manage/Controlpanels/Users/RenderUsers.jsx
```diff
@@ -67,7 +111,20 @@ class RenderUsers extends Component {
...
+          ({this.props.user.username})
+          {this.props.userschema && (
+            <>
+              <br />
+              <Button
+                onClick={() => {
+                  this.onClickEdit({ formData: this.props.user });
+                }}
+              >
+                {this.props.intl.formatMessage(messages.edit)}
```
nileshgulia1, line 124: A personal perspective: Can we move this button inside the options/Dropdown component for individual entry? Or maybe use a "linkedButton" like plone classic UI?

src/components/manage/Controlpanels/Users/RenderUsers.jsx
```diff
@@ -101,9 +158,31 @@ class RenderUsers extends Component {
             </Dropdown.Menu>
           </Dropdown>
         </Table.Cell>
+        {Object.keys(this.state.user).length > 0 &&
+          this.props.userschema.loaded && (
+            <ModalForm
```
nileshgulia1, line 163: I think the current `userschema` do respect the "Member fields" addition/deletion. So it's possible to populate with identical values both in user reg and user info. Perhaps those fields in the personal-information should be the same as user-info(clicking the edit button).

src/components/manage/Controlpanels/Users/RenderUsers.jsx
```diff
@@ -56,6 +66,40 @@ class RenderUsers extends Component {
...
+    // Do not handle groups and roles in this form
+    delete data.groups;
+    delete data.roles;
+    this.props.updateUserData(data.id, data);
+    this.setState({ user: {} });
+    this.props.listUsers();
+    return toast.success(
+      <Toast
+        success
+        title={this.props.intl.formatMessage(messages.success)}
```
nileshgulia1, line 80: I tried to change the full name of existing users via-edit button and the Toast message returns "undefined" somehow while the rest functionality works fine.

nileshgulia1 (commented):
A bit late to the party, Thanks @erral  for working on this. The user/groups controlpanel is huge and has been drastically improved by the efforts like this!

src/components/manage/Controlpanels/Users/RenderUsers.jsx
ionlizarazu: I've moved the edit button inside the actions @nileshgulia1

ionlizarazu (approved):

avoinea (approved):

nileshgulia1 (approved):

src/components/manage/Controlpanels/Users/RenderUsers.jsx
```diff
@@ -56,6 +67,40 @@ class RenderUsers extends Component {
...
+        content={this.props.intl.formatMessage(messages.updateUserSuccess)}
+      />,
+    );
+  }
```
ksuess, line 85: The change should be reflected after submit. Therefore the `listUsers` should be done after the update.
```suggestion
  onEditUserSubmit(data, callback) {
    // Do not handle groups and roles in this form
    delete data.groups;
    delete data.roles;
    this.props
      .updateUserData(data.id, data)
      .then((resp) => {
        this.setState({ user: {} });
        this.props.listUsers();
      })
      .then(() => {
        toast.success(
          <Toast
            success
            title={this.props.intl.formatMessage(messages.success)}
            content={this.props.intl.formatMessage(messages.updateUserSuccess)}
          />,
        );
      });
    return;
  }
```

ksuess: A nice to have would be: If a searchstring is given, then take it into account for `listUsers`.

ksuess (changes requested):
The change of user data should trigger an update of the list of users.

src/components/manage/Controlpanels/Users/RenderUsers.jsx
erral: I managed to do it (I think) :wink:

ksuess: Great that you found a way. Glad you took the time to figure out. Merci! Works fine now.

ksuess (approved):

nileshgulia1:
I am eagerly anticipating the merger!

erral:
If you need me to review something, I will be off until at least 11th of August

PR #4574, opened 2023-03-18
Update message Add-on control panel

ksuess:


erral (approved):

ksuess:
By now just english provides a detailed help message on installing backend addons.
Feel free to update your language translations.

I found this issue while updating my training.

erral:
Why is the default message so long and the existing translations so short? If the defaultMessage is such long, the translations should have been translations of that defaultMessage, right?

It's just a curiosity, because I see that the original defaultMessage is really longer than all the existing translations...

davisagli (approved):
@erral The existing short translations may have been added before the defaultMessage was included in the `.po` files -- for example it was not there when you added the Spanish translation in https://github.com/plone/volto/commit/b944a7d4880f5d50c69472c78ec775d4d6adb067#diff-631af0849c562ae85096b429d9c5d16c5d0ddde5aa14030bd9853f2c083da79cR64

The changes here look good to me.

erral:
Yes I am not saying that the changes are not good. It was just because when using any translation helping tools, usually they offer an option to do the translation based on the default message this was strange. Just that.

We can also mark the translations as fuzzy, which is what i18ndude does when the default message is changed, but that invalidates the translations usually, at least in Plone.

ksuess:
Eliminating the hint on buildout. Thanks for reviewing and approving.

The translations should be updated too. I will propose translations for german. Please check if an update for your languages could be reasonable too.

PR #4583, opened 2023-03-21
[WIP] Render span elements; wrap text inside lists in spans

tiberiuichim:


tiberiuichim:
[vokoscreenNG-2023-03-22_12-23-34.webm](https://user-images.githubusercontent.com/141568/226882697-b49a28e5-9386-484c-8342-e9584d6a9042.webm)

So the solution is to automatically fix the structure to the desired state in the normalizeNode.

tiberiuichim:
@davisagli Somehow the latest slate doesn't play well with cypress. See this PR, with just the lib upgrade, where a lot of tests fail: https://github.com/plone/volto/pull/4593

This was a solution suggested on the web somewhere.

packages/volto-slate/src/blocks/Table/deconstruct.js
```diff
@@ -19,7 +19,7 @@ import {
 export function syncCreateTableBlock(rows) {
   const id = uuid();
   const block = {
-    '@type': 'table',
+    '@type': 'slateTable',
```
tiberiuichim, line 22: I have no idea why this wasn't flagged before.

sneridagh: ups

cypress/support/e2e.js
```diff
@@ -12,6 +12,16 @@ before(function () {
 
 beforeEach(function () {
   cy.log('Setting up API fixture');
+
+  // avoid a mysterious test failure with upgrade to slate-react 0.91.4
+  const resizeObserverLoopErrRe = /^[^(ResizeObserver loop limit exceeded)]/;
```
sneridagh, line 17: wut? T_T

package.json
```diff
@@ -379,9 +379,9 @@
     "semantic-ui-react": "2.0.3",
     "semver": "5.6.0",
     "serialize-javascript": "3.1.0",
-    "slate": "0.84.0",
-    "slate-hyperscript": "0.81.3",
-    "slate-react": "0.83.2",
+    "slate": "0.91.4",
```
sneridagh, line 382: Could this affect existing sites in any way?

sneridagh (approved):
LGTM

package.json
tiberiuichim: @sneridagh no, I don't think so. 

The slate upgrade alone makes our tests fail, I have one PR with only that change. So, regardless of anything, if we want to make the upgrade, we need to do the test fix.

tiberiuichim:
@sneridagh don't merge yet, I need to run more tests

PR #4597, opened 2023-03-22
Refactor: Delete Class Component from Delete.jsx File

TajAlasfiyaa:
Hey team,

I've just submitted a pull request that deletes a React class component and replaces it with a function component. This change should help simplify our codebase and make it easier to maintain going forward.

Function components are generally more lightweight and easier to reason about than class components, so this should be a good improvement overall.

Please review the changes and let me know if you have any feedback or concerns. Thanks for your attention!

Best,
TajAlasfiyaa Ishag

### DIFF
--- a/src/components/manage/Delete/Delete.jsx
+++ b/src/components/manage/Delete/Delete.jsx
@@ -1,9 +1,9 @@
+
 /**
  * Delete container.
  * @module components/manage/Delete/Delete
  */
-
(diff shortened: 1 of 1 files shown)


sneridagh:
@TajAlasfiyaa Thanks for your contribution, since you didn't sign the contributor agreement we cannot go forward. Also, we will take care of this as a whole in the upcoming GSOC project.

PR #4598, opened 2023-03-22
DefaultView (blocks disabled): Show field name as tip on hover of label

ksuess:


tiberiuichim (approved):

davisagli (commented):
This might be okay for editors but I don't think it's something I would want to display in a tooltip for anonymous viewers.

tiberiuichim:
@davisagli that view is very rough. If you're exposing end users to it, you either have bigger problems, or they're the type of users that don't mind that info, or they find it useful.

davisagli:
@tiberiuichim I guess you are probably right in most cases.

sneridagh:
I agree that this view is kind of a rough fallback view, ideally an integrator should always provide a better tailored one.

PR #4605, opened 2023-03-24
fixed the Table background color when inverted and basic active at same time

### LINKED ISSUE #4603: Table block: "minimalistic design" doesn't work together with "table color inverted"
The text color is changed to white, but the background color is not changed. Probably the "minimalistic design" hardcodes background color?

Latest Volto master.

lord2anil:
closes #4603


tiberiuichim (approved):

stevepiercy:
@lord2anil can you open a PR on a branch in the volto repo? GitHub Actions will only run on branches inside the Plone organization. @fredvd @sneridagh would you please take a look at permissions for @lord2anil to open a PR in volto? We might need to make an adjustment.

fredvd:
@stevepiercy   It's primarily a workflow issue I think. The towncrier checks  for example have run for lord2anil's Pull Request from his own workspace, because the  on: definition explicitly mentions pull_request.  The other Volto workflows are registered  on push: for a branch on the main repo.  @sneridagh We need to take a look at the workflows so that the necessary ones like unit tests are also running on pull requests. .

JeffersonBledsoe:
@lord2anil Thanks for the contribution! On testing this PR, I've noticed a couple of issues:

- The 'black' colour doesn't match that of 'inverted'. This could cause an issue if the themeing system has been used to change the colour of 'inverted'.
  - I would suggest solving this in a `.less` file while you can check for `.inverted.basic` and read from the `@invertedBackground` and [other inverted table variables](https://github.com/Semantic-Org/Semantic-UI-LESS/blob/806bd67fcc445ce254a55a7dd153dae408f4485d/definitions/collections/table.less#L957-L961)
- The current fix only applies to edit mode and so it breaks again in view mode. The above suggestion about using the .less files should solve this.

Are you able to make these changes and we can work towards getting this PR merged?

news/4603.bugfix
```diff
@@ -0,0 +1 @@
+fixed the Table background color when inverted and basic active at same time @lord2anil
```
stevepiercy, line 1:
```suggestion
Fixed the table background color when its `inverted` and `basic` styles are active at the same time. @lord2anil
```

stevepiercy (changes requested):

lord2anil:
Thanks for the feedback. Due to time constraints, I cannot implement the changes right now.

ichim-david:
The fix isn't what is needed. Due to the time constraints of the author I am closing this pull request, we can open a new one with the solution proposed by @JeffersonBledsoe

PR #4629, opened 2023-03-31
Trigger CI on pull_request event

davisagli:
@fredvd @sneridagh @mauritsvanrees This is what I would prefer instead of #4621. I tried it in https://github.com/plone/volto/pull/4628 and it seems to work.

### DIFF
--- a/.github/workflows/acceptance.yml
+++ b/.github/workflows/acceptance.yml
@@ -1,5 +1,5 @@
 name: Acceptance Tests
-on: [push]
+on: [push, pull_request]
 jobs:
   core:
     runs-on: ubuntu-latest
--- a/.github/workflows/acceptance.yml
+++ b/.github/workflows/acceptance.yml
@@ -2,6 +2,7 @@ name: Acceptance Tests
 on: [push, pull_request]
 jobs:
   core:
+    if: github.event_name != 'pull_request' || github.event.pull_request.head.repo.full_name != github.event.pull_request.base.repo.full_name
     runs-on: ubuntu-latest
     name: Core Basic
     timeout-minutes: 45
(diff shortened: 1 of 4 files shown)


fredvd (approved):
approved, as discussed on #4621 we should only add on: pull_request to avoid security issues.

davisagli:
@wesleybl Good point! I think this is what prevents it: https://github.com/zopefoundation/Zope/blob/master/.github/workflows/tests.yml#LL47C5-L47C11

Let me try that...

davisagli:
Hmm, it works, but the duplicate jobs for the pull_request trigger still show up in the list of checks as "skipped", which is a bit noisy.

davisagli:
An alternative would be to limit builds for the `push` event to only key branches (`master`, `16.x`, etc). I think the only downside is that if someone pushes a commit to a work-in-progress branch that doesn't have a PR yet, the CI jobs would not run until they open a PR. Maybe that is actually what we want, to avoid unnecessary builds...

fredvd:
> Hmm, it works, but the duplicate jobs for the pull_request trigger still show up in the list of checks as "skipped", which is a bit noisy.

Yes, I have the same in another workflow where normal pushes trigger something for a testing environment and tag releases update the live environment. The best I could do there was look at github.ref_type to cancel al jobs for on: push, but they still show up in the actions run list.

wesleybl:
> An alternative would be to limit builds for the push event to only key branches (master, 16.x, etc).

I usually don't like that, because I want to run the tests, before opening a PR. Usually when I open the PR the tests are already passing.

mauritsvanrees (commented):
I approve as well.

PR #4637, opened 2023-04-03
fixed Responsiveness issue in login page

SayanDeveloper:
### Before:
oldPlone

### After:
newPlone

@fredvd @stevepiercy


news/3250.bugfix
```diff
@@ -0,0 +1 @@
+fixed Responsiveness bug in login page
```
stevepiercy, line 1: Give yourself credit!

```suggestion
Fixed responsiveness bug in login page. @SayanDeveloper
```

stevepiercy (changes requested):
This PR requires review by a core developer before being merged. I requested review from a few, as it requires write permission to do so (and the Contributors group does not have that access).

SayanDeveloper:
alright sir :-)

stevepiercy (approved):

theme/themes/pastanaga/extras/main.less
```diff
@@ -35,6 +35,10 @@ body.cms-ui .ui.basic.segment.content-area {
   cursor: text;
 }
 
+#page-login > .container {
+  margin: auto !important;
```
tiberiuichim, line 39: I'm not comfortable with the `!important` here. Could you point to the setting that it's trying to "fix"? Why is the container working in the other pages?

SayanDeveloper: because these stylings are getting inherited from semantic.less (which is not expected)
image

So i have done the override of styling for this specific login page container by specifying in ```#page-login > .container``` selector with !important

davisagli:
Already fixed

Issue #4638, opened 2023-04-03
sitemap.xml.gz needs to be split for Google


reebalazs:
The sitemap.xml.gz file can only contain 50000 items and be a maximum of 50k. For sites with a lot of content, the current xml is rejected by Google.

```
Sitemap size limits: All formats limit a single sitemap to 50MB (uncompressed) or 50,000 URLs. If you have a larger file or more URLs, you must break your sitemap into multiple sitemaps. You can optionally create a [sitemap index](https://developers.google.com/search/docs/crawling-indexing/sitemaps/large-sitemaps) file and submit that single index file to Google.
```

Reference: https://developers.google.com/search/docs/crawling-indexing/sitemaps/build-sitemap#:~:text=Sitemap%20size%20limits%3A%20All%20formats,your%20sitemap%20into%20multiple%20sitemaps.

reebalazs:
I'm working on this, PR in progress until we test it on another site.

ale-rt:
Maybe this is interesting for you: https://github.com/collective/collective.splitsitemap

tisto:
@reebalazs did we implement this feature in Volto core? I have an issue with plone.de where we the sitemap-index.xml is outdated and the sitemap.xml.gz is more recent. Both versions are around and that highly confuses Google.

PR #4639, opened 2023-04-03
Generate a split sitemap (also fix robots.txt)

reebalazs:
`sitemap.xml.gz` remains identical.

`sitemap-index.xml` can be used instead as an index file, which will link to `sitemap1.xml.gz`, `sitemap2.xml.gz`, ...

The default index size is 2000 which also considers the max file size to remain under Google's limit. (50k)

Also:

- Fix robots.txt to not contain an internal link

### DIFF
--- a/news/4638.bugfix
+++ b/news/4638.bugfix
@@ -1,0 +1,1 @@
+Generate a split sitemap @reebalazs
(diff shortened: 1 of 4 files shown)


reebalazs:
Fixes #4638

reebalazs:
Still testing this out on a new site.

PR #4642, opened 2023-04-03
Unified media widget

jackahl:
effort to unify different means of choosing media (Image/videos) in a Block to a unified widget with the same UX/UI across different Blocks


ichim-david:
@jackahl We now have the Image widget added in Volto 18. Would you still like to continue your MediaWidget or should this pull request be closed?

PR #4646, opened 2023-04-04
Added Compelling instead of very interesting

SajidCurious:
I think " Compelling " word is better to use instead of " very interesting"


stevepiercy:
> There is no option to Sign In

It says "Sign", not "Sign In".

SajidCurious:
I'll check it out

sneridagh:
I think the change is not that relevant, and since this is stuck here, closing for now. @stevepiercy please raise hand if you think it's an interesting change.

PR #4658, opened 2023-04-05
Added querystring search get option

robgietema:


erral:
This change should be properly documented.

We faced not being able to use the `@search` endpoint passing a lot of parameters because there is a limitation on the HTTP standard about the URL length, we had to change our calls to the `@querystring-search` because this endpoint uses POST instead of GETs.

robgietema:
> This change should be properly documented.
> 
> We faced not being able to use the `@search` endpoint passing a lot of parameters because there is a limitation on the HTTP standard about the URL length, we had to change our calls to the `@querystring-search` because this endpoint uses POST instead of GETs.

The change is not enabled by default, so by default this will still use POST and not GET. You can opt-in to this behavior if you choose to do so. Discussion about this implementation can be found here: https://github.com/plone/plone.restapi/issues/1252

tisto:
@erral I agree that this should be properly documented. As @robgietema already mentioned, we are going to fully document this in plone.restapi at first. This should go into the Volto docs as well.

Background: we are currently running a set of extensive load tests against a new Plone 6 / Volto 16 deployment of a high-profile website in Germany which will launch in April.

Our analysis revealed that the querystring-search endpoint is responsible for a considerable amount of cache misses. Therefore we plan to switch from using POST to using GET for this installation. We will closely monitor if this approach has any downsides when we run into HTTP limits. Since most requests come from search/listing endpoints we do not expect to run into any limitations. Though, you never know. If that happens we might have to re-think our approach.

Could you elaborate on what problems you exactly ran into? This might help us to plan our next steps...

PR #4661, opened 2023-04-06
fix sitemap.xml.gz not is not compressed #4622

### LINKED ISSUE #4622: sitemap.xml.gz is not compressed. Not gzip format
Download https://demo.plone.org/sitemap.xml.gz

<pre>
$ gzip -d sitemap.xml.gz 
gzip: sitemap.xml.gz: not in gzip format

$ gunzip sitemap.xml.gz 
gzip: sitemap.xml.gz: not in gzip format

$ file sitemap.xml.gz 
sitemap.xml.gz: XML 1.0 document, ASCII text
</pre>

dobri1408:
Fixes #4622


davisagli (commented):
Why does it work with `gz` and not `gzip`? https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding mentions `gzip` but not `gz`

davisagli:
@avoinea How did you retrieve the file for the test? The success may depend on the specifics of the tool you used and how it interprets the header.

avoinea:
@dobri1408 @davisagli :thinking: 

Maybe that's why it works now. As `gz` is not a possible value for `Content-Encoding`, it's ignored and fallbacks to whatever file provides.

davisagli:
@avoinea Ah, that's it.

`Content-Encoding` is for temporary encoding during transfer. `Content-Encoding: gzip` tells the HTTP client it needs to ungzip the body of the response in order to get the original payload. So it does that and the resulting file is no longer gzipped, even though it has the .gz extension.

We should just not set `Content-Encoding` here.

src/express-middleware/sitemap.js
```diff
@@ -5,7 +5,7 @@ export const sitemap = function (req, res, next) {
   generateSitemap(req).then((sitemap) => {
     if (Buffer.isBuffer(sitemap)) {
       res.set('Content-Type', 'application/x-gzip');
-      res.set('Content-Encoding', 'gzip');
+      res.set('Content-Encoding', 'compress');
```
davisagli, line 8: @dobri1408 This isn't correct either. We need to remove the Content-Encoding header entirely. Because the gzip compression is supposed to remain after the file is downloaded, not only during transfer.

dobri1408:
@davisagli Thank you for the review
I will test without the header and open a new pr, because I did a wrong commit, by mistake. I will put you and @avoinea  as reviewers.

dobri1408:
https://github.com/plone/volto/pull/4663

PR #4663, opened 2023-04-07
fix sitemap.xml.gz is not compressed #4622 (v2)

dobri1408:
I removed the res.set('Content-Encoding', 'gzip'); as it makes the archive not open.


dobri1408:
@davisagli I did as you said and the archive is opening now, I think that you are right! Thank you for enlightening me!

davisagli (approved):

avoinea (approved):

PR #4667, opened 2023-04-09
Make URL a literal string to fix broken link

stevepiercy:
This is the third time I have fixed this link to `ReactJS.org`. I don't have any clue why it does not stay fixed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -143,7 +143,7 @@
 
 ### Documentation
 
-- Fix broken links at ReactJS.org. @stevepiercy [#4569](https://github.com/plone/volto/issues/4569)
+- Fix broken links at `ReactJS.org`. @stevepiercy [#4569](https://github.com/plone/volto/issues/4569)
 - Fix video warnings and link errors. @stevepiercy [#4578](https://github.com/plone/volto/issues/4578)
 
 
--- a/news/4667.documentation
+++ b/news/4667.documentation
@@ -1,0 +1,1 @@
+Fix broken link to `ReactJS.org`. @stevepiercy


davisagli (approved):

stevepiercy:
Side note: I don't know if it is appropriate to have the same news item listed twice. Should I remove the most recent one from the 17.0.0-alpha.3 or leave both?

- https://github.com/plone/volto/blob/master/CHANGELOG.md#documentation
- https://github.com/plone/volto/blob/master/CHANGELOG.md#documentation-5

davisagli:
@stevepiercy I don't think we should edit release notes from a version that was already released. It's not a big deal; I wouldn't put effort into solving that problem.

stevepiercy:
In general, I agree with not editing release notes, but I think there are valid exceptions to the general rule, including:

- Error or omission.
- Links become invalid and need to be updated or removed.

davisagli:
@stevepiercy ok, that makes sense.

stevepiercy:
I also noticed there are a lot of duplicate news items between the 16 and 17-alpha releases. I don't know if that is intentional or a bug with towncrier. @mauritsvanrees @sneridagh what are your thoughts?

I'm going ahead with merging this PR for now, because without it, GitHub Actions on `plone/documentation` fail.

davisagli:
@stevepiercy It's an accurate record of what's happening. We're making changes for 17, backporting them to the 16.x branch, and then copying the release notes for 16.x releases back to the master branch after a release. When we port a change to another branch, there are multiple releases where the change is new and warrants being noted.

Issue #3716, opened 2022-10-04
Improve validation of IdWidget


stevepiercy:
This issue defines what we want the frontend to do based on what the backend allows, specifically the `and`ing of all these:

- No new lines (https://zopeschema.readthedocs.io/en/latest/fields.html#asciiline)
- Allow only 7-bit bytes characters (https://zopeschema.readthedocs.io/en/latest/fields.html#ascii)
- No characters that are illegal in URLs: `\x00-\x1F`, `\x7F`, `<`, `>`, and `&`. (https://github.com/zopefoundation/Zope/blob/d1814d0a6bddb615629b552de10e9aa5ad30a6da/src/OFS/ObjectManager.py#L81-L94)
- Not the empty string `""`. (https://github.com/zopefoundation/Zope/blob/d1814d0a6bddb615629b552de10e9aa5ad30a6da/src/OFS/ObjectManager.py#L90)
- Is not `.` or `..` (can be traversed). (https://github.com/zopefoundation/Zope/blob/d1814d0a6bddb615629b552de10e9aa5ad30a6da/src/OFS/ObjectManager.py#L98)
- Cannot start with `_`. (https://github.com/zopefoundation/Zope/blob/d1814d0a6bddb615629b552de10e9aa5ad30a6da/src/OFS/ObjectManager.py#L101)
- Cannot start with `aq_`. (https://github.com/zopefoundation/Zope/blob/d1814d0a6bddb615629b552de10e9aa5ad30a6da/src/OFS/ObjectManager.py#L105)
- Cannot end with `__`. (https://github.com/zopefoundation/Zope/blob/d1814d0a6bddb615629b552de10e9aa5ad30a6da/src/OFS/ObjectManager.py#L108)
- Cannot start with `@@` or `++`. (https://github.com/zopefoundation/Zope/blob/d1814d0a6bddb615629b552de10e9aa5ad30a6da/src/OFS/ObjectManager.py#L112)
- Cannot be `REQUEST`. (https://github.com/zopefoundation/Zope/blob/d1814d0a6bddb615629b552de10e9aa5ad30a6da/src/OFS/ObjectManager.py#L131)
- Cannot contain `/`. (https://github.com/zopefoundation/Zope/blob/d1814d0a6bddb615629b552de10e9aa5ad30a6da/src/OFS/ObjectManager.py#L133)

_Originally posted by @stevepiercy in https://github.com/plone/volto/issues/3704#issuecomment-1266373914_

[...]

PR #4686, opened 2023-04-12
3716 - improve validation of id widget

tedw87:
I modified the regex which validates the value in `IdWidget` to match all the requirements mentioned in the issue #3716


locales/de/LC_MESSAGES/volto.po
```diff
@@ -2300,9 +2300,9 @@ msgid "Ok"
...
-msgstr "Nur Kleinbuchstaben (a-z) ohne Umlaute oder Sonderzeichen, Zahlen (0-9) und die beiden Zeichen "-" und "_", und "." sind erlaubt."
+# defaultMessage: Only 7-bit bytes characters are allowed. Cannot contain characters like:"<", ">", "&", "#", "/", "?". Cannot start with: "_", "aq_", "@@", "++". Cannot end with "__". Cannot be ".", "..", double quotes(can be traversed). Cannot be "REQUEST" or any characters that are illegal in URLs. No newlines.
+msgid "Only 7-bit bytes characters are allowed. Cannot contain characters like:"<", ">", "&", "#", "/", "?". Cannot start with: "_", "aq_", "@@", "++". Cannot end with "__". Cannot be ".", "..", double quotes(can be traversed). Cannot be "REQUEST" or any characters that are illegal in URLs. No newlines."
+msgstr "Es sind nur 7-Bit-Byte-Zeichen zulässig. Darf keine Zeichen wie „<“, „>“, „&“, „#“, „/“, „?“ enthalten. Kann nicht beginnen mit: "_", "aq_", "@@", "++". Kann nicht mit "__" enden. Darf nicht ".", "..", doppelte Anführungszeichen sein (kann durchlaufen werden). Darf nicht „REQUEST“ oder irgendwelche Zeichen sein, die in URLs unzulässig sind. Keine Zeilenumbrüche."
```
stevepiercy, line 2305: Does German use a mixture of double quotation marks: `„`  and `“` versus `"`?

locales/es/LC_MESSAGES/volto.po
```diff
@@ -2305,9 +2305,9 @@ msgid "Ok"
...
-msgstr "Solo letras minúsculas (a-z) sin tildes, números (0-9), y los caracteres "-", "_", y "." están permitidos."
+# defaultMessage: Only 7-bit bytes characters are allowed. Cannot contain characters like:"<", ">", "&", "#", "/", "?". Cannot start with: "_", "aq_", "@@", "++". Cannot end with "__". Cannot be ".", "..", double quotes(can be traversed). Cannot be "REQUEST" or any characters that are illegal in URLs. No newlines.
+msgid "Only 7-bit bytes characters are allowed. Cannot contain characters like:"<", ">", "&", "#", "/", "?". Cannot start with: "_", "aq_", "@@", "++". Cannot end with "__". Cannot be ".", "..", double quotes(can be traversed). Cannot be "REQUEST" or any characters that are illegal in URLs. No newlines."
+msgstr "Solo se permiten caracteres de bytes de 7 bits. No puede contener caracteres como: "<", ">", "&", "#", "/", "?". No puede comenzar con: "_", "aq_", "@@", "++". No puede terminar con "__". No puede ser ".", "..", comillas dobles (se puede atravesar). No puede ser "SOLICITUD" ni ningún carácter que sea ilegal en las URL. Sin saltos de línea."
```
stevepiercy, line 2310: Does the API really not accept `SOLICITUD` versus `REQUEST`?

locales/eu/LC_MESSAGES/volto.po
```diff
@@ -2301,9 +2301,9 @@ msgid "Ok"
...
-msgstr "Letra xeheak (a-z), karaktere berezi gabe, zenbakiak (0-9) eta "-", "_" eta "." ikurrak bakarrik erabili daitezke."
+# defaultMessage: Only 7-bit bytes characters are allowed. Cannot contain characters like:"<", ">", "&", "#", "/", "?". Cannot start with: "_", "aq_", "@@", "++". Cannot end with "__". Cannot be ".", "..", double quotes(can be traversed). Cannot be "REQUEST" or any characters that are illegal in URLs. No newlines.
+msgid "Only 7-bit bytes characters are allowed. Cannot contain characters like:"<", ">", "&", "#", "/", "?". Cannot start with: "_", "aq_", "@@", "++". Cannot end with "__". Cannot be ".", "..", double quotes(can be traversed). Cannot be "REQUEST" or any characters that are illegal in URLs. No newlines."
+msgstr "7 biteko karaktereak soilik onartzen dira. Ezin ditu izan: "<", ">", "&", "#", "/", "?" bezalako karaktereak. Ezin da hasi: "_", "aq_", "@@", "++". Ezin da "__"rekin amaitu. Ezin da izan ".", "..", komatxo bikoitzak (pasatu daitezke). Ezin da izan "ESKAERA" edo URLetan legez kanpokoak diren karaktereak. Lerro berririk ez."
```
stevepiercy, line 2306: `ESKAERA` or `REQUEST`?

stevepiercy (commented):
A few questions. I don't know the answers, but the content seems a bit off to me.

locales/es/LC_MESSAGES/volto.po
tedw87: This is a mistake for sure. Thank you for pointing that. I will check again all translations.

locales/de/LC_MESSAGES/volto.po
tedw87: I made a research and I understand that `„ and “` is the traditional way but they accept also `""`

locales/ca/LC_MESSAGES/volto.po
```diff
@@ -2303,8 +2303,8 @@ msgid "Ok"
 msgstr "D'acord"
 
 #: components/manage/Widgets/IdWidget
-# defaultMessage: Only 7-bit bytes characters are allowed. Cannot contain characters like:"<", ">", "&", "#", "/", "?". Cannot start with: "_", "aq_", "@@", "++". Cannot end with "__". Cannot be ".", "..", double quotes(can be traversed). Cannot be "REQUEST" or any characters that are illegal in URLs. No newlines.
-msgid "Only 7-bit bytes characters are allowed. Cannot contain characters like:"<", ">", "&", "#", "/", "?". Cannot start with: "_", "aq_", "@@", "++". Cannot end with "__". Cannot be ".", "..", double quotes(can be traversed). Cannot be "REQUEST" or any characters that are illegal in URLs. No newlines."
+# defaultMessage: Only 7-bit bytes characters are allowed. Cannot contain characters like:<, >, &, #, /, ?. Cannot start with: _, aq_, @@, ++. Cannot end with __. Cannot be these specific keywords: REQUEST, ., ..,"", or any other characters that are illegal in URLs. No new lines allowed.
```
stevepiercy, line 2306: Sorry for my incomplete first pass. This time I am being thorough.

1. Should this keyword `""` be `"`, or is it intended to be empty string?
2. Similarly is `..` really a keyword?
3. Let's combine the two separate parts that define illegal characters into one sentence. This should be done everywhere, and retranslated.
4. Also I made the final sentence wording to be consistent with the others.
5. I corrected some separation and spacing of character lists.

```suggestion
# defaultMessage: Only 7-bit bytes characters are allowed. Cannot contain characters like: <, >, &, #, /, ?, or others that are illegal in URLs. Cannot start with: _, aq_, @@, ++. Cannot end with __. Cannot be these specific keywords: REQUEST, ., .., "". Cannot contain new lines.
```

FWIW, I think it is *much* easier to read without quoting all the entities. Thanks for doing this!

stevepiercy (changes requested):

locales/ca/LC_MESSAGES/volto.po
tedw87: Thank you also for the suggestions! I will update the text today. Regarding the `""`, if you write nothing in the input, it is still valid. Must be invalidated when you type specifically `""` like a single word. At least that's what I understood from the issue.

stevepiercy (approved):
LGTM, of course, I can only vouch for the English version. Thank you!

locales/de/LC_MESSAGES/volto.po
sneridagh: @tisto @steffenri @jackahl @ThomasKindermann @ksuess @pbauer I guess that here using the typographic version is not really necessary/mandatory, right? @tedw87 why did you put it there in the first place? Sorry, I don't have the full picture either but it would be nice if you take a look.

src/components/manage/Widgets/__snapshots__/IdWidget.test.jsx.snap
```diff
@@ -188,7 +188,7 @@ exports[`IdWidget renders an id widget with an reserved name 1`] = `
 exports[`IdWidget renders an id widget with invalid characters 1`] = `
 <div>
   <div
-    class="error inline field text field-wrapper-my-field"
+    class="inline field text field-wrapper-my-field"
   >
```
sneridagh, line 192: Why the test do not error now? If that was the intention, it should continue erroring :)

sneridagh (changes requested):

src/components/manage/Widgets/__snapshots__/IdWidget.test.jsx.snap
tedw87: In the issue, I saw this: `Allow only 7-bit bytes characters (https://zopeschema.readthedocs.io/en/latest/fields.html#ascii)`
This include uppercase letters, so I figured it must be permitted. Did I get it wrong? In case uppercase is valid now, I will change the test file to have an invalid input so it returns error.

locales/de/LC_MESSAGES/volto.po
tedw87: Not the empty string "".
https://github.com/plone/volto/issues/3716

erral:
Sorry, but I wouldn't try to translate the messages to other languages as a part of the PR that adds the functionality, I would leave this as a work for translators.

sneridagh:
@tedw87 status?

sneridagh:
@davisagli could you please take a second look?

src/components/manage/Widgets/__snapshots__/IdWidget.test.jsx.snap
sneridagh: @stevepiercy I think uppercase letters should not be allowed in an `id`, although I think that Plone then normalizes them, I guess we should not allow them to enter, as it can lead to confusion.

sneridagh:
@tedw87 sorry to have this abandoned. What's it's status? Can we resurect it? Thanks!

tedw87:
> @tedw87 sorry to have this abandoned. What's it's status? Can we resurect it? Thanks!

Hello, the updates will be as follows: 1. No uppercase letters are to be used; 2. 'contributors' should not be used as a shortname. I will update the validation criteria accordingly. Thank you.

sneridagh (approved):
LGTM, but I'd like other people with knowledge of regex take a look

/cc @davisagli @tiberiuichim

stevepiercy:
This tool might help: https://regex101.com/r/E5gJye/1

tiberiuichim:
The ideal scenario would be to make the validation error simpler, "Only alphanumeric characters . For a complete list of allowed characters, see /somepage". And we could have a static page with the rules for it.

I think the regex is fine, although I see that it allows for lf and cr/lf characters

stevepiercy:
> The ideal scenario would be to make the validation error simpler, "Only alphanumeric characters . For a complete list of allowed characters, see /somepage". And we could have a static page with the rules for it.

I disagree. As an editor, I don't want to go to another page to find acceptable input.

I agree that documentation of valid input for the idWidget should be added.

> I think the regex is fine, although I see that it allows for lf and cr/lf characters

I think neither are allowed. Have you verified? Using the [regex101 test](https://regex101.com/r/E5gJye/1), I found they were not allowed. Maybe I did not set it up right?

stevepiercy:
@tedw87 the best place is in the [Storybook for the IdWidget](https://6.docs.plone.org/storybook/?path=/story/edit-widgets-id--text). It already exists, but it might need to be updated. Here's documentation that I added just last week about it: https://6.docs.plone.org/volto/contributing/documentation.html#storybook-entry. However I have never created or edited a Storybook entry, but I'm sure if you need help, you can ask here, and someone more knowledgable than I will be more than happy to help.

sneridagh (approved):
LGTM but not a regex expert here! @stevepiercy @davisagli a final look?

stevepiercy:
@sneridagh the Storybook preview does not work. I have no idea why:

https://65b0360fb762f40008f1c5f1--volto.netlify.app/storybook

I thought I had fixed that in https://github.com/plone/volto/pull/5599#issuecomment-1877098700, but it appears to have been broken in this PR.

If it works locally, then I'm fine with Netlify not working and find where it went sideways later.

----

As far as the regex, it appears that it does not reject `\n` or `\r`. See regex tester at https://regex101.com/r/E5gJye/1.

stevepiercy:
It looks like the latest regex works: https://regex101.com/r/E5gJye/3.

That leaves the Storybook build failure to resolve. I forced it to rebuild with the latest commit, but it still does not load: https://deploy-preview-4686--volto.netlify.app/storybook

stevepiercy:
Thanks for checking @tedw87! @sneridagh let's go ahead and merge this PR, and if the Storybook does not build and deploy after I update git submodule tips on the Plone 6 documentation, we can open a new issue to deal with it there.

stevepiercy:
The Storybook build is fine in production at https://6.docs.plone.org/storybook/, so something with the Volto docs preview in Netlify might need some tweaking.

Issue #4692, opened 2023-04-13
A querystring appearing in the URL after accessing a page with a listing/search block


sneridagh:
After merging https://github.com/plone/volto/pull/4159, the site URL start to have a strange behavior, if a listing block is present, then the URL shows a querystring:

<img width="633" alt="image" src="https://user-images.githubusercontent.com/486927/231795879-8e0c81cf-a440-4d8e-98a6-336fe9d73c1b.png">

The interesting thing is that it appears everywhere once you click in another router link.

In the videos in https://github.com/plone/volto/pull/4159 it does not show, maybe something that was introduced lately in the PR?

/cc @bipoza @erral @ionlizarazu this is quite urgent, we would need to fix asap. If it's not possible, I'd better revert the change and proceed with another release.

sneridagh:
Also, I can see that the @querystring endpoint is being hit multiple times... this is another thing that we have to make sure that doesn't happen.

sneridagh:
@erral Thanks a lot! Let's see if we can push a fix asap! If not, let me know and we can revert and postpone it until we have a better one.

erral:
It looks like the code under the `useEffect` in `usePagination` is run no matter the user has gone away from the page where the pagination is shown.

Somehow this `useEffect` should be deactivated (sorry if the concept is not correct) when browsing away.

sneridagh:
@erral we just merged some fixes to the listing/search blocks in cases of two of them are present in the same page, see: 

https://github.com/plone/volto/pull/4694

Maybe now the situation is improved.

It's what I saw, the `usePagination` hook should be addressed.

erral:
I managed to remove the querystring before this merge, so I will test it again.

Anyway I have seen something strange going on, because I removed the querystring when going to any page except the home page. I need to test my fix further.

PR #4698, opened 2023-04-14
Pagination with router params

### LINKED ISSUE #3868: Pagination component does not keep state when navigating away from it

erral:
Fixes #3868 and #4692


erral:
@sneridagh we have reworked this PR and added cypress tests to show that the `?page=XX` functionality is working. We test clicking on the pagination numbers, going back, reloading the page and going to some other page too.

In the meantime we have fixed the `withQuerystringResults` function, which wasn't checking for the request being loaded, but being "not loading" which created additional queries to the endpoint.

ionlizarazu:
This is the cypress tests resulting video for the ListingBlock

https://user-images.githubusercontent.com/5443301/233557565-51fce414-f1fa-48d5-bcea-625ac4837f93.mp4

sneridagh:
@erral @ionlizarazu I've found a problem, see video. When I create a page, then go back to the listing, it's still empty :( so the query is not being triggered on route change for some reason.

https://user-images.githubusercontent.com/486927/236615730-ec90693c-b961-47e8-adc4-d12c83911ff5.mov

ionlizarazu:
> @erral @ionlizarazu I've found a problem, see video. When I create a page, then go back to the listing, it's still empty :( so the query is not being triggered on route change for some reason. I've tried on demo and works well.
> 
>  Screen.Recording.2023-05-06.at.11.23.48.mov

I've added a cypress test for this and fixed it.

robgietema:
How does this work when I add multiple listing blocks on 1 page?

ionlizarazu:
> How does this work when I add multiple listing blocks on 1 page?

https://cloud.cypress.io/projects/hvviu4/runs/df7a2ee0-472a-456a-8191-814a1b6c3b35/test-results/5fa7949f-3218-4883-a052-3adf784c06d4/video

robgietema (approved):

sneridagh (approved):

src/helpers/Utils/usePagination.js
```diff
@@ -1,25 +1,83 @@
...
+import { useSelector } from 'react-redux';
+import { slugify } from '@plone/volto/helpers/Utils/Utils';
+
+/**
+ * @function useCreatePageQueryStringKey
+ * @description A hook that creates a key with an id if there are multiple blocks with pagination.
+ * @returns {string} Example: page || page_012345678
+ */
+const useCreatePageQueryStringKey = (id) => {
+  const blockTypesWithPagination = ['search', 'listing'];
```
davisagli, line 13: Maybe this should come from the config so that addon blocks can use pagination.

davisagli:
I tried editing a search block and changing the number of items per page, then saving. The items shown per page updated correctly in edit mode, but after I saved, the view mode was still showing the old number of items per page until I reloaded.

src/helpers/Utils/usePagination.js
```diff
@@ -1,25 +1,83 @@
...
+      };
+      delete newParams[pageQueryStringKey];
+      history.replace({ search: qs.stringify(newParams) });
+      queryRef.current = qs.parse(location.search)?.query;
+    } else {
+      setCurrentPageState(
+        parseInt(
+          qs.parse(location.search)?.[pageQueryStringKey] || defaultPage,
+        ),
+      );
```
davisagli, line 74: The parsing of the page from the querystring needs to handle a couple edge cases better:
1. I navigated to `?page=x` and the pagination showed `NaN`
2. I navigated to `?page=-4` and the block showed no results.

If the value isn't a positive integer, this should probably just use the default page.

ionlizarazu:
> I tried editing a search block and changing the number of items per page, then saving. The items shown per page updated correctly in edit mode, but after I saved, the view mode was still showing the old number of items per page until I reloaded.

This is not an issue we are handling in this PR and it happens in the actual Volto version. Can we manage it in another issue or PR?

src/helpers/Utils/usePagination.js
ionlizarazu: You mean the 'blockTypesWithPagination'?

davisagli: @ionlizarazu Yes. But it can wait for a future PR, rather than delaying this one.

sneridagh:
@ionlizarazu I also think it's ready, I'll try to merge it this week.

PR #4717, opened 2023-04-19
Ensure SCSS dependencies are available in addon template

JeffersonBledsoe:
`razzle-plugin-scss` and `postcss-scss` don't seem to be installed by default in an app. While it is likely most developers will use an addon to set these dependencies, SCSS should still work out-the-box


sneridagh:
@JeffersonBledsoe I just did: 

`npm i -g @plone/generator-volto@alpha`

```
yo @plone/volto my-volto-app --description "test volto project" --skip-install --no-interactive --canary
```

then added a file `test.scss` and add it to `theme.js` in the project:

```
import 'semantic-ui-less/semantic.less';
import '@plone/volto/../theme/themes/pastanaga/extras/extras.less';
import './test.scss';
```

and worked like a charm... it should use the dependencies specified in Volto. Could be that for some reason the dependency is being hoisted and it's not present in your project main node_modules?

For stable versions of the generator:

`npm i -g @plone/generator-volto`

```
yo @plone/volto my-volto-app --description "test volto project" --skip-install --no-interactive
```

and then same as above with the scss file, and it does works well too.

Make sure that your razzle versions are in sync with the version of Volto you are using (4.2.17 for 16, 4.2.18 for 17alpha).

JeffersonBledsoe:
Hmm, I'll give this recommendations a try. I updated the templates based on the fix I found locally, so hoisting issues are highly probable.
Closing for now, will re-open if I don't get anywhere with your suggestions 🙂

PR #4719, opened 2023-04-19
Don't decode querystring while adding apiExpanders

### LINKED ISSUE #4718: API middleware mangles urlencoding in querystring
Steps to reproduce:

1. Go to https://demo.plone.org/
2. Open the network tab in dev tools
3. Do a search for `boots & cats`

Expected: volto makes a request for `https://demo.plone.org/++api++/@search?SearchableText=boots %26 cats*&use_site_search_settings=1`
Actual: volto makes a request for `https://demo.plone.org/++api++/@search? cats*=&SearchableText=boots &use_site_search_settings=1`

davisagli:
addExpandersToPath parses the current querystring, makes modifications, and then stringifies it again. stringify is called with `encode: false` so we need to call parseUrl with `decode: false` to match.

This should be backwards compatible unless there are expanders whose name includes an urlencoded character, or code somewhere which is already doing extra encoding to work around this bug.

Fixes #4718

### DIFF
--- a/news/4718.bugfix
+++ b/news/4718.bugfix
@@ -1,0 +1,1 @@
+Fix fetching API paths with urlencoded characters in the querystring. @davisagli
--- a/src/middleware/api.js
+++ b/src/middleware/api.js
@@ -43,7 +43,7 @@ export function addExpandersToPath(path, type, isAnonymous) {
   const {
     url,
     query: { expand, ...query },
-  } = qs.parseUrl(path);
+  } = qs.parseUrl(path, { decode: false });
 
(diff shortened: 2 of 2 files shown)


davisagli:
I don't understand the failing unit tests yet. They are failing only in CI on Node 18, they work for me locally.

davisagli:
Looks like the test failures are a regression in Node 18.16.0: https://github.com/nodejs/node/issues/47563

sneridagh:
@davisagli I added more tests just to be 100% sure.

sneridagh (approved):

PR #4720, opened 2023-04-20
Don't decode querystring while adding apiExpanders (#4719)

sneridagh:


reebalazs:
Reviewed, tested. Good and needed.

PR #4725, opened 2023-04-21
fix:blacklisted routes match check

andreiggr:
found a small issue with `isBlacklisted` method of checking an external to Volto route. 

Previous implementation would always show routes from `externalRoutes` as not-blacklisted resulting in non-volto routes being rendered by Volto

### DIFF
--- a/src/components/manage/UniversalLink/UniversalLink.jsx
+++ b/src/components/manage/UniversalLink/UniversalLink.jsx
@@ -66,10 +66,10 @@ const UniversalLink = ({
     }
   }
 
-  const isBlacklisted =
-    (config.settings.externalRoutes ?? []).find((route) =>
-      matchPath(flattenToAppURL(url), route.match),
-    )?.length > 0;
+  const isBlacklisted = !!(config.settings.externalRoutes ?? []).find((route) =>
+    matchPath(flattenToAppURL(url), route.match),
+  );
+
   const isExternal = !isInternalURL(url) || isBlacklisted;
   const isDownload = (!isExternal && url.includes('@@download')) || download;
   const isDisplayFile =


sneridagh (changes requested):
Could you please elaborate a bit more on this one? I don't understand the bug, nor the fix. We are using it actively so we should be extremely careful with this one. Also, add a test for the culprit and the fix.

sneridagh (changes requested):
LGTM but could you please add a unit test that proves that now it's working as expected? Maybe introducing also a new helper? It would be great if we use more neutral terms for this feature, as `isExcludedURL` or `isExternalRoute`... Thanks!

andreiggr:
> LGTM but could you please add a unit test that proves that now it's working as expected? Maybe introducing also a new helper? It would be great if we use more neutral terms for this feature, as `isExcludedURL` or `isExternalRoute`... Thanks!

Hi @sneridagh. Yes, I wanted to explain how this could be handled better with examples but didn't really had the time for it. Most likely will jump on this during the Beethoven sprint.

andreiggr:
As it was defined before I touched it, `isBlacklisted` method does not really do much as it will return undefined. `.find` will look for the first element that satisfies the testing function inside. `config.settings.externalRoutes` is an array of objects. Example: 

```
  config.settings.externalRoutes = [
    ...(config.settings.externalRoutes || []),
    {
      match: {
        path: 'condition-path',
        exact: false,
        strict: false,
      },
    },
  ];
```
So if it matches the condition in that `.find`, the element found will be an object...to which we check `?.length > 0`.
Example:
length-of-object

That's why I changed it in 
```
  const isBlacklisted = !!(config.settings.externalRoutes ?? []).find((route) =>
    matchPath(flattenToAppURL(url), route.match),
  );
```
that will actually say if the route is matching any of the `.externalRoutes`

But then, why do we have the `isBlacklisted` method when we can use `isInternalURL` https://github.com/plone/volto/blob/master/src/helpers/Url/Url.js#L215 ? Looks like `isInternalURL` is also checking the `.externalRoutes`. Why not let `isInternalURL` tell if a route is internal or not? 

https://github.com/plone/volto/commit/10c96dd5ffa0930f8994caa393f9bad09274d4b8

andreiggr:
https://github.com/plone/volto/pull/4725/commits/d73cfb380e83eea1536003a2ea0ab95529f74e1e#diff-c0021190896d8196577b4d304ac09c44abff5437fcf56791f7832c522f43ea01R73-R78

For demonstration purposes. I pushed a commit where I let only the original `isBlacklisted` method tell if the route is external or not. I did this because in tests, `isInternalURL` would catch it before `isBlacklisted`.

Added my own test with a more complex regex rule in `.externalRoutes` but there also is a previously wrote test for blacklisted url's. 

On local, with this modification, assuming the `isBlacklisted` is working correctly, we would expect that `external-route` would be rendered as an external route...

isBlacklisted-bug-test-fail

.. but it's not. On both tests. 

I let the previous commit there to check the unit tests for that case.

avoinea:
Yes, I think the isBlacklisted logic is already handled by the isInternalUrl but maybe someone else can take a closer look: @davisagli

davisagli (approved):
@avoinea @andreiggr @sneridagh I agree, the isBlacklisted logic was duplicating what already happens in isInternalURL. Thanks for the test to show that it still works. This looks like a good simplification.

avoinea (approved):

sneridagh (approved):

PR #4730, opened 2023-04-24
Docs backport master to 16.x.x

stevepiercy:
This PR suggests two commits, one for each file, that backports changes from `master` to `16.x.x` where I am not 100% certain whether these changes should in fact be backported. A careful review by a maintainer is required before merging. I can also make revisions as needed.


davisagli:
@stevepiercy Neither of these changes belongs in the docs for the 16.x branch, because the changes they are documenting were not backported to 16.x. I'll close the PR.

Sorry for the delay in reviewing, I was busy hosting family.

PR #4733, opened 2023-05-03
WIP: PoC AboveTitle

danalvrz:
Add conditional AboveTitle component to display properties above the Title. Addons and themes can register their own custom component, nothing will be rendered by default.


erral:
We already have [Pluggables](https://6.dev-docs.plone.org/volto/recipes/pluggables.html) and this looks like pretty much similar to it, right? Can we do it with that or it is a different concept?

erral:
I haven't used myself the Pluggables framework, I was just asking whether this could be implemented using those.

Perhaps something like ```<Pluggable name={`above-title-${type}`}``` could work?

sneridagh:
"Problem" with Pluggables is that they are not SSR friendly. So, they are good for CMSUI, or all client components/widgets, but not for data.

Back to the approach, I'm still not sure about how to go with this. There are drawbacks, and the inline approach is not the most friendly one. What about if the content type does not have above title? it will "save" space? 

Also, overloading the title block with something that it's not title, I do not know if it's the best approach. Where we will save the data? in the content type as DX field? in the block? I don't know if I would merge both ways in the same block. Specially in the title one.

davisagli:
> "Problem" with Pluggables is that they are not SSR friendly. So, they are good for CMSUI, or all client components/widgets, but not for data.

Got it. It should say that in the docs. ;-)

> There are drawbacks, and the inline approach is not the most friendly one. What about if the content type does not have above title? it will "save" space?

It's an optional slot. By default nothing is added above the title, but addons can add something if they want to. The goal is to avoid needing to shadow the title block view to do this.

> Also, overloading the title block with something that it's not title, I do not know if it's the best approach. Where we will save the data? in the content type as DX field? in the block? I don't know if I would merge both ways in the same block. Specially in the title one.

It's for the use case where an addon wants to always put something above the title. The right place to store the data will vary based on the specific use case. For mine, it would come from the effective date and head_title fields on the content type.

We can discuss it, of course.

news/4733.feature
```diff
@@ -0,0 +1 @@
+Add conditional AboveTitle component to display properties above the Title @danalvrz
```
davisagli, line 1:
```suggestion
Add optional AboveTitle component to display properties above the Title @danalvrz
```

src/components/manage/Blocks/Title/Edit.jsx
```diff
@@ -143,13 +143,33 @@ export const TitleBlockEdit = (props) => {
...
-        {children}
-      </h1>
-    );
-  }, []);
+  const renderElement = useCallback(
+    ({ attributes, children }) => {
+      const hasType = properties['@type'];
+      const AboveTitle =
+        hasType &&
+        config.getComponent({ name: 'AboveTitle', dependencies: [hasType] })
```
davisagli, line 151:
```suggestion
        config.getComponent({ name: 'AboveTitle', dependencies: [hasType] })?
```

src/components/manage/Blocks/Title/Edit.jsx
```diff
@@ -143,13 +143,33 @@ export const TitleBlockEdit = (props) => {
...
+              <AboveTitle item={properties} />
+              <h1 {...attributes} className="documentFirstHeading">
+                {children}
+              </h1>
+            </>
+          ) : (
+            <h1 {...attributes} className="documentFirstHeading">
+              {children}
+            </h1>
+          )}
```
davisagli, line 167: Is there any reason not to do it like this?

```suggestion
          <>
            {AboveTitle ? <AboveTitle item={properties} /> : null}
            <h1 {...attributes} className="documentFirstHeading">
              {children}
            </h1>
          </>
```

danalvrz: no reason, it looks a lot better now, thanks!

wesleybl:
@danalvrz @davisagli couldn't that be done with a block? In a case where we need to display dates, we create a block and place this block in the content layout template.

davisagli:
@wesleybl I can see how that might be a possible way to do it, but we don't really want to give editors the flexibility to add the date anywhere other than right above the title.

davisagli:
@wesleybl That is helpful -- but it doesn't prevent adding the block in other places, right?

wesleybl:
There is a condition for the block not to be addable. See:

https://github.com/plone/volto/blob/e5416ce00c3c19e671b326ac98a1bd421db9d448/src/config/Blocks.jsx#L201

You can simply return true.

davisagli:
Closing since this was never completed and we probably should use slots instead.

PR #4737, opened 2023-05-04
Fix whitespace in locales created by the generator

davisagli:
This avoids creation of changes when i18n is synced, which causes the i18n check in a plone cookiecutter project to fail. Fixes https://github.com/collective/cookiecutter-plone-starter/issues/63

Also removed the translator names from the empty locale files, since these are locales for a new project, not the existing translations that those people worked on.


davisagli:
The towncrier check shouldn't be failing; I added a news entry inside packages/generator-volto

sneridagh (approved):

Fosten:
Test i18n sync fails due to one extra whitespace character in the pt locale:
https://github.com/Fosten/plone-pimaker-site/actions/runs/5606557238/jobs/10256823783
Test i18n sync passes after fixing line 3 here:
https://github.com/Fosten/plone-pimaker-site/commit/56fa9688c1edf1ab3e012996c29ae310f9239fb3

PR #4767, opened 2023-05-15
Convert header class to function

Gomez:


ksuess (approved):
I like the `shallowEqual`, didn't know it before…

sneridagh:
@Gomez I took the liberty to amend the PR adding a façade hook instead of using directly the Redux hooks. This will help the transition to use the new modern data fetching when available.

@Tishasoumya-02 Please take a look at this pattern. We can discuss it during this week's meeting. I'll merge this PR when ready, since it was already in the pipeline.

/cc @JeffersonBledsoe @nileshgulia1 @jackahl

Issue #770, opened 2019-06-18
Accessibility acceptance tests for content-types


tisto:
cypress-axe

pnicolli:
Renamed to split content types and blocks (see #4770 )

ichim-david:
Added content-types  initial axe tests in https://github.com/plone/volto/pull/6339

PR #4773, opened 2023-05-16
Add Accessibility acceptance tests for content-types(#770)

iRohitSingh:
Fix: #770 

Screenshot 2023-05-16 at 20 10 42


davisagli:
This was updated and finished in #6339

PR #4774, opened 2023-05-16
UX affordance: show folder icon for folderish pages which contain folderish pages

jaroel:


jaroel:
Depends on/blocked by https://github.com/plone/plone.volto/pull/118

jaroel:
UX review feedback will be provided and we'll use that instead of this Folder icon.

davisagli:
@sneridagh @albertcasado What was the conclusion at the sprint about what icon to use here?

pnicolli:
@davisagli did you have any news about this eventually? Do you think this could go in Volto 19?

davisagli:
@pnicolli See related discussion in https://github.com/plone/plone.volto/pull/118. We need to make sure it doesn't hurt performance, and I didn't make time to take a close look at it yet. I'd like to see it happen, but I don't think it will be ready for 6.2

PR #4782, opened 2023-05-18
Misc fixes for image rendering

davisagli:


src/components/manage/Blocks/Image/ImageSidebar.jsx
```diff
@@ -19,50 +21,54 @@ const ImageSidebar = (props) => {
...
+              onClick={(e) => {
+                e.stopPropagation();
+                onChangeBlock(block, {
+                  ...data,
+                  url: undefined,
+                  image_scales: undefined,
+                });
+              }}
+            >
+              <Icon name={clearSVG} size="30px" />
```
davisagli, line 39: This still needs to be styled properly, but the point is to add a way to remove the currently selected image.

Issue #4339, opened 2023-01-30
PLIP: Link Integrity View


tisto:
## Abstract

Plone (Volto) should provide a dedicated view to show all items that are connected with a particular content object.

## Motivation

Volto provides a link integrity view that shows relations and links that would break when deleting the object:

![Inhalte (10)](https://user-images.githubusercontent.com/632105/215619999-c4401d7f-dbdc-421d-8f82-ff168ba38030.png)

Though, this overlay only shows up when the editor tries to delete an item. Editors usually hesitate to press the delete-button, just to check.

## Proposal & Implementation

Implement a dedicated view "related items" that is added to the "..." menu in Volto that shows a list of all content objects that reference the current object.

## Relation Types

- related items (we don't use it in Volto)
- link/referencing (deserializer with adaptors, stored in the relations catalog)
  - RichText / Slate / DraftJS
  - Slate Table
  - URL / href -> works for nested blocks
- custom relation field

## UX Questions / Tasks

- pop up on delete
  - how do we show many relations?
  - do we group them by relation type
- how can editors see up-front where an object is used?
  -  tab in edit view
  - actions item in the left sidebar

tisto:
@ksuess @pbauer this is an idea we came up with (relations view per content object). Maybe we can build this on top of the work that you folks plan to do at the Alpine City Sprint. Looking forward to seeing you there!

ksuess:
Superbe idea!

wesleybl:
The related items are not even shown at the end of the contents view, as in Plone Classic:

https://github.com/plone/volto/issues/3740

tisto:
Notes from Beethoven Sprint 2023 (Albert):

- In general, the goal is to (during save) detect everywhere that a content item links to another content item, and update the list of isReferencing relations
- Already handled:
  - Links in Dexterity richtext fields
  - Relation fields
  - Links in slate text block (and old draftjs text block)
  - Any url or href field in any block
  - Blocks nested inside columns, hrefList, or slides of another block (https://github.com/plone/plone.volto/blob/main/src/plone/volto/linkintegrity.py)
- Not handled:
  - Links in slateTable block and custom blocks that use slate
  - Blocks nested inside “blocks”
  - Blocks which store a url in a field with a different name (can be added with a custom IBlockFieldLinkIntegrityRetriever adapter)
- Challenges:
  - It’s difficult to automatically find links because the backend doesn’t know anything about the block schema, so has to recognize URLs from the data structure only
[...]

PR #4786, opened 2023-05-18
Update delete content modal for link integrity

danalvrz:
Regarding: https://github.com/plone/volto/issues/4339

Requires:
[#1636](https://github.com/plone/plone.restapi/pull/1636)
[#4787](https://github.com/plone/volto/pull/4787)

Screenshots:
Screenshot 2023-05-18 at 20 41 42
Screenshot 2023-05-18 at 20 42 03


locales/en/LC_MESSAGES/volto.po
```diff
@@ -3417,6 +3417,21 @@ msgstr ""
...
 msgstr ""
 
+#: components/manage/Contents/Contents
+# defaultMessage: This item is also a folder.
+msgid "This item is also a folder."
+msgstr ""
+
+#: components/manage/Contents/Contents
+# defaultMessage: This item is referenced by other items. By
+msgid "This item is referenced by other items. Bydeleting it {brokenReferences}"
```
pgrunewald, line 3427: The msgid seems to be out-dated. Please regenerate the *.po files.

src/components/manage/Contents/Contents.jsx
```diff
@@ -1196,82 +1184,180 @@ class Contents extends Component {
...
+                                  values={{
+                                    containedItemsToDelete: (
+                                      <span>
+                                        {this.state.containedItemsToDelete}
+                                      </span>
+                                    ),
+                                    variation: (
+                                      <span>
+                                        {this.state.containedItemsToDelete === 1
+                                          ? 'item'
```
pgrunewald, line 1216: This should be a translateable message as well. Sames goes for items.

src/components/manage/Contents/Contents.jsx
```diff
@@ -1196,82 +1184,180 @@ class Contents extends Component {
...
+                                      values={{
+                                        brokenReferences: (
+                                          <span>
+                                            {this.state.brokenReferences}
+                                          </span>
+                                        ),
+                                        variation: (
+                                          <span>
+                                            {this.state.brokenReferences === 1
+                                              ? 'reference'
```
pgrunewald, line 1239: These should be translateable, as well.

pgrunewald, line 1239: These should be translateable, as well. All further instances, too.

cypress/tests/core/basic/content.js
```diff
@@ -87,7 +87,7 @@ describe('Add Content Tests', () => {
     cy.findByLabelText('/my-page').children('td').eq(1).click();
 
     cy.get('.top-menu-menu .delete').click();
-    cy.get('.modal.active').contains('Potential link breakage');
+    cy.get('.modal.active').contains('View broken links list');
```
pgrunewald, line 90: Maybe we should reflect some of the other use-cases the new modal is able to handle in the tests.

pgrunewald (commented):
I have found some smaller things, that need a closer look.

cypress/tests/core/basic/content.js
danalvrz: @pgrunewald I understand the idea of testing other cases, but would be of the opinion of keeping it concise to avoid slowing down the CI with extra content creation and a longer cypress test. I consider that the implementation of both backend and frontend is demonstrated in the current use-case, but let me know what do you think

sneridagh:
@danalvrz what's the status of this one?

cypress/tests/core/basic/content.js
pgrunewald: Sorry for the late response. I was mentioning this point, because judging from the changed files it would seem that only some text in the modal has been changed. But again, testing too fine-grained on subtle text parts might lead to flaky tests, which are burdensome to fix. LGTM in that regard.

danalvrz:
@sneridagh I believe it is ready to be reviewed, considering it still needs [#1636](https://github.com/plone/plone.restapi/pull/1636) to be merged

tisto:
@sneridagh we need to release plone.restapi 9, then we can merge this. Though, this would introduce a hard requirement on plone.restapi 9 unless we can make this feature somehow backwards compatible.

davisagli:
@tisto @sneridagh I updated it so that it degrades gracefully when using an older version of plone.restapi that does not include `items_total` in the `@linkintegrity` response. In that case it will not be able to warn about deleting contained items, but the rest of the linkintegrity warnings should still work.

tisto:
@davisagli awesome! Thanks!

sneridagh (approved):

Issue #4208, opened 2023-01-01
PLIP: Teaser block in core (mark II)


sneridagh:
## Abstract

Include in core a Teaser block that showcases a site content in a summarised way (card, listing, etc). 

## Motivation

Historically, there has been a number of implementations of a block that it's capable of "tease" an arbitrary content from the site in a summarised and visually appealing way (as a card, listing, etc). It has been identified as a recurrent and a very valuable feature that is needed for every modern site.

## Assumptions

No assumptions are required.

## Proposal & Implementation

There is a PoC in the PR: https://github.com/plone/volto/pull/3706

It has been agreed to start the implementation with the Teaser block that `@kitconcept/volto-blocks-grid` provides, and add more features (some could be optional) to extend the feature set of the block.

## Deliverables

- [x] Standalone `teaser` block
- [x] Local cache of hardcoded `title`, `description` and `head_title` of the target element in block fields
- [x] These cached fields are the ones that are used in the `Teaser` rendering
- [x] The user can override them at will
- [x] The `Teaser` does not update until you reset it and choose another target
- [x] Uses component registry for defining templates for default content types teases
- [x] Alignment widget, left, right, center using the style wrapper
- [x] Add handler for resetting the teaser completely (target and overrides)

### Wishlist

- [ ] Extend the default local cache fields
- [ ] Sync the original teaser with the overrides
- [ ] The cache can be refreshed on demand by the user, overriding the current values (UX TBD)
- [ ] Optional switch that makes the data is always "live", unless override values are present (UX TBD)
- [ ] Images do not use "the shortcut" but the image cataloged data for the target object
- [ ] Extend the Teaser with "slots". These slots could be: bottom-left, bottom-middle and bottom-right.
- [ ] Improved serialization for teasers using transforms and the upcoming Relations serializers (as in the RelatedChoices fields.
- [ ] Able to extend (and show) metadata if the teaser is linked to an object then we would also be able to show Tags or any other object metadata
- [ ] Inline editing of the teaser text elements?
- [ ] Round image (variations?)
- [ ] Ribbons (variations?)
- [ ] Using SVG for images (to create a stripe of icons instead of images)

## Risks

The block is completely new, so no risks are involved.

## Participants

@sneridagh: I will champion this, but help is welcome!

davisagli:
@sneridagh @tisto @pbauer Is this what was already completed at the Beethoven Sprint for Plone 6.1, or does this go further?

sneridagh:
[...]

PR #4790, opened 2023-05-19
Teaser dynamic

steffenri:
Ticket: https://github.com/plone/volto/issues/4208

Description:
- Added 2 options for the teaser (live and overwritable)
- Live option always fetches fresh data from source (no option to overwrite when activated)
- Overwritable option works as the teaser does at the moment
- Add additional refresh button when the teaser is overwritable to get fresh content (button is only shown when overwritable option is chosen)

## Screenshot Live Sidebar
Edit-Page (1)

## Screenshot Overwritable Sidebar
Edit-Page (2)


tisto:
@steffenri @pbauer this would be a major change that would require a PLIP IMHO.

pbauer:
We discussed that at the Beethoven Sprint 2024:

* Performance: Not a problem because the block-serializer already returns brain-metadata for teaser-targets (as done in DLR). So there are no additional requests per teaser.
* We want live (without overrides) as default 
* This need usability-testing  because the override-feature it might be confusing. Maybe we make it conditional on a setting
* We need to add the head-title to the block-serializer
* Needs to be rebased, cleaned up and finished

pbauer:
Replaced by https://github.com/plone/volto/issues/6023

PR #4791, opened 2023-05-19
Move image block data to top level (new Image component)

pnicolli:
@davisagli  Leaving this here, ready to merge in case we actually move image_field and image_scales to top level.


pnicolli:
Merging this manually and continuing in #3337

PR #4792, opened 2023-05-19
Update relations control panel to use new result format from relations endpoint

davisagli:
@ksuess I tried to update the control panel to use the updated result format that you worked on this morning. Please check my work (especially the broken relations component, which I'm not sure how to test).

Edited to add: I was working on this quickly at the end of the day and I am not confident it is right yet.


ksuess (approved):
Thank you @davisagli !
With a fix on plone.restapi correcting the format for broken links according the format for valid relations, this is fine and the broken links show up in the panel as expected and like before. Yeah!

ksuess:
For testing purpose https://github.com/collective/example.contenttype can be installed. Then create two instances A and B. Create a relation. Delete B. Voilà a broken relation appears in the panel.

PR #4793, opened 2023-05-19
Accessibility tests in CMS-UI

Wagner3UB:
Initial tests inside Cypress for keyboard accessibility. 

This is a minimal part of the hole and long process. So far, we have conducted a test to replicate the behavior of a user inside the CMS-UI after creating a new component "news item", where the user modifies a content and save it using only the keyboard. In this process, we had fixed some A11Y problems:

- Now it is possible to leave the text block using the Tab key.
- Slate text containers re-arranged to give the tab a proper path (text input → drag & drop button → delete button) 
- Added a general ":focus-visible" to standardize the focus within the application.
- small a11y general fixes

@sneridagh We want to know two things:

- We have created a new folder to run the accessibility test, but we are not sure if this is the right approach.
- The idea is merge these small parts separately or do more work and merge all at once.


sneridagh:
@Wagner3UB all good for me, go ahead with the current implementation.

If it's possible do not introduce changes in the formatting of the less files (take a look at the diffs) probably an IDE formatter run on them and introduced some formatting changes along with legit ones. It's better to group them all in the same PR (if any). Also, these ones were intended, so not much of value in this PR.

Thanks for taking care! Please continue and let me know the outcome and the status.

Issue #4800, opened 2023-05-23
p.a.redirector redirections get no redirection for crawlers (via SSR) and volto redirects to incorrect ++api++ url depending on rewrite rule setup


djay:
## Describe the bug

Really this is two bugs
- if a crawler crawls a page with p.a.redirector redirection (such as manual url management or via changing the short url, it will get a 200 with the content of the target SSR page. ie it won't get a 301 and won't know the page has moved. 
   - this bug is outlined in comment in https://github.com/plone/plone.rest/pull/135#issuecomment-1359971729 by @davisagli but didn't find a ticket for it
- if accessed in a browser, under certain circumstances, the page will load but end up with the wrong url in the browser bar (++api++ added to the url) so if you reload the page it will break. Note this doesn't happen on out-of-the-box plone or 6.demo.plone.org
   - seems to depend on how you do your rewrite rules. externally vs internally. no rewrite rules is broken (results in ++api++ in the url)

Relevant tickets in volto https://github.com/plone/volto/issues/1095 and restapi https://github.com/plone/plone.restapi/issues/181 don't make it clear if the ++api++ is meant to be left in or not. but it makes most sense an api call should redirect to an api call.

## To Reproduce

Out of the box plone with volto

1.  Create a page
2. change the short name
3. access old url using curl
4. you will get 200 not 30x from SSR
5. if using a browser then volto will make a content api call that will then result in a 30x redirect
6. If you use the nginx and rewrite rules then you will get api redirect without ++api++ in it. e.g this reproduces this case      
      ```
      $ curl -H "Accept: application/json" -v http://localhost:8080/VirtualHostBase/http/www.blah.com/++api++/Plone/VirtualHostRoot/test1
      ...
      < HTTP/1.1 302 Found
      ...
      < Location: http://www.blah.com:8080/test2
      ```
7. but if no rewrite rules or internal VHM then redirect is *with* ++api++ e.g.
    ```
    $ curl -H "Accept: application/json" -v http://localhost:8080/++api++/Plone/test1
    ... 
    < HTTP/1.1 302 Found
    ...
    Location: http://localhost:8080/++api++/Plone/test2
    ```
8. This url is directly put into the window.location and if you refresh you get json

It seems that volto relies on the ++api++ being missing in the redirect and if it gets a redirect from an api with ++api++ then it puts ++api++ into the browser url causing the bug

## Expected behavior

- direct access to a changed url should result in a 30x from SSR, not 200.
[...]

Issue #1095, opened 2020-01-08
plone.app.redirector support for Volto


tisto:
It would be nice to have plone.app.redirector support for Volto.

plone.restapi currently returns the content object itself on a redirect path.

Say you have a page "my-page" and you rename it to "my-new-page", then plone.restapi will return the same content object for:

/my-page -> 200 OK

and

/my-new-page -> 200 OK

Volto currently has no way to detect that "/my-new-page" is a redirect.

plone.restapi could return a 301 instead:

/my-new-page -> 301 Redirect

Though, Volto would need to being able to handle this. Right now, as expected Volto show the content of the original object. Which might be fine for the user but causes a problem for search engines who will detect lots of duplicated content.

I am happy to work on the plone.restapi part of this but I need a Volto master like @robgietema or @sneridagh to help me with the Volto part. :)

I would also be fine if Volto would just show a page "This link is outdated, please click on <REDIRECT_LINK>". We do not need a super-sophisticated solution as a first step.

cekk:
It looks like restapi correctly respond with a 301 if you try to call a get with the old url on a renamed object..or am i wrong?

Probably there should be something to fix in Volto right now, because it doesn't seems to work properly: i get and error if i try to access the old url from volto. And there are some calls blocked by cors.

Does anyone started working/looking on this?

ale-rt:
Note that the proper codes are 302 and 307, see https://github.com/plone/plone.app.redirector/issues/8 and https://github.com/plone/plone.app.redirector/pull/22

tisto:
@cekk we started to look into this for a project and it kind of basically works. Though, there are still edge cases to fix and test in production. I am +1 on changing plone.restapi to return 302 instead of 301.

cekk:
ok, i tried last version of p.a.redirector but the api still return a 301..maybe there is still something to do in restapi?

@tisto is there something to see for Volto somewhere? We also need to start looking into it and don't want to do the fix twice if you already started it

tisto:
@cekk if I am not mistaken we used Volto out-of-the-box and just added some redirects and it "kind-of-worked" but not as reliable and consistent as we hoped. Also the current classic p.a.redirector controlpanel did not work at all for us. Therefore we dropped the idea for the time being. Though, I'd be happy to pick it up again and discuss how we can accomplish this.

cekk:
@tisto if you have time, this week i'd like to discuss about that

tisto:
[...]

PR #4821, opened 2023-05-26
SSR redirections

mamico:
this is an attempt to solve #4800 (the part "adjust SSR code to pass on the redirect that it gets from the api redirect") and #1095 issues


davisagli (commented):
307 and 308 are also redirect statuses, and we should probably handle them.

mamico:
> 307 and 308 are also redirect statuses, and we should probably handle them.

Probably. But at the moment I am focused on GET/HEAD and 301/302 which, IMHO, cover most of the cases in question.

davisagli:
Closing in favor of #4834

PR #4831, opened 2023-05-29
Fix search block input clear button doesn't reset the search

iFlameing:


sneridagh (changes requested):
Why can't we just clear the search results (from the store?). I think putting a ` ` space in there is not the way. Did you evaluate it?

iFlameing:
@sneridagh I thought about this we can do that. But there is some question. When we click on clear. It should clear our search query string store and then we reset the content. But I don't think we need that behavior. Once you visit the page there is already a querystring request which lists pages. So, we also have to handle that. I have set the searchtext= `  `  because there is a case where we are checking the value of the search text. Since an empty string is evaluated as boolean false. we keep getting old results. By passing the space string we are making it true and then fetching the result based on your state of the search block. We can discuss this once you are back from holiday.

ionlizarazu:
I think is not that complicated. The problem is here:
https://github.com/plone/volto/blob/master/src/components/manage/Blocks/Search/hocs/withSearch.jsx#L300

When we clear the input with the button the `toSearchText` param is '' so we get the `searchText` param value, wich is the previous value we had before clearing the input.

Try removing the searchText value there.

ionlizarazu:
Maybe we should add tests for this

iFlameing:
@ionlizarazu It is easy I also found out. But we have to change another function also.

I have created a sepearte function and created a new pr: https://github.com/plone/volto/pull/4837

PR #4837, opened 2023-06-01
Fix search block input clear button doesn't reset the search

iFlameing:


sneridagh (approved):

ionlizarazu:
I think this should be managed by the `onTriggerSearch` function. With this modifications the url doesn't change when the last char is removed from the input.

sneridagh:
@ionlizarazu good catch. The idea is to merge this one, then https://github.com/plone/volto/pull/4698, so both have to get along.

@iFlameing could you please take a look?

ionlizarazu:
@sneridagh I would reverse the merging order.
We have deeply tested  #4698 and i think is a good improvement wich is not directly related with this bugfix.
Also, it would be desirable to have more intense tests for the search block to prevent this kind of issues, right? (I'm working on some tests for this PR)

ionlizarazu:
I've pushed some tests that I think are the minimum for the seach block live mode input. We should also test the not live searching mode, but I haven't gotten that far :sweat_smile:

sneridagh:
@iFlameing @ionlizarazu the other PR has been around for a while, could we amend and finish this one?

ionlizarazu:
I've added the pending not live search block tests, and all are passing, but I don't know why is failing the grid-block test

ionlizarazu:
Ready for me @sneridagh @iFlameing @erral

Issue #4838, opened 2023-06-01
Allow deselecting color in ColorPickerWidget


ksuess:
I use the color picker widget for **styles** of the slate text block.
Therefore it is a necessity to be able to de-select a color/style.
I think it is not too bad to be able to de-select a color in a general color picker widget.

ksuess:
Please backport to Volto 16.

PR #4839, opened 2023-06-01
Allow to deselect color in ColorPickerWidget.

ksuess:
fix https://github.com/plone/volto/issues/4838

Would be nice to be able to de-select a color in a general color picker widget.


ksuess:
I think this is the right approach, would you check please, @tiberiuichim, @sneridagh, thanks.

sneridagh:
@ksuess Did you try how does this behave when you have a color by default set? I would do that under a prop, since could be that you really want to have one by default, and do not allow to remove it.

ksuess:
> @ksuess Did you try how does this behave when you have a color by default set? I would do that under a prop, since could be that you really want to have one by default, and do not allow to remove it.

[Respect missing_value of ColorPickerWidget on de-select](https://github.com/plone/volto/pull/4839/commits/2b71f125c0db60afcb56599df98f245f45811fd3)

ksuess:
Please backport to Volto 16.

ksuess:
> @ksuess I tested it out and the defaults work as expected. Nevertheless, what is for `missing_value`? 

With just default, the color value is default on deselect. Only with both default and missing_value it's possible to have a default on creating a new block, and having a fallback color value different from default value on de-select.

```
  const addMyCustomBlockColorStyling = ({ schema, intl }) => {
    addStyling({ schema, intl });

    schema.properties.styles.schema = {
      fieldsets: [
        {
          id: 'default',
          title: 'Default',
          fields: ['Background'],
        },
      ],
      properties: {
        Background: {
          title: 'Stil',
          widget: 'color_picker',
          default: 'blueBox',
          missing_value: 'grayBox',
          colors: [
            { name: 'greenBox', label: 'Ausserdem' },
            { name: 'blueBox', label: 'Bemerkung' },
            { name: 'grayBox', label: 'Wichtig' },
            { name: 'orangeBox', label: 'Achtung' },
          ],
        },
      },
      required: [],
    };

    return schema;
  };
```

PR #4841, opened 2023-06-01
Do not break requires validation on number fields with value 0

cekk:
old validation handle 0 as an empty value, and this breaks number fields that can set value as 0


davisagli (approved):
LGTM

pnicolli (approved):

sneridagh (approved):

Issue #4460, opened 2023-03-04
Refactor class components to functional components


sneridagh:
## Abstract

Refactor all the possible components in Volto that still are class-based components. Convert them to functional components, whenever possible.

## Motivation

Volto still have a number of core components that are React class-based components. This prevents the use of hooks in them, and newest DX on them. 

The idea is also while moving to them, use hooks to abstract the functionality present in Redux and in data fetching so in a near future we can easily switch using the hooks to use other props provider.

## Assumptions

No assumptions.

## Proposal & Implementation

Refactor should be done carefully and providing Cypress (acceptance) tests where they do not exist. The lifecycle of the components should be preserved while watching out for edge cases and UX problems.

## Deliverables

It should be delivered all at once, as a major breaking change. Ideally, we _won't_ change the contract of any of the components, so no new props passed down, no return anything to the tree that we were not returning before, unless it's justified. 

If the contract has to be broken, then it must be properly documented and exposed in the upgrade guide.

The refactored components. Here there is a preliminary list, subject to change:

- [ ] src/components/theme/Comments/CommentEditModal.jsx
- [ ] src/components/theme/Comments/Comments.jsx
- [ ] src/components/theme/App/App.jsx
- [ ] src/components/theme/ContactForm/ContactForm.jsx
- [ ] src/components/theme/Navigation/Navigation.jsx
- [ ] src/components/theme/Sitemap/Sitemap.jsx
- [ ] src/components/theme/Logout/Logout.jsx
- [ ] src/components/theme/Search/SearchTags.jsx
- [ ] src/components/theme/Search/Search.jsx
- [ ] src/components/theme/Register/Register.jsx
- [ ] src/components/theme/PasswordReset/PasswordReset.jsx
- [ ] src/components/theme/PasswordReset/RequestPasswordReset.jsx
- [ ] src/components/theme/View/View.jsx
- [ ] src/components/theme/View/AlbumView.jsx
- [ ] src/components/theme/View/LinkView.jsx
- [ ] src/components/theme/SearchWidget/SearchWidget.jsx
- [ ] src/components/theme/Anontools/Anontools.jsx
- [ ] src/components/theme/Breadcrumbs/Breadcrumbs.jsx
- [ ] src/components/theme/Login/Login.jsx
- [ ] src/components/theme/Header/Header.jsx
- [ ] src/components/manage/Blocks/Maps/Edit.jsx
- [ ] src/components/manage/Blocks/LeadImage/Edit.jsx
[...]

PR #4845, opened 2023-06-06
Refactor Anontools

Tishasoumya-02:
#4460


sneridagh (changes requested):

sneridagh:
@Tishasoumya-02 @nileshgulia1 @JeffersonBledsoe @jackahl I've refactored the `useContent` hook. I'll explain later, but take a look at it.

Tishasoumya-02:
When useContent is destructed , unit test case is giving error , I believe because of the Json structure in the test file 
```
  content: {
        data: { '@id': 'myid' },
        get: {
          loading: false,
          loaded: true,
          error: null,
        },
      },
```

Tishasoumya-02:
When destructing the return value of useContent ,unit tests were failing so reverted it back . I guess the prop Types are doing the work for it 🤔

src/hooks/content/useContent.js
```diff
@@ -0,0 +1,22 @@
+import { useSelector, shallowEqual } from 'react-redux';
+
+export function useContent() {
+  const data = useSelector((state) => state.content.data, shallowEqual);
```
sneridagh, line 4: @tiberiuichim what do you think of this?

tiberiuichim: @sneridagh I'm not sure, in principle is fine, but I have some concerns. 

I'm genuinely curious about all the "ceremony", with multiple hook calls and shallowEqual. Why not just return state.content?

```
const content = useSelector((state) => state.content);
return {
  data: content.data,
  loading: content.get.loading,
  ...
}
```

I think that at this point, while we wait for useQuery, the hook is underpowered. It can't be used for content subrequests, which highlights the other problem: the hook right now only takes care of yielding content info, it doesn't trigger a `getContent` action.

Following this train of thought, I think it's better to either go full useQuery or not at all. If we introduce this, we'll have multiple variations of this type of code in the codebase (old style, useContent, etc), which will all have to be migrated to the new variation of `useContent`.

The counter-argument is that the `useContent` with subrequests is not needed in vanilla Volto, so it's fine to introduce it right now as it is and enhance it with params and options later.

sneridagh: Thanks for the review! Let's try to talk today!

nileshgulia1 (approved):
@Tishasoumya-02 @sneridagh Nice!

sneridagh (approved):

sneridagh:
Congrats! @Tishasoumya-02 The first one! 🎉 Let's continue for the rest!

PR #4848, opened 2023-06-06
Use proper heading tag (depending on the headline) in default listing template

sneridagh:


erral:
https://github.com/plone/volto/blob/master/src/components/manage/Blocks/Listing/schema.js#L35

https://github.com/plone/volto/blob/master/src/components/manage/Blocks/Listing/ListingBody.jsx#L48

stevepiercy (approved):

davisagli (commented):
Semantically, should it even be a heading? It is just an item in a list, not part of the document structure.

ionlizarazu:
> Semantically, should it even be a heading? It is just an item in a list, not part of the document structure.

Even the element should be inside an `<article>` tag for good a11y terms

sneridagh:
The classical problem with headings and a CMS. I'd say that putting it in place a "smart" component won't solve the problem,  we could have edge cases out there. Thinking about it I'd say I lean forward to what @davisagli said. It's a list of elements, which are not part of the doc structure, so they shouldn't be a heading. We can solve it using just classnames for styling.

https://www.a11yproject.com/posts/how-to-accessible-heading-structure/#headings-and-aria%2C-some-discussion

Although it seems sensible indeed that we use the article tag:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/article

it also needs a h* ... so we are in the same place as before... 

TBH, I'm kind of torn in this one.

erral (approved):

davisagli:
There are failing tests

sneridagh:
Next will be green.

davisagli (approved):

PR #4852, opened 2023-06-07
fix(styleMenu): highlight selected block styles

### LINKED ISSUE #4851: Stylemenu styles are not highlighted
**Describe the bug**
When you select a style option in slate styleMenu, The corresponding slate Inlinetoolbar button is not highlighted indicating the styles are applied.
**Bug gif**:

https://github.com/plone/volto/assets/22280901/97f87d77-bc4b-4442-a917-2979358aca9f

**Expected behavior**
It used to work a while ago. The styleMenu button should be highlighted indicating the styles applied.

- Volto Version: latest master

nileshgulia1:
fixes #4851

### DIFF
--- a/packages/volto-slate/src/editor/plugins/StyleMenu/utils.js
+++ b/packages/volto-slate/src/editor/plugins/StyleMenu/utils.js
@@ -93,8 +93,8 @@ export const isBlockStyleActive = (editor, style) => {
       keyName.split('-').filter((x) => x === style).length > 0
     )
       return true;
-    else return false;
   }
+  return false;
 };
 
 export const isInlineStyleActive = (editor, style) => {
--- a/packages/volto-slate/src/editor/plugins/StyleMenu/utils.js
+++ b/packages/volto-slate/src/editor/plugins/StyleMenu/utils.js
@@ -72,20 +72,29 @@ export const toggleInlineStyle = (editor, style) => {
 };
 
 export const isBlockStyleActive = (editor, style) => {
+  const keyName = `style-${style}`;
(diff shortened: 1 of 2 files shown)


nileshgulia1:
Gif: 

https://github.com/plone/volto/assets/22280901/56cc7d29-7d73-4d64-8ad7-efd76ddc909e

packages/volto-slate/src/editor/plugins/StyleMenu/utils.js
```diff
@@ -72,20 +72,29 @@ export const toggleInlineStyle = (editor, style) => {
...
+      if (n.styleName.split(' ').filter((x) => x === style).length > 0) {
+        return true;
+      }
+    } else if (
+      n[keyName] &&
+      keyName.split('-').filter((x) => x === style).length > 0
+    )
       return true;
-    }
+    else return false;
```
avoinea, line 96: @nileshgulia1 I think this else return will make the for at line 86 useless as it will return after the first iteration no matter what :thinking:

avoinea (changes requested):

packages/volto-slate/src/editor/plugins/StyleMenu/utils.js
nileshgulia1: @avoinea Nice catch! I noticed the matched elements in the form of `sn` list has always length one. In that case we won't even need a loop. maybe I'm missing some use case here. 

Anyway, `return false` should be outside of loop.

avoinea (approved):

PR #4854, opened 2023-06-07
Server-side redirects for plone.app.redirector aliases

JeffersonBledsoe:
Solves the Volto part of https://github.com/plone/volto/issues/4800. We're stripping the ++api++ if it is in a location change as we probably don't want to ever navigate directly to a page with ++api++ in it in the browser. Change the action causes the redux store to be correct, but the URL updating occurs earlier in the react-router stack, and so we also manually update the URL to not have ++api++ in it. History is still preserved correctly (++api++ pages aren't in the history stack but we can still go back and forwards normally).

The logic is currently mixed in with the blacklistRoutes middleware while I get some feedback, but this logic should probably be its own middleware for readability.

Note that the `<Redirect>` at https://github.com/plone/volto/blob/87ff1f561788c40809934edc12e2c0c683736522/src/components/theme/View/View.jsx#L217 is causing SSR'd pages to send a 302 rather than forwarding on the status code sent by the REST API. Applying the changes mentioned in the [react-router-dom docs for StaticRouting](https://v5.reactrouter.com/web/guides/server-rendering) to pass the status along to the `<StaticRouter>` should resolve this and can be done in a follow-on PR.


JeffersonBledsoe:
Hi @mamico I've incorporated your changes from https://github.com/plone/volto/pull/4821 into this PR and fixed some of the [issues mentioned by @davisagli](https://github.com/plone/volto/pull/4821#pullrequestreview-1447552722) as well as resolved some other issues I had with client-side and SSR'd content while trialing your PR. I don't think either PR makes sense without the other really which is why I've merged them into 1.

davisagli:
@JeffersonBledsoe This is the new one you opened today. Did you mean to close it?

JeffersonBledsoe:
@davisagli Not sure what happened there, I didn't close it. I had a GitHub bug where i ended up with two branches of the same name which is why i created a second copy of this PR. Mind reviewing this one?

src/components/theme/View/View.jsx
```diff
@@ -203,8 +203,17 @@ class View extends Component {
...
+      : this.props.error?.code;
+    // Checking to see if it's a 3XX HTTP status code. error.status only works on the server
+    if (status?.toString()?.[0] === '3') {
+      const redirectUrl = __SERVER__
+        ? this.props.error.response.headers.location
+        : this.props.error.url;
+      let redirect = flattenToAppURL(redirectUrl);
+      // We can hit situations where we end up being redirected to an api route. We don't want that so lets remove ++api++.
+      redirect = redirect.replace('/++api++', '');
       return <Redirect to={`${redirect}${this.props.location.search}`} />;
```
djay, line 217: you will need to set a redirect code which is the same the one you got from the api call since later restapi will give different codes depending if you are logged in or not

djay (changes requested):

src/components/theme/View/View.jsx
JeffersonBledsoe: Yep, this will be done in a later PR as mentioned in the description

mamico:
> Hi @mamico I've incorporated your changes from #4821 ...

Thanks. I have a busy few days, but on Monday I can help and review everything. Although I think you have a complete view on the whole problem, having experienced it myself, I hope I can make my own small contribution. Talking about `legacy traversing`, (those without `++api++` to be clear) I guess we can consider them unsupported here, right?

src/components/theme/View/View.jsx
```diff
@@ -203,8 +203,17 @@ class View extends Component {
    */
   render() {
     const { views } = config;
-    if (this.props.error && this.props.error.code === 301) {
-      const redirect = flattenToAppURL(this.props.error.url).split('?')[0];
+    const status = __SERVER__
+      ? this.props.error?.status
+      : this.props.error?.code;
+    // Checking to see if it's a 3XX HTTP status code. error.status only works on the server
+    if (status?.toString()?.[0] === '3') {
```
mamico, line 210: Not all the `3XX` are redirects (i.e. 304 Not modified), IMHO is better to use something like  `[301, 302, 307, 308].includes(status)`

src/helpers/Api/Api.js
```diff
@@ -104,6 +108,14 @@ class Api {
                 url: request.xhr.responseURL,
               });
             }
+
+            if (err?.status[0] === 3) {
```
mamico, line 112:
```suggestion
            if ([301, 302, 307, 308].includes(err?.status)) {
```

mamico:
@JeffersonBledsoe @djay @davisagli there are any news on this? thx.

JeffersonBledsoe:
@mamico We’ve been running this in production for a while now, and while it’s mostly been okay, we’ve noticed some weird URLs appearing. I need to verify that this PR isn’t the cause before I’d feel comfortable merging it

djay:
@mamico what we are seeing is that google search console is ending up indexing urls like

https://digitalnsw.pretagov.com.au/components/cards?expand=actions,breadcrumbs,navigation&expand.navigation.depth=2

(where there was an old url of https://digitalnsw.pretagov.com.au/demo-pages/cards)

it might not be change to blame though. We just haven't had the time yet to investigate where these are coming from.

djay:
ok I looked into it. 

This code needs a change to strip off expand or volto related parts of the api call when a redirect happens and only keep the arguments that were part of the original call. I think we can safely say that we don't allow p.a.redirect to turn a url without arguments into one with arguments right @davisagli ?

so
- [ ] get the original request (when in node/ssr) and get the original querystring
- [ ] discard any querystring if an api call resulted in a redirect and replace it with original querystring (if there was one)

mamico:
> ok I looked into it.

Great.

> This code needs a change to strip off expand or volto related parts of the api call when a redirect happens and only keep the arguments that were part of the original call. I think we can safely say that we don't allow p.a.redirect to turn a url without arguments into one with arguments right @davisagli ?
> 
> so
> 
>     * [ ]  get the original request (when in node/ssr) and get the original querystring
> 
>     * [ ]  discard any querystring if an api call resulted in a redirect and replace it with original querystring (if there was one)

Looking at the code, the original query string already seems to be added, probably we just need to remove the query string returned from the backend?

I'll give it a try.

https://github.com/plone/volto/blob/5b17e9c3e44ec1204541192f3875dfcca5082e67/src/components/theme/View/View.jsx#L217
```
      let redirect = flattenToAppURL(redirectUrl);
      // We can hit situations where we end up being redirected to an api route. We don't want that so lets remove ++api++.
      redirect = redirect.replace('/++api++', '');
      return <Redirect to={`${redirect}${this.props.location.search}`} />;
```

src/components/theme/View/View.jsx
```diff
@@ -203,8 +203,17 @@ class View extends Component {
...
+      const redirectUrl = __SERVER__
+        ? this.props.error.response.headers.location
+        : this.props.error.url;
+      let redirect = flattenToAppURL(redirectUrl);
+      // We can hit situations where we end up being redirected to an api route. We don't want that so lets remove ++api++.
+      redirect = redirect.replace('/++api++', '');
```
mamico, line 216: Strip querystring (see https://github.com/plone/volto/pull/4854#issuecomment-1689163422)

```suggestion
      redirect = redirect.replace('/++api++', '').replace(/\?.*/, '');
```

davisagli (approved):
@sneridagh I think this is ready now. The fix is basically the same as Rob's from #5999 but it strips /++api++ from the redirect location if it is there.

JeffersonBledsoe:
Thanks for finishing this!

PR #4867, opened 2023-06-13
Fix exclude from nav widget

cekk:
previous widget was an array and you can't submit it because the field is boolean.


cekk:
damn, you're right! I'll close this one

pnicolli:
I will move my review to the other one then :D

PR #4868, opened 2023-06-13
Enforce max upload size

davisagli:
Add a config setting `maxFileUploadSize`, and check it when files are selected for upload.

If anyone has an idea about how to do this in a more centralized way, please let me know!

### DIFF
--- a/src/helpers/FormValidation/FormValidation.js
+++ b/src/helpers/FormValidation/FormValidation.js
@@ -2,7 +2,7 @@ import { map, uniq, keys, intersection, isEmpty } from 'lodash';
 import { messages } from '../MessageLabels/MessageLabels';
 import config from '@plone/volto/registry';
 import { toast } from 'react-toastify';
-import { Toast } from '@plone/volto/components';
+import Toast from '@plone/volto/components/manage/Toast/Toast';
 
 /**
  * Will return the intl message if invalid
--- a/src/helpers/FormValidation/FormValidation.js
+++ b/src/helpers/FormValidation/FormValidation.js
@@ -1,5 +1,8 @@
 import { map, uniq, keys, intersection, isEmpty } from 'lodash';
 import { messages } from '../MessageLabels/MessageLabels';
+import config from '@plone/volto/registry';
+import { toast } from 'react-toastify';
+import { Toast } from '@plone/volto/components';
 
 /**
  * Will return the intl message if invalid
@@ -369,3 +372,29 @@ class FormValidation {
 }
 
 export default FormValidation;
+
+/**
+ * Check if a file upload is within the maximum size limit.
+ * @param {File} file
+ * @param {Function} intlFunc
+ * @returns {Boolean}
+ */
+export const validateFileUploadSize = (file, intlFunc) => {
+  const isValid =
+    !config.settings.maxFileUploadSize ||
+    file.size <= config.settings.maxFileUploadSize;
+  if (!isValid) {
+    toast.error(
+      <Toast
+        error
+        title={intlFunc(messages.error)}
+        content={intlFunc(messages.fileTooLarge, {
+          limit: `${Math.floor(
+            config.settings.maxFileUploadSize / 1024 / 1024,
+          )}MB`,
+        })}
+      />,
+    );
+  }
+  return isValid;
+};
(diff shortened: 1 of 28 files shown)


src/helpers/FormValidation/FormValidation.js
```diff
@@ -369,3 +372,29 @@ class FormValidation {
...
+        content={intlFunc(messages.fileTooLarge, {
+          limit: `${Math.floor(
+            config.settings.maxFileUploadSize / 1024 / 1024,
+          )}MB`,
+        })}
+      />,
+    );
+  }
+  return isValid;
+};
```
ksuess, line 400: 1. May I suggest to distinguish between image files and other files. 
- Two limits: one for image files, one for other files

2. I think limits should be attributes of the Plone instance, rather than frontend configuration. This info would be available with https://github.com/plone/volto/pull/3537 (assuming that every website has a logo ;-) ) together with an enhancement of the site attributes and an enhancement of the restapi endpoint @site.

ksuess (commented):
my two cents

src/helpers/FormValidation/FormValidation.js
erral: Moreover, if this is to be a frontend configuration, I think this setting should be overrideable per-block type too.

ksuess: > Moreover, if this is to be a frontend configuration, I think this setting should be overrideable per-block type too.

Do you think this is necessary? I think restricting the upload size is just for preventing the users to upload big high resolution photos for small teaser images. It is unrelated to the actual scaling for rendering.

davisagli: @ksuess @erral My goal here is to implement a simple, optional catch-all limit on file uploads for the entire volto installation. I agree there are lots of possible use cases where limits with a more specific scope might be needed, but frankly, my client just wants one limit everywhere and I don't have time to implement all the complexity you are bringing up at the moment.

Fortunately, I think what I've implemented here is forward-compatible with adding more specific limits later, as people have time to work on it. I think the roadmap would look something like this:
1. In the backend, add a way to specify a limit on a specific NamedBlobFile/NamedBlobImage field. If it's not specified, fall back to a sitewide limit from the registry. Expose the field-specific limit in the schema API service, and the sitewide limit in the site API service.
2. Make the volto FileWidget use the limit from the schema when it is specified, otherwise fall back to the sitewide limit from the site service, otherwise fall back to the limit from config.settings.
3. Add a block-level limit to blocks that accept file uploads without using the FileWidget. Use it if specified, otherwise fall back to site and global settings in the same way.

With that plan in mind, are you okay with adding what I have in this PR, as a first step?

sneridagh (approved):

src/helpers/FormValidation/FormValidation.js
wesleybl: @davisagli @ksuess @erral I had to set a limit for the entire portal in the backend. I created a configuration record for this. At first, I wanted to return the limit in something like `@site`. But I realized that if the manager changed this value, some already logged in users would still have the old value. So I created a synchronous request, to be called in `validateFileUploadSize`:

```javascript
const maxImageSize = () => {
  const request = new XMLHttpRequest();
  request.open('GET', '/++api++/@image-size', false);
  request.send(null);
  if (request.status === 200) {
  return JSON.parse(request.response)['max_image_size'];
  }
  return 0;
};
```

Any other better ideas?

I would have to at least use [formatUrl](https://github.com/plone/volto/blob/7694b0f232cc6e23d760bc316e92e9df1a133362/packages/volto/src/helpers/Api/Api.js#L20), but it is not exported. Can we export it?

davisagli: @wesleybl For a file upload limit in the backend, I'm thinking it would be ideal if the backend returned a 413 response (https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/413) and the frontend showed a helpful error message for this. However:
1. It will take quite some care to make sure that the backend checks the Content-Length header before reading the entire request.
2. It's a complex topic because there are often multiple processes handling the request (traefik/nginx, varnish, volto, Plone) and each one might have its own limits on size and time.

I think we can make improvements but it's not a simple topic.

PR #4890, opened 2023-06-16
Refactor Delete

Tishasoumya-02:
#4460


src/components/manage/Delete/Delete.jsx
```diff
@@ -1,22 +1,17 @@
-/**
- * Delete container.
- * @module components/manage/Delete/Delete
- */
-
-import React, { Component } from 'react';
+import { useEffect, useState } from 'react';
 import PropTypes from 'prop-types';
 import { Helmet } from '@plone/volto/helpers';
```
nileshgulia1, line 3: Fix imports please.

src/components/manage/Delete/Delete.jsx
```diff
@@ -32,172 +27,110 @@ const messages = defineMessages({
...
-    pathname: PropTypes.string.isRequired,
-    content: PropTypes.shape({
-      title: PropTypes.string,
-    }),
-    returnUrl: PropTypes.string,
-  };
+const Delete = ({ location, history }) => {
+  const dispatch = useDispatch();
+  const intl = useIntl();
+  const [isClient, setisClient] = useState(false);
```
nileshgulia1, line 33: You can move isClient logic to a new hook as suggested by @JeffersonBledsoe

src/components/manage/Delete/Delete.jsx
```diff
@@ -32,172 +27,110 @@ const messages = defineMessages({
...
-    if (this.props.deleteRequest.loading && nextProps.deleteRequest.loaded) {
-      this.props.history.push(
-        this.props.returnUrl ||
-          this.props.pathname.replace('/delete', '').replace(/\/[^/]*$/, ''),
-      );
-    }
-  }
+  useEffect(() => {
+    dispatch(getContent(pathname.split('/delete')[0]));
+    setisClient(true);
```
nileshgulia1, line 42: set clientState on separate hook on first mount.

src/components/manage/Delete/Delete.jsx
```diff
@@ -32,172 +27,110 @@ const messages = defineMessages({
...
+      </div>
+    );
   }
-}
+  return <div />;
+};
+
+Delete.propTypes = {
+  deleteContent: PropTypes.func.isRequired,
+  getContent: PropTypes.func.isRequired,
```
nileshgulia1, line 119: These proptypes can be removed.

src/hooks/content/useContent.js
```diff
@@ -0,0 +1,8 @@
+import { useSelector, shallowEqual } from 'react-redux';
+
+export function useContent() {
```
nileshgulia1, line 3: You can rebase and fix conflicts as the other one is already merged :)

src/components/manage/Delete/Delete.jsx
Tishasoumya-02: I was thinking of creating a separate PR , if everything is OK then it can be merged and I can pull the change in other working branches

src/components/manage/Delete/Delete.jsx
```diff
@@ -32,172 +25,95 @@ const messages = defineMessages({
...
-   */
-  constructor(props) {
-    super(props);
-    this.onCancel = this.onCancel.bind(this);
-    this.onSubmit = this.onSubmit.bind(this);
-    this.state = { isClient: false };
-  }
+  useEffect(() => {
+    setisClient(true);
+  }, [isClient]);
```
nileshgulia1, line 41: You should remove this dependency. It can cause infinite calls to useEffect.

src/components/manage/Delete/Delete.jsx
```diff
@@ -32,172 +25,95 @@ const messages = defineMessages({
...
-   * Component did mount
-   * @method componentDidMount
-   * @returns {undefined}
-   */
-  componentDidMount() {
-    this.props.getContent(this.props.pathname.split('/delete')[0]);
-    this.setState({ isClient: true });
-  }
+  useEffect(() => {
+    history.push(
```
nileshgulia1, line 48: The if `deleteRequest.loading` and `deleteRequest.loaded` ?

sneridagh:
@tiberiuichim it would be great if we could have your opinion on https://github.com/plone/volto/pull/4890/commits/22a850eaaca3dcce696b776f884aea53d33a75c3 too.

davisagli:
@sneridagh @JeffersonBledsoe My initial gut reaction is that I don't see the benefit of creating individual hooks like useContent, useDeleteContent, etc. It feels like boilerplate that just adds another layer of indirection, and I don't really understand how it would help with moving to useQuery/useMutation in the future.

nileshgulia1:
> 

I would vote for creating a single trigger hook for all API calls that accepts a callback for each action. Though I don't know if it should come with the refactoring process or be the thing of future.

sneridagh:
@davisagli @tiberiuichim @nileshgulia1 @JeffersonBledsoe I'm quite torn on this one.

The initial idea for the hooks façade was from @tiberiuichim who mention it at some point in one of our countless discussions and musing on Volto things. I thought that it was the right moment to implement it since it would allow us to a swift and soft transition when we move to the new data fetching architecture.

Again, the idea is that if the façade exists, then the changes to the components when the new data fetching library is ready, will be as easy as make an add-on that shadows all the façade and replace them with the RQ ones. No other intervention required.

This is, of course, in theory... coming up with a 1:1 API architecture for the façade is complex. But on the paper, it should work. However, this might also change if Suspense is around (then the way of writing the components will also be different), and we will have to face a component change again... 😫

Maybe the reality is that it's too overkill... and that the React ecosystem will force us to rewrite the components again, so the effort is futile.  Coming up with a sane API is complex (as expected), and not without doubts.

<aybe we will have to live with the idea that when we made the change, the components should be also changed (maybe using the `@plone/components` ones)... 

I'd like to know your opinion as well...

/cc @plone/volto-team

src/hooks/content/useContent.js
```diff
@@ -1,21 +1,63 @@
...
+/**
+ *
+ * @param {string|Array} urls Content url(s).
+ * @todo Support URL object being passed in here
+ */
+export function useDeleteContent(urls) {
+  const dispatch = useDispatch();
+  const mutate = useCallback(
+    (urls) => {
+      dispatch(deleteContent(urls));
```
tiberiuichim, line 54: I think, with a return, it would provide more value. 

We need an "interface" to access the inner request state. Did it finish? Was there an error? Is it still loading? etc.

sneridagh:
I'd like to get to an agreement soon. I plan to discuss about it tomorrow during the Volto Team Meeting. This is right now a blocker for @Tishasoumya-02 to continue with the project. 


/cc @plone/volto-team

nileshgulia1 (approved):

PR #4892, opened 2023-06-16
fix: override of breadcrumps, navigation and types when subrequest getContent triggers

### LINKED ISSUE #4891: Get content for subrequests overrides breadcrumbs, navigation and types
If we have apiExpanders for breadcrumbs, navigaton or types and we make a subrequest getContent this will override the states of breadcrumbs, navigation and types

razvanMiu:
Closes #4891

### DIFF
--- a/news/4891.bug
+++ b/news/4891.bug
@@ -1,0 +1,1 @@
+- Fix override of breadcrumps, navigation and types when subrequest getContent triggers - @razvanMiu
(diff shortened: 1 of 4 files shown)


avoinea (approved):

sneridagh:
@pnicolli @davisagli didn't we fix this already?

pnicolli:
Yes, it was indeed already fixed here: https://github.com/plone/volto/pull/4761/files

razvanMiu:
Ohh, wow, I had `volto@16.20.6` in my frontend and in that version there was no check for [subrequests](https://github.com/plone/volto/blob/16.20.6/src/reducers/breadcrumbs/breadcrumbs.js#L48) and when I made my change in e new clone of volto I didn't look again at that [line](https://github.com/plone/volto/blob/f25a6e66df2b226a0606d53639fc42869792ec53/src/reducers/breadcrumbs/breadcrumbs.js#L49). My bad, sorry.

PR #4901, opened 2023-06-20
Support volto 16 on Node 18

davisagli:
Here's a workaround for the hash function issue with webpack 4 on Node 17+ (https://github.com/webpack/webpack/issues/14532). Let's see if the tests say...

If the tests are good, this also still needs a few small updates to docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,20 +53,16 @@ First get all the requirements installed on your system.
 
 ### Prerequisites
 
-- [Node.js LTS (16.x)](https://nodejs.org/)
+- [Node.js LTS (16.x or 18.x)](https://nodejs.org/)
 - [Python](https://python.org/) - See below for specific versions.
 - [Docker](https://www.docker.com/get-started) (if using the Plone docker images)
-
-*UPDATE 2022-10-25*: Since 2022-10-25, NodeJS 18 is in LTS state (https://github.com/nodejs/release#release-schedule). However, due to changes in internal SSL libraries, some Volto dependencies have been deprecated and need to be updated in order to continue working in NodeJS 18, mainly Webpack 4 (see: https://github.com/webpack/webpack/issues/14532#issuecomment-947525539 for further information). You can still use it, but NodeJS should be run under a special flag: `NODE_OPTIONS=--openssl-legacy-provider`. See also Volto's PR: https://github.com/plone/volto/pull/3699 for more information.
 
 The versions of Python that are supported in Volto depend on the version of Plone that you use.
 
 | Plone | Python | Volto |
 |---|---|---|
 | 5.2 | 2.7, 3.6-3.8 | 15.0 |
-| 6.0 (beta) | 3.8-3.10 | 16.0 (alpha) |
-
-At the time of this writing, Volto 16 is still in alpha status, and Plone 6 is in beta status.
+| 6.0 | 3.8-3.11 | 16.0 |
 
 ### Create a Volto project using the generator
 
@@ -229,10 +225,11 @@ JavaScript-centered trainings.
 
 ## Node Support
 
+- Node 18: Supported since Volto 16
 - Node 16: Supported since Volto 14
-- Node 14: Supported since Volto 8.8.0
-- Node 12: Deprecated from Volto 16 onwards. It was supported since Volto 4
-- Node 10: Deprecated from Volto 13 onwards. It was supported since Volto 1 (and its predecessor "plone-react")
+- Node 14: Supported from Volto 8.8.0 - 16
+- Node 12: No longer supported. It was supported from Volto 4 - 15
+- Node 10: No longer supported. It was supported from Volto 1 - 12
 
 ## Browser support
 
--- a/.github/workflows/acceptance.yml
+++ b/.github/workflows/acceptance.yml
@@ -7,7 +7,7 @@ jobs:
     timeout-minutes: 45
     strategy:
       matrix:
-        node-version: [14.x, 16.x]
+        node-version: [14.x, 16.x, 18.x]
     steps:
       - uses: actions/checkout@v3
 
@@ -61,7 +61,7 @@ jobs:
     timeout-minutes: 35
     strategy:
       matrix:
-        node-version: [14.x, 16.x]
+        node-version: [14.x, 16.x, 18.x]
     steps:
       - uses: actions/checkout@v3
 
@@ -115,7 +115,7 @@ jobs:
     timeout-minutes: 45
     strategy:
       matrix:
-        node-version: [14.x, 16.x]
+        node-version: [14.x, 16.x, 18.x]
     steps:
       - uses: actions/checkout@v3
 
@@ -168,7 +168,7 @@ jobs:
     name: Core Basic - Plone 5
     strategy:
       matrix:
-        node-version: [16.x]
+        node-version: [16.x, 18.x]
     steps:
       - uses: actions/checkout@v3
 
@@ -222,7 +222,7 @@ jobs:
     timeout-minutes: 35
     strategy:
       matrix:
-        node-version: [16.x]
+        node-version: [16.x, 18.x]
         python-version: [3.7]
     steps:
       - uses: actions/checkout@v3
@@ -300,7 +300,7 @@ jobs:
     timeout-minutes: 35
     strategy:
       matrix:
-        node-version: [16.x]
+        node-version: [16.x, 18.x]
     steps:
       - uses: actions/checkout@v3
 
@@ -355,7 +355,7 @@ jobs:
     timeout-minutes: 35
     strategy:
       matrix:
-        node-version: [16.x]
+        node-version: [16.x, 18.x]
     steps:
       - uses: actions/checkout@v3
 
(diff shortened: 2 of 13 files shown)


mauritsvanrees:
My thoughts on this.

-------------

For simplicity let's ignore Plone backend and Classic UI for a moment.  This means that once Volto 17.0.0 final is released, we will release Plone 6.1 at the same time.

Reminder: here is the Plone release schedule: https://plone.org/download/release-schedule
**We currently do not mention Volto there, and we totally should change that.**

What do we say there currently?

1. Each minor release gets at least one year of maintenance support.

Plone 6.0 was released in December 2022 with Volto 16.  So Volto 16 should get maintenance support until at least December 2023.

2. On top of this, after the next minor version is released, there is an overlap of half a year where the previous release still gets maintenance support.

Say Plone 6.1 / Volto 17 is released in October 2023.  Volto 16 should then be maintained with bugfix releases until April 2024.

3. Each minor release gets at least two years of security support.

So Volto 16 should get security support until at least December 2024.

4. There is an overlap of at least one year between minor versions. This gives you time to migrate to a new minor version.

If Plone 6.1 / Volto 17 is released in October 2023, this last rule makes no difference: Volto 16 should get security support until December 2024.

If Plone 6.1 / Volto 17 is released in March 2024, this means Volto 16 should get security support until March 2025.

This is how I understood and intended what we agreed around the time that Plone 6.0 was released.  **Does this still sound reasonable?**

------

I *would* say this means that if Volto 17 is ready, **we should not delay a 6.1 release for months** waiting for some changes in backend or Classic UI.  A new Volto is a fine reason to have a 6.1 release.
This goes the other way as well: if enough backend/classic changes are available to consider making a 6.1, we should not wait months until Volto 17 is ready.

The other side of the coin is that if this means a new minor Plone release every 4 or 5 months, then people will get confused, and we will have too many parallel branches to maintain.  So a bit of waiting is stil okay.

----------

On the backend, if Python itself drops support for Python 3.8 (in October 2024), then Plone drops support for it as well.  This means Plone 5.2 is end of life in October 2024.

I expect the same to be true for Volto.  **If a node/npm version is no longer supported upstream, then Volto can drop support as well.**

The other side of that coin is that this should not be an excuse to actively break Plone/Volto on unsupported upstream versions.  For example I do not want to add `python_requires>=3.9` in `setup.py` in package branches that are used on Plone 6.0.  And if unintentionally syntax is introduced that only works on Python 3.9 or higher, then I would consider that a regression that should be fixed.

---------

About LTS, Long Term Support.

Years ago some people in the Plone community decided that we should not use the term "LTS".  We used it for Plone 6 anyway, I think with as main reason: this is Plone, and we have always supported Plone versions for way longer than other similar systems, so we have earned the right to use this term.

We *did* mean LTS to mean: the Plone 6 series is supported, **not** the 6.0 series per se.  I do not remember if we said this clearly during the State of Plone talk at the Plone conference 2022 in Namur.
It *is* what we say in the release schedule: major releases get long support, but minor versions get less long support.  **You should be on the latest minor version to have the longest support.**

In my mind it is basically the [Django release schedule](https://www.djangoproject.com/download/): the Django 4 series as a whole is supported for a bit over 4 years, but to get this full range of support, you need to be on the latest minor release, Django 4.2, which is an LTS release.  Of course Django has a much smaller core, making it easier to time box releases every nine months.

-----

Summary:

**For Volto I would say it should have the same maintenance/security support as Plone minor versions.**

And once we know that for example Plone 6.2 will be the last minor release in the Plone 6 series, then I hope that for the Volto version that belongs to 6.2, we can support it longer.

stevepiercy (approved):
Thank you for the docs!

sneridagh (approved):

tisto:
Regarding this PR: @mauritsvanrees already pointed out that semantic versioning allows dropping dependencies that run out of security updates. We should do this for Volto and merge this PR. If dropping a dependency requires upgrading a core component (e.g. having to upgrade webpack because of a new Node version), we should do a breaking change within the existing Volto version. Not sure if semantic versioning officially allows this. Though, this is one of the known shortcomings of semantic versioning (if you are on Volto 16 and Volto 17 is out, there is no way you can do a breaking change in Volto 16).

I agree with @mauritsvanrees, @sneridagh, and @davisagli that in this case, there is no other way than to introduce a breaking change and to "violate" semantic versioning. The alternative would be to not do anything at all and leave a branch without security patches (that we promised to support).

We should try to avoid any breaking changes if possible, which is exactly what @davisagli did in this PR.

--

My thoughts on the overall release schedule topic (after a discussion with @sneridagh and @davisagli yesterday):

Before going into the details, we should discuss the goals of our release schedule and policy:

1) protect the release team from burning out by having to support multiple versions of our stack at the same time, we have to be clear about what we are able to do and what not
2) provide our users (agencies/clients) with a clear and reliable roadmap they can rely on and plan for the future, without forcing them to one single option

Those two goals are somehow contradictory. The ideal world for the release team and the developers would be that we only support the latest version and we do not care about anything else (developing new stuff is fun!). The ideal world for the agencies/clients would be that we support all versions (or at least give them a wide variety of choices they can freely choose from).

My gut feeling is that the best we can do as the Plone release team is offering a choice between two "flavors" (let's ignore the naming for now):

- a "latest" version where you get the latest features, the most recent node/python version
- a "stable" or "conservative" version that gives you time to upgrade (you could call this LTS)

If we take into account, that we plan to do a new Volto release every half year, or at least every year (@sneridagh we need to come up with a rough forecast of how often we want/need to do a breaking Volto release) we will end up with 1-2 Volto releases per year. Note that with semantic versioning, we are not in full control of when a new version will have to be released. We just bump a version if we introduce a breaking change.

If we are serious about a new Plone minor release for every Volto release, this means in a three-year period (that we and our clients are used to so far)we will have between three and six Volto releases.

This means in a two to three-year period of a major Plone release, the release team would have to support between two and six Volto versions at the same time.

I don't see how this could work if we tightly couple Plone releases to Volto releases. Maybe it is a bit less than what I calculate because the development will slow down a bit. Though, we have to take into account that the development speed of Volto and the Plone backend will be different and that we are using semantic versioning to indicate breaking changes properly.

This means that there could be lots of Volto releases in a short period of time and there might be Volto releases where we don't want to have a Plone minor release (e.g. maybe because we do not have any feature that we can advertise).

As said many times before. We have to accept the fact that Volto will evolve faster than Plone used to, and we shouldn't tightly couple Volto releases to Plone releases. We have plone.restapi to give us the flexibility that we did not have in the past.

For clients/agencies we would currently offer them two options (with the proposed two years release schedule):

a) stay with the current release and run out of security updates within two years
b) upgrade every two years to the latest version

One of the key USPs when selling Plone in the past was stability. We were able to provide more stability and longer release cycles than any competitor on the market. When it comes to larger clients (e.g. government agencies/universities/research institutions) they plan in time frames of 5-10 years and on a large code base, an upgrade is an expensive and error-prone process. Smaller clients often do not have the budget to upgrade every two years.

One could argue that having security patches on the backend is sufficient from a security point of view. Though, I wouldn't want to bet on this when it comes to decision-makers or government security agencies (that scan websites regularly for outdated components and then tell our clients about it).

There are clients that are used to upgrades of their websites or even full relaunches every 2-3 years. I never encountered a single client like this in the Plone world (no matter if they are small or big). When I approach a client after 2-3 years and offer them an upgrade, the normal reaction is "we just finished our relaunch". This is fine, as a company we do not have to or want to offer our clients frequent upgrades. Though, if the Plone release schedule forces us to do so, and does not give us and our clients any other options, we will lose those clients.

I think it is essential for our current client base and our marketing that we keep our stability USP when pitching Plone.

My fear is that if we apply the current Plone approach of release management to Plone 6 with Volto, we will end up in a situation where we the release team has to support more versions than ever and we force our clients to do upgrades in a way higher schedule than what they are used to. We would also limit the development of Volto to fit into what we do for Plone Classic packages, which makes semantic versioning (bumping a version if you are breaking something) very hard.

We discussed "the business side of Plone" at Beethoven Sprint and it turns out Eau de Web and Red Turtle have a different approach than kitconcept, when it comes to upgrades. Most agencies currently just "stick with the version" they started with. This means if we do not come up with a clear schedule what version we consider "LTS" (or any other name) and each company decides on its own. We might end up with Plone companies working on their own branches of Volto on their own (seems this is already the case somehow).

This is an option and companies can offer their commercial service to do upgrades on outdated Plone/Volto versions. Though, I would prefer to have a release schedule that would allow us to work together on an LTS version, rather than having company-specific LTS versions (and not calling them LTS because we want to avoid the term). I want a community-driven mutual effort (e.g. on an LTS release) and exposing our strengths rather than a commercially driven effort and making us look worse than our competitors.

We could drop the LTS term and go back to what we are used to. Meaning, we support two major versions at the same time. Though, we do not have this option for Plone 6 because Plone 5.2 will run out of security support soon.

If we want to continue to support two major versions, we would need a replacement for Plone 5.2. Soon we will only support one major version with a possible two years upgrade cycle. This is way faster than what we are used to and we cut the option to just stick with the "older version" for fewer upgrades that we had in the past.

This topic is too complex and important to discuss in a Volto ticket. @mauritsvanrees @sneridagh can we get together this week or next week to schedule a call to write down what we discussed at Beethoven Sprint and come up with a proposal that we can discuss with the community?

README.md
```diff
@@ -229,10 +225,11 @@ JavaScript-centered trainings.
 
 ## Node Support
 
+- Node 18: Supported since Volto 16
```
wesleybl, line 228: This needs to be updated in master.

PR #4905, opened 2023-06-22
Remove mention of LTS in Volto

### LINKED ISSUE #4904: Remove any mention of LTS
Instances of LTS still exist in the Volto docs and need to be replaced with new language to better describe our true intent. See the note in:

https://6.docs.plone.org/volto/upgrade-guide/index.html#upgrading-to-volto-16-x-x

There may be other instances throughout the docs. @sneridagh

stevepiercy:
Fixes #4904


mauritsvanrees (approved):
Fine for me, but it is up to Victor.

One way that we could still regard Volto 16 as LTS, is when we make Volto 17 a developer/unstable release to try new things out, and then Volto 18 a stable LTS release again, like node and Linux do: uneven versions are unstable, even are stable.  I think this idea was floated briefly during the Beethovensprint, but Volto has not done this so far, and I don't think it will.  Unless this changes, LTS seems not an ideal term.

tisto (changes requested):
Please see my comments: https://github.com/plone/volto/pull/4901#issuecomment-1603893233

stevepiercy:
Do we have a decision on using the term "LTS"?

sneridagh:
@stevepiercy not yet, I don't expect this to happen before we see each other again in the PC I'd say...

sneridagh (approved):
However, and for the record, I approve this.

PR #4662, opened 2023-04-06
TypeScript support in core

sneridagh:
So, let's put this on the table and discuss it during the Beethoven Sprint.

To be clear: I'm (still) far from being sold on TypeScript, and have profound doubts on how it could benefit Volto in short/mid term (as I always did). For a number of reasons:

- Plone/Volto developers base are not familiar with it,
- let alone old Plone community members who wants to adopt Volto for their next projects.
- The community lacks of "TypeScript Wizards(TM)" to rely on when things get weary.
- Problems derived from adding an additional abstraction layer to the equation outweighs all the possible benefits that it could bring, and a whole new category of errors (typings) to solve, instead of focusing in features and getting things done.
- "Type safety" is still a concept that, personally speaking, is overrated and often misunderstood by "hurray, my IDE autocompletes everything". Unit, acceptance testing and documentation are as important as "type safety".

_However_, once named the "cons", there are also some "pros". 

- Best way to document code internally, signatures, props, config object, etc.
- Could be the entry point and the enabler of some other complex things.
- The whole ecosystem is migrating to it, and the typings permeate the IDEs and your code.
- Volto is becoming stable enough to start introducing it.

Summarising, the proposal is *not* to migrate Volto to TypeScript, but having the ability to use it here and there in core if appropriate in some specific areas to gain knowledge and confidence as a tool, and artefact, not as a must.

I know could be a drawback and might be chaotic to have some things in JS and others in TS. As said, let's discuss it during the sprint, and see how it goes.

/cc @mtoepfl I'd like to know your feedback about the PR, if you please could take a look, it would be great!

Regarding implementation details, I added a `tsconfig.json`, adding TypeScript support for `.ts` files only as Razzle already supports it. In the ESlint side, I replaced the (deprecated) `babel-eslint` with `@babel/eslint-parser`, and use the TypeScript ESlint parser only on the `.ts` files. So effectively this way Volto supports both JS and TS at the same time.

*UPDATE*: I fixed some ESlint issues that the new parser and rules was complaining about. The important changes related to this PR are in the root.

*UPDATE (2023-04-17)*: I've managed to make the change non-breaking for 16 series: https://github.com/plone/volto/pull/4705 Undoubtedly, there could be backport issues still, but it's doable.

PR #4917, opened 2023-06-24
Remove anonymous function calls. Remove default exports from.

sneridagh:
backported from #4662


davisagli (approved):

davisagli:
There's an acceptance test which failed; trying it again

Issue #4645, opened 2023-04-03
UI Bug : After clicking on the submit button in user management module ,the submit button remains in focus


hritikd3:
**Describe the bug**
When submitting the CMS user management form, the submit button remains in focus and an additional outline appears around it. This creates a visual inconsistency and may give the impression that the button is still active or require additional action, leading to confusion for users

**To Reproduce**

1. Go to 'Go to 'https://volto.kitconcept.com'
2. Click on 'login-> profile icon->site setup'
3. Scroll down to 'users section and try submitting the button , the submit button will still on focus '

**Expected behavior**
When submitting the CMS user management form, the submit button should lose focus and any additional outline should disappear. This would create a consistent and clear visual feedback that the form has been successfully submitted and any further action is not required.

**Screenshots**
![image](https://user-images.githubusercontent.com/86347286/229590810-bd373526-85de-4856-ad3c-67176dd02d6b.png)

**Software (please complete the following information):**

- OS: [Windows 11]
- Browser [chrome]
- Volto Version [e.g. 8.5.0]
- Plone Version [e.g. 5.2.2]
- Plone REST API Version [e.g. 7.0.1]

hritikd3:
Working on this issue.

PR #4922, opened 2023-06-27
 #4645 issue: Updated button.overrides

Shivaji2808:
Updated the code so that the submit button will become red when clicked instead of the previous buggy blue circular shadow it was showing before.


davisagli:
This is not the correct fix. We need to remove the box-shadow from the focus style for circular buttons only. This is removing the box-shadow from all buttons, and adding a new color which is not wanted.

PR #4946, opened 2023-07-05
Use all the apiExpanders in use, so we perform a single request for getting all the required data.

sneridagh:


tiberiuichim (approved):

tiberiuichim:
I don't know if there's an extra check for duplicate appextras, some projects may have added themselves. A word in the upgrade guide is needed, for sure.

news/4946.feature
```diff
@@ -0,0 +1 @@
+Use all the apiExpanders in use, so we perform a single request for getting all the required data. @sneridagh
```
pnicolli, line 1: I am no Steve, but I will still suggest a minor change just to avoid the repetition of the word "use" and to clarify 😛 

```suggestion
Use all the available apiExpanders, so we perform a single request for getting all the required data. @sneridagh
```

src/config/index.js
```diff
@@ -76,6 +76,8 @@ let config = {
     publicURL,
     apiPath,
     apiExpanders: [
+      // Added here for documentation purposes, addded at the end because it
```
pnicolli, line 79:
```suggestion
      // Added here for documentation purposes, added at the end because it
```

src/config/index.js
```diff
@@ -213,6 +215,22 @@ let config = {
...
+  ...config.settings.apiExpanders,
+  {
+    match: '',
+    GET_CONTENT: ['breadcrumbs', 'actions', 'types'],
+  },
+  {
+    match: '',
+    GET_CONTENT: ['navigation'],
+    querystring: {
+      'expand.navigation.depth': config.settings.navDepth,
```
pnicolli, line 229: Does this work if final projects change `navDepth`? I am not entirely sure, or I don't get why it works at the moment.

sneridagh: You are correct, it won't... since it's passed once, with the "old", default value... Not an easy one :/

Suggestions?

davisagli: @sneridagh An idea: make it possible to pass a function that constructs the querystring, instead of a prebuilt querystring object.

sneridagh: @davisagli see https://github.com/plone/volto/pull/4946/commits/3bcffd6ef10b98754149f4f5b2651606a379d4ec

sneridagh:
@davisagli for some reason the Plone 5 expander for types is not working? :/ How that could be?

src/config/index.js
sneridagh: @pnicolli it turns out we can't because we haven't set `config.settings.navDepth`, unless we assume we are going to sync them manually (I guess that the default is not going to change that much).

sneridagh:
@davisagli @pnicolli ready.

PR #4947, opened 2023-07-05
Experimental feature flag for new add block button enabled by default.

sneridagh:
How we did we would do this?

- Next major (17) we flip the experimental flag to `true`
- In two majors (18) we make it default? (I hope that by then we have Quanta Toolbar around, tbh).


rotavio:
We are using some addons that allows adding blocks inside other blocks, like columns, tabs and accordion. There are some issues with these blocks when the experimental feature is enabled.

Is the "block inside block" feature native or created by the addons team?

Is there something like this native to Volto? If no, are there plans for it?

rotavio:
Thank you for the information. I understand.
What I mean by issues are not exactly errors, but little annoyances when adding blocks. You have to add a text block. The add button appears only on the last text block of the array.
But i understand how this is not a volto issue. Thanks!

sneridagh:
@rotavio It would be great if you could provide exact instructions for reproducing the annoyance, and screenshots/video. Which add-on are you working with? etc. Thanks!

avoinea (approved):

sneridagh:
@plone/volto-team I had further discussions regarding this matter with a couple Team members who brought up good points:

1.- We really haven't yet a proper alpha testing for including the feature by default

This is utterly true, and I didn't realise this has to be the case always. So doing it now at the end of the alpha phase of 17 is useless.

2.- The feedback is still quite few, and a proper alpha phase would have helped. Mostly because of 1.

3.- We still are unsure of the final shape that the Experimental / Future / Feature flags should have, we decided not to change the current shape

So I propose to slate this until 18, the first alpha would have it enabled by default (in the shape of the current experimental flag, but turned on). During beta/RC/final release, and if we decide to have it as a default feature, then it should abandon the form of the "experimental" flag and be just another "settings" key (turnable off).

PR #4955, opened 2023-07-08
Refactor Toolbar/More component

Tishasoumya-02:
#4460


src/components/manage/Toolbar/More.jsx
```diff
@@ -209,346 +169,285 @@ class More extends Component {
...
+                      color="#e40166"
+                      title={intl.formatMessage(messages.removeWorkingCopy)}
+                    />
+                  </button>
                 </li>
               </Plug>
             )}
-            <Plug pluggable="toolbar-more-menu-list" id="rules">
-              {rulesAction && (
+            {content.working_copy && !content.working_copy_of && (
```
nileshgulia1, line 418: I'm a bit concern about these changes. The rendered jsx should not differ in any case. Please make a side by side comparison. Also note, I just rebased with master.

src/hooks/content/useContent.js
```diff
@@ -0,0 +1,28 @@
+import { useSelector, shallowEqual } from 'react-redux';
```
nileshgulia1, line 1: Also, I think we agreed to remove custom hooks.

nileshgulia1 (changes requested):

PR #4960, opened 2023-07-08
Refactor Blocks/Video/Edit component 

Tishasoumya-02:
#4460


src/components/manage/Blocks/Video/VideoSidebar.stories.jsx
```diff
@@ -0,0 +1,60 @@
...
+          locale: 'en',
+          messages: {},
+        },
+      }}
+    >
+      <div id="toolbar" style={{ display: 'none' }} />
+      <IntlVideoSidebarComponent
+        {...args}
+        data={{}}
+        properties={{
```
nileshgulia1, line 22: @Tishasoumya-02 I'm not sure this is the correct implementation. The preview_image comes from block schema in this case but not from content-type metadata. Also a storybook for Video block is sufficient in this case instead of Sidebar.

nileshgulia1 (changes requested):

nileshgulia1 (approved):

iFlameing:
@sneridagh I think we should add default props and prop types to the functional components also. What do you think?

packages/volto/src/components/manage/Blocks/Video/Edit.jsx
```diff
@@ -29,183 +22,122 @@ const messages = defineMessages({
...
-   * @method onChangeUrl
-   * @param {Object} target Target object
-   * @returns {undefined}
-   */
-  onChangeUrl({ target }) {
-    this.setState({
-      url: target.value,
-    });
-  }
+const Edit = React.memo((props) => {
```
davisagli, line 25: I think using React.memo here is wrong, because it does a shallow compare on the arguments, so it will miss re-rendering sometimes when it is needed.

davisagli (changes requested):

davisagli (approved):

PR #4963, opened 2023-07-10
Refactor Block/Table Cell component

Tishasoumya-02:
#4460 
error->draftJsLibIsSoftNewlineEvent is undefined


src/components/manage/Blocks/Table/Cell.jsx
```diff
@@ -1,188 +1,126 @@
...
-   * @param {object} editorState Editor state.
-   * @returns {undefined}
-   */
-  onChange(editorState) {
-    this.setState({ editorState }, () => {
-      this.props.onChange(this.props.row, this.props.cell, editorState);
-    });
-  }
+  const onChange = (editorState) => {
+    setEditorState(onChange(row, cell, editorState));
```
nileshgulia1, line 74: @Tishasoumya-02 It's an unnecessary state reset here. The setEditorState doesn't accept a callback like setState. You need to explicitly listen for state change inside a useEffect.

nileshgulia1 (changes requested):

nileshgulia1:
TODO: all other components in Blocks/Table/ should be refactor in this PR

davisagli:
This was the DraftJS-based table block, which was removed from Volto, so I'll close the PR.

PR #4966, opened 2023-07-11
Exclude video markup from text output

stevepiercy:
For the Plone Nuclia project, we are experimenting with various outputs for easier indexing. This PR excludes videos from only the text output from the `make text` command. It affects only the text builder.


stevepiercy:
I'm going to break the rule of requiring an approval before merge. Apologies, but this PR can't wait, and it has no affect on docs other than `make text` will now work.

davisagli:
@stevepiercy Makes sense to me

PR #4967, opened 2023-07-12
Upgrade bundlewatch to 0.3.3

wesleybl:
This is needed to update `axios` and avoid vulnerabilities in `axios` < 0.21.3. See:

https://security.snyk.io/package/npm/axios


davisagli (approved):
This looks fine, but we should move this to `devDependencies`. There's no reason to require everyone using volto to install a package that is used as an aid for development of volto.

wesleybl:
> This looks fine, but we should move this to devDependencies.

@davisagli done.

davisagli (approved):

PR #4983, opened 2023-07-18
Move eslint to devDependencies

wesleybl:


sneridagh:
@wesleybl this being in deps has a reason :) How are you going to install ESlint now in the projects if it's not as a direct dependency? Like this one, there is a lot of deps in the same situation... the duality library/framework of Volto is difficult, and there is nothing out there like it. We could move it to its own package (as Razzle and others do), then install that package in the project, but we never made that move.

wesleybl:
@sneridagh @davisagli wouldn't it be the case if you put these packages in the generator and put an update note?

sneridagh:
The thing is that maintain the ones in the generator is a pain. The ones that are there is because it's strictly necessary for other reasons. We would like to reduce that list, instead of increasing it.

The ideal scenario would be to have such a dev package that we mentioned, and a code-mod like script/process to update projects automatically.

wesleybl:
Ref: https://github.com/plone/volto/pull/5872

PR #4984, opened 2023-07-18
Add XSendfile headers to files and images middleware

instification:
Existing middleware for files and images only add specific headers to the response.

When using [collective.xsendfile](https://github.com/collective/collective.xsendfile), upstream servers rely on the X-Accel-Redirect or X-Sendfile headers.

This PR:
 - adds `x-sendfile` and `x-accel-redirect` headers to the list that are added to the response. 
 - normalises `files.js` to use `request.headers[header]` instead of `request.get(header)` which ensures comparisons are case insensitive.

Main reason for using xsendfile is that middleware responses get buffered in memory by node, which for large files creates 2 issues:

 - Increased memory usage
 - Noticeable delay before the file is served to the client while it is buffered from the backend

xsendfile means that only a tiny response is sent and the serving of large files can be handled by nginx or other XSendfile capable proxies.

### DIFF
--- a/news/4984.bugfix
+++ b/news/4984.bugfix
@@ -1,0 +1,1 @@
+Add XSendfile headers to files and images middleware @instification
--- a/src/express-middleware/files.js
+++ b/src/express-middleware/files.js
@@ -2,13 +2,13 @@ import express from 'express';
 import { getAPIResourceWithAuth } from '@plone/volto/helpers';
 
 const HEADERS = [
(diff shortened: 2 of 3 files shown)


sneridagh (approved):
LGTM, @davisagli ?

davisagli (approved):
Good idea.

Though, I guess it depends on the frontend webserver and the backend being on the same server, or at least having access to files at the same path.

Issue #4982, opened 2023-07-18
Listing block default summary view uses headline html tags for its listing


tiberiuichim:
**Describe the bug**

We should use a `<ul>` tag for the listing block, rather then the semantically-significant hx tags, as they can cause confusion in people relying on the page structure with headlines that function of section headers.

Another issue: in Plone classic the listing used to have significant colors, based on review state. It's useful, we should bring that back.

davisagli:
> We should use a <ul> tag for the listing block

Agree -- but it's a breaking change.

> Another issue

Please create another github issue then. It's unlikely that they will be fixed at the same time, and unlikely that someone who is working on one of them wants to follow the discussion of both in the same place.

JeffersonBledsoe:
Just looking over #6318 which is linked to solve this. I agree on the use of the `<ul>` tag for the summary, however, I think being able to set a heading for the title of each element in the listing should be a function we retain. It's fairly common practise (e.g. see the [search results page on the W3C website](https://www.w3.org/WAI/search/?q=test) or the [services and information section on the GOV.uk website](https://www.gov.uk/)) as the title and descriptive text are related. I see we already have a "Headline level" setting, could we add another option for "item heading level"? We could compute the item heading level from the headline level automatically (it should always be one below), but we may not always have a headline to actually compute this from

PR #4994, opened 2023-07-20
#4982 - replaced listing block summary template view <hx> tags with <ul> tags

hazigabriel:
As [mentioned here](https://github.com/plone/volto/issues/4982) I have updated the html tags for the listing items when using the Summary variation for the Listing block and updated the styling to match the previous <hx> styling. I have also added a changelog entry due to this being a breaking change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,10 @@
 
 ## 17.0.0-alpha.20 (2023-07-18)
 
+### Breaking
+
+- Replaced `h3` html tags from the Listing Block Summart variation with `ul` tags as can cause confusion in people relying on the page structure with headlines that function of section headers. @tiberiuichim [#4982](https://github.com/plone/volto/issues/4982)
+
 ### Feature
 
 - Use all the apiExpanders in use, so we perform a single request for getting all the required data. @sneridagh [#4946](https://github.com/plone/volto/issues/4946)
--- a/src/components/manage/Blocks/Listing/SummaryTemplate.jsx
+++ b/src/components/manage/Blocks/Listing/SummaryTemplate.jsx
@@ -27,7 +27,9 @@ const SummaryTemplate = ({ items, linkTitle, linkHref, isEditMode }) => {
             <ConditionalLink item={item} condition={!isEditMode}>
               <Component componentName="PreviewImage" item={item} alt="" />
               <div className="listing-body">
-                <h3>{item.title ? item.title : item.id}</h3>
+                <ul className="listing-item-ul">
+                  <li>{item.title ? item.title : item.id}</li>
+                </ul>
                 <p>{item.description}</p>
               </div>
             </ConditionalLink>
--- a/theme/themes/pastanaga/extras/main.less
+++ b/theme/themes/pastanaga/extras/main.less
@@ -645,6 +645,14 @@ body.has-toolbar-collapsed .mobile-menu {
   transition: transform 0.5s cubic-bezier(0.09, 0.11, 0.24, 0.91);
 }
 
+.listing-item-ul {
+  padding: 0;
+  margin: 0;
+  font-size: 1.3em;
+  font-weight: 500;
+  list-style: none;
+}
+
 // Deprecated as per https://github.com/plone/volto/issues/1265
 // @import 'utils';
 @import (multiple) '../extras/fonts';


stevepiercy (changes requested):

CHANGELOG.md
```diff
@@ -10,6 +10,10 @@
 
 ## 17.0.0-alpha.20 (2023-07-18)
 
+### Breaking
+
+- Replaced `h3` html tags from the Listing Block Summart variation with `ul` tags as can cause confusion in people relying on the page structure with headlines that function of section headers. @tiberiuichim [#4982](https://github.com/plone/volto/issues/4982)
```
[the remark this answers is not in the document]
hazigabriel: Thank you for the documentation provided, I will be back with updates once I will complete the process of singing the Plone Contributor Agreement.

src/components/manage/Blocks/Listing/SummaryTemplate.jsx
```diff
@@ -27,7 +27,9 @@ const SummaryTemplate = ({ items, linkTitle, linkHref, isEditMode }) => {
             <ConditionalLink item={item} condition={!isEditMode}>
               <Component componentName="PreviewImage" item={item} alt="" />
               <div className="listing-body">
-                <h3>{item.title ? item.title : item.id}</h3>
+                <ul className="listing-item-ul">
+                  <li>{item.title ? item.title : item.id}</li>
+                </ul>
```
erral, line 32: this doesn't work.

There should be just one `<ul></ul>` and each item rendered should be a `<li></li>`

Here we have a <ul> per item.

The <ul> should go above the `.items.map((item)` and everything inside it should be a `<li>`

erral (changes requested):

src/components/manage/Blocks/Listing/SummaryTemplate.jsx
erral: Moreover should we use plain <ul> <li>-s or should we use [Semantic UI react Lists](https://react.semantic-ui.com/elements/list/)?

this way we can even have a `<List.Header>` with the title and so on.

We can even use the [Component Registry](https://6.docs.plone.org/volto/configuration/component-registry.html) register the List component there, and then the customization is way easier.

hazigabriel: thank you for the feedback, I have applied changes using Semantic UI components

news/4994.breaking
```diff
@@ -0,0 +1 @@
+ Replaced `h3` html tags from the Listing Block Summart variation with `ul` tags as can cause confusion in people relying on the page structure with headlines that function of section headers. @tiberiuichim [#4982](https://github.com/plone/volto/issues/4982)
```
stevepiercy, line 1: Keep it brief and give yourself credit.

```suggestion
Replaced `h3` HTML tags from the Listing Block Summary variation with `ul` tags. Fixes #4982. @hazigabriel 
```

stevepiercy (changes requested):

stevepiercy (approved):
Thank you! A maintainer should review the code. I reviewed only this news entry.

theme/themes/pastanaga/extras/main.less
```diff
@@ -645,6 +645,12 @@ body.has-toolbar-collapsed .mobile-menu {
   transition: transform 0.5s cubic-bezier(0.09, 0.11, 0.24, 0.91);
 }
 
+.summary-listing-item-header {
+  color: #007eb1 !important;
```
tiberiuichim, line 649: please use less vars. Also, absolutely no `!important`. See https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity for more info on how to solve the problem.

tiberiuichim (changes requested):

theme/themes/pastanaga/extras/main.less
hazigabriel: I have added a new variables in main.variables for the title color and fixed this, thank you

tiberiuichim:
- try not to force push, it's a smell if you have to do it
- solve conflicts

tiberiuichim:
Nice. If possible, let's go one step further. The state colors are defined here 

https://github.com/plone/volto/blob/03cd36dfa82de3a03b2f88a6b59ceeea1bc99292/src/config/Workflows.js#L2-L9

available as `config.settings.workflowMapping`. Each item in the listing may have a `review_state` field. Then you can use that to set a style prop.

tiberiuichim:
Following discussions with Victor, we should take inspiration from his other PR https://github.com/plone/volto/pull/4848 and use unstyled classnames instead of hardcoded colors

theme/themes/pastanaga/extras/main.less
```diff
@@ -650,6 +650,12 @@ img.responsive {
   height: auto;
 }
 
+#main .summary-listing-item-header {
+  color: @summaryItemTitle;
```
tiberiuichim, line 654: See if it's possible to use some form of less inheritance for this selector (inheriting from the original h3). https://coderwall.com/p/4dwk_q/inheritance-in-less

tiberiuichim (changes requested):

theme/themes/pastanaga/extras/main.less
hazigabriel: I have tried extending the `.summary-listing-item-header` selector to match the original `h3` styling however I found that unless I specifically define `h3` styling in a .less file I would not be able to extend the `.summary-listing-item-header` styling.

In my case, if I were to extend `.summary-listing-item-header` with predefined `h3` styling, like in the following code, `&:extend` works as expected.
```
h3 {
  color: @summaryItemTitle;
  font-size: 1.25em;
  font-weight: 500;
}

#main {
  .summary-listing-item-header {
    &:extend(h3 all);
  }
}
```
If I would delete the hard coded `h3` style  then `.summary-listing-item-header` would not inherit any styling from the original `h3`.


Should I make any changes?

tiberiuichim: https://github.com/Semantic-Org/Semantic-UI-LESS/blob/806bd67fcc445ce254a55a7dd153dae408f4485d/definitions/globals/site.less#L55

the h3 is defined in semantic. Can you try to identify the problem?

hazigabriel: Hello, the changes were added via commit f58e96a. It seems that  `.summary-listing-item-header ` would not inherit the `<a>` color and I have to override it  on line [659](https://github.com/plone/volto/blob/f58e96aae947a1af2efb1b597e909fe07affb7b0/theme/themes/pastanaga/extras/main.less#L659)

PR #4997, opened 2023-07-20
make selectedView and className props available for Search block

danalvrz:
Made `selectedView` and `className` props available to the `SearchBlockView.jsx`, following the behavior of Listing Block, to improve styling development.


davisagli (approved):
LGTM

sneridagh (approved):

Issue #4581, opened 2023-03-21
Environment variables need to be properly documented


tiberiuichim:
https://github.com/plone/volto/blob/d3865b3ac9e5c1213058373e6ecf5422060d7ab0/src/config/index.js#L91-L93

tiberiuichim:
https://6.dev-docs.plone.org/volto/configuration/environmentvariables.html

KIET7UKE:
Hey @tiberiuichim , I'm interested to work on this. Can you please explain to me in easy, what to change?

Hello-Qasim:
Hi @tiberiuichim, I am Qasim Javed a student of Computing and in the final year of my degree . I have just started looking to contribute for open source . At this stage i am leaning web development and already made many projects on JavaScript and React. 
I am interested to work on this issue,Can you please me in the understanding of this issue assign to me ?

Your response will be highly appreciated,

Thanks

stevepiercy:
@KIET7UKE @Hello-Qasim please see Volto and general contributing guidelines.

- https://6.docs.plone.org/volto/developer-guidelines/contributing.html
- https://6.docs.plone.org/contributing/index.html#making-contributions-on-github 

I have also updated the issue description, and will delete both of your comments to reduce clutter. If you are still not clear on what needs to be done and how we expect contributors to work in Plone, please ask specific questions. Thank you!

harshlancer:
As these are runtime variable and as the docs suggest these should be documented like
_yarn build && RAZZLE_API_PATH=https://plone.org yarn start:prod_
Right?

AAYUSH-droid:
hey! i would like to contribute can i get assigned!

stevepiercy:
@AAYUSH-droid see https://github.com/plone/volto/issues/4581#issuecomment-1484224952

hasnaat-iftikhar:
If that issue is still open. So can I start working on that?

stevepiercy:
@hasnaat-iftikhar please read and follow [Contributing to Plone](https://6.docs.plone.org/contributing/index.html) and [First-time contributors](https://6.docs.plone.org/contributing/first-time.html).

Akshat2Jain:
Hey Steve can you guide me a little in this?

stevepiercy:
@Akshat2Jain sorry, I don't know. I would suggest you look at what @tiberiuichim posted and referenced, and try to figure it out from there, and if it is still not clear, then ask him as he created this issue.

[...]

PR #5012, opened 2023-07-21
#4581 Environment variables documented properly

PrasannaGadalGIT:


stevepiercy:
@PrasannaGadalGIT also see https://6.docs.plone.org/contributing/index.html#change-log-entry

davisagli:
This isn't addressing issue 4581, which is about adding missing documentation for how to use these environment variables. The change here is wrong; the code runs in Node where there is no window object.

PR #5016, opened 2023-07-24
Fix not able to edit second column of table block.

### LINKED ISSUE #5015: Not able to edit second column of table block.
When I added the multiple line of text in first column in table block. I am not able to add text into second column because editor is present in middle without any indication.

iFlameing:
Fixes #5015

### DIFF
--- a/news/5015.bugfix
+++ b/news/5015.bugfix
@@ -1,0 +1,1 @@
+Fix not able to edit second column of table block. @iFlameing
--- a/packages/volto-slate/src/blocks/Table/TableBlockEdit.jsx
+++ b/packages/volto-slate/src/blocks/Table/TableBlockEdit.jsx
@@ -9,7 +9,7 @@ import { map, remove } from 'lodash';
 import { Button, Table } from 'semantic-ui-react';
 import cx from 'classnames';
 import { defineMessages, injectIntl } from 'react-intl';
-
+import { Node } from 'slate';
 import Cell from './Cell';
 import { BlockDataForm, Icon, SidebarPortal } from '@plone/volto/components';
 import TableSchema from './schema';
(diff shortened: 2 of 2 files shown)


tiberiuichim (approved):

tiberiuichim:
I think the changelog entry is specific to your project.

davisagli:
I'm confused about the purpose of this change. It looks like it is setting the vertical alignment based on whether the cell has contents, but not doing anything that affects whether the cell can be edited. Was the problem that if you click inside the cell but not inside a text node, nothing happens?

I think:
1. We should always set the vertical alignment to `top`; it looks better than `middle`.
2. We should make sure that a click inside the cell moves the cursor to the closest text node, even if the click wasn't inside one.
Thoughts, @tiberiuichim?

tiberiuichim:
> 2\. We should make sure that a click inside the cell moves the cursor to the closest text node, even if the click wasn't inside one.

Yes. 

With volto-slate and the rest of the code we have based on slatejs, sometimes this is hard because we have multiple slate editors on the screen, which aren't "active". Sometimes we have code that forbids a component (a block, for example) from being included in the react batch refresh updating, so when you click in a text area you first have to activate/select its parent component, then rely on slate properly detecting the dom node that was clicked, to translate it into "cursor position inside my internal tree structure". This is a probable explaination for why sometimes slate crashes: if, by clicking, I create/mutate dom nodes, slate can't translate my click to a slate node.

This problem, "I click on a text block and my cursor position is not lost" has been a really big challenge with volto-slate.

davisagli:
@iFlameing

1. I think `top` looks better than `middle` in view mode also.
2. Yes. But the problem still exists regardless of where the text is positioned, because the user doesn't know there is a special place they need to click which is smaller than the entire cell.

Issue #5022, opened 2023-07-26
Cannot read property of undefined error. screen errors out when i choose no value option in block on adding the table of contents field.


satyam4p:
**Describe the bug**
When I add table of contents field in page and in block section select entries as no value following error shows up:

TypeError: Cannot read properties of undefined (reading 'formatMessage')

**To Reproduce**

1. Go to 'page and add table of contents'
2. in right panel of block for table of content select entries
3. from drop down of entries choose 'no value' option

**Expected behavior**
On Choosing no value the normal text showing 'No Value' should be shown as tag in Entries field.

**Screenshots**

**Software (please complete the following information):**

- OS: IOS
- Browser Chrome
- Volto Version 16.22.0
- Plone Version 6.0.6

**Additional context**

satyam4p:
I'm currently working on this one.

PR #5027, opened 2023-07-26
Fix: added fix for 'no value' entry select in table of content field

satyam4p:
fixes: #5022


davisagli (approved):
@satyam4p Thanks for this fix!

Issue #5028, opened 2023-07-26
Hover and focused border for block child


claudiaifrim:
**Describe the bug**
Hover and focused border for block child of accordion (as eg.) do not work properly due to z-index = -1 set on blocks css: https://github.com/plone/volto/blob/master/theme/themes/pastanaga/extras/blocks.less#L25

**To Reproduce**

1. Create a volto project and include "@eeacms/volto-accordion-block"
2. Create a page and add accordion block
3. Add a text block (or any other block) in accordion
4. Hoover on any block added in the accordion.

**Expected behavior**
Hover and focused border for any block child should work. See screenshoot:
![image](https://github.com/plone/volto/assets/4428527/6be982c2-783c-4b76-a0fc-e5575c6564b8)

**Screenshots**
![image](https://github.com/plone/volto/assets/4428527/09ae26f8-b9cf-4b81-b847-b779830d8424)

**Software (please complete the following information):**

- OS: Linux Mint, macOS
- Browser chrome, safari, firefox
- Volto Version "@plone/volto": "16.21.3" and "@plone/volto": "17.0.0-alpha.17"
- Plone Version Plone 6.0.0 (6011)
- Plone REST API Version 8.32.6

**Additional context**
Issue available for all complex blocks that have another blocks in them, like: columns, section, tabs, accordion.

PR #5029, opened 2023-07-26
fix: #5028 - fix hover and focused border for block child

claudiaifrim:
Fix for issue #5028

### DIFF
--- a/cypress/tests/core/basic/actions.js
+++ b/cypress/tests/core/basic/actions.js
@@ -1,0 +1,57 @@
+describe('actions Tests', () => {
+    beforeEach(() => {
+        cy.autologin();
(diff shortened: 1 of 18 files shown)


avoinea (approved):

davisagli (approved):
It looks like this was because z-index: -1 placed the border below the `<div class="accordion">` element which has an opaque white background. So, I'm not sure that the fix is relevant to every block inside another block. But, the fix seems okay to me.

davisagli:
@claudiaifrim Unfortunately I need to revert this change in #5039. Setting the z-index of the block outline to `auto` means that it was sometimes getting in the way of clicking on a selected block.

PR #5036, opened 2023-07-28
Fix selecting the grid block when it contains other blocks

### LINKED ISSUE #5035: Not possible to select Grid block after adding an item to it
**Describe the bug**
As an editor, when you add a Grid block and then any block inside of it, it is not possible to select the Grid block again, only its contained blocks.

**To Reproduce**

1. Go to '/add?type=Document'
2. Add a Grid block (any number of columns)
3. Add a block into a column (any block type)
4. Try to select the Grid block (e.g. to add a block title)

**Expected behavior**
As an editor, you should be able to select the Grid block, even if it contains other blocks.

**Screenshots**
Screenshot 2023-07-28 at 10 07 04

- OS: MacOS
- Browser: Firefox
- Volto Version: 17.0.0-alpha.20
- Plone Version 6.0.0b1
- Plone REST API Version 8.23.0

davisagli:
Fixes #5035

### DIFF
--- a/news/5036.bugfix
+++ b/news/5036.bugfix
@@ -1,0 +1,1 @@
+Fix selecting grid block when a sub-block is selected. @davisagli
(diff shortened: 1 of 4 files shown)


danalvrz (commented):
thanks!

PR #5041, opened 2023-07-31
Refactor relations actions: Spin off relation stats.

ksuess:


news/5041.internal
```diff
@@ -0,0 +1 @@
+Refactor relations actions: Spin off relation stats. @ksuess
```
davisagli, line 1: @ksuess This involves a breaking change to the shape of the redux state...i.e. breaking if someone has written custom components that use these actions. I think that is relatively unlikely, and I think the changes are worth it to keep the stats more separate from the relation list. But we should rename this file from .internal to .breaking and mention that change.

davisagli (changes requested):

davisagli (approved):

PR #5044, opened 2023-08-01
Refactor Change password 

Tishasoumya-02:
#4460


packages/volto/src/components/manage/Preferences/ChangePassword.jsx
```diff
@@ -70,169 +61,97 @@ const messages = defineMessages({
...
-  onSubmit(data) {
+  const onSubmit = (data) => {
     if (data.newPassword === data.newPasswordRepeat) {
-      this.props.updatePassword(
-        this.props.userId,
-        data.oldPassword,
-        data.newPassword,
-      );
+      dispatch(updatePassword(userId, data.oldPassword, data.newPassword));
       toast.success(
```
davisagli, line 81: This should be dispatch(updatePassword( ... )).then(() => {
    toast.success( ... );
}

(We only want to show the toast message after the API call was completed successfully.)

PR #5046, opened 2023-08-01
Refactor Personal Information

Tishasoumya-02:
#4460


src/components/manage/Preferences/PersonalInformation.jsx
```diff
@@ -1,131 +1,65 @@
...
-  onSubmit(data) {
-    // We don't want the user to change his login name/username or the roles
-    // from this form
-    // Backend will complain anyways, but we clean the data here before it does
+  const onSubmit = (data) => {
     delete data.id;
     delete data.username;
     delete data.roles;
-    this.props.updateUser(this.props.userId, data);
+    dispatch(updateUser(userId, data));
```
tiberiuichim, line 32: maybe we should use the promise from `dispatch(updateUser(userId, data)).then(() => toast.success())`

davisagli (approved):

PR #5053, opened 2023-08-02
User / Groups search control panel / sharing panel improvements #4551

tedw87:
When searching for users (Users Control Panel, Page sharing, etc.) it takes some time for the results to return from the server (LDAP). The problem is that there is no visual feedback for the end-user to show this, also the form is not blocked and the user can click multiple time on the search button, stressing the server.

[users-loading.webm](https://github.com/plone/volto/assets/323385/3a22264a-2618-440e-a4bf-8db9d4133780)


erral:
This is partially addressed also in #4572

I won't be able to look at this at least in 10 days, sorry about that.

src/components/manage/Sharing/Sharing.jsx
```diff
@@ -145,6 +145,7 @@ class SharingComponent extends Component {
     this.onToggleInherit = this.onToggleInherit.bind(this);
     this.state = {
       search: '',
+      isLoading: false,
```
avoinea, line 148: @tedw87 This should also be used to block the search button or the form re-submit as right now it looks like I am able to DDOS attack the backend RestAPI endpoint by clicking the search button or by hitting the `Enter` key

Screenshot from 2023-09-07 17-47-32

avoinea (approved):

sneridagh:
@erral @tedw87 ok, so we have https://github.com/plone/volto/pull/4572 and this one, which one goes first, and does they collide at some point?

erral:
I talked to @jackahl about the GSOC, and he was going to try not to collide with my changes.I haven't checked this PR though, and I can't say... I need to take some time to checK

sneridagh:
@erral @tedw87 @robgietema @jackahl Please guide me on this one whenever you can, it would be great to have both for 17 final.

erral:
I have checked this PR, and it applies cleanly on top of mine, as it is complementary, it adds the "loading" spinner while the search is done or the users/groups are loaded.

So I would say to merge first #4572 and right after that, this one the #5053.

@avoinea 
@tedw87 
@sneridagh

erral (approved):

PR #5054, opened 2023-08-03
Listing headlines (continuation of #4994 )

tiberiuichim:


tiberiuichim:
@hazigabriel the Default template should be changed as well, not just the Summary

PR #5055, opened 2023-08-03
Fix search block showing no option select in sort on property

iRohitSingh:


davisagli (changes requested):
@iRohitSingh It is not working. I set up a search block which only has the option to sort by effective date, and it is still showing the dropdown:

Screenshot 2023-08-08 at 10 40 38 AM

iRohitSingh:
@davisagli @sneridagh can you please review this pr once again.

iRohitSingh:
Screenshot: 

Screenshot 2024-07-15 at 8 02 16 PM


edit: 

Screenshot 2024-07-15 at 8 02 58 PM


Screenshot 2024-07-15 at 8 03 29 PM

iRohitSingh:
@sneridagh I have implemented what we have discuss into the call.
I have opened the issue for not syncing the top sorton with facet sort on property.
https://github.com/plone/volto/issues/6169. Once it is fixed we don't have to think about anything.

Screenshot 2024-07-15 at 8 20 01 PM

packages/volto/locales/ca/LC_MESSAGES/volto.po
```diff
@@ -1729,6 +1729,11 @@ msgstr "Si esteu segur que aquest usuari ha abandonat l'objecte, podeu desbloque
 msgid "If you are certain you have the correct web address but are encountering an error, please contact the {site_admin}."
 msgstr "Si esteu segur que teniu l'adreça web correcta però trobeu un error, poseu-vos en contacte amb el {site_admin}."
 
+#. Default: "If you have only one sort on a property, please place it in the top sort field."
+#: components/manage/Blocks/Search/schema
+msgid "If you have only one sort on a property, please place it in the top sort field."
```
stevepiercy, line 1734: I don't understand this message. And if I don't understand it, then an editor won't either. It needs to use language that the end user will understand.

In the UI, `property` is not used. Under `Controls`, I see `Sort on options`. Additionally if there is only one option selected, it is already at the top.

Finally, this should be a drag and drop field to allow the user to reorder the selection of sort on options, without having to remove them, then add them back.

I think this is the intent:

```suggestion
msgid "Select the options by which you want to sort items. Sorting by multiple options will be performed in order from first to last selected options."
```

If that's not correct, please elaborate.

stevepiercy (changes requested):

packages/volto/locales/ca/LC_MESSAGES/volto.po
iRohitSingh: @stevepiercy We have two sort on field in search
 block see screenshots. For second one I want to add a description that if they only have only one sorton options then please select the value in upper sorton field.

Screenshot 2024-07-19 at 8 25 14 PM

stevepiercy: I don't think you understand what these two options actually do. See https://6.docs.plone.org/volto/user-manual/blocks.html#search-block. If that does not help, please ask for clarification.

Ultimately I think a new message is not necessary. However perhaps other elements of the UI could be improved to make their function clearer to newcomers.

sneridagh:
@iRohitSingh what's the status of this one?

sneridagh:
@stevepiercy I will review again this, but when I looked at it with @iRohitSingh We only saw this option. The UI is a bit convoluted, but the idea is to come up with something that is consistent and makes sense without explaining too much. And, ultimatelly, it continues working in the same way it does.

iRohitSingh:
@sneridagh @davisagli I have tested this on four parameters and adding the screenshot. I want your review regarding the ux and everything is correct. I will add the cypress once the functionality is correct.

1. select 1 sort on in listin criteria sorton

Screenshot 2024-07-30 at 9 31 30 PM

Edit

Screenshot 2024-07-30 at 9 32 00 PM


2. Only one soron options below.


Screenshot 2024-07-30 at 9 33 11 PM


Edit

Screenshot 2024-07-30 at 9 33 29 PM


3. Only selected down sort on options
Screenshot 2024-07-30 at 9 34 21 PM

Edit mode
Screenshot 2024-07-30 at 9 34 32 PM


4. Select both Listing sorton and Sorton options


Screenshot 2024-07-30 at 9 35 36 PM


Edit mode

Screenshot 2024-07-30 at 9 35 50 PM

stevepiercy:
As I understand the existing functionality, the first `Sort on` establishes a default sort order for objects, and the second `Sort on` displays controls to the end user for sorting. Please correct me if I am wrong, or confirm my understanding, then I can provide a useful review.

davisagli:
@stevepiercy That's right.

@iRohitSingh This is pretty good, but I don't like the button that is used when there is no default sort (top "sort on" setting) and only one sort option (bottom "sort on" setting). It is not obvious that it is a button the user can interact with. I am going to change it so that we also show a select box in that case.

davisagli (approved):

PR #5058, opened 2023-08-04
Fix toc accessibility issue

dobri1408:
There is a bug in volto that can be tested on https://demo.plone.org/toc-issue
If you press on a Toc element( with the keyboard) the view will change on the element, but the focus will not move
Actual behaviour:
[bug.webm](https://github.com/plone/volto/assets/50819975/70b2c08e-d56f-4688-9d19-3f6e4e23033d)
Expected behaviour
[my fix.webm](https://github.com/plone/volto/assets/50819975/1792b50c-e64f-431e-8d6a-304190f94315)


tiberiuichim:
I have a feeling that the solution is too much and it can potentially cause issues. Better to use react-router-hash-link with smooth scroll, on the TOC links?

dobri1408:
@tiberiuichim Thank you for your input, I like your solution so much. I didn't know about react-router-hash-link, and I checked online, and it s also moving the focus. It is perfect! I will start implementing this first thing on Monday.

dobri1408:
@tiberiuichim I have implemented the approach that you suggested.

tiberiuichim (approved):

packages/volto-slate/src/editor/render.jsx
```diff
@@ -174,20 +174,20 @@ export const renderLinkElement = (tagName) => {
...
-      <Tag {...attributes} className={className}>
+      <Tag {...attributes} className={className} tabIndex={0}>
         {children}
         {mode === 'view' && slug && (
           <UniversalLink
             className="anchor"
             aria-hidden="true"
             tabIndex={-1}
             href={`#${slug}`}
+            aria-label="anchor"
```
davisagli, line 190: Is it meaningful to use `aria-label` on an element that also has `aria-hidden="true"`?

davisagli (commented):
@dobri1408 I tested this. In Chrome it is a clear improvement. In Firefox I cannot tab to the entries in the ToC. (That is probably an existing problem before your PR, but maybe you have an idea how to fix it?)

davisagli (approved):

PR #5064, opened 2023-08-05
Fix creating user issue 5031

mehedikhan72:
When email was set to be the username field, the actual username field still appeared and the registration was broken.

I fixed it by showing the username field conditionally based on whether this setting was enabled.

I used the following to get our desired boolean value and stored it in a state, which was used to conditionally render the username input element.
```
  checkLoginUsingEmailStatus = async () => {
    await this.props.getControlpanel('security');
    this.setState({
      loginUsingEmail: this.props.controlPanelData.data.use_email_as_login,
    });
  };
  ```
  
  Feel free to let me know if I need to make changes regarding this PR. Thanks!

### DIFF
--- a/src/components/manage/Controlpanels/Users/UsersControlpanel.jsx
+++ b/src/components/manage/Controlpanels/Users/UsersControlpanel.jsx
@@ -123,7 +123,7 @@ class UsersControlpanel extends Component {
     }
   };
 
-  // Because username field needs to be disabled if email login is enabled so we need a way of knowing if email login is enabled.
+  // Because username field needs to be disabled if email login is enabled.
   checkLoginUsingEmailStatus = async () => {
     await this.props.getControlpanel('security');
     this.setState({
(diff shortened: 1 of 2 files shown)


ksuess:
@mehedikhan72 This change request looks pretty good. Please come back with it, after your Plone Contributor Agreement process is done. See https://6.docs.plone.org/contributing/first-time.html#plone-contributors-team

For cross referencing issue and PR, please note the issue number in the PR description. For example "Fix #xyz".

'yarn.lock' should be excluded from your changed files.

mehedikhan72:
Hi @ksuess, thanks for the suggestions. Although I do have signed the plone contributor agreement but when I made the initial commit, the commit wasn't made by my github account. I will look into this and make another PR.

stevepiercy:
@mehedikhan72 I see that you are now a member of the plone/contributors team. If mister-roboto continues to state that there is a problem with your email, then you need to make sure that the email you used in the Plone Contributor Agreement and the email you use for GitHub match.

PR #5071, opened 2023-08-07
fix: Fix querystringResults subrequests id, to work properly in dupli…

giuliaghisini:
…cate pages where blocks id's are the same.


pnicolli (approved):

src/components/manage/Blocks/Listing/withQuerystringResults.jsx
```diff
@@ -23,7 +23,7 @@ export default function withQuerystringResults(WrappedComponent) {
     } = props;
     const { settings } = config;
     const querystring = data.querystring || data; // For backwards compat with data saved before Blocks schema. Note, this is also how the Search block passes data to ListingBody
-
+    const subrequestID = content.UID + '-' + id;
```
davisagli, line 26: I was a bit worried in case there is ever content without a UID -- but in that case we should end up with "undefined-someid" which is no worse than before.

davisagli (approved):
Approving, but there are some acceptance tests which failed. I'll retry them...

davisagli:
It looks like the tests have failed multiple times and might need a closer look.

giuliaghisini:
> It looks like the tests have failed multiple times and might need a closer look.

acceptance tests fails not because this pr, i don't know why..

davisagli:
@giuliaghisini The failing tests are related to this change. If you watch the video you can see there is an error because `content` is undefined.

sneridagh (approved):

ionlizarazu:
This changes has break the listing block search results:
image

There is no number for the results because content is undefined.

PR #5106, opened 2023-08-21
Fixes #5105; For the Sitemap component, don't hardcode root as navigation action path

### LINKED ISSUE #5105: Sitemap component is not properly rooted in navigation root
**Describe the bug**

In Plone classic the /sitemap page is generated based on the navigation root. This is not so in Volto.

https://github.com/plone/volto/blob/4523b13d6586891b9dad7db1caedaab5f6f7dd4b/src/components/theme/Sitemap/Sitemap.jsx#L115C21-L115C21

https://github.com/plone/volto/blob/4523b13d6586891b9dad7db1caedaab5f6f7dd4b/src/actions/navigation/navigation.js#L21

**To Reproduce**

1. Set a folder as a navigation root (assign INavigationRoot interface to it)
2. Go to address/of/folder/sitemap
3. The sitemap is generated with the navigation generated from the Plone root

tiberiuichim:
The plone.restapi navigation endpoint will compute a navigation root from a context path, but the problem is that Volto will trigger the navigation endpoint either on root or on the lang folder. 

https://github.com/plone/plone.restapi/blob/ed070e0769c4ac3efda05448db8846f61dae255c/src/plone/restapi/services/navigation/get.py#L40

https://github.com/plone/volto/blob/4523b13d6586891b9dad7db1caedaab5f6f7dd4b/src/components/theme/Sitemap/Sitemap.jsx#L115C21-L115C21

https://github.com/plone/volto/blob/4523b13d6586891b9dad7db1caedaab5f6f7dd4b/src/actions/navigation/navigation.js#L21

### DIFF
--- a/news/5106.bugfix
+++ b/news/5106.bugfix
@@ -1,0 +1,2 @@
+For folders inside navigation roots, properly fetch navigation from the
+navroot, rather then the site root  @tiberiuichim
--- a/src/components/theme/Sitemap/Sitemap.jsx
+++ b/src/components/theme/Sitemap/Sitemap.jsx
@@ -22,6 +22,13 @@ const messages = defineMessages({
     defaultMessage: 'Sitemap',
   },
 });
+
+export function getSitemapPath(pathname = '', lang) {
+  const prefix = pathname.replace(/\/sitemap$/gm, '').replace(/^\//, '');
+  const path = prefix || lang || '';
+  return path;
+}
+
 /**
  * Sitemap class.
  * @class Sitemap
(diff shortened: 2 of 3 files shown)


erral:
this should also be taken into account in https://github.com/plone/volto/pull/3537

erral:
@tiberiuichim revisiting this PR in order to do a more complete use of the navroot endpoint (#3537), I see quite easy to change the usage of `isMultilingual` and `toBackendLang` here. 

But I don't get at all how this could be done for the SSR part (I guess, that's what it's done in the `asyncConnect` part, right?

Perhaps, if #3537 is merged, I can prepare a PR bringing the usage of `@navroot` here and we can discuss there how to bring it also to the SSR part.

PR #5111, opened 2023-08-22
Links references view cypress tests

jackahl:


tisto:
@jackahl what's the status here? Can someone take this over during your holiday?

tisto:
@jackahl @sneridagh the Makefile pins plone.restapi 8.42.0:

https://github.com/plone/volto/blob/63fbe9179d534b8b57d118fcb29e23a16510e271/Makefile#L18

jackahl:
> @jackahl @sneridagh the Makefile pins plone.restapi 8.42.0:
> 
> https://github.com/plone/volto/blob/63fbe9179d534b8b57d118fcb29e23a16510e271/Makefile#L18

@tisto the info that we should wait for the next Plone release is from @davisagli. So maybe we should ask him.

sneridagh:
@jackahl my last commit to this PR will do to use the latest p.restapi in CI.

sneridagh:
@jackahl should we align this with @jaroel PR? or merge them after it? Or @jaroel already join this one to there?

jackahl:
Test for images linked via the override field of a teaser block is currently still failing. This probably needs fixing in p.restapi

sneridagh:
@jackahl @tisto this was never in the spec, in fact an override is not a link, is a `preview_image` field, which is not "smart" since it is supposed not to be treated as a "linked" element. We can discuss how to add it to the linkintegrity catalog but in fact, it's not a link...

cypress/tests/core/basic/linkIntegrityView.js
```diff
@@ -0,0 +1,457 @@
...
+              title: 'Document linked',
+              // preview_image: [
+              //   {
+              //     '@id': '/image-linked',
+              //     '@type': 'Image',
+              //     image_field: 'image',
+              //     Title: 'Image linked',
+              //     Type: 'Image',
+              //   },
+              // ],
```
davisagli, line 304: @jackahl @sneridagh Is the comment here the reason for the continued test failure?

sneridagh: @jackahl Could you please take a look, so we can close this? The strange thing is that https://github.com/kitconcept/fzj-internet/pull/1488 is passing, and this one, do not.

jackahl: @davisagli it should not be. This is commented out, as the `createContent()` function seems to have trouble to add preview images. There for I am adding the image "by hand" afterwards.

jackahl: But I can confirm the image is added as it can be seen in cypress after this line has been run 
Screenshot 2023-11-27 at 15 53 26
https://github.com/plone/volto/pull/5111/files/5c54795589b8166d9aa91cb602fe7eba96b90041#diff-3fc8e2fbc20dc1ed7d495f717dc8c3cc8b774e3b8050a11f5655297eb3885db4R321

Makefile
```diff
@@ -15,13 +15,13 @@ MAKEFLAGS+=--no-builtin-rules
...
 DOCKER_IMAGE_ACCEPTANCE=plone/server-acceptance:6.0.8
-KGS=
+KGS=plone.restapi==9.2.0
 NODEBIN = ./node_modules/.bin
 SCRIPTSPACKAGE = ./packages/scripts
 
 # Plone 5 legacy
 DOCKER_IMAGE5=plone/plone-backend:5.2.12
-KGS5=plone.restapi==8.43.3 plone.volto==4.1.0 plone.rest==3.0.1
+KGS5=plone.restapi==9.2.0 plone.volto==4.1.0 plone.rest==3.0.1
```
jackahl, line 24: @sneridagh @davisagli from what I could find out, this  seems not to be sufficient to bump restapi up to 9.2.0 in the test backend. At least in my local test runs on this branch it stayed at 9.1.2

jackahl: I think we need a new release of https://hub.docker.com/r/plone/server-acceptance/tags here with the appropriate restapi version

sneridagh: mmm I see, that is a problem :/ 
The acceptance images right now do not "compile" and are not `ADDONS` env aware... It would be great that they would be... For now, we would have to wait until 6.0.9, and have enough with FZJ ones.

sneridagh:
@jackahl Could you please take a final look at this one and find out why it's failing? Let's see if we can finally wrap up this one.

jackahl:
It seems the CI here is still using an older version of p.restapi then the local test setup

Molochem:
All Checks pass when I check them

sneridagh:
@jackahl @Molochem probably a flaky test due to network latency in CI. I added intercept/wait here and there. Let's see how it goes.

sneridagh:
This is super weird, locally it passes, but in CI the response for that test gets empty it seems :/

sneridagh:
Finally...

Issue #4422, opened 2023-02-22
Sphinx toctree warnings from included CHANGELOG.md


stevepiercy:
Ever since we started to include the `CHANGELOG.md` in the documentation, we had to resolve its path between both `plone/documentation` and `plone/volto`.

At first we attempted to use the Sphinx extension `ifconfig`, but we found that it generated misleading warnings that occasionally caught the attention of authors. `ifconfig` was never designed for this use case.

Next we removed `ifconfig` from Volto to silence the warnings in `plone/documentation`, but that resulted in dozens of warnings in Volto, like this:

`documentation/submodules/volto/docs/source/index.md:38: WARNING: toctree contains reference to document 'release-notes/index' that doesn't have a title: no link will be generated`

These warnings are just noise and prevent authors from spotting useful warnings.

@ksuess suggested creating a duplicate of `CHANGELOG.md` in `docs/`, and instead updating the reference there. That will remove the warnings.

However, as the generation of `CHANGELOG.md` is a step in the release of Volto, the release manager would need to agree that this would be a good step to automate, if possible. I can see that it could be useful for other repositories that include a change log in their docs.

@sneridagh @tiberiuichim @mauritsvanrees @ksuess @nileshgulia1 and any others, do you have thoughts about this proposal?

I would also like to update the comments in the `CHANGELOG.md` to update the URL to the docs and to add some clarity, per https://github.com/plone/documentation/issues/1431#issuecomment-1437514155

See also: https://github.com/plone/volto/pull/4404#issue-1588864779

nileshgulia1:
@stevepiercy @ksuess Since we are updating the tips of volto submodule, wouldn't the changelog automatically get fetched when we are building the docs from "plone/documentation"?  I agree with https://github.com/plone/documentation/issues/1431#issue-1538209052 I don't mind "not seeing the changelog" when building Volto docs.

stevepiercy:
The entire volto repository gets pulled in as a git submodule. Updating the tips just means pulling in the volto repo at that hash.

> Next we removed ifconfig from Volto to silence the warnings in plone/documentation, but that resulted in dozens of warnings in Volto, like this

IOW, that is the same as not including the changelog when building volto docs only. See https://github.com/plone/documentation/issues/1431#issuecomment-1434212761 

[...]

PR #5135, opened 2023-08-27
Fix Sphinx toctree warnings from included CHANGELOG.md

stevepiercy:
See https://github.com/plone/volto/issues/4422
- Prepend CHANGELOG.md with html_meta stuff for SEO.
- Manually copy CHANGELOG.md into docs/source/release-nots/index.md. This step might be automated via release-it.
- Use a relative symlink from inside the volto docs directory to the news directory at the volto root.
- Only check changelog entries when running docs-linkcheckbroken in volto.
- Find news items for docs-linkcheckbroken only in volto
- Remove non-existent excluded files.
- Suppress warning for news items

This PR needs to be merged before its counterpart in `documentation` will work: https://github.com/plone/documentation/pull/1527

To test locally, checkout the branch `linkcheck-news-volto` in both `documentation` and `volto`.


stevepiercy:
linkcheckbroken should have failed. I think it is not reading the files in `news`. I need to dig into this more later.

stevepiercy:
Sphinx does not read the files in /news. It appears that the required symlink is never created. Would it be better to commit a symlink to the repo (probably not) or find a GitHub workflow step that will create the symlink? My GitHub Workflow DuckDuckFu has not turned up anything obvious.

stevepiercy:
w00t! Got the symlink to work to news. Now I need to make the workflow fail when linkcheck fails.

stevepiercy:
YAY! It fails! https://github.com/plone/volto/actions/runs/5995760347/job/16259268890?pr=5135#step:7:242

Now to clean them up so it should pass on the next run.

stevepiercy:
Shoot, it was failing for something besides a bad link in news. I could not spot the difference between `plone/documentation` and `volto` workflows and Makefiles that would cause this. So close!

docs/source/index.md
```diff
@@ -51,3 +51,10 @@ user-manual/index
 contributing/index
 release-notes/index
 ```
+
+````{ifconfig} context in ("volto")
```
davisagli, line 55: Is this not executed because `in ("volto")` evaluates to iterating over the characters in a string?

```suggestion
````{ifconfig} context in ("volto",)
```

davisagli:
@stevepiercy Aha, the link check is finding the broken link, but it's not breaking the build:

docs/source/index.md
stevepiercy: `context` evaluates to the string `volto`, so:

```pycon
>>> "volto" in ("volto")
True
>>> "volto" in ("volto",)
True
```

It makes no difference, although it was a good idea. I have been bitten by omitting that trailing comma many times in other code.

stevepiercy:
Exactly. When linkcheckbroken fails as it does now, then the Documentation workflow should also fail, but it passes. I am again staring at the comparison between documentation and volto repos, but nothing jumps out at me.

davisagli:
@stevepiercy Yeah, this is the point at which using `make` always feels like a deal with the devil. Trying a fix now...

davisagli:
@stevepiercy 🎉

stevepiercy:
Yeah, my `make` skillz are "enough to be dangerous". Fixed news entry, let's see what happens.

stevepiercy:
Phooie! Time to sacrifice a 🐔 and utter some incantations.

davisagli:
@stevepiercy I'm striking out, going to have to come back to this another time.

stevepiercy:
@davisagli for trying. I will try again when I have a few cycles.

stevepiercy:
docs-linkcheckbroken now passes as it should with this make command.

`cd $(DOCS_DIR) && $(SPHINXBUILD) -b linkcheck $(ALLSPHINXOPTS) $(BUILDDIR)/linkcheck | GREP_COLORS='0;31' grep -wi "broken\|redirect" --color=always | GREP_COLORS='0;31' grep -vi "https://github.com/plone/volto/issues/" --color=always && if test $$? = 0; then exit 1; fi || test $$? = 1`

Let's see if it breaks next with a broken link.

stevepiercy:
@sneridagh @davisagli this is ready for review. Note that I modified `package.json`, which requires extra care.

docs/source/index.md
```diff
@@ -38,3 +38,10 @@ user-manual/index
 contributing/index
 release-notes/index
 ```
+
+````{ifconfig} context in ("volto",)
```
davisagli, line 42: Is it worth adding some sort of comment to remind us what this is for?

package.json
```diff
@@ -187,7 +187,7 @@
         "yarn i18n",
         "git add locales"
       ],
-      "after:bump": "pipx run towncrier build --draft --yes --version ${version} > .changelog.draft && pipx run towncrier build --yes --version ${version} && make corepackagebump VERSION=${version}",
+      "after:bump": "pipx run towncrier build --draft --yes --version ${version} > .changelog.draft && pipx run towncrier build --yes --version ${version} && make corepackagebump VERSION=${version} && cp CHANGELOG.md docs/source/release-notes/index.md && git add docs/source/release-notes/index.md && git commit -m 'copy CHANGELOG.md' && git push",
```
davisagli, line 190: I think you can leave out the git add, git commit, and git push here. release-it already pushes a commit with the changes that are made during this hook.

davisagli (commented):
One main comment here -- I think we can simplify what is done in the "after:bump" hook, since release-it already commits changes that are made while bumping the version.

I'll leave final approval for @sneridagh, since it's important for him to be familiar with the approach here in case something goes wrong when he actually makes a release.

package.json
sneridagh: I'd move it to a makefile command (as the `corepackagebump`). In fact, the `corepackagebump` actions are similar to these ones.

sneridagh (approved):
@stevepiercy @davisagli please a final look before I merge.

davisagli (approved):
LGTM -- hopefully

stevepiercy:
I forgot that there is a counterpart in Documentation that needed to be merged at https://github.com/plone/documentation/pull/1527. Merged now. Let's see what happens.

PR #5152, opened 2023-09-01
Don't show ``No value`` option in SelectWidget and ArrayWidget if default value is 0

### LINKED ISSUE #5151: Problem with default value 0 in Select Widget
**Describe the bug**
If we have a schema in Plone with a field of type `schema.Choice` and with a default value of `0`, Volto adds a "No value" option to the select, even though the field is required.

**To Reproduce**

An example schema where this problem occurs:

```python
def _my_vocabulary():
    levels = {
        0: _('None'),
        1: _('One'),
    }
    for level, name in levels.items():
        term = SimpleTerm(
            value=level,
            token=level,
            title=name,
        )
        yield term


my_vocabulary = SimpleVocabulary(list(_my_vocabulary()))

class ISettings(Interface):

    field = schema.Choice(
        vocabulary=my_vocabulary,
        title=_("Field"),
        default=0,
    )

```

If you create a controlpenel with this interface the problem will occur.

**Expected behavior**
I hope that Volto does not create a "No value" option in the select

Volto 17.0.0-alpha.25
Plone 6.0.6
plone.restapi 8.40.0
CMF 3.0
Zope 5.8.3
Python 3.11.4 (main, Aug 19 2023, 06:03:52) [GCC 13.2.0]
PIL 9.5.0 (Pillow)

**Additional context**

Voto does this test to define whether to add the option:

https://github.com/plone/volto/blob/cafad1c570195023a82007e31207c78f97ca4177/src/components/manage/Widgets/SelectWidget.jsx#L193

But `if (!0)` is `true`

wesleybl:
Fixes #5151

### DIFF
--- a/src/components/manage/Widgets/ArrayWidget.jsx
+++ b/src/components/manage/Widgets/ArrayWidget.jsx
@@ -325,7 +325,9 @@ class ArrayWidget extends Component {
               : this.props.choices
               ? [
                   ...choices,
-                  ...(this.props.noValueOption && !this.props.default
+                  ...(this.props.noValueOption &&
+                  (this.props.default === undefined ||
+                    this.props.default === null)
                     ? [
                         {
                           label: this.props.intl.formatMessage(
--- a/news/5151.bugfix
+++ b/news/5151.bugfix
@@ -1,0 +1,1 @@
+Don't show ``No value`` option in SelectWidget and ArrayWidget if default value is 0. @wesleybl
--- a/src/components/manage/Widgets/ArrayWidget.test.jsx
+++ b/src/components/manage/Widgets/ArrayWidget.test.jsx
@@ -2,7 +2,7 @@ import React from 'react';
 import renderer from 'react-test-renderer';
 import configureStore from 'redux-mock-store';
 import { Provider } from 'react-intl-redux';
-import { waitFor } from '@testing-library/react';
+import { waitFor, render, fireEvent } from '@testing-library/react';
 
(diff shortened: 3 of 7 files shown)


src/components/manage/Widgets/ArrayWidget.jsx
```diff
@@ -325,7 +325,8 @@ class ArrayWidget extends Component {
               : this.props.choices
               ? [
                   ...choices,
-                  ...(this.props.noValueOption && !this.props.default
+                  ...(this.props.noValueOption &&
+                  typeof this.props.default === undefined
```
davisagli, line 329:
```suggestion
                  this.props.default === undefined
```

`typeof undefined` returns the string `"undefined"`, which does not equal the object `undefined`

davisagli (changes requested):

src/components/manage/Widgets/ArrayWidget.jsx
wesleybl: @davisagli I fixed it. Thanks!

davisagli (approved):
Thanks! LGTM

davisagli:
@wesleybl Before the change, a default of `null` would also show the `no value` option; after the change, it does not. Do you think we should preserve the old behavior for `null`?

wesleybl:
> Do you think we should preserve the old behavior for null?

I think so. I'll take care of it.

davisagli (approved):
@wesleybl Thanks!

PR #5153, opened 2023-09-01
fix preview image

giuliaghisini:
Fixed PreviewImage component, because it was working only if item.image_scales where setted.
Added show_default prop to display or not default image


src/components/theme/PreviewImage/PreviewImage.jsx
```diff
@@ -7,13 +7,23 @@ import DefaultImageSVG from '@plone/volto/components/manage/Blocks/Listing/defau
 /**
  * Renders a preview image for a catalog brain result item.
  */
-function PreviewImage({ item, alt, ...rest }) {
+function PreviewImage({
+  item,
+  alt,
+  image_field,
+  show_default = true,
```
davisagli, line 14: We use camel case for names defined in javascript, so this should be showDefault. (`image_field` uses an underscore because it's the name of a field from the backend)

davisagli (commented):
It would be nice to have a new test in PreviewImage.test.js for the showDefault=false case.

giuliaghisini:
> It would be nice to have a new test in PreviewImage.test.js for the showDefault=false case. Also some of the existing tests are failing.

updated tests

davisagli (approved):

sneridagh (approved):

PR #5156, opened 2023-09-04
Fix empty slate text block in table of contents

kreafox:
toc


tiberiuichim (approved):

src/components/manage/Blocks/ToC/View.jsx
```diff
@@ -54,6 +54,7 @@ const View = (props) => {
         const level = entry[0];
         const title = entry[1];
         const items = [];
+        if (!title?.trim()) return;
```
davisagli, line 57: Below, we use `title || block.plaintext` as the title of the ToC entry, so I think we should also trim the same thing here, and not only the title?

kreafox: @davisagli Indeed, thank you for noticing this, I've fixed it.

src/components/manage/Blocks/ToC/View.jsx
```diff
@@ -101,6 +101,8 @@ const View = (props) => {
     return entries;
   }, [data, levels, properties, blocksFieldname, blocksLayoutFieldname]);
 
+  // console.log(tocEntries);
```
tiberiuichim: can be removed.

davisagli (approved):

PR #5157, opened 2023-09-04
Added hide and show feature of volto blocks

SaraBianchi:
New feature to hide and show Volto blocks.

In Edit: the hide/show icons have been added below the "delete" button, if a block is hidden but selected, it has an opacity of 65% and "eye" icon keep red, except for the "Description" block which has a different text color and isn't accessible. If the blocks aren't selected they remain opaque by 30%.

screen-edit-show-hide

In View: `RenderBlocks` has been changed, if the `hidden` property exists, the block doesn't render.

screen-view-show-hide

### DIFF
--- a/locales/ca/LC_MESSAGES/volto.po
+++ b/locales/ca/LC_MESSAGES/volto.po
@@ -4458,6 +4458,11 @@ msgstr "S'ha enviat la confirmació de restabliment de la contrasenya"
 msgid "hero"
 msgstr "Heroi"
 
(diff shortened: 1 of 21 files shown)


sneridagh:
@SaraBianchi since this is a major UX change introduction, this should be discussed by the @plone/volto-team 

What's the main use case and motivation for this? Would you mind writing a quick PLIP?

pnicolli:
@sneridagh I can explain tomorrow morning and yes we should definitely write a PLIP for this
@SaraBianchi Let's talk and I can explain the process

PR #5159, opened 2023-09-05
Fix default template for content type which contains block with schem…

### LINKED ISSUE #5158: Adding default template for content type with block(having schemaEnhancer) throwing error.
If you have a schema enhancer for a particular block and then you try to add this block in the layout options of content type will throw error.

Steps to reproduce the error.

1. Create a schemaEnhancer for a particular block
2. Go to any content type and dexterity content layout and select this block.

You will get error.

iFlameing:
…aEnhancer.

Fixes #5158

### DIFF
--- a/news/5158.bugfix
+++ b/news/5158.bugfix
@@ -1,1 +1,1 @@
-Fix default template for content type which contains block with schemaEnhancer. @iFlameing
+Fix editing layout for blocks using schema enhancers. @iFlameing, @davisagli
--- a/news/5158.bugfix
+++ b/news/5158.bugfix
@@ -1,0 +1,1 @@
+Fix default template for content type which contains block with schemaEnhancer. @iFlameing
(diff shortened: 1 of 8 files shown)


davisagli (changes requested):
@iFlameing This doesn't look like the right fix to me. We need to figure out why the schema enhancer breaks and fix it, not disable the enhancers.

davisagli:
@iFlameing Oh, now I understand. You are turning off schema extenders only for the block settings schema. Let me check whether this is the best place to do it, but that makes more sense.

davisagli (commented):
@iFlameing I refactored so that the logic for whether to apply the enhancers is in BlockDataForm, not withVariationSchemaEnhancer. The job of withVariationSchemaEnhancer is to actually apply the enhancers, not decide whether it's appropriate to do that.

news/5158.bugfix
```diff
@@ -0,0 +1 @@
+Fix default template for content type which contains block with schemaEnhancer. @iFlameing
```
davisagli, line 1:
```suggestion
Fix editing layout for blocks using schema enhancers. @iFlameing, @davisagli
```

davisagli (approved):
@iFlameing I also added tests. I think it's ready now.

avoinea (approved):

sneridagh (approved):

PR #5160, opened 2023-09-05
feat: proxy response headers

mamico:


JeffersonBledsoe:
@mamico Could it be worth making these headers configurable through Volto config/ environment variables?

PR #5165, opened 2023-09-08
Fix use of CSS modules in webpack 5

### LINKED ISSUE #5019: less modules not working on Volto 17
I have an addon that uses `less` modules. In `Volto` 16, the CSS is generated normally. In `Volto` 17, in production mode, the `less` CSS is not added to the bundle. The strange thing is that in development mode everything works ok. Does anyone have any idea what might be going on? I also tried CSS modules and SCSS modules and had the same problem.

wesleybl:
In `webpack 5`, when we use the plugin `mini-css-extract-plugin`, the variable `loaderContext._module.matchResource` is filled. This variable is considered when generating the hash of the CSS class name. However, in Volto, the `mini-css-extract-plugin` plugin is only used in the client build. So the CSS class name looked different on the client and the server.

Then, the function that generates the class name was customized, so as not to consider the `loaderContext._module.matchResource` variable.

fixes #5019

### DIFF
--- a/webpack-plugins/webpack-less-plugin.js
+++ b/webpack-plugins/webpack-less-plugin.js
@@ -4,6 +4,24 @@ const MiniCssExtractPlugin = require('mini-css-extract-plugin');
 const PostCssFlexBugFixes = require('postcss-flexbugs-fixes');
 const postcssLoadConfig = require('postcss-load-config');
 
+const interpolateName = require('loader-utils').interpolateName;
+
+function normalizePath(file) {
+  return path.sep === '\\' ? file.replace(/\\/g, '/') : file;
+}
+
+// Custom function to not use 'loaderContext._module.matchResource' in hashing CSS class name.
+function getLocalIdent(loaderContext, localIdentName, localName, options) {
+  const relativeResourcePath = normalizePath(
+    path.relative(options.context, loaderContext.resourcePath),
+  );
+
+  // eslint-disable-next-line no-param-reassign
+  options.content = `${options.hashPrefix}${relativeResourcePath}\x00${localName}`;
+
+  return interpolateName(loaderContext, localIdentName, options);
+}
+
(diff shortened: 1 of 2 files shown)


tiberiuichim (approved):

webpack-plugins/webpack-less-plugin.js
```diff
@@ -4,6 +4,26 @@ const MiniCssExtractPlugin = require('mini-css-extract-plugin');
...
+// Custom function to not use 'loaderContext._module.matchResource' in hashing CSS class name.
+function getLocalIdent(loaderContext, localIdentName, localName, options) {
+  let relativeMatchResource = '';
+
+  const relativeResourcePath = normalizePath(
+    path.relative(options.context, loaderContext.resourcePath),
+  );
+
+  // eslint-disable-next-line no-param-reassign
+  options.content = `${options.hashPrefix}${relativeMatchResource}${relativeResourcePath}\x00${localName}`;
```
davisagli, line 22: I don't see a compelling reason to interpolate the empty string

```suggestion
  const relativeResourcePath = normalizePath(
    path.relative(options.context, loaderContext.resourcePath),
  );

  // eslint-disable-next-line no-param-reassign
  options.content = `${options.hashPrefix}${relativeResourcePath}\x00${localName}`;
```

davisagli (approved):
LGTM, but one tiny suggestion

webpack-plugins/webpack-less-plugin.js
sneridagh: @wesleybl please take a look at David's suggestion.

wesleybl: @davisagli @sneridagh I had left it as it was, to be as similar as possible to the original code, in order to facilitate future comparisons. But I don't know if that makes a lot of sense. So, I made the changes suggested by David.

PR #5167, opened 2023-09-09
Add dockerized approach to add-on generator

sneridagh:
Needs docs, I will update it while writing my training.


packages/generator-volto/generators/addon/templates/Makefile
```diff
@@ -34,35 +21,131 @@ GREEN=`tput setaf 2`
 RESET=`tput sgr0`
 YELLOW=`tput setaf 3`
 
+PLONE_VERSION=6
+VOLTO_VERSION=17.0.0-alpha.25
```
davisagli, line 25: @sneridagh shouldn't this be configurable?

sneridagh: What do you have in mind? In all our projects we configure it in the Makefiles themselves.

davisagli: @sneridagh Never mind, I was thinking about the app generator (which takes a command line option with the volto version), not the addon generator.

PR #5180, opened 2023-09-14
16.24.0 backports

sneridagh:


avoinea:
@sneridagh Does this include also #5053? Would be nice to have it also in 16.x

sneridagh:
@avoinea I just added them. It seemed to me too much but they merged correctly.

PR #5185, opened 2023-09-14
Use navroot to render the sitemap (targets `17.x.x`)

erral:
In #5105 we changed the sitemap component to be rooted to the language.

Instead doing that, in the PR we use the available navroot to do it.

Moreover I have extracted the hardcoded value of 4 used to call the `getNavigation` action to a `siteMapDepth` setting.


src/components/theme/Sitemap/Sitemap.jsx
```diff
@@ -106,25 +102,20 @@ export default compose(
...
       promise: ({ location, store: { dispatch, getState } }) => {
         if (!__SERVER__) return;
-        const { settings } = config;
 
-        const path = getSitemapPath(
-          location.pathname,
-          settings.isMultilingual
-            ? toBackendLang(getState().intl.locale)
-            : null,
+        const navroot = getState().navroot.data?.navroot?.navroot?.['@id'];
```
tiberiuichim, line 115: Is the navroot available in both 7.x and 8.x branches of plone.restapi?

erral: No right now it is not. But I can try to add a PR to backport them .

davisagli: Why does it need to be backported? This fix will go in Volto 17 and maybe Volto 16, which both expect plone.restapi 8.x, even on Plone 5.

tiberiuichim: https://github.com/plone/plone.restapi/blob/ed192208499989b379de8534a3aa12d037316db4/setup.py#L11

Volto 16 works fine with Plone 4, we're using it in production on a complex site. If this PR is backported to Volto 16, it will be a breaking change.

davisagli: Okay, I didn't know you are using Volto 16 with Plone 4. That is not officially supported, as far as I am aware, and there are no automated tests for it in volto. So this is a risk you took on in using Volto 16 with an unsupported Plone version. But, I agree we should avoid breaking that, unless it's really hard to avoid.

erral: The navroot endpoints are not that complex, I would say that they could be backported to 7.x and 8.x easily. But let's see what the tests say :sweat_smile:

erral: See: https://github.com/plone/plone.restapi/pull/1702

src/components/theme/Sitemap/Sitemap.jsx
```diff
@@ -23,12 +23,6 @@ const messages = defineMessages({
   },
 });
 
-export function getSitemapPath(pathname = '', lang) {
```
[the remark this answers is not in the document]
erral: I have kept the function and the tests, but mark it as deprecated. This way, as you say if anyone is using the function, it keeps working.

Would that be enough?

sneridagh (approved):
LGTM! 

/cc @davisagli @tiberiuichim ?

tiberiuichim (approved):

sneridagh:
@erral Sorry, apparently, this got lost in the sea of PRs, merging now!

PR #5191, opened 2023-09-17
Fixes #5128 (sharing page - Save/cancel buttons - cms ui)

### LINKED ISSUE #5128: a11y - sharing page - Save/cancel buttons - cms ui
**Describe the bug**
- When navigating using the keyboard, the 'Save' button appears before the 'Cancel' button. This is not an expected behavior.

**To Reproduce**
- Run through the page using the keyboard 

**Expected behavior**
The cancel button need to appear first and after that the save one.

**Screenshots**
Schermata 2023-08-24 alle 17 14 57

Volto Version 17.0.0-alpha.19
Plone Version 6.0.4

ZubairImtiaz3:
Fixes #5128

**Issue:**
The initial HTML structure had the "Submit" button before the "Cancel" button. However, with the use of float: right; visually, the "Cancel" button appeared first and the "Submit" button afterward. This caused an accessibility issue where, despite the visual order, the tab order followed the HTML structure, leading to the "Submit" button being focused before the "Cancel" button.

**Resolution:**
To address the focus order discrepancy without changing the visual order, the structure was modified to use Semantic UI's grid system, ensuring that the buttons are right-aligned. By adjusting the HTML order to have the "Cancel" button first, both the visual display and tabbing order were aligned, fixing the focus issue.


ZubairImtiaz3:
@mister-roboto Already signed the Plone Contributor Agreement.

stevepiercy:
@ZubairImtiaz3 please read and follow [Contributing to Plone](https://6.docs.plone.org/contributing/index.html) and [First-time contributors](https://6.docs.plone.org/contributing/first-time.html).

stevepiercy:
Also please edit the PR description to include "Fixes #5128" so that when this PR is merged, then the linked issue will close automatically. Thank you!

ZubairImtiaz3:
@stevepiercy Signed the contract and joined the plone organization.

stevepiercy:
@ZubairImtiaz3 you edited the title of the issue, not the description, to include "Fixes #5128". GitHub only cares about the description.

Also you need to add a change log entry.

Both of these are covered in https://6.docs.plone.org/contributing/index.html#change-log-entry and Step 2 in https://6.docs.plone.org/contributing/first-time.html#create-a-pull-request-from-your-fork

ZubairImtiaz3:
@stevepiercy Thanks for guiding me through the process! I've now updated the description to include "Fixes #5128", added a change log entry, and ensured all checks have passed.

stevepiercy:
Thank you, @ZubairImtiaz3. The next step is to request a review. I usually look at the suggestions that GitHub provides in the upper right, and the GitHub usernames that recently edited files, for reviewers.

sneridagh:
@ZubairImtiaz3 
I don't think that put that into a grid is the solution... Can't we find a better one? Removing the float and reorder the buttons might be enough.

I'd like also to have a Cypress a11y test in place that assess that this is fixed. There are some in place, maybe you can take a look.

sneridagh (changes requested):
@ZubairImtiaz3
I don't think that put that into a grid is the solution... Can't we find a better one? Removing the float and reorder the buttons might be enough.

I'd like also to have a Cypress a11y test in place that assess that this is fixed. There are some in place, maybe you can take a look.

ZubairImtiaz3:
@sneridagh In the original implementation, the buttons were positioned at the end of the container. By removing the float and reordering them without additional styles, they naturally positioned at the start of the container. To ensure they align with the original design, by being at the end of the container, I've now leveraged Semantic UI's flex utilities.

Regarding the Cypress a11y test, I've added a test under cypress/tests/core/basic/a11y.js that ensures there are no accessibility violations on the sharing page. The test passed successfully on my local setup, validating that the issue is now addressed in an accessible manner.

<a href="https://ibb.co/7rY4g42">Screenshot-2023-10-01-211357</a>

src/components/manage/Sharing/__snapshots__/Sharing.test.jsx.snap
```diff
@@ -154,59 +154,63 @@ exports[`Sharing renders a sharing component 1`] = `
         </p>
       </div>
       <div
-        className="ui clearing attached segment actions"
+        className="ui clearing attached segment ui"
```
sneridagh, line 157: Duplicate `ui` className.

src/components/manage/Sharing/Sharing.jsx
```diff
@@ -460,29 +460,31 @@ class SharingComponent extends Component {
...
-                  secondary
-                  aria-label={this.props.intl.formatMessage(messages.cancel)}
-                  title={this.props.intl.formatMessage(messages.cancel)}
-                  floated="right"
-                  onClick={this.onCancel}
-                >
-                  <Icon className="circled" name={clearSVG} size="30px" />
-                </Button>
+              <Segment className="ui" attached clearing>
+                <div className="ui right aligned container">
```
sneridagh, line 464: This is not how you use SemanticUI. You don't use the semantic classes outside the SemanticUI components (which uses props to handle them). I'd prefer even to use bare CSS to miss how SemanticUI works.

sneridagh (changes requested):
Sorry for the late review.

Also, if we have to get to a fix to this, all the forms have a centralized way to put the actions (see https://github.com/plone/volto/blob/d2e7e574f7b3f4e12714b975f5cc306981acd7de/src/components/manage/Form/Form.jsx#L755).
If we are to fix this, we should do it for all forms.
There are exceptions, because of the nature of the form (if they are heavily customized), like the sharing one.

ichim-david:
Closing this pull request as abandoned, it's been quite a while since last change attempt, if there is a desire to be continued the conflicts must be fixed and the feedback taken into consideration.

PR #5192, opened 2023-09-18
Remove integration with volto-style-block, as it's not working anymore

dobri1408:
This was left from the days when volto-slate was an addon. I see that volto-style-block doesn't work natively with slate, so this doesn't have to remain.  The volto-style-blocks can add the CSS by itself.


avoinea (approved):

sneridagh (approved):
@ksuess double checking with you since I know you use it.

ksuess (approved):

Issue #5126, opened 2023-08-24
a11y - sharing page - table labels - cms ui


Wagner3UB:
**Describe the bug**
- Checkboxes: Unassociated label with name and function

**To Reproduce**
- go with the screen-reader to any table checkbox item

**Expected behavior**
- A checkbox label must inform the row/column labels and the current status

**Screenshots**
<img width="1309" alt="Schermata 2023-08-24 alle 16 58 44" src="https://github.com/plone/volto/assets/60133113/018f4b0d-e4fb-47ef-91bd-1561d50cfe3d">

**Software (please complete the following information):**

Volto Version 17.0.0-alpha.19
Plone Version 6.0.4

**Additional context**

ShivangRawat30:
I would like to work on this issue.

Wagner3UB:
Hello @ShivangRawat30 sure, go ahead

stevepiercy:
@ShivangRawat30 please follow https://6.docs.plone.org/contributing/index.html and https://6.docs.plone.org/contributing/first-time.html.

abhinaisai2002:
Hey, I didnt understood this issue correctly, can you provide some more info about it.

stevepiercy:
@abhinaisai2002 what specifically is not clear about the issue description?

Perhaps reading this reference, which describes 4 ways of hiding label text, may help: https://www.w3.org/WAI/tutorials/forms/labels/#hiding-label-text

abhinaisai2002:
> @abhinaisai2002 what specifically is not clear about the issue description?
> 
> Perhaps reading this reference, which describes 4 ways of hiding label text, may help: https://www.w3.org/WAI/tutorials/forms/labels/#hiding-label-text

What is the need of the function for the label? Could you more explain me about that

stevepiercy:
@abhinaisai2002 are you familiar with web accessibility (a11y)? I don't know what else I can say, beyond that link I provided.

abhinaisai2002:
@stevepiercy actually i was referring the function keyword in the context of the bug
```
Describe the bug
Checkboxes: Unassociated label with name and **function**
```

That's where i was confused, will check a11y and try to solve the issue

Wagner3UB:
Hey @abhinaisai2002, I rephrase my sentence, the checkboxes should associate the 'name' column with the top row that describes it. For example, the first checkbox should read 'name' + 'logged-in users,' and make a relation with 'Can add' and indicate whether the checkbox is selected or not.

[...]

PR #5201, opened 2023-09-20
a11y for checkboxes on the sharing page issue #5126

Wagner3UB:
Fixing the a11y issue [#5126](https://github.com/plone/volto/issues/5126)


src/components/manage/Sharing/Sharing.jsx
```diff
@@ -412,8 +416,21 @@ class SharingComponent extends Component {
...
+                              )}
+                              aria-label={this.props.intl.formatMessage(
+                                messages.assignNewRoles,
+                                {
+                                  entry: entry.title,
+                                  role: role.id,
+                                },
+                              )}
                               onChange={this.onChange}
-                              value={`${entry.id}:${role.id}`}
```
giuliaghisini, line 416: why you have removed property 'value'? 
i think it's wrong

giuliaghisini (changes requested):

src/components/manage/Sharing/Sharing.jsx
Wagner3UB: @giuliaghisini By mistake, I didn't want to delete.

giuliaghisini (approved):

src/components/manage/Sharing/Sharing.jsx
```diff
@@ -83,6 +83,10 @@ const messages = defineMessages({
     id: 'Permissions have been updated successfully',
     defaultMessage: 'Permissions have been updated successfully',
   },
+  assignNewRoles: {
+    id: 'Assign the {entry} to the {role} role',
+    defaultMessage: 'Assign the {entry} to the {role} role',
```
davisagli, line 88: This seems backwards in English; the checkbox assigns a role to a user or group. I think it should be:

```suggestion
    id: 'Assign the {role} role to {entry}',
    defaultMessage: 'Assign the {role} role to {entry}',
```

Wagner3UB: Thanks @davisagli. I made the changes you suggested.

davisagli: @Wagner3UB It's better, but in English we would say "Assign the Editor role to davisagli", not "Assign the Editor role to the davisagli"

davisagli (approved):

PR #5205, opened 2023-09-20
fixed deprecated buffer.from

### LINKED ISSUE #5198: create-addon-loader.js uses deprecated buffer api
**Describe the bug**
For fun I wanted to see if Volto would run with bun.sh, which results in an
error from create-addon-loader.js
```
fs.writeFileSync(addonsLoaderPath, new Buffer.from(code));                                     ^
TypeError: function is not a constructor (evaluating 'new Buffer.from(code)')
```

It seems that we are using a deprecated call even when using Node 16.
https://nodejs.org/docs/latest-v16.x/api/buffer.html#new-bufferarray

**Expected behavior**
We should use Buffer.from(code) instead of new Buffer.from(code)

Junko-Takeguchi:
Closes #5198 
Updated the deprecated package.
image


ichim-david:
@Junko-Takeguchi you need to add news entry otherwise the pull request fails, see other pull requests on how todo it

Junko-Takeguchi:
@ichim-david thanks for the help

PR #5206, opened 2023-09-21
document the querystringSearchGet setting

erral:
See https://github.com/plone/volto/pull/4658 and https://github.com/plone/plone.restapi/issues/1252

The setting is not documented in the setting reference, and it is worth to have it there.

### DIFF
--- a/docs/source/configuration/settings-reference.md
+++ b/docs/source/configuration/settings-reference.md
@@ -445,7 +445,8 @@ querystringSearchGet
     Volto uses `HTTP POST` requests to query the `@querystring-search` endpoint.
     This can create a lot of traffic between Volto and the backend, and can also create a lot of cache misses.
 
-    By modifying this configuration setting and setting it to `true` the endpoint queries will be executed as `HTTP GET` requests and thus any proxy-cache in between Volto and the backend may cache those queries making your site performance better.
+    By modifying this configuration setting and setting it to `true`, the endpoint queries will be executed as `HTTP GET` requests.
+    Thus any proxy cache in between Volto and the backend may cache those queries, improving your site performance.
 
     Please, be aware that this could break some other functionality in your site or some of your queries may break because they contain a lot of parameters (there is an HTTP spec that puts a limit on the lenght of a URL). Please test this setting properly before enabling in a production site.
 
--- a/docs/source/configuration/settings-reference.md
+++ b/docs/source/configuration/settings-reference.md
@@ -448,7 +448,9 @@ querystringSearchGet
     By modifying this configuration setting and setting it to `true`, the endpoint queries will be executed as `HTTP GET` requests.
     Thus any proxy cache in between Volto and the backend may cache those queries, improving your site performance.
 
-    Please, be aware that this could break some other functionality in your site or some of your queries may break because they contain a lot of parameters (there is an HTTP spec that puts a limit on the lenght of a URL). Please test this setting properly before enabling in a production site.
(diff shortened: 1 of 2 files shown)


docs/source/configuration/settings-reference.md
```diff
@@ -441,6 +441,14 @@ siteTitleFormat
         }
     ```
 
+querystringSearchGet
+    Volto uses `HTTP POST` requests to query the `@querystring-search` endpoint.
+    This can create a lot of traffic between Volto and the backend, and can also create a lot of cache misses.
+
+    By modifying this configuration setting and setting it to `true` the endpoint queries will be executed as `HTTP GET` requests and thus any proxy-cache in between Volto and the backend may cache those queries making your site performance better.
```
stevepiercy, line 448:
```suggestion
    By modifying this configuration setting and setting it to `true`, the endpoint queries will be executed as `HTTP GET` requests.
    Thus any proxy cache in between Volto and the backend may cache those queries, improving your site performance.
```

docs/source/configuration/settings-reference.md
```diff
@@ -441,6 +441,14 @@ siteTitleFormat
...
+
+    Please, be aware that this could break some other functionality in your site or some of your queries may break because they contain a lot of parameters (there is an HTTP spec that puts a limit on the lenght of a URL). Please test this setting properly before enabling in a production site.
```
stevepiercy, line 450:
```suggestion
    Please be aware that this could break some other functionality in your site, or some of your queries may break, when they contain more than 2000 characters.
    [See an explanation of character limits in URLs](https://stackoverflow.com/a/417184/2214933).
    Please test this setting properly before enabling in a production site.
```

stevepiercy (changes requested):
Minor grammar, and added a reference to SO post. What do you think?

erral:
Thanks @stevepiercy !!!!

docs/source/configuration/settings-reference.md
```diff
@@ -441,6 +441,17 @@ siteTitleFormat
...
 
+querystringSearchGet
+    Volto uses `HTTP POST` requests to query the `@querystring-search` endpoint.
+    This can create a lot of traffic between Volto and the backend, and can also create a lot of cache misses.
+
+    By modifying this configuration setting and setting it to `true`, the endpoint queries will be executed as `HTTP GET` requests.
+    Thus any proxy cache in between Volto and the backend may cache those queries, improving your site performance.
+
+    Please be aware that this could break some other functionality in your site, or some of your queries may break, when they contain more than 2000 characters.
+    [See an explanation of character limits in URLs](https://stackoverflow.com/a/417184/2214933).
```
stevepiercy, line 452: Fix a redirect caught by linkcheckbroken.

```suggestion
    [See an explanation of character limits in URLs](https://stackoverflow.com/questions/417142/what-is-the-maximum-length-of-a-url-in-different-browsers/417184#417184).
```

stevepiercy (changes requested):
I gave you a bad link. This should fix linkcheckbroken.

stevepiercy (approved):
LGTM!

davisagli (approved):
Thanks!

davisagli:
@wesleybl We are using it on a large site and haven't had any problems, but there could be a problem if there are too many criteria.

stevepiercy:
@wesleybl I just pushed updated docs with an explanation about the hazards of GET requests with a large querystring:

https://6.docs.plone.org/volto/configuration/settings-reference.html#main-settings

Scroll to `querystringSearchGet`.

PR #5207, opened 2023-09-21
fix: restrict moment.js locales to supported languages

mamico:
The restriction of moment locales in webpack no longer works, probably after an update of moment.

This is the detail of the current bundle

image

According to the documentation I added a webpack specific plugin, this is the result

image

### DIFF
--- a/news/5207.bugfix
+++ b/news/5207.bugfix
@@ -1,0 +1,1 @@
+restrict moment.js locales to supported languages @mamico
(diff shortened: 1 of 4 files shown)


sneridagh (approved):

davisagli (commented):
@mamico Good catch! It might be something I missed with the webpack 5 upgrade

PR #5208, opened 2023-09-21
(feat):Put default alt as description than image title, because this is the most probably scenario

dobri1408:
With this change editors, will have the alt option already set to the description or title of the image, because this is the most probably scenario.


sneridagh (changes requested):
@dobri1408 this is a breaking change... for existing sites that already expect it to be the title, and suddenly it changes to the description after an upgrade.

I'd make it configurable with a switch in the config. 

/cc @plone/volto-team we would have to discuss this, isn't the description be potentially too long? I don't know enough a11y but could be that there are also restrictions on this.

giuliaghisini:
i think to be consistent, it has to be data.alt, like here
https://github.com/plone/volto/blob/f0284f14742fcdeddfc9f6735c1fcad81f20099d/src/components/manage/Blocks/Image/View.jsx#L80

I also request, for accessibility, if it's possible to set alt attribute as  when alt prop is empty, and not as  because for screen readers and accessibility it's not correct the second way.

Issue #5174, opened 2023-09-12
UX Feedback: Anchor Links / Slug / Linked Headline


tisto:
Volto 17 implements linked headlines:

- PR: https://github.com/plone/volto/pull/3540
- PLIP: https://github.com/plone/volto/issues/4298

I just document the feedback that we received from one of the UX experts from our client:

# Discoverability of the "copy anchor link" feature

- A user has no way to understand how to copy an anchor link
- Functionality to copy anchor links is exposed to viewers of the website (regular users might consider this to be a glitch or bug in the website)
- Copying UX patterns (cargo cult) is not necessarily a good idea. Especially when copying a pattern for a system for developers (GitHub/GitLab)

<img width="1466" alt="Screenshot 2023-09-12 at 08 21 53" src="https://github.com/plone/volto/assets/632105/4524be43-a1da-4a80-9227-745d36069b76">

Note Timo: Even though I know this feature from Github and GitLab, I reported a bug to my team when the new Anchor feature of Volto was introduced in one of our projects. I though it was a visual glitch or regression.

# Fragile

- There is no way for an editor to know if changing any content of a page (e.g. a headline) could break an existing anchor link (this is considered to be a KO criteria for the adoption)
- Making an anchor target explicit (anchor block) or auto-generating the anchors would not completely avoid this problem but both solutions would make it less likely that an editor breaks existing anchors

--

I am just sharing this here with the community for now and I honestly don't know which way to go without further consideration and discussions. This is a hard problem. We might want to do some more research of how other systems do it.

We (kitconcept) have a budget to implement the anchor feature in Volto 17 or in an add-on product. Our clients would be willing to pay us to Open Source the development and they want it to be Open Source. Though, they need to feature in the next few weeks and the current implementation in Volto 17 does not meet their needs.

cc @davisagli @sneridagh 

@sneridagh one question: is it possible to make the "show anchor icon on hover" feature optional or hide it from anonymous viewers? I honestly think that this is highly confusing for a public website and does not provide any useful feature for anonymous users.

# Implementation Options for setting Anchors

- Option A: editors create anchors manually
  - Option A1: editors explicitly add an anchor block, enter a unique ID for the anchor
  - Option A1: each block has an "ID" or "Anchor ID" field where editors can set the ID, independently of the actual content
- Option B: the CMS automatically generates unique anchor IDs (e.g. https://www.plone.org/about-us#s89S92)
[...]

PR #5225, opened 2023-09-25
Improve linked headlines

sneridagh:
Closes: #5174
As discussed in https://github.com/plone/volto/issues/5174

- [x] disable "copy anchor link to clipboard" functionality for anonymous users of the website (this is a feature that no public website has anywhere in the world)
- [x] keep the feature for now for logged-in users, integrators can disable it with a feature toggle
- [x] change the wording of the notification from "Link copied to clipboard" to "Anchor link copied to the clipboard" to make it more obvious that an anchor link has been copied (again, this only works for editors that know the feature)
- [x] fix the bug that the anchor link does not work if your headline has a special character (e.g. "Überschrift")

Unfortunately this is a Chrome issue, it does works well in FF and Safari :( This only happens in SSR. At least, on links, and when the router is present, it does well.

- [x] disable the "copy to anchor link to clipboard" for the page title

### DIFF
--- a/cypress/tests/core/blocks/block-anchors.js
+++ b/cypress/tests/core/blocks/block-anchors.js
@@ -3,28 +3,6 @@ import { slateBeforeEach } from '../../../support/volto-slate';
 describe('Block Tests: Anchors', () => {
   beforeEach(slateBeforeEach);
 
-  it('Add Block: Links', () => {
-    // Change page title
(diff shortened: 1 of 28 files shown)


tiberiuichim (approved):

sneridagh:
Letf TODO: try to normalize the links that have special characters, so we make sure that all of them will work in all situations.

PR #5230, opened 2023-09-26
Remove node requirements in generator-volto, add try-catch clause in code that seems to break in release mode

sneridagh:
Let's see if that works (blind here, since locally, generating the app folder works well). Should be a released package thing.

### DIFF
--- a/packages/generator-volto/generators/app/index.js
+++ b/packages/generator-volto/generators/app/index.js
@@ -243,8 +243,10 @@ Run "npm install -g @plone/generator-volto" to update.`,
       this.globals,
     );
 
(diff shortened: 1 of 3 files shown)


davisagli (commented):
@sneridagh For the alpha version of the generator I adjusted things so that no files need to be deleted: https://github.com/plone/volto/commit/b780de65a4f25d6c1d59b79675a19818c940f27d#diff-918b4e82877daa8aee62ce41977634a3a267f26f503562e1355c416ff643713b

I'd suggest backporting that rather than do it differently.

sneridagh:
closing in favor of:
https://github.com/plone/volto/pull/5232

PR #5240, opened 2023-09-29
documented runtime enviroment variabels

Akshat2Jain:
This PR is related to #4581 
@tiberiuichim is this what you are looking for? 
I did what I understood from the comments.


tiberiuichim:
@Akshat2Jain I meant on this page: https://6.docs.plone.org/volto/configuration/environmentvariables.html

which is generated from https://github.com/plone/volto/blob/master/docs/source/configuration/environmentvariables.md

PR #5241, opened 2023-09-29
mention ipn website

pbauer:


tisto (approved):
@pbauer awesome!!!

news/5241.documentation
```diff
@@ -0,0 +1,2 @@
+Add new website IPN
```
tisto, line 1: @sneridagh don't we have an "internal" towncrier option?

stevepiercy: Yes: https://6.docs.plone.org/contributing/index.html#contributing-change-log-label

pbauer: yes but that's docs, not internal imho

PR #5256, opened 2023-10-01
Bypass reviewdog and install Vale directly

stevepiercy:
Resolve failing CI build due to reviewdog breakage. See https://github.com/plone/volto/pull/5253


stevepiercy:
@sneridagh this is ready for review. I pulled configuration from `plone/documentation`. It now runs Vale, but with the flag `--no-exit` which means "Don't return a nonzero exit code on errors." It also eliminates reviewdog as the test runner.

davisagli (approved):
@stevepiercy Thanks, LGTM

PR #5245, opened 2023-09-29
Fix error: no such file or directory, open 'addon-testing-project/jsconfig.json' in addon clone command

wesleybl:
In `@plone/generator-volto` > 7.0.0-alpha.6, we no longer have `jsconfig.json`, but `tsconfig.json`. So let's test if `jsconfig.json` exists and use it if it does. If it doesn't exist, we use `tsconfig.json`

Fixes #5239

PR #5257, opened 2023-10-02
Support Volto 17

pnicolli:
Use tsconfig by default, fallback to jsconfig if present.
Should I make this a feature? Or internal? Ideas?


wesleybl:
@pnicolli see: #5245

Maybe your implementation is slightly better than mine.

pnicolli:
@wesleybl Oooooops, sorry I didn't see that! It looks basically the same to me, I'll go ahead and close this one.

PR #5258, opened 2023-10-02
Ensure .gitignore gets copied when running `yo @plone/volto-addon`

instification:
Minor bugfix. When running `yo @plone/volto-addon` the `.gitignore` file does not get copied.

I tried to use the suggestions in https://github.com/yeoman/generator/issues/812 but adding `.npmignore` did not solve the problem.

However, iteems like it was already fixed for the app generator: 

https://github.com/plone/volto/blob/afadd4b3b8f4c69f4cef705430d28cb56f0c7b1f/packages/generator-volto/generators/app/index.js#L230-L243

So this PR is porting that fix to the addon generator.


packages/generator-volto/generators/addon/index.js
```diff
@@ -207,11 +207,6 @@ Run "npm install -g @plone/generator-volto" to update.`,
   }
 
   install() {
-    this.renderTemplate(
-      `${this.templatePath()}/**/*`,
-      this.destinationPath(),
-      this.globals,
-    );
```
davisagli, line 214: @instification I think this means that `<%= addonName %>` will no longer be interpolated in Makefile.

davisagli (changes requested):

packages/generator-volto/generators/addon/index.js
instification: Thanks! I've refactored to keep the globals so I think this is going to work ok now.

davisagli (approved):
LGTM now

tiberiuichim (approved):
It would be good to have a comment in the code for this, to explain the WTF for the next developer that looks at the code.

Issue #5214, opened 2023-09-21
a11y - edit page - add block/blocks-chooser search field - cms ui 


Wagner3UB:
**Describe the bug**
After typing something in the "search" field inside the "block chooser" modal an "X" appears to cancel the typed text, but this button has an aria-label "search". Another problem is when the button is focused, it only changes color. It is insufficient for accessibility. The same occours with the input search field, doesn't have a proper focus.

- [x] Button label
- [x] Button focus
- [x] input search field focus

**To Reproduce**

1- Go to https://demo.plone.org/edit
2- enter in edit mode by clicking on the edit button (top-left)
3- Press enter to create a new content
4- Click on the "add block" icon
5- Type some text, the "X" will appear inside the input

**Expected behavior**
Give a coherent name/aria-label to the button
Give proper focus to the button; the minimum requirement is to have borders.
Give proper focus to the input search field

**Screenshots**
<img width="453" alt="Schermata 2023-09-21 alle 16 57 51" src="https://github.com/plone/volto/assets/60133113/94ba3335-567d-4b73-acc3-681c57c1b190">

vikaskumar2703:
I would like to work on this issue

rishuraj1:
Hey @Wagner3UB I really want to contribute, please assign me to this issue and can you give me an heads up. Thanks.

stevepiercy:
@rishuraj1 please read and follow [Contributing to Plone](https://6.docs.plone.org/contributing/index.html) and [First-time contributors](https://6.docs.plone.org/contributing/first-time.html).

rishuraj1:
Hey @Wagner3UB  I identified the issue and was able to work on it. Please look into #5254  . Thanks

PR #2787, opened 2021-10-28
Block chooser search

bipoza:
As a first step to this issue https://github.com/plone/volto/issues/1488, I have created a search input in the BlockChooser component to ease selecting blocks.

This seach box, filters the available blocks filtering by title.

https://user-images.githubusercontent.com/26112509/139233792-9a62e5bb-4416-4bbb-b842-4bd66514bdd5.mov

I think that activating this the component using a shortcut would be nice, but it required some thinking.
What do you think?

I keep working on doing some improvements.

PR #5259, opened 2023-10-02
5214.bugfix - Fixed the aria-label and button accessibility 

aadarsh-nagrath:
Fixed the aria-label for close/clear search button and added circled border to it.

Issue reference #5214

### DIFF
--- a/src/components/manage/BlockChooser/BlockChooserSearch.jsx
+++ b/src/components/manage/BlockChooser/BlockChooserSearch.jsx
@@ -9,6 +9,10 @@ const messages = defineMessages({
     id: 'Search',
     defaultMessage: 'Search',
   },
+  closeButton: {
+    id: 'Close',
+    defaultMessage: 'Close',
+  },
 });
 
 const BlockChooserSearch = ({ onChange, searchValue }) => {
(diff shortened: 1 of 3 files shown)


src/components/manage/BlockChooser/BlockChooserSearch.jsx
```diff
@@ -36,13 +40,13 @@ const BlockChooserSearch = ({ onChange, searchValue }) => {
...
             className="clear-search-button"
-            aria-label={intl.formatMessage(messages.search)}
+            aria-label={intl.formatMessage(messages.closeButton)}
             onClick={() => {
               onChange('');
               searchInput.current.focus();
             }}
           >
-            <Icon name={clearSVG} size="18px" />
+            <Icon className='circled' name={clearSVG} size="18px" />
```
Wagner3UB, line 49: The use of the className here is incorrect, we don't need a circle around the icon all the time, just when focused.

Wagner3UB:
@aadarsh-nagrath[ Look at this PR](https://github.com/plone/volto/pull/2787/files#diff-179af92bdb87b141d715e97149be7c6ab65f4a53d08c6ad225ed13dc6844be9eR511), please. Something was done here to improve accessibility using the mouse; now, we need to find a way to make it work with the keyboard as well

PR #5261, opened 2023-10-03
Prefer resolving DNS to ipv4

davisagli:
If the OS returns both ipv4 and ipv6 addresses when resolving a hostname, prefer the ipv4 one.

For example, this fixes connection errors that happen when volto is connecting to the backend on `localhost` (the default) but `/etc/hosts` has `::1` listed before `127.0.0.1` as an IP address for localhost.

### DIFF
--- a/news/5261.bugfix
+++ b/news/5261.bugfix
@@ -1,0 +1,1 @@
+Adjust DNS resolution to prefer IPv4 addresses when both IPv4 and IPv6 are resolved. @davisagli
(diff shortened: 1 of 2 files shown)


fredvd (commented):
Will this only be active in development setup/startups of the razzle server?

sneridagh (approved):

sneridagh:
@fredvd Nope, this will be active always. But since it's "smart", I'd say to go with it.

PR #5262, opened 2023-10-03
(Fix): Sorting Order and Sorting On are not working in Search Block edit mode

dobri1408:
[screen-capture.webm](https://github.com/plone/volto/assets/50819975/3aa83444-4427-4819-a307-67879485f7c0)


dobri1408:
Hi @sneridagh, the PR is ready for review. In this PR, I have solved the following two problems: if you are on edit, and you have a preexistent search block ( so not created in this instance), and you are clicking on the reversed order, or sort on, the search results are not updated. This is because on the edit, there is a function called onTriggerSearch, that before didn't receive any parameters.  I have also included a test.

tiberiuichim (approved):

avoinea (approved):

ichim-david:
@dobri1408 I've tested your work and indeed the sorting works now when selecting a sort option.
I would like for you to also take into consideration that if you select "No selection" there is no call to
querystring search to get the preview of this query without any sort options passed.
When you save you will see different results from the edit page since the sorting option changed.
Once that is treated if it can be treated we are good to go

ichim-david (approved):

dobri1408:
Hi @ichim-david I have implemented what you have asked.

PR #5263, opened 2023-10-03
Emit a proper URL for the Plone root in a listing

tiberiuichim:
Not that I think there's any valid use case for it, but at least we avoid a nasty warning


sneridagh (commented):
Shouldn't be

```
url = '/'
```

PR #5266, opened 2023-10-03
Normalize the shape of the image component `item` prop if it contains the serialized object after creation to match the one in the catalog.

sneridagh:


davisagli (commented):
I think it makes sense.

src/helpers/Url/Url.test.js
```diff
@@ -350,4 +351,101 @@ describe('Url', () => {
       );
     });
   });
+  describe('flattenScales', () => {
+    it('flattenScales test from serialization', () => {
```
pnicolli, line 355:
```suggestion
    it('flattenScales test from the catalog', () => {
```
This is actually the catalog test

src/helpers/Url/Url.test.js
```diff
@@ -350,4 +351,101 @@ describe('Url', () => {
...
+              '@@images/image-1600-188968febc677890c1b99d5339f9bef1.jpeg',
+            height: 665,
+            width: 1182,
+          },
+        },
+        size: 319364,
+        width: 1182,
+      });
+    });
+    it('flattenScales test from catalog', () => {
```
pnicolli, line 402:
```suggestion
    it('flattenScales test from serialization', () => {
```
This is actually the serialization test

pnicolli (changes requested):

PR #5271, opened 2023-10-04
Accessibility management for Slate blocks and keyboard navigation focus

Wagner3UB:
- Now it is possible to leave a block using the Tab key.
- Slate text containers re-arranged to give the tab a proper path (text input → drag & drop button → delete button)
- Added a general ":focus-visible" to standardize the focus within the application.
- a11y -  general fixes

### DIFF
--- a/packages/volto-slate/src/blocks/Text/keyboard/traverseBlocks.js
+++ b/packages/volto-slate/src/blocks/Text/keyboard/traverseBlocks.js
@@ -74,8 +74,8 @@ export function goDown({ editor, event }) {
 }
 
 export function traverseBlocks(opts) {
-  const { event } = opts;
-  event.preventDefault();
+  // const { event } = opts;
+  // event.preventDefault();
   // return event.shiftKey ? goUp(opts) : goDown(opts);
-  return true;
+  return false;
 }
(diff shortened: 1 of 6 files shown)


tiberiuichim:
@Wagner3UB did you try the list indenting behavior with the tab key? Also, I'd rather consider the Volto blocks editor as the equivalent of the "Word editor", where lines navigation is done using the arrow keys. It's not a form, so navigation does not have to be exclusively with the Tab key. Also, it's still accessible, AFAIK, if you can navigate with the arrow keys rather then Tab.

Wagner3UB:
@tiberiuichim Interesting way of thinking, I will remember this in the future. And no, I hadn't seen the list indenting behavior.

I'm going to reintroduce that part of the code, however, I need to make some osservations:

With the current Volto code
- We can only use the arrow down key twice. After that, the focus is lost, and no further actions are possible. This issue occurs within the first text block. You can test it by opening the editor and pressing the arrow down key three times on the demo site. If you click inside any other block, this behavior cannot be observed anymore, it only happens when you open the editor page. At the moment I don't know how to fix this behavior.
- The trash can and the drag-and-drop buttons are not accessible via the keyboard (or perhaps I am unaware of how to do it)

PR #5288, opened 2023-10-07
Fix page changes not being announced by assistive technology when navigating using the client-side router

### LINKED ISSUE #4724: Page changes are not announced over Screen reader (WCAG 2.4.2 failure)
**Describe the bug**
When changing between pages using a normal browser page transition, the title of the new page is announced to screen readers once the page has loaded. This does not happen with Volto due to the page content change happening client side, skipping the default browser behaviour. This is a common problem in the world of SPAs. See [next.js routing issues](https://github.com/vercel/next.js/issues/7681) and [gatsby routing issues](https://github.com/gatsbyjs/gatsby/issues/5581) for the same problems in other React frameworks.

Here's some other good articles that sum up the problem: https://hidde.blog/accessible-page-titles-in-a-single-page-app/ https://www.gatsbyjs.com/blog/2019-07-11-user-testing-accessible-client-routing/#common-accessibility-barriers-in-client-rendered-web-apps.

**To Reproduce**

1. Visit a page with screen reader enabled (VoiceOver was used by the issue author)
2. Add a new page 
3. Return to the homepage
4. Click the link to go to the page you created 

**Expected behavior**
The contents of the `<title>` element in the head is read aloud so users know what the page they have just navigated to is

**Actual behaviour**
Nothing is announced

**Screenshots**

https://user-images.githubusercontent.com/30210785/233435012-8049cc65-c644-4d90-99f8-c9f64ff97270.mov

** Proposed solution**

- Add an off-screen assertive live region which mirrors the contents of the `<title>` element in the head
- Reset the page focus to the 'body'

***Implementation***

- ~~Update the `LOAD_PROTECTOR` API to wait for content to have been fully loaded~~
- ~~Ensure the `Helmet` component uses this load protector~~
  - Some `Helmet` calls have i18n so the data in the load protector won't match the data in the title
- Update `Helmet` to render the announcer using the passed in title
- Use the load protector
- Update the previously mentioned live region on load protector change
  - We use the title if it's available (We might not have one, but we should do)
  - We use the first `<h1>` if we can't find an element
  - We use the path if we can't find anything else
- We don't want to trigger the live region on page load, as the normal browser behaviour will occur
- Use `document.activeElement.blur()` to reset the focus to the browser default
  - Needs testing, we might need a 'wrapper' element depending on whether the above works correclty

**Solution options**

Taken from Marcy Sutton's article (linked above)

- Dynamically set focus to an HTML wrapper element on page change, to both move focus to the new content and make an announcement in assistive technology.
  - This pattern often uses tabindex="-1" on a DIV or other block-level element to allow focus to be placed on an otherwise non-interactive element.
- Dynamically set focus to a h1-h6 heading element instead of a wrapper to move focus to new content and make a shorter screen reader announcement.
  - This also typically requires tabindex="-1" to focus the heading with JavaScript in a cross-browser way.
- Dynamically set focus to an interactive element like a button to put keyboard users on an operable button/UI control in the correct part of the app and announce it to screen reader users.
  - The name of the button matters a lot here.
- Leave focus where it is and make an [ARIA Live Region](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions) announcement instead.
- Reset focus to the top of the application (i.e. a parent wrapper element) to mimic a traditional browser refresh and announce new content in assistive technology.
- Turn on focus outlines for keyboard and screen reader users while suppressing them for the mouse using [CSS :focus-visible and polyfill](https://github.com/WICG/focus-visible) or the [What Input](https://github.com/ten1seven/what-input) library.
- Any combination of the above

JeffersonBledsoe:
Create a mutation observer on client-side load that tracks when the `<title>` element contents changes and updates an aria live region to announce the page change. Needs cross-browser and cross-screen reader testing

Fixes #4724

## TODO

- [x] How should we handle focus after page change?
- [x] History API isn't listened to.


JeffersonBledsoe:
@plone/volto-accessibility Could we test this / discuss the implementation?

ichim-david:
hey @JeffersonBledsoe I don't have time until the wk busy with some bugs :( I added a calendar entry so I don't forget about it :)

ichim-david:
Status report on testing the feature as part of the Salamina sprint:
1. We've tested https://volto.demo.plone.org/ navigating different urls using the keyboard
 and notice the lack of announcement that we have changed url.
 This was using VoiceOver on Mac and NVDA on Windows
2. We've tested the change on Windows and Mac and heard the announcement of the title
which worked until we played with the browser history navigation that Jeff handled today in
https://github.com/plone/volto/pull/5288/commits/86d5b6b5abc9372f0e4ad0658637801d057bc70a
3. We tested https://www.gov.uk where we heard the location change although it was challenging at times
due to the combination of browsers or screen reader technology, it was easier to hear on NVDA then VoiceOver

Tomorrow we will hopefully talk with the A11y expert that Red Turtle has asked to participate and hopefully get some
feedback also from him on this proposal.

JeffersonBledsoe:
After further discussion with an accessibility consultant at the Salamina sprint, it was concluded that the current behaviour is acceptable. Only with more user testing can we concretely decide what to do with focus and so leaving the focus where it is on page change is an acceptable behaviour for now.

ichim-david (changes requested):
@JeffersonBledsoe you have locking changes that doesn't belong to this pull request and you also set it.only on locking.js file

ichim-david:
> @JeffersonBledsoe you have locking changes that doesn't belong to this pull request and you also set it.only on locking.js file

@JeffersonBledsoe you probably needed to change the selector to be more specific, but it.only shouldn't be there still

JeffersonBledsoe:
Good catch on the `.only`! The other changes are relevant. There are multiple alert on the page at times now so the tests that assumed there was only one need updated

JeffersonBledsoe:
@ichim-david I applied the fix we discussed at the sprint. It doesn't seem to impact anything but is a good safeguard anyway incase somebody does end up with focus on it somehow

packages/volto/src/components/theme/RouteAnnouncer/RouteAnnouncer.jsx
```diff
@@ -0,0 +1,65 @@
...
+      childList: true,
+      subtree: true,
+    });
+    window.addEventListener('popstate', handlePop);
+
+    return () => {
+      observer.disconnect();
+      window.removeEventListener('popstate', handlePop);
+    };
+  }, []);
```
pnicolli, line 40: By just looking at the code I would say that eslint should complain that this array should at least contain the `updatePage` function. Would it be possible to move it inside the `useEffect` next to `handlePop`?

ichim-david (approved):

Wagner3UB (approved):

PR #5289, opened 2023-10-07
move the function to get the lead image sizes to the block configuration

erral:
This way the sizes function can be overridden in a custom project like it can be for the Image block


libargutxi (approved):

sneridagh (approved):

sneridagh:
@erral Could we have some related docs?

pnicolli (approved):

Issue #5268, opened 2023-10-04
a11y - edit page - skip links - cms ui


Wagner3UB:
UPDATE 25/10/2023: This issue was modified to align the content with the landmark modifications created by David, something I was not aware of until I saw the pull request. The skip links are a useful tool to navigate, but aren't the only solution to navigate between important "points of interests". The skip links must work inside the CMS-UI, this is important, and this was done.

**Describe the bug**
This issue will tackle two problems, essentially, but it can serve as a solution for other accessibility situations in the future:

1. When you click the 'edit' button, the default behavior is to focus on the first editable block, which is usually the title block. However, when using assistive technology, this behavior can be inconsistent, leaving users unsure of their location and what action to take. To improve this, the fieldset containing all blocks should be made accessible to screen readers and provide clear information about the editable content before moving to the first element. - **Another issue (https://github.com/plone/volto/pull/5392) to solve this problem**
2. Accurate link identification is crucial for a smooth navigation experience. This means that link names and anchor points should align with the editing page's content. The current "footer/navigation/content" labels aren't entirely accurate and should instead reference terms like "edit main content," "toolbar configurations," or "editable content sidebar."

**To Reproduce**

1- Go to https://demo.plone.org/edit
2- enter in edit mode by clicking on the edit button (top-left)
3- `shift + tab` and navigate to the skip links "skip to footer", "skip to main content" and "skip to navigation"

**Expected behavior**
The expected behavior needs to be discussed, and various options are available. I will provide some previous information, but there are many ways to set this behavior.

- ~~[ ] Make the first focus more informative for assistive technologies~~ - **Another issue (#5392) to solve this problem**
- [ ] The skip links within the CMS-UI can serve as a 'point of interest and redirect.' This means that if the user needs to change their screen position (e.g., from the sidebar to the main content or from the toolbar to the sidebar), they can do so by navigating to the skip links and finding what they need with a proper label ("edit main content," "toolbar configurations," or "editable content sidebar."). This will work to improve the UX and give the user another useful way to access those parts of the CMS-UI besides using the landmarks.
- ~~[ ] Skip links should be accessible early on. I see two possibilities: either the focus shifts to the first skip link after clicking the 'Edit' button or, even before that, when users access the CMS (after logging in), the first click on the edit page should activate a skip link~~. The skip links can be accessed via landmarks
[...]

PR #5290, opened 2023-10-07
A11y - Updated aria-labels for landmarks, added landmark for sidebar and plug skiplinks

ichim-david:
This pull request addresses some of the issues mentioned in the following issue
https://github.com/plone/volto/issues/5268

1. "The second option brings us to another scenario. The CMS's main page doesn't have a sidebar, which means the skip links should also be different from the 'edit' page."
Added a pluggable section to skiplinks this way if you have other links that you want to skip to you can add extra links
in a Plug.

2. There is now a complementary role to the sidebar so you can easily focus on it using a screen reader with the landmark
section

3. Updated aria-label for Footer, Navigation, Sidebar and Skip links:
Everything is now Capitalized based on examples I've seen on other websites such as Apple.com or looking at the documentation examples of https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/complementary_role

Here is a before and after landmark reading out of a Mac

current-aria-label
modified-aria-label


sneridagh (approved):
@Wagner3UB Could you please also take a look?

Wagner3UB:
I'm not familiar with the pluggable component, so I can't help with that. 

However, the result of these changes efficiently solves the sidebar problem.

I have a small question: When we are on the home edit page, without the sidebar, the landmarks guide shows 'sidebar empty complementary'. This way, it kinda helps show that the sidebar is empty and "doesn't work for nothing", but it's not quite accurate because it's not that there's no content in the sidebar, we don't have the sidebar. And the fact that it's clickable but doesn't take us anywhere can be confusing.

However, I'm not sure if is something possible.

Wagner3UB (approved):

PR #5293, opened 2023-10-07
Initial structure for Volto docs reorganization

stevepiercy:


docs/source/README.md
```diff
@@ -0,0 +1,56 @@
...
+They provide practical steps that serve our study.
+
+Most of Volto's tutorials are in the [GitHub repository `plone/training`](https://github.com/plone/training), and are served from the [Plone Training website](https://training.plone.org/).
+
+There may be some tutorials buried deep inside Volto documentation.
+Additionally some content from the Trainings may be well-suited for a tutorial.
+Both of these should be moved to the directory `/tutorials/`.
+
+
+## Guides
```
tiberiuichim, line 21: More info from Steve:

> I want to avoid more than 2 levels. However, Guides does not count as one of those levels. It will be only an indicator, not displayed as a folder, like this: https://docs.readthedocs.io/en/stable/ even though it may reside in a folder. Hope that is clear.
> As far as concatenating guides, it depends. If they all are closely related, then maybe concatenation makes sense. But at the same time, one long topic is not as easy to use.

Issue #3901, opened 2022-11-16
Can't change between block and page settings tabs with a screen reader


JeffersonBledsoe:
**Describe the bug**

When using a screenreader, the 'Page' and 'Block' tabs in the sidebar aren't clickable. Instead, pressing enter or space makes the sidebar go fullscreen.

**To Reproduce**

1. Open a screen reader
2. Visit an edit page
3. Focus on a block so the block settings appear
4. Use the screen reader to navigate to the Page settings
5. Press enter or space

**Expected behavior**
 The sidebar changes the tab to show the page settings

**Actual behavior**
The sidebar expands to be fullscreen

**Screenshots**
The below video shows me using the screen reader to navigate to the block settings and trying to press <kbd>Enter</kbd> or <kbd>Space</kbd> to change the tab in the sidebar. Instead, the `Fullsize sidebar` button is pressed.

https://user-images.githubusercontent.com/30210785/202165651-af5ae863-83f5-46f6-a666-5d1476023e0b.mov

**Software (please complete the following information):**

- OS: macOS 12.16.1 (using VoiceOver)
- Browser: Google Chrome & Safari
- Volto Version: [16.0.0-alpha.50](https://github.com/plone/volto/releases/tag/16.0.0-alpha.50)
- Plone Version: 6.0.0b3

**Additional context**

Viewing the current activeElement from the browser console shows that the focused element is the fullsize button even though I'm focused on the tabs with the screen reader. The tabs are reported to the screenreader as links, but they don't seem to get focused. This may be a problem with Semantic UI rather than the Volto implementation, I've yet to investigate further.

Vsoni27:
Hello, I'm interested in working on this issue for Plone as part of the Google Summer of Code program. As a beginner in open source, I was hoping to get more information about the problem to better understand it.

Could you provide more details about the issue, such as the source code files or functions that are causing the problem? I'm still exploring the Plone documentation and familiarizing myself with the technologies and tools used by the CMS, so any guidance or resources you can share would be appreciated.

ichim-david:
@JeffersonBledsoe I've added the pull request that is ready for testing https://github.com/plone/volto/pull/5294

PR #5294, opened 2023-10-07
Use buttons for sidebar formtabs for proper keyboard and screen reader support

ichim-david:
Will fix https://github.com/plone/volto/issues/3901


sneridagh (changes requested):
LGTM! Could you please add the on hover change to pointer cursor? I just checked and it's missing. Thanks!

ichim-david:
@sneridagh I've added the ui button class thus gaining the cursor on hover as seen in my screenshot 
menu-item-button

@sneridagh is this a breaking change because the buttons are now buttons instead of empty links and have some extra classes? They look the same and function better but I leave it to you to make a decision if we need to announce anything about this change

sneridagh (approved):

PR #5295, opened 2023-10-07
Bundle optimization

pnicolli:
Reduce bundle size by splitting code. Work started from editor/admin interfaces.

Already split as of this writing:
- Folder Contents
- Rules Controlpanel
- Users Controlpanel
- Relations Controlpanel
- Form components
- Diff

Reduced bundle size by ~7% (gzipped), from 582.92 to 542.6.
A lot more can be done.

### DIFF
--- a/news/5282.feature
+++ b/news/5282.feature
@@ -1,0 +1,1 @@
+Upgrade to @plone/scripts 3.0.1 @sneridagh
--- a/package.json
+++ b/package.json
@@ -241,7 +241,7 @@
     "@loadable/component": "5.14.1",
     "@loadable/server": "5.14.0",
     "@loadable/webpack-plugin": "5.15.2",
-    "@plone/scripts": "3.0.0",
+    "@plone/scripts": "3.0.1",
     "@testing-library/cypress": "9.0.0",
     "@testing-library/jest-dom": "5.16.4",
     "@testing-library/react": "12.1.5",
--- a/package.json
+++ b/package.json
@@ -412,6 +412,7 @@
     "@storybook/builder-webpack5": "^6.5.15",
     "@storybook/manager-webpack5": "^6.5.15",
     "@storybook/react": "^6.5.15",
+    "@types/loadable__component": "5.13.5",
     "@types/react-test-renderer": "18.0.1",
     "@typescript-eslint/eslint-plugin": "6.7.0",
     "@typescript-eslint/parser": "6.7.0",
--- a/packages/coresandbox/src/components/Blocks/TestBlock/Data.jsx
+++ b/packages/coresandbox/src/components/Blocks/TestBlock/Data.jsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import { useIntl } from 'react-intl';
-import { BlockDataForm } from '@plone/volto/components';
+import { BlockDataForm } from '@plone/volto/components/manage/Form';
 
 const TestBlockData = (props) => {
   const { block, blocksConfig, data, onChangeBlock } = props;
--- a/packages/volto-slate/src/blocks/Table/TableBlockEdit.jsx
+++ b/packages/volto-slate/src/blocks/Table/TableBlockEdit.jsx
@@ -11,7 +11,8 @@ import cx from 'classnames';
 import { defineMessages, injectIntl } from 'react-intl';
 
 import Cell from './Cell';
-import { BlockDataForm, Icon, SidebarPortal } from '@plone/volto/components';
+import { Icon, SidebarPortal } from '@plone/volto/components';
+import { BlockDataForm } from '@plone/volto/components/manage/Form';
 import TableSchema from './schema';
 
 import rowBeforeSVG from '@plone/volto/icons/row-before.svg';
--- a/packages/volto-slate/src/blocks/Text/DefaultTextBlockEditor.jsx
+++ b/packages/volto-slate/src/blocks/Text/DefaultTextBlockEditor.jsx
@@ -12,11 +12,7 @@ import {
   validateFileUploadSize,
 } from '@plone/volto/helpers';
 import config from '@plone/volto/registry';
-import {
-  BlockDataForm,
-  SidebarPortal,
-  BlockChooserButton,
-} from '@plone/volto/components';
+import { SidebarPortal, BlockChooserButton } from '@plone/volto/components';
 
 import { SlateEditor } from '@plone/volto-slate/editor';
 import { serializeNodesToText } from '@plone/volto-slate/editor/render';
@@ -34,7 +30,7 @@ import { handleKey } from './keyboard';
 import TextBlockSchema from './schema';
 
 import imageBlockSVG from '@plone/volto/components/manage/Blocks/Image/block-image.svg';
-
+import { BlockDataForm } from '@plone/volto/components/manage/Form';
 import './css/editor.css';
 
 // TODO: refactor dropzone to separate component wrapper
--- a/packages/volto-slate/src/elementEditor/PluginEditor.jsx
+++ b/packages/volto-slate/src/elementEditor/PluginEditor.jsx
@@ -3,9 +3,10 @@ import { isEqual } from 'lodash';
 import React from 'react';
 import { useDispatch } from 'react-redux';
 import { ReactEditor } from 'slate-react';
-import { Icon as VoltoIcon, BlockDataForm } from '@plone/volto/components';
+import { Icon as VoltoIcon } from '@plone/volto/components';
 import { setPluginOptions } from '@plone/volto-slate/actions';
 import BaseSchemaProvider from './SchemaProvider';
+import { BlockDataForm } from '@plone/volto/components/manage/Form';
 
 import briefcaseSVG from '@plone/volto/icons/briefcase.svg';
 import checkSVG from '@plone/volto/icons/check.svg';
--- a/src/components/index.js
+++ b/src/components/index.js
@@ -77,7 +77,6 @@ export { default as Actions } from '@plone/volto/components/manage/Actions/Actio
 export { default as Add } from '@plone/volto/components/manage/Add/Add';
 export { default as AddonsControlpanel } from '@plone/volto/components/manage/Controlpanels/AddonsControlpanel';
 export { default as UndoControlpanel } from '@plone/volto/components/manage/Controlpanels/UndoControlpanel';
-export { default as Contents } from '@plone/volto/components/manage/Contents/Contents';
 export { default as Circle } from '@plone/volto/components/manage/Contents/circle';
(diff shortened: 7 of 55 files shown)


pnicolli:
As of today, all controlpanels, form components and widgets are code split. We are down to 496kb gzipped (started from 582).

pnicolli:
@tiberiuichim I added the babel plugin but I have a problem.

I added it to the razzle config and it works fine for building Volto with yarn, but it breaks jest (try to comment out the Header component in `src/components/index.js` to reproduce)

If I add the same configuration in babel.js then jest works fine but webpack breaks while building Volto.
Any ideas or pointers?

pnicolli:
After a discussion held during today's volto meeting, we decided to try to skip the babel transform imports plugin and moving all lazy loaded components back to the main components index and lazy load those there. I will try in the next days and let you know, so we can see how the result would be.

pnicolli:
@sneridagh @tiberiuichim  Removed babel-plugin-transform-imports and readded imports in components index. I realized I could import the lazyloaded components from their own index, and this looks a lot cleaner to me. This should also make this PR a non breaking change, actually. If it looks good for you, I will rebase onto the newest main branch for the final review.

sneridagh:
@pnicolli ok! let's talk about it tomorrow in the meeting. I have a couple of questions.

packages/volto/cypress/tests/core/controlpanels/upgrade.js
```diff
@@ -39,7 +39,7 @@ describe('Upgrade Site Tests', () => {
       body: {
         ...getUpgradeNeedsUpgrade,
       },
-    }).as('getSystemNeedsUpdate');
+    }).as('getUpgradeNeedsUpgrade');
```
sneridagh, line 42: Renamed, but then not used? watch out

pnicolli: @sneridagh You have a point here. I changed it because the way I understand this file, there's a name clash. The original version of the test is:

```javascript
  it('As manager, I can click upgrade site, the upgrade control panel shows', function () {
    cy.intercept('GET', '/**/@system', {
      statusCode: 200,
      body: {
        ...getsystemNeedsUpgrade,
      },
    }).as('getSystemNeedsUpdate');
    cy.intercept('GET', '/**/@upgrade', {
      statusCode: 200,
      body: {
        ...getUpgradeNeedsUpgrade,
      },
    }).as('getSystemNeedsUpdate');
    cy.navigate('controlpanel');
    cy.wait('@getSystemNeedsUpdate');

    cy.findByText('Please continue with the upgrade.').click();
    cy.get('.content-area').contains(
      'The site configuration is outdated and needs to be upgraded.',
    );
  });
```

If I understand correctly there are two different calls intercepted with the alias `getSystemNeedsUpdate` and I doesn't feel right. On the other hand, if I change it where do I wait for the second one? Maybe after the click?

pnicolli: I tried adding an additional `wait` after the click, I am not sure it is needed at all, but the request is actually made to the backend so I think it is worth waiting for it to actually happen, at least in the current state of things

pnicolli:
@sneridagh @tiberiuichim @stevepiercy 
Marking this as ready for review. I added the docs we discussed a few months ago for core development, if you feel other docs are needed let me know. Should we mention this in the upgrade docs? It is transparent to the integrators imho, or at least it is aimed to be transparent and non-breaking.

I see a couple cypress tests are still not green, I'll keep looking into it but help is appreciated since I can't reproduce these locally :(

ichim-david:
@pnicolli try to wrap the test in _.times, sometimes the tests are flaky and you might get the error only after x try.
also try with electron and not a browser as I usually run cypress with Firefox and it's better than electron but since the tests run in electron you also need todo so in order to try to get the failures.
  
// run only this test 20 times
  Cypress._.times(20, (k) => {
   it.only('As editor, I can unlock a locked page', function () {
     console.log('test');
   });
});

docs/source/contributing/performance-guidelines.md
stevepiercy, line 1: I am not clear who the audience is for this document. Is the audience a Volto developer or a contributor to Volto?

Does it belong under Contributing as in this PR, or under a new section in [Frontend > Development > Lazy loading](https://6.docs.plone.org/volto/development/lazyload.html)?

"Performance guidelines" is too vague as a description of what this feature does. I would rename it to "Bundle optimization", "Build size optimization", or "Lazy load", whichever you think is best. The end effect is improved performance, but performance can also be improved at several other points in the entire stack, such as caching or smaller images.

Once I have a better understanding of your intent and the context, then I can do a good review of the docs.

stevepiercy (changes requested):
I need more information before I can adequately review the docs.

Side note, do the tests use valid syntax? It looks ugly, but maybe it is valid.

packages/volto/src/components/manage/Blocks/Block/__snapshots__/Settings.test.jsx.snap
```diff
@@ -3,15 +3,18 @@
...
+  \\"fieldsets\\": [
+    {
+      \\"title\\": \\"Default\\",
+      \\"id\\": \\"default\\",
+      \\"fields\\": []
+    }
+  ],
+  \\"properties\\": {},
+  \\"required\\": []
+}"
```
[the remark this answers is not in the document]
pnicolli: It's ugly, but valid.

The lazy-loaded components, when used nested inside another component, give us more challenges when rendering for unit tests. Since they are unit tests, we don't really need to test nested components, because they have their own tests, so I worked around the rendering issues by mocking the lazy-loaded components and making it so that we just render the props that are passed to them. In other words, we don't render nested forms, but just the fact that we are rendering a form with the given schema.

The ugly syntax that showed up is how a json schema is rendered with snapshot testing.

docs/source/contributing/performance-guidelines.md
pnicolli: The audience is the developer adding or changing stuff in core Volto code, that's why I thought it belonged in Contributing. The `Development > Lazy Loading` section is targeted at people developing their own project using Volto, not people contributing directly to Volto itself, or at least this is my understanding of the docs structure, but let me know if I'm wrong here.

I am fine with changing name to the new page. My feeling was that this could be the start of a section dedicated to performance optimizations, but I see your point. We could call it "Bundle size optimization" for now and discuss a title change if broader topics are added in the future, what do you think?

pnicolli:
@ichim-david @tiberiuichim @sneridagh I managed to replicate the issue by using Electron (thanks for the pointer!) and I found that it is because of this line here: https://github.com/plone/volto/blob/main/packages/volto/cypress/support/commands.js#L707

I am not sure why typing a blank space then deleting it is needed, any ideas? Because only in electron and of course only in this branch the blank space is not deleted so the test breaks because of the extra space. I have no clue why changing the imports to lazy-loaded ones in slate (or in volto) would cause this behavior. On the other hand, I tried running all the tests that use the `pasteClipboard` command in this branch and they are all green by removing this line... but I'm not comfortable with just removing it without any confirmation, I don't know why it's there :)

docs/source/contributing/performance-guidelines.md
stevepiercy: Thanks for the explanation. Let's leave this file where it is, but rename it to focus on what it does. I'll try to put in a review this week at the Axolote Sprint.

stevepiercy (approved):
FYI, in narrative text, we use one sentence per line. Also VSCode is overly aggressive with formatting .md files and should be turned off.

stevepiercy:
This is really good work. After this is merged, then https://github.com/plone/documentation/pull/1651 needs to be merged, so that an error about the missing term "lazy loading" will not throw an error. That PR also needs a quick review.

stevepiercy:
I restarted the failed jobs, which might be due to flaky tests.

stevepiercy:
It looks like a couple of Slate acceptance tests are broken.

pnicolli:
Yeah I am trying to fix these tests that didn't break a few days ago, and I can't replicate this locally even using Electron :( I tried restarting these multiple times but with no luck.
The test says `cy.type('Colorless ...')` and then checks that it is correctly there, but in the slate field you actually have the string `Clorless`... it happened locally once and never again, and even if it did I really think I would need help by slate-savvy people here. I keep getting errors with slate tests, but since they are really related to weird stuff I am wondering: am I really the only one getting these??

ksuess:
I can reproduce https://github.com/plone/volto/actions/runs/8726201362/job/23941446900?pr=5295 with Electron.
It types the wrong string 'Clorless'.
I cannot reproduce it with Chrome, which is used by the test workflow.
Nonetheless I suggest to try the fix that eliminates the failure locally.
I will do a PR with the fix of the test 'As editor I can add links' by using `cy.getSlateEditorAndType`. Also because the use of `cy.get('#toolbar').click();` before typing does not make sense to me.

ksuess:
@pnicolli You may want to see that https://github.com/plone/volto/pull/5966 fixes the failing test https://github.com/plone/volto/actions/runs/8726201362/job/23941446900?pr=5295

packages/volto/cypress/tests/core/controlpanels/upgrade.js
sneridagh: I see, maybe we need to revisit this at some point.

sneridagh (approved):
LGTM, I just tried with VLT and looks good too!

@tiberiuichim @davisagli @plone/volto-team please I'd like to merge this asap, it's that kind of PRs that enter in the merge hell due to the large file changeset.

ichim-david:
@pnicolli this is awesome work that should be merged as soon as possible and improved in future alpha releases in my opinion as @sneridagh says the more is left here the more work there is todo afterwards to merge the main branch here.

I do have one aspect that I would like to bring to your attention @plone/volto-team regarding the chunks names and I wonder if we shouldn't try to have a naming prefix, something like 'volto-core-' + 'chunk name'.
chunk-names
There is a big difference between the manual chunk name we give and what is assigned when there is chunk name comment as seen in the screenshot I took from the bundle analyzer.

I think the non named chunk naming is overkill and too verbose if I compare to the naming of other packages, yet I feel like having names such as Widgets, Contents is not very indicative that this is the volto core bundle and should have a prefix.

I leave this comment as a starting ramp for a potential brainstorming or opinion flood on the bundle naming as we all know that naming is one of the hardest problems in computer science :)

ichim-david (approved):

pnicolli:
@ichim-david fully agree that a more consistent naming should be adopted, my decision for this PR was to leave it for later since it would require more changes and I kind of wanted to draw the line at some point :D but I already planned more work on this topic

sneridagh:
Merged! @pnicolli @deodorhunter @mamico thanks a lot for this huge effort!! 💚

PR #5299, opened 2023-10-08
feat!: new implementation for the proxy middleware

mamico:
After changing the reverse proxy rules that pointed directly to the backend, to utilize Volto proxy middleware, for binary downloading (i.e., images and files), we noticed a significant drop in performance. 

I tried to analyze it and improve it by changing the middleware implementations. This is the checklist that I think needs to be verified/completed:

* [ ] have performance comparable to that of Plone.
* [ ] add x-forwarded headers in the request for the backend.
* [ ] properly proxying important request/response headers (cache, range, ...) to the backend and browser
* [ ] correctly handle authenticated requests
* [ ] correctly handle p.a.redirectors.
* [ ] handle errors correctly (not found, not authorized, ...)
* [ ] Add documentation and configurability for integrators

What was tested/measured.
 
I loaded a 92 MB file into Plone (size is not important, but large files make analysis easier) in a Plone vanilla 6.0.6 installation, waitress, without Zeo server.

Plone apache benchmark: http://localhost:8080/Plone/discord-0-0-29.deb/@@download/file/discord-0.0.29.deb,

Volto main (yarn build and node build/server.js), same Plone site
http://localhost:3000/discord-0-0-29.deb/@@download/file/discord-0.0.29.deb

Volto this branch (yarn build and node build/server.js), same Plone site
http://localhost:3000/discord-0-0-29.deb/@@download/file/discord-0.0.29.deb

image

Plone has the better performance here, this branch is not so far.

Even comparing the memory usage of the Volto process during testing, a much higher consumption is evident in the current implementation, compared to the proposed one.

image

IMO, the main problem with the current implementation is that all the file content is downloaded into memory before being transmitted to the user. This implementation, on the other hand, streams data from the backend to the client.

### DIFF
--- a/src/config/server.js
+++ b/src/config/server.js
@@ -36,6 +36,11 @@ const settings = {
       },
     },
   ],
+  // configs for extends/overrides the http-proxy-middleware options for files and images
+  // see the defaults in src/helpers/Proxy/Proxy.js
+  httpProxyOptions: {},
+  httpProxyOptionsImages: {},
+  httpProxyOptionsFiles: {},
 };
 
 export default settings;
--- a/src/express-middleware/files.js
+++ b/src/express-middleware/files.js
@@ -1,30 +1,13 @@
 import express from 'express';
-import { getAPIResourceWithAuth } from '@plone/volto/helpers';
+import { defaultHttpProxyOptions } from '@plone/volto/helpers';
+import { createProxyMiddleware } from 'http-proxy-middleware';
+import config from '@plone/volto/registry';
 
-const HEADERS = [
-  'accept-ranges',
-  'cache-control',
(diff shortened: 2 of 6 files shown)


davisagli:
@mamico Looks like a really good idea at first glance! I will be traveling home and taking a couple days off, but will try to review this soon.

mamico:
@davisagli have you had time to try it?

Regarding "properly handling errors (not found, not authorized, ...)," this does not seem easy with `http-proxy-middleware`, see https://github.com/chimurai/http-proxy-middleware/issues/179.

On the other hand `express-http-proxy` has an implementation that enables this functionality, see https://github.com/villadora/express-http-proxy/issues/248.

PR #5305, opened 2023-10-11
(FIX): put padding so the text is not clipped

dobri1408:
BEFORE: 
Captură de ecran din 2023-10-11 la 10 14 36

AFTER:
Captură de ecran din 2023-10-11 la 10 13 20

### DIFF
--- a/src/components/manage/Blocks/Search/components/SelectStyling.jsx
+++ b/src/components/manage/Blocks/Search/components/SelectStyling.jsx
@@ -15,7 +15,7 @@ export const customSelectStyles = {
     borderBottom: '1px solid #c7d5d8',
     boxShadow: 'none',
     borderBottomStyle: state.menuIsOpen ? 'dotted' : 'solid',
-    minHeight: '40px',
+    minHeight: '37px',
   }),
   menu: (styles, state) => ({
     ...styles,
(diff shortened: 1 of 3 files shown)


src/components/manage/Blocks/Search/components/SelectStyling.jsx
```diff
@@ -32,6 +32,7 @@ export const customSelectStyles = {
   valueContainer: (styles) => ({
     ...styles,
     padding: '0px',
+    paddingTop: '4px',
     paddingLeft: 0,
```
ichim-david, line 36: @dobri1408  this is redundant given that at line 34 you have padding 0

ichim-david: @dobri1408 this doesn't fix the fact that you can also have bottom text clipping when you use p or g
see the screenshot
4px-2-rows-demo

ichim-david (changes requested):

dobri1408:
I made sure that the padding is right now. I have put an 8 px padding on all sides.

dobri1408:
I think the best solution is to increase the height of the container as the rows of the content are growing. Because the problem is now shifted to a text of 3 lines, instead of two.  I will try to approach this solution.

ichim-david:
> I think the best solution is to increase the height of the container as the rows of the content are growing. Because the problem is now shifted to a text of 3 lines, instead of two. I will try to approach this solution.

I've seen that this fixes only 2 lines but you also have to think that it is a placeholder text, at some point, you have to keep it simple enough to be easily translatable and useful, I doubt that someone will use 3 lines of placeholder text.

sneridagh (approved):

PR #5328, opened 2023-10-20
Fix reference to installation

stevepiercy:


davisagli (approved):
Sure, I guess it's fine to link to the top of that page since it starts with the requirements/prereqs.

PR #5329, opened 2023-10-20
Visible focus semantic-ui reset

Wagner3UB:
Fix all the problems for issue #5214

- Default toolbar CMS-UI buttons outline focus working
- Edit toolbar cms-ui buttons outline re-done - semantic-ui problem
- Searchbox and clear button inside blocks-chooser with visible focus
- Clear button with the correct label

Schermata 2023-10-20 alle 15 19 17
Schermata 2023-10-20 alle 15 19 22
Schermata 2023-10-20 alle 15 19 34
Schermata 2023-10-20 alle 15 24 34
Schermata 2023-10-20 alle 15 24 45


ichim-david:
@Wagner3UB You should use the aria-label that you had in this commit:
https://github.com/plone/volto/pull/5329/commits/0afea9494640b7b893f79a63da37a67bf3a1573e#diff-5161834df5f3e3c2c0ccdf7e45d12b4489bf8aad97f2e804a551900c9e0b22f7R14

If someone with sight issues relies on the VoiceOver to speak about what that element is it makes more sense
to give a greater context about what that button clears.
Apple.com uses the same label **Clear Search** not just **Clear**
clear-search

We need to be more inclusive with Aria labels and not less, so I hope that any future efforts regarding labeling
will also focus on improving our current labels to have more meaningful text rather than "Add", "Edit" or "Clear"

PR #5334, opened 2023-10-24
Move some files around

tiberiuichim:


docs/source/guides/contextnavigation.md
stevepiercy, line 1: This file contains both guide and reference material. The list of parameters might belong in a separate reference file. What do you think?

docs/source/guides/express.md
stevepiercy, line 1: This file has some `config.settings` that might need to be moved into `configuration/settings-reference.md`.

stevepiercy: Also there are two different settings keys, and I am not sure whether that is correct:

- `config.settings.serverConfig.staticFiles`
- `config.settings.staticFiles`

docs/source/guides/ie11compat.md
stevepiercy, line 1: I think this should be deleted as we don't want to support an EOLed browser, as of 2022-06-15.

docs/source/guides/index.md
```diff
@@ -13,4 +13,20 @@ myst:
...
+creating-views
+images
+i18n
+express
+lazyload
+appextras
+contextnavigation
+pluggables
+widget
+ie11compat
```
stevepiercy, line 31: Buh-bye!

```suggestion
```

docs/source/reference/environment-variables.md
stevepiercy, line 1: This page actually is a how-to guide, not a reference.

Also let's insert missing MyST `html_meta` stuff.

```markdown
---
myst:
  html_meta:
    "description": "How to use environment variables in Volto"
    "property=og:description": "How to use environment variables in Volto"
    "property=og:title": "How to use environment variables in Volto"
    "keywords": "Volto, Plone, frontend, React, environment variables"
---

```

stevepiercy (changes requested):
This is a good start. I made a couple of comments and suggestions. For this first pass, I think moving the files with only a few changes, such as moving chunks of content into a possibly more appropriate place and fixing errors and warnings, is a good strategy. Thank you!

sneridagh:
@stevepiercy your comments on this one are more headed to do more in-depth changes (more than just moving some files around). Shouldn't we do this in the other long-lived branch/PR? I'm concerned that we don't step on each other in separated PRs.

stevepiercy:
@sneridagh this _is_ the long-lived branch. Although these changes would be good to backport to main, too.

docs/source/guides/ie11compat.md
stevepiercy: One more thing to consider, although we are hesitant to say what browsers we _do_ support, this would be an example of what we do _not_ support. This would be good to mention in requirements and support policies.

"Volto does not support technologies that are in end-of-life (EOL) or otherwise no longer supported by their maintainers."

stevepiercy: See https://github.com/plone/volto/issues/5380

docs/source/reference/environment-variables.md
stevepiercy: OK to leave it where it is, and later extract the guide stuff into a guide.

PR #5335, opened 2023-10-24
Visible focus semantic UI - reset

Wagner3UB:
Fix all the problems for issue https://github.com/plone/volto/issues/5214

- Searchbox and clear button inside blocks-chooser with visible focus
- Clear button with the correct label

Schermata 2023-10-20 alle 15 19 17
Schermata 2023-10-20 alle 15 19 22
Schermata 2023-10-20 alle 15 19 34

**I removed the toolbar modifications to open a PR specifically for this change


giuliaghisini (approved):

theme/themes/pastanaga/elements/input.overrides
```diff
@@ -95,8 +95,19 @@ of an error is present, it overrides a default from SemanticUI grid definitions.
     cursor: pointer;
     text-align: initial;
 
-    &:focus {
-      outline: none;
+    &:focus,
```
ichim-david, line 98: This change will apply the pink color also to other buttons such as the search icon from the header not just the x from the block chooser search.
focus-hover-search-icon

ichim-david (commented):
@Wagner3UB the toolbar buttons still have no focus, look at FF vs Edge.
Your issue was only concerning the search input however you've mentioned
that you've also tackled the toolbar buttons and this is not the case
buttons-focus

ichim-david (changes requested):
@Wagner3UB see my comments for what I think needs to be changed

theme/themes/pastanaga/elements/input.overrides
Wagner3UB: @ichim-david Ok, I made a modification to that CSS, now it uses only the add block search.

theme/themes/pastanaga/extras/blocks.less
```diff
@@ -783,6 +775,14 @@ body.has-toolbar.has-sidebar-collapsed .ui.wrapper > .ui.inner.block.full {
     // reseting the default Accordion behavior
     display: initial;
   }
+  .ui.form .ui.input input {
+    //reset for semantic-ui outline:none
+    border-width: @1px;
+    border-color: transparent;
+    &:focus {
+      border-color: @blueFocus;
```
ichim-david, line 783: @Wagner3UB @sneridagh for the focus of the input I would use
@focusedFormBorderColor instead of @blueFocus
https://github.com/plone/volto/blob/main/theme/themes/default/globals/site.variables#L96
this is because Volto uses this rule for the sidebar textarea and inputs
so it would make sense to keep it consistent.
focus-of-inputs

ichim-david:
@sneridagh @plone/volto-team 
I don't have a problem with the search input as it is with full border but I wonder if it wouldn't be more in the pastanaga
design to have only the bottom border  
search-inactive
search-active

We are using a few search input styles:
1. Search block
search-input
2. Search for sharing or control panels 
search-sharing
3. Search user groups 
search-user-groups
4. Contents select objects 
search-styles

Which style do you think would work best with our block chooser while retaining the pastanaga theme cohesion?

stevepiercy:
@ichim-david I think this issue is about `:focus`, not `border`.

As an aside, I strongly dislike `border-bottom` only for field inputs, as the default for browsers is to have borders all around, and it just looks weird to me.

Wagner3UB:
I share Steve's viewpoint, having only the border-bottom might not suffice. Another aspect that concerns me is that, in order to achieve sufficient contrast, the color must literally change, rather than just a shift in shade or contrast. The bar is quite thin. Making this adjustment, as I did it, the 'pink' color seems incorrect, potentially implying an error in the field.

ichim-david (approved):

ichim-david:
> @sneridagh @plone/volto-team I don't have a problem with the search input as it is with full border but I wonder if it wouldn't be more in the pastanaga design to have only the bottom border search-inactive search-active
> 
> We are using a few search input styles:
> 
>     1. Search block
>        search-input
> 
>     2. Search for sharing or control panels
>        search-sharing
> 
>     3. Search user groups
>        search-user-groups
> 
>     4. Contents select objects
>        search-styles
> 
> 
> Which style do you think would work best with our block chooser while retaining the pastanaga theme cohesion?

Personally, I don't like the UX design of this search input and how it looks when we have the focus due to missing padding but my issue with the design doesn't invalidate the current work as such I have accepted these changes.

Once this is accepted I will add a pull request myself where I propose some changes to the search
and I will gather feedback on my proposed changes there since we are in Volto 18 alpha where we can propose
changes that will make the interface look different.

ichim-david:
@plone/volto-team I am merging this pull request since it has been through a few rounds of testing, we are in the alpha phase so we will have many more releases before the final release in case we need to tweak these changes even further.

PR #5339, opened 2023-10-25
[Search Block] do not send sorting information by default

### LINKED ISSUE #5338: [Search Block] if the configuration does not set a sort_on parameter, search block requests to sort on effective.
**Describe the bug**

**To Reproduce**

1. Create a new page
2. Add a search block and add a preconfigured search (type = Document for instance). Do not set any sorting configuration
3. Do a full text search

**Expected behavior**
I would expect that the search query sent to the backend includes only the preconfigured search (portal_type=Document) and the SearchableText with the entered search

** What happens **

The search query executed is the following

POST https://demo.plone.org/++api++/page-with-search-block/@querystring-search

Body:
```json
{
"metadata_fields":"_all",
"b_size":25,
"query":[
  {"i":"portal_type","o":"plone.app.querystring.operation.selection.any","v":["Document"]},
  {"i":"SearchableText","o":"plone.app.querystring.operation.string.contains","v":"ima"}
],
"sort_on":"effective",
"sort_order":"ascending",
"b_start":0
}
```
Neither sort_on or sort_order are configured in the search block, so it shouldn't send anything to the backend.

erral:
Fixes #5338


ionlizarazu (approved):

tiberiuichim (approved):

sneridagh:
I have a deja-vu on this one... didn't we have to setup the defaults for the p.a.q. to behave like it's intended? (eg. the default controls). Didn't Plone classic also sets it in the collections by default?

erral:
This is a search block which is not feature-paired with a Collection.

When having a full text search you don't want to sort by effective date by default, it goes against the meaning of 'fulltext' which is sorted by relevance.

If you want to set a default sorting criteria you can do it: set it in the configuration of the block, if I don't set anything don't think replacing me.

This happens, and I know it is not a restapi or Volto problem, because we don't have a key in ZCatalog to 'soft by relevance'.

davisagli (commented):
It makes sense to me to remove the default sort_on by effective IF the block has no configured sort AND the user has entered a SearchableText query.

If the block has no configured sort but the user hasn't entered a SearchableText query, it probably makes more sense to reverse-sort by effective than to do no sort at all (which gives results in an indetermistic order based on internals of the ZCatalog implementation).

tl;dr: This is heading in the right direction, but I would add a condition based on whether the user has entered a text query.

sneridagh:
@erral I just dig into it and my concerns are gone now... somehow I remembered that defaults were in there for a reason. 
Could you please go ahead with David's suggestions? Thanks!

sneridagh:
Oh, and this is a change of behavior that could potentially affect existing projects so it must be a breaking change...

erral:
Well, I have no problem on adding the behavior that @davisagli suggests, but I think that's wrong. We already have a setting in the search block to let the user select the sorting of the results, if the user hasn't selected anything, I think that we are not entitled to decide which is the *correct* sorting. If so we would have implemented it in ZCatalog right?

And breaking... well it breaks the current way of working, which is wrong :)

tiberiuichim:
> It makes sense to me to remove the default sort_on by effective IF the block has no configured sort AND the user has entered a SearchableText query.

Yes, this should be the default behavior. Relevance based search when there's searchable text, all the time.

ichim-david:
@davisagli @sneridagh this looks like something that is ready to merge and could make it in 18, are there any further blockers from merging this given that 18 is, either way, a breaking change release?

davisagli (approved):
LGTM

Issue #5219, opened 2023-09-22
[Search Block]: Live search triggers several requests to `@querystringsearch`


sneridagh:
**Describe the bug**
The search block triggers several requests on actions

eg. using the searchbox:
It now triggers three times, one of them being "wrong" (not matching the current query elements), so there's one moment that the result is incorrect. This could affect acceptance tests too.

**To Reproduce**

1. Add a search block with a custom query
2. Search for a term in the searchbox
3. Monitor the network tab, it triggers several requests to `@querystringsearch`

**Expected behavior**
We should only trigger one request per action/state change.

Also, when typing a debounce should be added too.

satyam4p:
Picking this one up

satyam4p:
Working on this one

JASIM0021:
can you describe it with more details including file or page url if posible

ionlizarazu:
I found to a way to test the amount of the querystring-search queries with cypress. We can work on it in this branch https://github.com/plone/volto/tree/ionlizarazu-search-block-querystring-search @sneridagh

PR #5340, opened 2023-10-25
[Search Block] Improvements to make just 1 query to the backend

erral:
- cypress tests to check that search block issues just 1 query to the backend
- change composing order to query the bakckend once


ionlizarazu:
fixes #5219

src/components/manage/Blocks/Search/SearchBlockView.jsx
```diff
@@ -106,4 +106,6 @@ export const SearchBlockViewComponent = compose(
   (Component) => React.memo(Component, blockPropsAreChanged),
 )(SearchBlockView);
 
-export default compose(withQueryString, withSearch())(SearchBlockViewComponent);
+export default compose(withSearch())(
+  compose(withQueryString)(SearchBlockViewComponent),
```
tiberiuichim, line 110: I don't understand this diff. In effect, the execution order is the same, right?

src/components/manage/Blocks/Search/hocs/withSearch.jsx
```diff
@@ -333,7 +333,8 @@ const withSearch = (options) => (WrappedComponent) => {
       setSearchData(
         getInitialState(data, facets, urlSearchText, id, sortOn, sortOrder),
       );
-    }, [deepFacets, facets, data, urlSearchText, id, sortOn, sortOrder]);
+      // eslint-disable-next-line react-hooks/exhaustive-deps
```
tiberiuichim, line 336: maybe use JSON.parse(deepFacets) so you can avoid the eslint-disable line?

src/components/manage/Blocks/Search/SearchBlockView.jsx
davisagli: @tiberiuichim `compose` operates from right to left. So, before the change, withSearch() was applied first, but now it is applied second.

There's no need to use compose here; it could be written as withSearch()(withQueryString(SearchBlockViewComponent))

@ionlizarazu It would be helpful if you could explain a bit about how changing the order solves the problem.

ionlizarazu: @tiberiuichim I found that this wasn't really the problem, but as @davisagli said, the order is different and it makes sense to me to execute first the withQueryString and after the withSearch.

I'll remove the composes to make it more readable.

src/components/manage/Blocks/Search/hocs/withSearch.jsx
ionlizarazu: it makes sense, I also have to add a deepData to remove the eslint-disable line.

src/components/manage/Blocks/Search/SearchBlockView.jsx
```diff
@@ -106,4 +106,4 @@ export const SearchBlockViewComponent = compose(
   (Component) => React.memo(Component, blockPropsAreChanged),
 )(SearchBlockView);
```
sneridagh, line 108: Why did you remove the compose?

sneridagh (commented):
Just one comment, the rest LGTM if @tiberiuichim approves.

sneridagh:
Merging, since it seems it fixes the flaky test problem we have in main.

src/components/manage/Blocks/Search/SearchBlockView.jsx
ionlizarazu: https://github.com/plone/volto/pull/5340#discussion_r1375475109

PR #5347, opened 2023-10-30
Fix the right order of parameters in normalizeExternalData.js

dobri1408:
Because of this bug, if you try to copy the text:
This indicator is a headline indicator for monitoring progress towards
In a richtext widget, an error will come.

### DIFF
--- a/packages/volto-slate/src/editor/extensions/normalizeExternalData.js
+++ b/packages/volto-slate/src/editor/extensions/normalizeExternalData.js
@@ -2,7 +2,7 @@ import { normalizeExternalData as normalize } from '@plone/volto-slate/utils';
 
 export function normalizeExternalData(editor) {
   editor.normalizeExternalData = (fragment) => {
-    return normalize(fragment);
+    return normalize(editor, fragment);
   };
   return editor;
 }


avoinea (approved):

avoinea:
@sneridagh We will need this also in 16.x :see_no_evil:

tiberiuichim (approved):
@sneridagh let's get this rolling, seems important.

davisagli (commented):
@dobri1408 At the moment I don't understand the situation where the bug occurs, so I am not sure how to test the fix. I tried pasting "This indicator is a headline indicator for monitoring progress towards" into a slate block on demo.plone.org but got no error. Can you explain the steps to reproduce the issue better?

dobri1408:
Hi @davisagli. First of all, you can see [here](https://github.com/plone/volto/blob/5d7b7c1aef101a8dab340f1ced43969ac1546bf1/packages/volto-slate/src/utils/blocks.js#L70) that the normalizeExternalData function expects the following arguments: editor, nodes (or fragments). The error doesn't come in the slate block, but in the richtext widget. From what I know, this widget is not used in demo.plone.org, but was used in one of the EEA projects.

davisagli:
@dobri1408 Okay, I tested it by creating a Dexterity content type with a rich text field, then trying to paste this text there. This PR seems to be successful at avoiding the error in the JS console. However, the text is still not pasted.

davisagli:
@dobri1408 @tiberiuichim I did some more testing and made some changes in https://github.com/plone/volto/pull/5347/commits/4bbebea467bffcff96c7c04b29bb5da4a8e2d8d2 that seem to make it work for me. Caveat: I don't really know what I'm doing.

sneridagh:
@dobri1408 we need now a towncrier entry for each package we "touch". Could you please add one there too? (it's ok to be the same.

sneridagh:
@dobri1408 Ah, I'll fix it it's the dangling old one.

Issue #5351, opened 2023-10-31
Fix condition for default view translation, for missing default vies case


reebalazs:
In a site, clicking on the three dots in the portal root in edit mode failed with a translation error message.

Code review of `Display.jsx` reveals that this is caused because the asyncconnect code returns empty string "" in case the view is not set. But this causes a translation error because then it tries to translate the empty string, which fails.

It's unclear what causes the layout to be missing. However from the asyncconnect code it's obvious that this is a case to be prepared for, therefore we must be tolerant in the case the layout to be unset, instead of letting
the translations engine fail - which caused this error.

PR #5352, opened 2023-10-31
WIP Fix condition for view translation, for missing view case #5351

reebalazs:
Fixes clicking on three dots, which causes an error in case of layout is missing,

It's unclear what causes the layout to be missing. However from the asyncconnect code it's obvious that this is a case to be prepared for, therefore we must be tolerant in the case the layout to be unset, instead of letting the translations engine fail - which caused this error.


reebalazs:
Remark: I've checked manually that if there is <=1 views, the selector does not show, iow the current behaviour is preserved together with the fix, too.

sneridagh:
@reebalazs unit tests fail

PR #5356, opened 2023-11-02
In slate, render span nodes as spans rather than paragraphs.

davisagli:
@tiberiuichim Curious for your thoughts on this. Currently https://github.com/plone/blocks-conversion-tool produces span nodes for text inside a table cell, but perhaps it should not do that.

### DIFF
--- a/packages/volto-slate/src/editor/config.jsx
+++ b/packages/volto-slate/src/editor/config.jsx
@@ -248,6 +248,7 @@ export const elements = {
 
   div: ({ attributes, children }) => <div {...attributes}>{children}</div>,
   p: ({ attributes, children, element }) => <p {...attributes}>{children}</p>,
+  span: ({ children }) => <span>{children}</span>,
 
   // While usual slate editor consider these to be Leafs, we treat them as
   // inline elements because they can sometimes contain elements (ex:


davisagli:
I'm going to try to fix this in the blocks-conversion tool instead, I don't think we want extra spans in the markup.

davisagli:
Hmm, adding the spans was done as a workaround for this issue: https://github.com/plone/blocks-conversion-tool/issues/31

tiberiuichim:
@davisagli In the block conversion tool, the cell value should be: `<div>text</div>`. I think that might work.

https://github.com/plone/volto/blob/c8c0b5c8c4a6430222989be282d9e8e2772439a3/packages/volto-slate/src/editor/config.jsx#L249

davisagli:
@tiberiuichim The cell I'm converting is like this: `<td>text<strong>strong text</strong></td>` so there's a text node and an inline node, but no paragraph or div. Is there a way to represent this in slate?

tiberiuichim:
@davisagli I'm not too familiar with the table plugin for volto-slate (unfortunately), but the editor inside a single cell should "see" this markup: `<div>text<strong>strong text</strong></div>`

I guess, in the converter, always wrap the cell content in a `div` element.

tiberiuichim:
I think this should also be fixed and use a div: https://github.com/plone/volto/blob/c8c0b5c8c4a6430222989be282d9e8e2772439a3/packages/volto-slate/src/blocks/Table/TableBlockEdit.jsx#L46

davisagli:
@tiberiuichim Thanks, making the converter wrap the cell contents in a div seems to work decently. https://github.com/plone/blocks-conversion-tool/pull/38

PR #5375, opened 2023-11-07
Check for existence of news symlink

stevepiercy:
When running `make docs-html`, the build will fail because a symlink exists. This pull request checks for the existence of a good symlink and creates one if it does not exist. Whether it exists or not, an appropriate statement is echoed.


davisagli (approved):
I think you could have also just added the `-f` flag to `ln` to always recreate the link even if it already exists. But this looks fine too.

stevepiercy:
TIL. I actually like that option better. Oh, well, next time.

PR #5384, opened 2023-11-07
Update internal Plone version to 6.0.8

sneridagh:


davisagli (commented):
@sneridagh There are still references to 6.0.6 and 6.0.7 in the generator. Please always search for all occurrences of the old version.

PR #5392, opened 2023-11-10
First focus edit page

Wagner3UB:
Fix part of this issue https://github.com/plone/volto/issues/5268

- Make the first focus more informative for assistive technologies.

quote:
When you click the 'edit' button, the default behavior is to focus on the first editable block, which is usually the title block. However, when using assistive technology, this behavior can be inconsistent, leaving users unsure of their location and what action to take. To improve this, the fieldset containing all blocks should be made accessible to screen readers and provide clear information about the editable content before moving to the first element.

### DIFF
--- a/locales/ca/LC_MESSAGES/volto.po
+++ b/locales/ca/LC_MESSAGES/volto.po
@@ -776,6 +776,11 @@ msgstr ""
 msgid "Content that links to or references {title}"
 msgstr ""
 
+#: components/manage/Blocks/Title/Edit
+# defaultMessage: Content title
+msgid "Content title"
+msgstr ""
+
 #: components/manage/Controlpanels/ContentTypes
 # defaultMessage: Content type created
 msgid "Content type created"
--- a/locales/de/LC_MESSAGES/volto.po
+++ b/locales/de/LC_MESSAGES/volto.po
@@ -773,6 +773,11 @@ msgstr "Inhaltsregeln von übergeordneten Seiten"
 msgid "Content that links to or references {title}"
 msgstr "Zu {title} verlinkende oder referenzierende Inhalte"
 
+#: components/manage/Blocks/Title/Edit
+# defaultMessage: Content title
+msgid "Content title"
+msgstr ""
+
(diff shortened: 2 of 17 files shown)


Wagner3UB:
@ichim-david Any suggestions to improve ever more this first focus? I thought about using "aria-description" to put more information, but I'm not sure if it's more correct than using only "aria-label".

ichim-david:
> @ichim-david Any suggestions to improve ever more this first focus? I thought about using "aria-description" to put more information, but I'm not sure if it's more correct than using only "aria-label".

@Wagner3UB I've tried aria-describedby and its support is inconsistent from browser to browser
https://a11ysupport.io/tech/aria/aria-describedby_attribute

https://github.com/plone/volto/assets/152852/f0c7db9e-aee9-4470-8473-259a63524bec


https://github.com/plone/volto/assets/152852/7de276d2-036a-4362-a372-6af3f89777b9


https://github.com/plone/volto/assets/152852/d64c6216-fd14-4cf8-87fa-f649597725a5

What I noticed that worked was when enabled on an element with role button which we don't
use on our input.

Gutenberg editor has all of the blocks set with role of "document", see 
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/document_role
https://wordpress.org/gutenberg/

It's kind of nice because you also get a warning when you go outside of the block area.
I find it a bit strange though this notion of "document" although looking at the documentation
they mention that it is used for widgets and you could say that a block is an application widget.

If you play with Gutenberg you get nice feedback suck as entering Block: Heading then when you go inside
you get the text found inside it, then it says exiting Block: Heading region.

Unless we also make this kind of aria support to all of our blocks which would require a PLIP I don't know
how useful this label "Content title" will be in the end.

Personally, I don't think people would know what content title means when it's read like "Welcome to plone 6, Content Title, edit text".

It would make more sense to add some description for the field that is read after reading the title, but the right technique would need to be applied in order for it to be read by the screen reader.

Wagner3UB:
PR closed. The delta between the versions is too large to justify keeping this branch open. I will re-apply the changes in a new PR.

PR #5397, opened 2023-11-13
Transfer of main Plone types definitions to its own package

sneridagh:
It's time that we share them through projects, we have some constraints due to the nature of the scenarios we want them to work (isolation-Volto vanilla).


tiberiuichim (approved):

mtoepfl (approved):
LGTM

PR #5408, opened 2023-11-16
Add postcss-less to the default addon

sneridagh:


wesleybl:
@sneridagh I had the error:
```
Error: Cannot find module 'postcss-less'
```

when running `yarn stylelint` in an app. But I don't use the default addon. So I'm wondering if we can have it in the app generator template as well.

PR #5409, opened 2023-11-16
All in with monorepo

sneridagh:
Note for Volto core developers (/cc @plone/volto-team) 

Once this is merged, please take a look at: https://github.com/plone/volto/blob/a5401295af0a4fb799401102788ab446f4ad6276/docs/source/contributing/developing-core.md

It is recommended that you remove your `node_modules` from your local Volto cloned repository, then run `pnpm i`. You cannot run the repo with other thing than `pnpm` anymore.

If you have any PR opened these days, you'll have to update it using the command line. `git` is able to detect the moves, but you probably will have to confirm them. Also, the Volto `news` items now live in `packages/volto/news` but `git` is also detecting them.

We will be improving and polishing in the next days in case that there's something still not adjusted.


.vscode/settings.json
```diff
@@ -0,0 +1,23 @@
+{
```
tiberiuichim, line 1: Is this file intended to be here? I see some customizations, which may not be to everyone's liking.

sneridagh: The colors? :) It is required for the eslint config to get it right. Don't know if that config can be "global". I will remove the colors.

sneridagh: FYI, The colors are only "the shell" of VScode customizations though, using the Peacock official extension.

stevepiercy:
@sneridagh we need to be careful about the terms we use and not invent new ones. There are two frontends for Plone, even though Volto will dominate the world. See 6c5c92390f55e9fc378d9a862e65890c646e1f98

sneridagh:
> @sneridagh please take a look at [6f79fbf](https://github.com/plone/volto/commit/6f79fbf47f29323f2dd13ed230b34c58823d007a).

LGTM

sneridagh:
@plone/volto-team I think this is ready. 

All is green, except the Towncrier checks will fail because it does not have "reference" with main. Once it's merged, it will be fine. I'd like to merge as soon as possible because it's already a mess to look at the changes, since it detects more than 3000 and Github file diff view just dies. We can refine in subsequents PRs.

I've tested to merge PRs and all file movements are detected well. So existing PRs will behave correctly (the merge has to be done via command line, locally). We'll have to see how difficult the backports to 16 are. We could also move 16 build to a monorepo approach, shouldn't be that hard, and pretty straightforward following the changes in main.

tiberiuichim (approved):

davisagli (commented):
@sneridagh When I switched to this branch to take a look, I got an error from pnpm:
```
% git checkout gotomonorepo                                                                                                                                                                                                          branch 'gotomonorepo' set up to track 'origin/gotomonorepo' by rebasing.
Switched to a new branch 'gotomonorepo'
Changes to lockfile found, running `pnpm install`
 ERROR  Unknown option: 'prefer-frozen-shrinkwrap'
Did you mean 'prefer-frozen-lockfile'? Use "--config.unknown=value" to force an unknown option.
For help, run: pnpm help install
Running pnpm pnpm install,--prefer-offline,--prefer-frozen-shrinkwrap,--no-optional failed
```

.github/workflows/acceptance.yml
```diff
@@ -415,9 +682,83 @@ jobs:
           name: cypress-videos
           path: cypress/videos
 
-  guillotina:
+  # guillotina:
```
davisagli, line 685: @sneridagh Is it intended to disable the guillotina tests?

README.md
```diff
@@ -80,9 +81,9 @@ On Plone 6, we recommend using the known good set (KGS) of package versions that
 
 On Plone 5, Volto is currently tested with the following packages pinned to specific versions, and we recommend using the same versions, which are:
```
davisagli, line 82: @sneridagh The versions we give here need to match the versions in KGS5 in the Makefile, if we want to make the claim that that's what is tested. Currently they do not match.

SECURITY.md
davisagli, line 1: This file is recognized by github and linked as the "security policy" on the right side of the repository's landing page. (https://docs.github.com/en/code-security/getting-started/adding-a-security-policy-to-your-repository) We should update but not remove it.

TODO-Monorepo.md
davisagli, line 1: This list looks somewhat out of date?

apps/plone/package.json
davisagli, line 1: What is the purpose of adding apps/plone to the repository?

packages/generator-volto/generators/app/index.js
```diff
@@ -123,7 +121,7 @@ Run "npm install -g @plone/generator-volto" to update.`,
 
     if (!this.voltoYarnLock) {
       this.log(chalk.red("Retrieving Volto's yarn.lock"));
-      this.voltoYarnLock = await utils.getVoltoYarnLock(voltoVersion);
+      // this.voltoYarnLock = await utils.getVoltoYarnLock(voltoVersion);
```
davisagli, line 124: Does the generator still need some work? We shouldn't be logging that it is retrieving yarn.lock if that's not what it is doing.

sneridagh:
> @sneridagh When I switched to this branch to take a look, I got an error from pnpm:

Updated `yarnhook` to latest allows full support to `pnpm`: https://github.com/plone/volto/pull/5444

.github/workflows/acceptance.yml
sneridagh: Undecided on this one, thinking in remove them, since there was no activity on that side for years. It will be easy to re-enable them, maybe postpone the decision for later.

SECURITY.md
sneridagh: Brought back from `packages/volto`.

apps/plone/package.json
sneridagh: It's part of the reorganization, as you already read in the docs, the idea of the `apps` folder is having all the support apps in there as in `docs`, `api`. 

The other idea is a playground with demostration of frameworks using the libraries, ideally having them tested too in CI.

If so, having a Volto project in there, and being able to test it in full seems "natural" to have it too. Also to test new things in there, like in the near future `pnpm` for projects too, or play with the idea of having a more "light" project setup for Volto as a library, without relying on Razzle, something I discussed so many times with @tiberiuichim 
This will allow also a better CI story on Volto projects too.

packages/generator-volto/generators/app/index.js
sneridagh: True, I forgot about this one. It should be removed, we don't need that anymore, although more tests to ensure that all is in place (eg. linting and prettifying) and we don't have hoisting problems should be in place.

.github/workflows/acceptance.yml
davisagli: @sneridagh I don't feel strongly about it, but I would put it in a separate PR so that it shows up a bit more clearly in the release notes.

apps/plone/package.json
davisagli: I mentioned this in a private conversation -- I think we'll need to follow up soon with some checks to make sure `apps/plone` doesn't fall out of sync with the project generator.

packages/generator-volto/generators/app/index.js
davisagli: I agree, the important thing is that it works and that we have tests to make sure it keeps working.

.vscode/settings.json
sneridagh: Removed! no longer needed.

Issue #5407, opened 2023-11-15
Edit breaks if broken short name is used


UnaiEtxaburu:
**Describe the bug**
When adding/editing a page if a broken short name (short name with white spaces) is used, after saving the page its impossible to edit it again. 

**To Reproduce**

1. Create a new page
2. Fill the title and add a short name with white spaces  
3. Save the changes
4. Try to edit the page

**Expected behavior**
Saving the page should be impossible if broken short name is used.

**Screenshots**
![broken-short-name test](https://github.com/plone/volto/assets/80053871/3a6d7183-9f31-4525-927d-79c3e3474796)

**Software (please complete the following information):**

- OS: [ubuntu 20.04]
- Browser [chrome, firefox]
- Volto Version [16.24.0, 17.0.0]
- Plone Version [6.0.6, 6.0.7]
- Plone REST API Version [8.40, 8.43]

**Additional context**

UnaiEtxaburu:
Related to #5179

IshaanDasgupta:
The problem only seems to occur when we are creating the page because when I create test then tried to set a broken short name it replaced spaces with "-" but when creating a new page this doesn't happen it is instead replaced with "%20" (same behaviour if using special characters expect when creating they are not replaced with anything).

Also if we don't put a extra space at the end, we can still make edits. The space at the end causes the edit just just not work

Just putting a validator for the short name before saving would work but I am new to volto so, I am not sure if it's better to put those checks on the backend. If putting a validator on the frontend is an option just let me know, I would like to work on this.

[Screencast from 17-11-23 06:00:44 AM IST.webm](https://github.com/plone/volto/assets/62870756/0bea7568-9669-43f6-9abe-035203a27863)

pranayjoshi:
I am working on it in the pull #5413

IshaanDasgupta:
Hi @pranayjoshi, just went through your PR for this issue.

pranayjoshi:
Hey @IshaanDasgupta ! I updated the repo with change logs and tests. I appreciate your suggestion, but I am not sure if it actually requires an error toast. However, if the maintainers suggests to implement it, I will be happy to integrate that as well. And yeah thanks for reviewing the code changes.🤝

IshaanDasgupta:
@pranayjoshi that's great, if a feature request is raised(for an error toast) in the future we will look at it then.

PR #5414, opened 2023-11-20
Adding validation for short name

IshaanDasgupta:
Solved #5407 
If a user tries to create/save a form containing invalid short name, the user will get an error toast and the form will not be submitted until they use a valid short name.


IshaanDasgupta:
@stevepiercy sorry didn't see that a pull request was already created for the same issue. Although that pull request failed should I still drop the issue or keep working on it ?

stevepiercy:
@IshaanDasgupta you should read and follow [Contributing to Plone](https://6.docs.plone.org/contributing/index.html) and [First-time contributors](https://6.docs.plone.org/contributing/first-time.html), as well as [Contributing to Volto](https://6.docs.plone.org/volto/contributing/index.html), especially the Code Quality section. It's all about saving time and frustration for you and other contributors.

Ultimately it is up to you to coordinate with other contributors to avoid duplicating effort. However, if you think that the other developer's proposed fix could be improved (hint: it can because it fails tests and lacks a change log entry), then it is fine to either collaborate with the other developers or make your own contribution.

IshaanDasgupta:
Thanks a lot, I'll coordinate with the other person also thanks for your time.

Issue #5416, opened 2023-11-20
YouTube video embed results to error if video url contains /live/


rioksane:
YouTube video embed results to error if video url contains /live/

```
Sorry, something went wrong with your request
t.url.match(...) is null
```

Tested at https://demo.plone.org

**To Reproduce**

1. Go to https://www.youtube.com/watch?v=ISdHvS6Ck3k&list=PLGN9BI-OAQkSXMXVBXLWQAQr0AF2xM_NU&index=19
2. Click on Share
3. Copy link https://www.youtube.com/live/ISdHvS6Ck3k?si=COeVakmC1lI6jQy3
4. Go to Plone 6 Volto site add a new page and add video embed block. 
5. Paste url and save page.

**Expected behavior**
Video embed should work

**Screenshots**
<img width="1160" alt="Screenshot 2023-11-20 at 19 15 00" src="https://github.com/plone/volto/assets/5127672/0a0e1086-2550-42a5-9d2a-3c5be2da3097">
<img width="995" alt="Screenshot 2023-11-20 at 19 17 50" src="https://github.com/plone/volto/assets/5127672/de2da89e-f4da-437d-9d9e-61bd8c67585b">

**Software (please complete the following information):**

- OSX
- Firefox
- Volto Version 
    plone6-demo 1.0.0
    Volto 17.5.0
    Plone 6.0.8
    plone.restapi 9.1.2
    CMF 3.2
    Zope 5.8.6
    Python 3.11.6 (main, Nov 1 2023, 14:02:22) [GCC 10.2.1 20210110]
    PIL 9.5.0 (Pillow)

rioksane:
@ericof this is the video embed issue mentioned

pranayjoshi:
Where is the Video embed option. I can only see the option to add page links etc.

IshaanDasgupta:
I am working on this

kuv2707:
Is the issue resolved?

stevepiercy:
@IshaanDasgupta no worries. We're all learning how to improve and use GitHub features here.

PR #5426, opened 2023-11-23
Adding extraction of videoID from youtube videos url which contain '/live/'

IshaanDasgupta:
Solved #5416 
Added an extra conditional statement for extracting 'videoID' from youtube video urls containing '/live/'

### DIFF
--- a/news/5416.bugfix
+++ b/news/5416.bugfix
@@ -1,1 +1,1 @@
-Adding a conditional statement to extract the 'videoID' from a youtube video which contains '/live/' in its url which previously used to throw an error. @IshaanDasgupta
+Refactoring the code for extraction of videoDetails from the video URL, adding code for extracting videoDetails from youtube video URLs with '/live/' in its URL which previously used to throw an error and adding jest tests for same. @IshaanDasgupta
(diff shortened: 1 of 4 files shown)


ericof (changes requested):
Could you, please, add a test for it? I believe we should have this in a function that can be properly tested

IshaanDasgupta:
Sure, I am not familiar with adding tests but I'll look into this.

IshaanDasgupta:
@ericof I have added the tests could you please go through them once as this my first time writing jest tests so if there is something wrong do let me know.

Also I have moved the videoDetails extraction code to a separate function and there was also a issue to video's thumbnails not begin displayed so I also changed the logic to bit to get the correct thumbnail URLs.

IshaanDasgupta:
The first screenshot is from https://demo.plone.org/ where the videos in which the url contains "list" or ".be" dont have their thumbnails displayed, the second one is from my local server. There was a problem with the extraction of the videoID from the urls with "list" and ".be".

Screenshot from 2023-11-25 21-14-00
Screenshot from 2023-11-25 21-17-03

ericof (approved):
Thanks for taking care of this!

PR #5429, opened 2023-11-23
Fix image paths in development mode

robgietema:
This fix changes how images are fetched from the backend. By default in development mode the images are fetched using the backend url throught the middleware, then ++api++ is added. This url suffix should only be used in either the frontend or in production. When using a different backend which doesn't use the ++api++ suffix this will break otherwise.


sneridagh (commented):
LGTM

PR #5434, opened 2023-11-24
Rationalize readmes

sneridagh:


stevepiercy (approved):
This is better than the idea to merge the readmes.

Eventually we should remove the marketing stuff from the main readme and focus on the essential information that is needed to develop in Plone.

PR #5445, opened 2023-11-27
Initialize data in form before the checks for the `blocks` and `blocks_layout` are done

sneridagh:
This fix an edge case when the data from the server content is empty, then the fields are populated, but the initialized data is snapshot after the check (and amendments) are done.


davisagli (commented):
No idea how to reproduce the bug for testing, but it looks reasonable.

tiberiuichim (approved):
I see it, as we're baking in default empty form in that initial data, so it's no longer the same as the data that comes from the server. But this has me thinking: it's a smell of something rotten somewhere else. How about we don't initialize an empty form if the server is still pending data?

sneridagh:
The edge case: A content type created programatically, having `blocks` and `blocks_layout` empty. Then, this code was in place to catch it, but if the snapshot happens after the init, then the comparison that rips out all non-changed data using the snapshot is wrong, then it does not includes the `blocks_layout`, so it keeps being empty. This was not happening, if you add further blocks than the default in the first edit.

Agreed that this should happen in the server probably if the data exists and it's wrong.

PR #5463, opened 2023-11-30
Fix paths present in `compilerOptions.paths` registered as Volto add-ons

sneridagh:
(when they should not!)


davisagli (commented):
@sneridagh makes sense, needs changelog

sneridagh:
@tiberiuichim @davisagli this one is ready too.

packages/registry/src/addon-registry.js
```diff
@@ -155,56 +155,75 @@ class AddonConfigurationRegistry {
...
+    const [, pathsConfig] = this.getTSConfigPaths();
+    if (pathsConfig) {
       // We only initialize the development addons present in `compilerOptions.paths`
       // configured by `mrs-developer` filtered by the add-ons registered (via config)
       // since we can have other paths there that are not addons
       const developmentAddons = this.addonNames.filter((key) =>
         pathsConfig.hasOwnProperty(key),
       );
-
+      // Extract this one and call it from both published nad here
```
tiberiuichim, line 194: comment not needed.

packages/registry/src/addon-registry.js
```diff
@@ -218,6 +237,13 @@ class AddonConfigurationRegistry {
   }
 
   initPublishedPackage(name) {
+    // I am in the paths list, if so, register it as a development package
+    // instead than a published one
+    const [, pathsConfig] = this.getTSConfigPaths();
+    if (pathsConfig.hasOwnProperty(name)) {
+      this.initDevelopmentPackage(name);
```
tiberiuichim, line 244: I think you can `return` here? The code under this line will not execute because of the `if`, anyway

packages/registry/src/registry.test.jsx
```diff
@@ -1,4 +1,5 @@
 import config from './index';
+import { describe, expect, it } from 'vitest';
```
tiberiuichim, line 2: vitest? no more jest?

PR #5470, opened 2023-12-01
Delete wrongly placed news items, add a check for them, issue a nice message

sneridagh:


.github/workflows/changelog.yml
```diff
@@ -89,3 +91,9 @@ jobs:
           towncrier check --compare-with origin/main --dir packages/volto-slate
         env:
           BASE_BRANCH: ${{ github.base_ref }}
+
+      - name: Wrong news changelog check
```
stevepiercy, line 95:
```suggestion
      - name: Wrong location of news changelog check
```

.github/workflows/changelog.yml
```diff
@@ -89,3 +91,9 @@ jobs:
...
+        if: steps.filter.outputs.wrongNews == 'true'
+        run: echo "Wrongly placed news item in root. Please place it in the appropiate folder." && exit 1
```
stevepiercy, line 97:
```suggestion
        run: echo "News items should be moved from the repository root to the appropriate package root in `packages/package-name`." && exit 1
```

stevepiercy (changes requested):
Make error message more helpful

PR #5492, opened 2023-12-04
Fixed `preview_image_link` branch use case in metadata tags in head

sneridagh:


davisagli (commented):
@sneridagh I'm not at all sure this is right. The serialization of `preview_image_link` is different on 2024.kitconcept.com (e.g. https://2024.kitconcept.com/++api++/en/blog-1/plone-conference-2023-eibar) than on the DLR site, and I don't remember why.

sneridagh:
@davisagli is that so? I took DLR as pattern, oh true, the behavior is slightly different. I will take a look.

sneridagh:
superseeded: #5939

PR #5498, opened 2023-12-06
Fix path for cypress artifacts

davisagli:
and use v3 of actions/cache, since v2 logs a warning about using an old Node version


packages/volto/news/5498.internal
davisagli, line 1: @sneridagh Is this the right place for a changelog entry about a change that is in the repository root and not inside any package?

PR #5506, opened 2023-12-07
get multilingual settings from backend

erral:
To be tested with https://github.com/plone/plone.restapi/pull/1738


sneridagh:
@erral the server has to know in advance this setting too. If not, all SSR requests will be "not multilingual". A golden rule is that the (nowadays static) config should be the same in server and in the client. Sometimes, one tend to forget that they are "detached" physically.

Knowing this setting in advance, is part of what I've come to call "bootstrap config info" service that should be retrieved from the server as first thing (there are several others, like is working copy enabled, or this site accept registrations), before everything else happens, first in SSR, passed then to the client. We've postponed it a lot of times because pulling this off won't be easy in a performant way.

sneridagh:
@erral I took a look again to this PR and now I saw the `getMultilingualInstalledAsyncPropExtender` solution. Could be that I missed that from your initial implementation, or you added it later? I think this will solve my initial concerns.

I am wondering if we could have a full-backend implementation for this use case, so we can even spare the `getMultilingualInstalledAsyncPropExtender` solution (eg. included in the `@site` endpoint). I wanted also to hold off this one also in case we started using `@plone/client` in Volto, but since this is most likely never happen, I am wondering if it's now ok to push again for this. What do you think?

/cc @davisagli

erral:
See #6818 for the new version

PR #5514, opened 2023-12-11
Make sure that the dependencies are build and available on all processes

sneridagh:


sneridagh:
@davisagli I realised that we have to maintain the versions in two `Makefile`. I think we need to make sure that running from "inside" `packages/volto` still works seamlessly (to help the transition, for a while). Not sure about it, my Makefile-fu is not strong, but how about using a centralised `.env` and include it from both?

PR #5518, opened 2023-12-12
Put cloneData function for slate block, in order to change the uuids of fragments

dobri1408:


davisagli (commented):
@dobri1408 Could you please add a description of the problem you are solving to the PR description? Without it, it is hard to review.

dobri1408:
Slate lets you put on some of its words(fragments) uuids, that can be used for add-ons, for example for volto-slate-footnotes from EEA. But, if you use the copy and paste of Volto from the sidebar, it will copy the fragments with the same uuids, and this is not ok.  @davisagli

packages/volto/news/5518.feature
```diff
@@ -0,0 +1 @@
+Put cloneData function for slate block in order to change tu uuids of fragments when copy @dobr1408
```
stevepiercy, line 1: I'm not sure what "tu" is supposed to be. Is that a typo?

Based on what I read in the discussion, and what little I can decipher in the code, here is what I think is the intent. Please correct as needed.

```suggestion
When copying a text block—implemented in Slate—using the sidebar that contains UUIDs attached to elements in the data, the UUIDs get copied into the new block, making them not unique. Now instead of duplicating the UUIDs, a unique nanoid will be generated and assigned to the copy, preserving uniqueness. @dobr1408
```

stevepiercy: Also if this fixes a bug, and does not introduce a new feature, then the file should be renamed to `5518.bugfix`.

dobri1408:
@stevepiercy yes indeed

packages/volto-slate/src/blocks/Text/index.js
```diff
@@ -118,6 +121,27 @@ export default function applyConfig(config) {
...
+    cloneData: (data) => {
+      return [
+        uuid(),
+        {
+          ...data,
+          value: [
+            ...(data?.value || []).map((c) => {
+              return {
+                ...c,
+                children: (c?.children || []).map((childrenData) => {
```
davisagli, line 133: Won't this add `children` to nodes that should not have it, like text nodes?

I would really like to see a test for this feature.

packages/volto-slate/src/blocks/Text/index.js
```diff
@@ -118,6 +121,27 @@ export default function applyConfig(config) {
...
+                  return {
+                    ...childrenData,
+                    data: { ...childrenData.data, uid: nanoid(5) },
```
davisagli, line 136: This will add a uid even if there wasn't one there before; is that what we want?

packages/volto-slate/news/5518.bugfix
```diff
@@ -0,0 +1 @@
+When copying a text block—implemented in Slate—using the sidebar that contains UUIDs attached to elements in the data, the UUIDs get copied into the new block, making them not unique. Now instead of duplicating the UUIDs, a unique nanoid will be generated and assigned to the copy, preserving uniqueness. @dobr1408
```
davisagli, line 1: Not sure what the sidebar bit is about.

```suggestion
When copying a text block—implemented in Slate—that contains UUIDs attached to elements in the data, the UUIDs get copied into the new block, making them not unique. Now instead of duplicating the UUIDs, a new UUID will be generated and assigned to the copy, preserving uniqueness. @dobr1408
```

davisagli (changes requested):

packages/volto-slate/src/blocks/Text/index.js
dobri1408: I think that all children have uid by default

packages/volto-slate/src/blocks/Text/index.js
dobri1408: Yes this is true, I will modify, and add tests

packages/volto-slate/news/5518.bugfix
dobri1408: "sidebar"  is there to emphasies that is about copying using volto, not the slate default copying method

packages/volto-slate/src/blocks/Text/index.js
davisagli: No, I don't think so. I have not seen it before.

packages/volto-slate/news/5518.bugfix
davisagli: I think that part of the volto UI is called the toolbar, not the sidebar

stevepiercy: Given that we are "copying a text block", I assume the context is obvious that it is a Volto block, not a Slate element, that is being copied, so we don't need to name whatever that thing is called. 😁 I agree with @davisagli's suggestion here.

Side note, after a quick search in the docs, the thingie on the right-hand side in Volto is usually called "sidebar". The thingie on the left-hand side is usually called "Plone toolbar" or "toolbar", and the thingie in a text editor is called just "toolbar" and sometimes "editor toolbar" or "Slate toolbar". It would be good to officially name these thingies. Great, I just made more documentation work for me. 🙄

stevepiercy (approved):
I approve docs. Code needs review and approval from a core contributor.

dobri1408:
@davisagli what do you think about this way?

dobri1408:
@davisagli

avoinea (approved):

ichim-david:
@tiberiuichim maybe you can also give you input on this change.
@davisagli @sneridagh are there other concerns that we need to address in order to get this pull request merged?

tiberiuichim:
It's missing a test. This thing would be merged a lot faster if it proved that it did what it claims to do.

ichim-david:
Awesome, were getting somewhere :) If the user knows what he needs todo then he will do it but I would like for us to avoid letting pull requests stay for months until they get so old that it takes a considerable effort to bring up to date.

I will try to be more helpful as well to push for my colleagues to ask quicker for feedback so that we can add quickly what is  necessary to get a change in a mergeable state and finally merged.

Thanks to you both for the feedback, we will see about getting those tests added.

davisagli:
@ichim-david For what it's worth, I [asked for tests](https://github.com/plone/volto/pull/5518#discussion_r1424918098) in December, and @dobri1408 [said he would add them](https://github.com/plone/volto/pull/5518#discussion_r1424920073), so I thought the next step was clear.

I agree that it would be good if we can find a way to respond to pull requests faster, but I'm not sure what to do about the current backlog. Realistically I only have time and energy to do a full review of one or two per day, and not every day.

ichim-david:
> @ichim-david For what it's worth, I [asked for tests](https://github.com/plone/volto/pull/5518#discussion_r1424918098) in December, and @dobri1408 [said he would add them](https://github.com/plone/volto/pull/5518#discussion_r1424920073), so I thought the next step was clear.
> 
> I agree that it would be good if we can find a way to respond to pull requests faster, but I'm not sure what to do about the current backlog. Realistically I only have time and energy to do a full review of one or two per day, and not every day.

@davisagli I thought the test was already written due to the last comment from @dobri1408 https://github.com/plone/volto/pull/5518#issuecomment-1860642389 and due to @avoinea approving the pull request as well.

If I can make a small personal request, when asking for further todo's before merging, do it in a dedicated comment, that small comment was in a thread and it seemed like something solved already, but when seeing a test requirement comment you can see if there was any reply given on that individual comment if done or not.

davisagli (commented):
@dobri1408 (+ /cc @ichim-david @tiberiuichim)

I'm looking at this again and have some questions:
1. What's the goal for how this interacts with the volto-slate-footnote addon? Is the cloned block supposed to refer to the same footnote, or a new copy of the footnote, or no footnote? If no footnote, then I think it would make more sense to remove the uid instead of generate a new one that is not used for anything?
2. I'm having a doubt about whether volto should try to anticipate everything that might need to happen to adjust a cloned block. Maybe it should just provide a setting that addons can use to register their own cleanup function.

PR #5532, opened 2023-12-13
Support for cross-JS/TS aliases in the add-on registry aliases generator

sneridagh:
I realised this morning that we can do this... and it does work, back and forth...
So let's open the veto and allow to add/convert code to TS anywhere in the Volto core.


tiberiuichim:
One potential problem that I can think of: we have React components in .js files, which would map incorrectly to `.ts`. One fix would be to go through the repo, identify all such cases and move them to `.jsx`

tiberiuichim:
Here's an example: https://github.com/plone/volto/blob/06f17199614ab1f601d6bcac30b1c4d635add56f/packages/volto/src/components/theme/LanguageSelector/LanguageSelector.js#L42

sneridagh:
Uh, I never realised! We should fix it, for sure. It would be a breaking. However, I assume that it would continue working if the counterpart is consistent with the mapping.

tiberiuichim:
@sneridagh I mean, it's possible to amend the code and look for the other filename variations.

sneridagh:
Won't we run possibly into unwanted and wrong mappings then (maybe we have other edge cases in there)?

If we have the convention of every file that contains JSX has to have the especific extension, let's enforce it, and fix violations.

PR #5533, opened 2023-12-14
Remove outdated acceptance tests docs

sneridagh:
The "good" ones are in `docs/source/contributing/acceptance-tests.md`


ichim-david (approved):

stevepiercy (changes requested):
LGTM. Out of curiosity, do we have current docs for acceptance tests somewhere?

stevepiercy (approved):

stevepiercy:
nm, answered my own question: https://6.docs.plone.org/volto/contributing/acceptance-tests.html

But that begs the question, why do we break apart testing into separate pages?

ichim-david:
> nm, answered my own question: https://6.docs.plone.org/volto/contributing/acceptance-tests.html
> 
> But that begs the question, why do we break apart testing into separate pages?
@sneridagh you make a good point maybe the better enhancement would be to bring the acceptance testing
to replace the section that is deleted so that the testing is in one place only.
The acceptance section is larger but we have the TOC to the right so I guess navigation wouldn't be a problem

sneridagh:
@stevepiercy I think acceptance testing is big (and important in Volto) enough to have its own section.

PR #5541, opened 2023-12-15
Fix autopopulated value of facet when settings the value for another

### LINKED ISSUE #5432: Search block facet value autopopulated with listing criteria value when you select other facet value.
The value of the facet is filled automatically. I haven't selected the facet value and it appears automatically when I selected the other facet value.

1. Create a search block.
2. In listing criteria select the review state and it's value. You can see the screenshot below.
3. After that in first facet select the review state once again.
4. In the second facet select the Type and save the page.
5. Go in view mode and change the Type facet to select a value a value like `Event`.

You will see the review state facet is filled automatically. Which is the bug.

https://github.com/plone/volto/assets/33936987/b45fe2c9-8d22-4ee7-8f51-80e7d214ad41

iFlameing:
…one.

Fixes #5432 

@tiberiuichim Can you please review it.


sneridagh:
@iFlameing still a draft and WIP or is it ready to review?

iFlameing:
@sneridagh  @tiberiuichim Ready for review now.

packages/volto/src/components/manage/Blocks/Search/hocs/withSearch.jsx
```diff
@@ -96,9 +96,26 @@ function normalizeState({
   const { types: facetWidgetTypes } =
     config.blocks.blocksConfig.search.extensions.facetWidgets;
 
+  // Here we are removing the QueryString of the Listing ones which is persent in the Facet
```
tiberiuichim, line 99: typo

tiberiuichim: Please make this comment more intelligible

packages/volto/src/components/manage/Blocks/Search/hocs/withSearch.jsx
```diff
@@ -96,9 +96,26 @@ function normalizeState({
...
+  // because we already initialize the facet with those value.
+  const configuredFacets = facetSettings
+    ? facetSettings.map((facet) => facet?.field?.value)
+    : [];
+
+  let copyOfQuery = query.query ? [...query.query] : [];
+
+  for (let value of configuredFacets) {
```
tiberiuichim, line 107: If I understand correctly, I think we can use a list filter for this whole code block, which would simplify it and make  easier to understand.

packages/volto/src/components/manage/Blocks/Search/hocs/withSearch.jsx
```diff
@@ -288,8 +325,17 @@ const withSearch = (options) => (WrappedComponent) => {
     );
     const previousUrlQuery = usePrevious(urlQuery);
 
+    // During first render the previousUrlQuery is undefined and urlQuery
+    // is empty so it ressetting the facet when you are navigating but during reload we have urlQuery and we need
+    // to set the facet at first render.
+    const preventOverrideOfFacetState =
+      previousUrlQuery === undefined && urlQuery.length === 0;
```
tiberiuichim, line 332: I think trying to make decision here based on some previous state is not a good way to do things. In general. It makes the working of the block harder to reason about (because you introduce time-traveling by looking at the previous state), and harder to replicate, etc.

Now, I am guily of doing this as well, in other cases. But I always feel that it's not a good path to go.

tiberiuichim, line 332: I think trying to make decision here based on some previous state is not a good way to do things. In general. It makes the working of the block harder to reason about (because you introduce time-traveling by looking at the previous state), and harder to replicate, etc.

Now, I am guily of doing this as well, in other cases. But I always feel that it's not a good path to go.

iFlameing: Please suggest to me or give me a pointer to avoid this. I am not able to think of any solution.

tiberiuichim (approved):

PR #5542, opened 2023-12-15
i18n polishing

erral:
I have been working on polishing the po file generation to align with Weblate. They are mainly cosmetic changes, but this way we create the same po file structure that we are using in plone.app.locales, and we avoid Weblate doing a lot of changes each time there is a PR (changes like putting first the file the msgid comes from instead of the default message)

- Changed the `# defaultMessage: xxxx` header with `#. Default: "xxx"`.
- Changed the order of the default message and the "source" of the msgid
- Changed the order of the `MIME-Version` header.
- Fixed 2 msgids and default values that in Volto had newlines, so they couldn't be properly translated.

Generated PO files are still valid and can be edited with other tools like Poedit, for instance.

If you find the PR OK, after it's merged I will work on Weblate integration.


stevepiercy:
Will this PR result in a change to how developers currently work on translations, specifically in [Internationalization](https://6.docs.plone.org/volto/recipes/i18n.html)?

erral:
No, they can continue working like they work today. 

There are no changes on the way Volto code is internationalized.

They can even continue working translating Volto editing directly the PO files. 

No changes on that.

stevepiercy:
Is there a plan to ask folks to translate Volto through Weblate? I just hunted down this https://hosted.weblate.org/projects/plone/volto-main/.

This looks to be an easier way for people to contribute translations, but would it bypass the Plone Contributor Agreement?

erral:
Yes the plan is to integrate Weblate too, but it is not mandatory, users can keep working on PO files too.

To tell the truth I didn't realize about the contributor5agreement to translate Volto. As it is not needed for plone.app.locales (it lives in collective), I didn't even thought about this.

But, i the CA is required, I would either retire Volto from Weblate or check if any User validation can be added there. Anyway, rigth now the contributions for Volto in Weblate are stopped.

stevepiercy:
I'm not a Contributor Agreement expert. Let's get @polyester's or another Plone Foundation member's opinion on making contributions of translations through Weblate. Weblate has stuff about agreeing to the license of the project, but nothing about a contributor agreement. If we could automate MrRoboto to check a contributor through Weblate just as it does now for GitHub, then that might be acceptable.

davisagli (approved):
Looks good to me.

davisagli:
@sneridagh Okay to merge?

sneridagh (approved):

sneridagh:
@erral we will need a backport PR for 17 and 16, I guess.

erral:
Yes, I plan to connect Weblate only to `main`.

I can check the backport if the cherry-picking does not work.

sneridagh:
@erral @davisagli don't know if issue this one as breaking... since people using it are going to see changes in their .po (although they will be innocuous). What do you think?

davisagli:
@sneridagh Not breaking. The old .po format still works if they do not update it.

PR #5581, opened 2023-12-27
Added support for custom CSS properties in the StyleWrapper

sneridagh:
This is a pattern I use lately, and it makes a lot of sense to me.

A couple of questions:

- Would it be useful/needed to also inject a className in order to hook to? I was about to do it, but I don't know if that would be required... maybe to know if a value is set or not... but you can always have a default fallback value.
- Naming... what not! I chose `CSSProperty` as the sufix... any other suggestions?

I'm well aware that this opens the door to fine grained style customizations, if we extend the pattern not only to use custom CSS properties but also use random CSS properties (another converter should be created, then do not append the -- and disable nested support, then) I leave it up to the community. However, the initial feeling when we came up with the style wrapper stands: I wouldn't like the wrapper to be missunderstood, and that people are able to easy create blocks that resemble WordPress Elementor...

Let's iterate over this, in the meanwhile, I will work on integrate the PoC into VLT.

**Update 01/01/2024**: Answer to questions:

- It's not really needed, since one could build a selector like this:

```css
 .block.teaser[style*='--background-color'] {
    padding: 20px 0;
}
```

if required.

- This PR is not using a suffix discriminator to differentiate from the others. Now it detects if the property starts by `--` see the documentation.

### DIFF
--- a/docs/source/blocks/block-style-wrapper.md
+++ b/docs/source/blocks/block-style-wrapper.md
@@ -151,6 +151,9 @@ Then it's at your discretion how you define the CSS class names in your theme.
 
 ## Customize the injected class names
 
+```{versionadded} release-version-number
+```
+
 If you need other style of classnames generated, you can use the classname
 converters defined in `config.settings.styleClassNameConverters`, by
 registering fieldnames suffixed with the converter name. For example, a style
@@ -169,10 +172,13 @@ will generate classnames `primary inverted`
 
 ## Injecting custom CSS properties
 
+```{versionadded} release-version-number
+```
+
 The style wrapper also allows to inject custom CSS properties, using the converter syntax.
 This use case is useful in some scenarios where the property that you are injecting is generic, customizable per project.
 
-```
+```css
 {
   "styles": {
     "backgroundColor:CSSProperty": "#222",
@@ -180,7 +186,7 @@ This use case is useful in some scenarios where the property that you are inject
 }
 ```
 
-will inject a style object prop in the component:
+The above style will inject a style object property in the following component:
 
 ```html
 <div class="block teaser" style="--background-color: #222">
@@ -211,7 +217,7 @@ const BlockView = (props)=> (
 ```
 
 ```{note}
-You need to manually add the above code in your view component block code in order to benefit from the style injection.
+You need to manually add the above code in your view component block code to benefit from the style injection.
 The styles in the block edit component are injected automatically into the blocks engine editor wrappers, so you don't have to take any action.
 ```
 
--- a/docs/source/blocks/block-style-wrapper.md
+++ b/docs/source/blocks/block-style-wrapper.md
@@ -182,6 +182,14 @@ Then in your customized theme, you could set CSS attributes for this property, c
 The key feature of custom CSS properties is that they can be applied also using the cascade.
 This means that they can be placed anywhere in either CSS definitions or HTML structure, and they will be applied only in the context where they are defined.
 
+As an example, first define the style of a teaser block image as follows.
+
+```css
+.block.teaser img {
+  aspect-ratio: var(--image-aspect-ratio, 16 / 9);
+}
+```
+
 Next, you can enhance a block's schema by injecting the custom CSS property as follows.
 
 ```js
@@ -208,14 +216,6 @@ Finally the markup of the block will contain the custom property as shown.
 <div class="block teaser" style="--image-aspect-ratio: 1">
 ...
 </div>
-```
-
-As an example, first define the style of a teaser block image as follows.
-
-```css
-.block.teaser img {
-  aspect-ratio: var(--image-aspect-ratio, 16 / 9);
-}
 ```
 
 The custom CSS property definition will only apply within the div that it's defined.
--- a/docs/source/blocks/block-style-wrapper.md
+++ b/docs/source/blocks/block-style-wrapper.md
@@ -175,32 +175,50 @@ will generate classnames `primary inverted`
 ```{versionadded} 17.8.0
 ```
 
-The style wrapper also allows to inject custom CSS properties, using the converter syntax.
+The style wrapper also allows to inject custom CSS properties.
 This is useful where the property that you want to inject is customizable per project.
+For example, imagine we have in place a global CSS custom property `--image-aspect-ratio` so all the images in blocks (teasers, listings too) use it. The idea was that someone in his customized theme could set it and change them all in runtime, because this is how custom CSS properties work.
+The key feature of custom CSS properties is that they can be applied also using the cascade.
+This means that they can be placed anywhere in the CSS definitions or in the HTML structure, and they will be applied only in the context they are defined.
 
-```css
-{
-  "styles": {
-    "--background-color": "#222",
-  }
-}
+When the feature in this PR is used combined with the above you can enhance a block's schema (eg. teaser) as follows:
+
+```js
+  schema.properties.styles.schema.fieldsets[0].fields = [
+    ...schema.properties.styles.schema.fieldsets[0].fields,
+    '--image-aspect-ratio',
+  ];
+
+  // We use a select widget and set a default
+  schema.properties.styles.schema.properties['--image-aspect-ratio'] = {
+    widget: 'select',
+    title: 'Aspect Ratio',
+    choices: [
+      ['1', '1:1'],
+      ['16 / 9', '16/9'],
+    ],
+    default: '1',
+  };
 ```
 
-The above style will inject a style object property in the following component:
+Then, the markup of the block will contain the custom property:
 
 ```html
-<div class="block teaser" style="--background-color: #222">
+<div class="block teaser" style="--image-aspect-ratio: 1">
+...
+</div>
 ```
 
-Then, provided that you have the following CSS in place:
+and if we have this CSS definitions in place:
 
 ```css
-.block.teaser {
-  background-color: var(--background-color);
+.block.teaser img {
+  aspect-ratio: var(--image-aspect-ratio, 16 / 9);
 }
 ```
 
-It will apply the variable value injected by the style wrapper for the teaser block.
+The custom CSS property definition will only apply within the div that it's defined.
+Then the CSS where it's used will apply that custom CSS property value for only that div.
 
 If you want to use it in your custom components, you need to apply it in the root of your block's view component as follows:
 
--- a/docs/source/blocks/block-style-wrapper.md
+++ b/docs/source/blocks/block-style-wrapper.md
@@ -177,11 +177,12 @@ will generate classnames `primary inverted`
 
 The style wrapper also allows to inject custom CSS properties.
 This is useful where the property that you want to inject is customizable per project.
-For example, imagine we have in place a global CSS custom property `--image-aspect-ratio` so all the images in blocks (teasers, listings too) use it. The idea was that someone in his customized theme could set it and change them all in runtime, because this is how custom CSS properties work.
+For example, imagine you have an existing global CSS custom property `--image-aspect-ratio` that you use for all images in all blocks.
+Then in your customized theme, you could set CSS attributes for this property, changing it in runtime.
 The key feature of custom CSS properties is that they can be applied also using the cascade.
-This means that they can be placed anywhere in the CSS definitions or in the HTML structure, and they will be applied only in the context they are defined.
+This means that they can be placed anywhere in either CSS definitions or HTML structure, and they will be applied only in the context where they are defined.
 
-When the feature in this PR is used combined with the above you can enhance a block's schema (eg. teaser) as follows:
+Next, you can enhance a block's schema by injecting the custom CSS property as follows.
 
 ```js
   schema.properties.styles.schema.fieldsets[0].fields = [
@@ -201,7 +202,7 @@ When the feature in this PR is used combined with the above you can enhance a bl
   };
 ```
 
-Then, the markup of the block will contain the custom property:
+Finally the markup of the block will contain the custom property as shown.
 
 ```html
 <div class="block teaser" style="--image-aspect-ratio: 1">
@@ -209,7 +210,7 @@ Then, the markup of the block will contain the custom property:
 </div>
 ```
 
-and if we have this CSS definitions in place:
+As an example, first define the style of a teaser block image as follows.
 
 ```css
 .block.teaser img {
@@ -218,7 +219,7 @@ and if we have this CSS definitions in place:
 ```
 
 The custom CSS property definition will only apply within the div that it's defined.
-Then the CSS where it's used will apply that custom CSS property value for only that div.
+As you can see, the custom CSS property applies only within the `div` in which it is defined.
 
 If you want to use it in your custom components, you need to apply it in the root of your block's view component as follows:
 
--- a/docs/source/blocks/block-style-wrapper.md
+++ b/docs/source/blocks/block-style-wrapper.md
@@ -151,7 +151,7 @@ Then it's at your discretion how you define the CSS class names in your theme.
 
 ## Customize the injected class names
 
-```{versionadded} release-version-number
+```{versionadded} 16.0.0
 ```
 
 If you need other style of classnames generated, you can use the classname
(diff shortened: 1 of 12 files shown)


docs/source/blocks/block-style-wrapper.md
```diff
@@ -165,6 +167,54 @@ data like:
 
 will generate classnames `primary inverted`
 
+## Injecting custom CSS properties
+
+The style wrapper also allows to inject custom CSS properties, using the converter syntax.
+This use case is useful in some scenarios where the property that you are injecting is generic, customizable per project.
```
stevepiercy, line 173: I'm not sure whether "generic" and "customizable" should be used in the same sentence. Either the class is generic or customized, but not both at the same time. Can you clarify which it is or is it either one or the other?

```suggestion
This use case is useful in the scenario where the property that you want to inject is either generic or is customizable per project.
```

docs/source/blocks/block-style-wrapper.md
```diff
@@ -165,6 +167,54 @@ data like:
...
+
+```
```
stevepiercy, line 175:
````suggestion
```css
````

File: docs/source/blocks/block-style-wrapper.md [hunk 3]
```diff
@@ -165,6 +167,54 @@ data like:
...
+{
+  "styles": {
+    "backgroundColor:CSSProperty": "#222",
+  }
+}
+```
+
+will inject a style object prop in the component:
```
stevepiercy, line 183:
```suggestion
The above style will inject a style object property in the following component:
```

docs/source/blocks/block-style-wrapper.md
```diff
@@ -165,6 +167,54 @@ data like:
...
+
+```html
+<div class="block teaser" style="--background-color: #222">
+```
+
+```{note}
+Please notice that the resultant variable is transformed from camel case to kebab case.
```
stevepiercy, line 190: "Note" is inserted automatically in the admonition.

```suggestion
The resultant variable is transformed from camel case to kebab case.
```

docs/source/blocks/block-style-wrapper.md
```diff
@@ -165,6 +167,54 @@ data like:
...
+```jsx
+const BlockView = (props)=> (
+  <div style={props.style}>
+    // Block's view component code
+  </div>
+)
+```
+
+```{note}
+You need to manually add the above code in your view component block code in order to benefit from the style injection.
```
stevepiercy, line 214:
```suggestion
You need to manually add the above code in your view component block code to benefit from the style injection.
```

packages/volto/news/5581.feature
```diff
@@ -0,0 +1 @@
+Added support for custom CSS properties in the StyleWrapper @sneridagh
```
stevepiercy, line 1:
```suggestion
Added support for custom CSS properties in the `StyleWrapper`. @sneridagh
```

docs/source/blocks/block-style-wrapper.md
```diff
@@ -149,6 +149,8 @@ The resultant HTML would be the following:
 ```
 Then it's at your discretion how you define the CSS class names in your theme.
 
+## Customize the injected class names
+
```
stevepiercy, line 153: I don't know in which version this will be released, but this is a helpful admonition for developers. See also https://myst-parser.readthedocs.io/en/latest/syntax/admonitions.html#admonition-types. There is also `versionchanged` if you think that is more appropriate, but this smells like a new feature, so I went with `versionadded`.

````suggestion

```{versionadded} release-version-number
```

````

File: docs/source/blocks/block-style-wrapper.md [hunk 8]
```diff
@@ -165,6 +167,54 @@ data like:
 
 will generate classnames `primary inverted`
 
+## Injecting custom CSS properties
+
```
stevepiercy, line 171:
````suggestion

```{versionadded} release-version-number
```

````

**Review by stevepiercy — CHANGES_REQUESTED** [review]
Although not part of this PR, I found `classname` and `fieldname` and their plural forms to be weird. Are they supposed to be inline literals? If yes, then they need to be wrapped with single backticks and use the correct casing, else they should be "class name" and "field name".

File: docs/source/blocks/block-style-wrapper.md [hunk 9]
```diff
@@ -165,6 +167,54 @@ data like:
...
+The style wrapper also allows to inject custom CSS properties, using the converter syntax.
+This use case is useful in some scenarios where the property that you are injecting is generic, customizable per project.
+
+```
+{
+  "styles": {
+    "backgroundColor:CSSProperty": "#222",
```
davisagli, line 178: Could this be written as 
```suggestion
    "--background-color": "#222",
```
instead? Neither one can be accessed without square brackets in javascript, so I don't see a benefit to transforming the name to something else. Writing it the same way as the actual CSS property seems clearer to me.

sneridagh: The converters detection system now works with sufixes, colon separated. Are you saying that you'd prefer to add another detection system based on if '--' is present?

sneridagh: It's a good point though...

docs/source/blocks/block-style-wrapper.md
sneridagh: The classnames injection was released somehwere in middle 16 I think.

docs/source/blocks/block-style-wrapper.md
sneridagh: I did frown too but it's what it is... something ready to be customized, then it's generic, right?

docs/source/blocks/block-style-wrapper.md
sneridagh: @davisagli done, it was easy than expected.

docs/source/blocks/block-style-wrapper.md
sneridagh: Removed, no longer applies.

docs/source/blocks/block-style-wrapper.md
stevepiercy: Oh, I thought you are adding the custom CSS properties feature with this PR?

But for the original version, at the top of the file just below the title, we should have:

````
```{versionadded} 16.0.0
```
````

Comment (stevepiercy) [reply docs/source/blocks/block-style-wrapper.md:153]: Or are you referring to the interim, where something changed?

https://github.com/plone/volto/commits/5b4ec3f31ed20a12fc3eb25ffc49c9dee11bb4eb/docs/source/blocks/block-style-wrapper.md

If so, then wherever that changed, it should have the same syntax:

````
```{versionchanged} 16.x.x-or-whatever
```
````

File: docs/source/blocks/block-style-wrapper.md (hunk 1 above)
Comment (stevepiercy) [reply docs/source/blocks/block-style-wrapper.md:179]: Something ready to be customized is "customizable". Generic means general or lacking specificity. I edited my suggestion accordingly and made it more concise.

File: docs/source/blocks/block-style-wrapper.md [hunk 10]
```diff
@@ -165,6 +170,53 @@ data like:
 
 will generate classnames `primary inverted`
 
+## Injecting custom CSS properties
```
stevepiercy, line 173:
```suggestion
## Inject custom CSS properties
```

stevepiercy (changes requested):
Fixed a heading and follow up comments.

docs/source/blocks/block-style-wrapper.md
sneridagh: Moved to the top. The thing added now is only the custom properties, so it's also noted above it with another versionadded.

stevepiercy: OK. That works for the title, but we need a specific release number for this new feature here. That could be tricky if you backport it, but you would just need to remember to do the chore.

stevepiercy: Oh, nm, I see you did that below. I had not scrolled down yet.

stevepiercy (approved):
Thank you!

tiberiuichim:
The slope is steep and full of grease. Pretty soon we'll end up here: https://raw.githubusercontent.com/eea/volto-block-style/master/docs/screenshot.png

When I started volto-block-style I had the same intention: let me create a way to centralize block "themes" and make it possible to assign those. That's why the "Style presets" widget is placed first in that volto-block-style sidebar. Pretty soon it devolved into whatever this is: https://github.com/eea/volto-block-style/blob/1fd970881ca1d816f8db764cdb2498348f10f191/src/StyleWrapper/schema.js#L3

sneridagh:
@tiberiuichim I know... it's not easy. It's very hard to predict which techniques and patterns are the best for the use case. 

When I first pushed for the wrapper I had one thing clear: The intention was not to provide an easy way to plague Volto blocks settings with Elementor-like controls... because it's not what we are aiming for in Volto. `volto-style-block` was a bit of that. It's absolutely legit if you have a client that explicitly asks for this, and it's editors are real power users who knows how the web works and are capable of handling this complexity. So not enabling doing it in an easy way was important for me.

This is something in between, still with a clear purpose, and documented so people can know how to take advantage of it. A width value, a color, so you can then take it in the CSS and use it right away, one shoot only CSS. 

As I said, this opens the pandora's box by perverting it could be used to really inject arbitrary CSS properties. If you really wanted to do it, you can, but on your own. Although, given the power of using custom CSS properties instead, who wants this nowadays?

davisagli (approved):
Thanks for changing it so the developer can just write the property name the same way it is rendered; I think that makes it easier to use.

I do have some of the same doubts about whether people will use the feature judiciously, but the potential problem of adding too many options is there whether or not this feature exists. Would it help to give an example in the docs of a situation where using this would be appropriate and one where it would not?

docs/source/blocks/block-style-wrapper.md
```diff
@@ -175,32 +175,50 @@ will generate classnames `primary inverted`
 ```{versionadded} 17.8.0
 ```
 
-The style wrapper also allows to inject custom CSS properties, using the converter syntax.
+The style wrapper also allows to inject custom CSS properties.
 This is useful where the property that you want to inject is customizable per project.
+For example, imagine we have in place a global CSS custom property `--image-aspect-ratio` so all the images in blocks (teasers, listings too) use it. The idea was that someone in his customized theme could set it and change them all in runtime, because this is how custom CSS properties work.
+The key feature of custom CSS properties is that they can be applied also using the cascade.
+This means that they can be placed anywhere in the CSS definitions or in the HTML structure, and they will be applied only in the context they are defined.
```
stevepiercy, line 182:
```suggestion
This means that they can be placed anywhere in either CSS definitions or HTML structure, and they will be applied only in the context where they are defined.
```

docs/source/blocks/block-style-wrapper.md
```diff
@@ -175,32 +175,50 @@ will generate classnames `primary inverted`
...
 
-```css
-{
-  "styles": {
-    "--background-color": "#222",
-  }
-}
+When the feature in this PR is used combined with the above you can enhance a block's schema (eg. teaser) as follows:
```
stevepiercy, line 184: Again, I'm not clear, so here's my best guess:

```suggestion
Next, you can enhance a block's schema by injecting the custom CSS property as follows.
```

docs/source/blocks/block-style-wrapper.md
```diff
@@ -175,32 +175,50 @@ will generate classnames `primary inverted`
...
+    choices: [
+      ['1', '1:1'],
+      ['16 / 9', '16/9'],
+    ],
+    default: '1',
+  };
 ```
 
-The above style will inject a style object property in the following component:
+Then, the markup of the block will contain the custom property:
```
stevepiercy, line 204:
```suggestion
Then the markup of the block will contain the custom property:
```

docs/source/blocks/block-style-wrapper.md
```diff
@@ -175,32 +175,50 @@ will generate classnames `primary inverted`
...
 
 ```html
-<div class="block teaser" style="--background-color: #222">
+<div class="block teaser" style="--image-aspect-ratio: 1">
+...
+</div>
 ```
 
-Then, provided that you have the following CSS in place:
+and if we have this CSS definitions in place:
```
stevepiercy, line 212: I think that we should move the CSS code-block and this sentence to line 183. It flows better to define CSS first, then inject it through JS, and finally show the resultant HTML.

```suggestion
As an example, first define the style of a teaser block image as follows.
```

docs/source/blocks/block-style-wrapper.md
stevepiercy, line 204:
```suggestion
Finally the markup of the block will contain the custom property as shown.
```

docs/source/blocks/block-style-wrapper.md
```diff
@@ -175,32 +175,50 @@ will generate classnames `primary inverted`
...
-.block.teaser {
-  background-color: var(--background-color);
+.block.teaser img {
+  aspect-ratio: var(--image-aspect-ratio, 16 / 9);
 }
 ```
 
-It will apply the variable value injected by the style wrapper for the teaser block.
+The custom CSS property definition will only apply within the div that it's defined.
+Then the CSS where it's used will apply that custom CSS property value for only that div.
```
stevepiercy, line 221: I read the above two sentences a half-dozen times, and I am not sure why both are needed.

```suggestion
As you can see, the custom CSS property applies only within the `div` in which it is defined.
```

stevepiercy (changes requested):
I am unsure of a couple of passages, and noted them as such.

I also suggested reordering the code blocks and their narrative text to improve flow.

docs/source/blocks/block-style-wrapper.md
sneridagh: Not really, the above explanation was about how custom CSS properties work, the next is how you take advantage of it. The custom CSS property

docs/source/blocks/block-style-wrapper.md
sneridagh: @stevepiercy I'm just stating how custom CSS properties work, in case that somebody didn't realise that this can be done.

tiberiuichim (approved):

docs/source/blocks/block-style-wrapper.md
```diff
@@ -165,6 +170,72 @@ data like:
...
+    ],
+    default: '1',
+  };
+```
+
+Finally the markup of the block will contain the custom property as shown.
+
+```html
+<div class="block teaser" style="--image-aspect-ratio: 1">
+...
```
stevepiercy, line 217: Without an `img`, I don't see how the property will be applied because the selector does not actually match anything in the HTML as it is.

```suggestion
<img src="example.png">
```

docs/source/blocks/block-style-wrapper.md
stevepiercy: I'm confused because this appears to be just how CSS works, and we really don't want to reinvent a [tutorial for basic CSS](https://developer.mozilla.org/en-US/docs/Learn/Getting_started_with_the_web/CSS_basics). Maybe instead we add a seealso admonition?

````

```{seealso}
[CSS basics](https://developer.mozilla.org/en-US/docs/Learn/Getting_started_with_the_web/CSS_basics)
```

````

Also these two sentences as they are now don't make things clear to me. If we use the official W3C terms, maybe this conveys your intent?

````suggestion
As you can see, the custom CSS declaration applies only in the `div` where you inject the property.
````

File: docs/source/blocks/block-style-wrapper.md [hunk 17]
```diff
@@ -165,6 +170,72 @@ data like:
...
+    title: 'Aspect Ratio',
+    choices: [
+      ['1', '1:1'],
+      ['16 / 9', '16/9'],
+    ],
+    default: '1',
+  };
+```
+
+Finally the markup of the block will contain the custom property as shown.
```
stevepiercy, line 213:
```suggestion
Finally, assuming that you select the default value for the {guilabel}`Aspect Ratio` for the custom CSS property, then the markup of the block will contain the custom property as shown.
```

stevepiercy (changes requested):
I think we're getting closer. I rearranged the CSS to be first, so it flows better, and filled in a couple of pieces I thought were missing. Finally, I took another stab at those two sentences that confuse me more than help me understand.

ichim-david (approved):

davisagli (approved):

stevepiercy:
[These two sentences got missed again in the review](https://github.com/plone/volto/pull/5581#discussion_r1438402042). I'll create a separate PR to fix it.

PR #5584, opened 2023-12-29
Add robots.txt setting.

### LINKED ISSUE #5580: Fix `robots.txt` feature
It seems that the `robot.txt` feature is in a sorry state right now, so I'd like to be able to expose the robots.txt feature from the control panel as in classic.

In addition, to fix these issues that we have opened.

I would remove (at least in docs and code) the feature that the one in "public" folder is used (if everybody agrees), so we reduce the way of doing things in Volto instead of exposing yet another one. It seems it's not working since long time ago either way. One could still make it work if the middleware is disabled.

- [ ] https://github.com/plone/volto/issues/2166
- [ ] https://github.com/plone/volto/issues/5386
- [ ] https://github.com/plone/volto/issues/4934
- [ ] Update [`VOLTO_ROBOTSTXT`](https://6.docs.plone.org/volto/configuration/environmentvariables.html#volto-robotstxt)

robgietema:
Fixes #5580


packages/volto/news/5580.feature
```diff
@@ -0,0 +1 @@
+Expose robots.txt setting in Volto control panel and render robots.txt based on restapi call. @robgietema
```
stevepiercy, line 1:
```suggestion
Expose robots.txt setting in Volto control panel, and render robots.txt based on REST API call. @robgietema
```

stevepiercy (changes requested):
Just a minor grammar fix to the change note, otherwise I approve the docs change only. A core developer should review this as well.

packages/volto/src/helpers/Robots/Robots.js
```diff
@@ -15,41 +17,23 @@ import { addHeadersFactory } from '@plone/volto/helpers/Proxy/Proxy';
...
     if (authToken) {
       request.set('Authorization', `Bearer ${authToken}`);
     }
     request.use(addHeadersFactory(req));
-    request.end((error, { text }) => {
+    request.end((error, { text, body }) => {
       if (error) {
         resolve(text || error);
       } else {
-        // It appears that express does not take the x-forwarded headers into
```
sneridagh, line 31: @robgietema This was added by @reebalazs at some point, can't remember why, maybe it was relevant for something. Could you please ask him?

packages/volto/src/helpers/Robots/Robots.js
```diff
@@ -15,41 +17,23 @@ import { addHeadersFactory } from '@plone/volto/helpers/Proxy/Proxy';
  */
 export const generateRobots = (req) =>
   new Promise((resolve) => {
-    const internalUrl =
```
sneridagh, line 18: Also this line, make sure that the replacement with the publicURL is fine.

sneridagh (commented):
LGTM, but only make sure that the thing removed in the middleware is ok.

packages/volto/src/helpers/Robots/Robots.js
```diff
@@ -15,41 +17,23 @@ import { addHeadersFactory } from '@plone/volto/helpers/Proxy/Proxy';
...
-        const url = `${proto}://${host}${port}`;
-        text = text.replace(internalUrl, url);
-        // Replace the sitemap with the sitemap index.
-        text = text.replace('sitemap.xml.gz', 'sitemap-index.xml');
-        resolve(text);
+        resolve(
+          body['plone.robots_txt'].replace(
+            '{portal_url}',
+            config.settings.publicURL,
+          ),
```
davisagli, line 35: This replacement should be documented. Also, I think it would be good to keep the replacement of the internal URL with the public URL, in case the robots.txt from the backend includes an (internal) sitemap URL.

davisagli (changes requested):

packages/volto/src/helpers/Robots/Robots.js
stevepiercy: Where should this be documented? I'd suggest:

- A new page under [User Manual](https://6.docs.plone.org/volto/user-manual/index.html), called "Site Configuration" with the filename `site-configuration.md`.
- [Settings reference guide](https://6.docs.plone.org/volto/configuration/settings-reference.html)

sneridagh:
@robgietema could you take a look at the suggestions and move this forward? Thanks!

packages/volto/src/helpers/Robots/Robots.js
sneridagh: @davisagli I might be wrong, I haven't checked, but it seems that what the endpoint returns is the variable ready for interpolation `{portal_url}`?

ericof (approved):
LGTM

packages/volto/src/helpers/Robots/Robots.js
davisagli: @sneridagh Okay yes, I see that that's what is there in the default value for the robots_txt setting: https://github.com/plone/plone.base/blob/fcc67703ba65f52f6dfb49ce5e3ed90291a4e285/src/plone/base/interfaces/controlpanel.py#L26

packages/volto/src/helpers/Robots/Robots.js
```diff
@@ -15,43 +17,23 @@ import { addHeadersFactory } from '@plone/volto/helpers/Proxy/Proxy';
...
-          (proto === 'https' && '' + portNum === '443') ||
-          (proto === 'http' && '' + portNum === '80')
-            ? ''
-            : `:${portNum}`;
-        // Plone has probably returned the sitemap link with the internal url.
-        // If so, let's replace it with the current one.
-        const url = `${proto}://${host}${port}`;
-        text = text.replace(internalUrl, url);
-        // Replace the sitemap with the sitemap index.
-        text = text.replace('sitemap.xml.gz', 'sitemap-index.xml');
```
davisagli, line 53: @sneridagh @ericof Removing this replacement is a breaking change. We should keep it here, or handle it instead in https://github.com/plone/plone.volto/pull/183

packages/volto/news/5580.feature
```diff
@@ -0,0 +1 @@
+Expose robots.txt setting in Volto control panel, and render robots.txt based on REST API call. @robgietema
```
davisagli, line 1: Where is it exposed in the Volto control panel? I don't see it there. I expect it needs to be removed from https://github.com/plone/volto/blob/5c88b0b767bbca7edf36467e16faf827f288c02c/packages/volto/src/config/ControlPanels.js#L79

packages/volto/src/express-middleware/robotstxt.js
```diff
@@ -4,13 +4,12 @@ import { generateRobots } from '@plone/volto/helpers/Robots/Robots';
 /*
 robots.txt - priority order:
 
-1) robots.txt in /public folder
```
davisagli, line 7: Note to self: It wasn't obvious at first, but this is only updating the comment to match existing behavior, and not changing what the code does.

davisagli (changes requested):
@sneridagh @ericof For the most part this is just updating the implementation details, but I see one breaking change that needs to be resolved, and the changelog says it does something it doesn't.

packages/volto/src/helpers/Robots/Robots.js
sneridagh: Fixed, however I think it has to go to plone.volto too. Although that can be a breaking change, since it can mean an action needed in Google Search Consoles of people.

packages/volto/news/5580.feature
sneridagh: Done, but related to the other comment: 

image

The backend answer with the .gz, which we will override it afterwards in a shady way in code (btw, we are doing it right now)

davisagli: @sneridagh @ericof Let's change the sitemap to use sitemap-index.xml in the backend setting in https://github.com/plone/plone.volto/pull/183

packages/volto/src/helpers/Robots/Robots.js
davisagli: @sneridagh No, it's not a breaking change. It was getting replaced before, and it still will be.

packages/volto/news/5580.feature
stevepiercy, line 1:
```suggestion
Expose `robots.txt` setting in Volto control panel, and render `robots.txt` based on REST API call. @robgietema
```

stevepiercy (changes requested):
Just a sprinkle of Markdown syntax.

I haven't been following the flurry of robots.txt and sitemap.xml things, so I'm not qualified to give a technical review, and I got my hands full with $WORK.

stevepiercy:
The readme-link-check failure should be temporary. I restarted it to verify.

docs/readthedocs.org will fail until I have a moment to fix it. Ignore for now.

sneridagh:
@robgietema @ericof @davisagli a final blessing, please.

packages/volto/src/config/ControlPanels.js
```diff
@@ -76,7 +76,6 @@ export const unwantedControlPanelsFields = {
     'site_favicon_mimetype',
     'exposeDCMetaTags',
     'enable_sitemap',
-    'robots_txt',
```
davisagli, line 79: @sneridagh Should we keep hiding it here if the VOLTO_ROBOTSTXT environment variable is set? Because then robots.txt is served based on the environment variable and not the setting.

davisagli (approved):
One new doubt, but I think we could merge this anyway.

PR #5585, opened 2024-01-02
Enhance `ColorPickerWidget`

sneridagh:
Now it accepts also colors definitions like:

```ts
type Color =
  | {
      label: string;
      style: Record<`--${string}`, string> & { name: string };
    }
  | {
      name: string;
      label: string;
    };

const colors: Color[] = [
  { style: { name: 'red', '--background-color': 'red' }, label: 'red' },
  {
    style: { name: 'yellow', '--background-color': 'yellow' },
    label: 'yellow',
  },
  { name: 'green', label: 'green' },
]
```

So now it supports to save an object containing a list of custom CSS properties definitions.

See documentation for more background.


giuliaghisini (approved):

docs/source/recipes/how-to-use-color-picker-widget.md
```diff
@@ -0,0 +1,139 @@
...
+      label: string;
+      style: Record<`--${string}`, string> & { name: string };
+    }
+  | {
+      name: string;
+      label: string;
+    };
+
+const colors: Color[] = [
+  { style: { name: 'red', '--background-color': 'red' }, label: 'red' },
```
davisagli, line 84: Why is `name` included as part of `style` when it's not a CSS property?

davisagli (changes requested):

docs/source/recipes/how-to-use-color-picker-widget.md
sneridagh: I know, I'm not 100% sure too... We need a string value still, for backwards compatibility since we need it here and there in the rendered HTML in the widget. It also serves a double purpose, since it also turns into an injected class name, in case you need it too. The code to comply with BW compat was also simpler, so I left it there. We could lift it to the main object, although and if you still need the class name for some reason, then manually add it.

davisagli: Oh, it's a class name? That would be more obvious if it were called `className`. It still feels a bit weird to me to have it inside `style`, but if it's needed for backwards compatibility...

sneridagh: not a class name, but as it's not a custom CSS property (it does not start with `--`), it comes for free... because of the stylewrapper (which takes all what's inside and does its thing). It's more a convention... I'd be ok and try to move it to top. I can work on this these days.

stevepiercy:
In [this comment](https://github.com/plone/volto/pull/5585#issuecomment-1873950677) is the link for the Deploy Preview correct?

https://deploy-preview-5585--plone-components.netlify.app/

I expected to see something like this:

https://6.docs.plone.org/storybook/?path=/story/edit-widgets-colorpicker--default

If it is not correct, then is this a configuration item in GitHub Workflows or Netlify?

docs/source/recipes/how-to-use-color-picker-widget.md
```diff
@@ -0,0 +1,139 @@
...
+  { name: 'green', label: 'green' },
+]
+```
+
+## Using basic definitions
+
+The basic color definition is the one that saves a string label as the widget value.
+You can use it on your own code by reading it from the resultant data and use it according your designed solution.
+
+In the case of using it along with the StyleWrapper, the value will be injected as a class name of the form:
```
stevepiercy, line 98: This sentence needs to provide an abstraction of the form or pattern. I am totally guessing that this is the correct pattern. Please correct it as needed.

```suggestion
When combined with the `StyleWrapper`, the value will be injected as a class name of the form `has--PROPERTY_NAME--PROPERTY_VALUE`:
```

docs/source/recipes/how-to-use-color-picker-widget.md
```diff
@@ -0,0 +1,139 @@
...
+
+```html
+<div class="has--backgroundColor--ee22ee">
+  ...
+</div>
+```
+
+Then you should create the CSS rules according to these injected class names.
+
+## Using custom CSS properties as color definitions
```
stevepiercy, line 108: Another child.

```suggestion
### Custom CSS properties as color definitions
```

docs/source/recipes/how-to-use-color-picker-widget.md
```diff
@@ -0,0 +1,139 @@
...
+
+The `style` key defines a set of custom CSS properties to be saved as widget value.
```
stevepiercy, line 110: I'm not sure what you intend here with "to be saved as widget value", but I think you meant this.

```suggestion
The `style` key defines a set of custom CSS properties to be added as the value to the HTML attribute `style`.
```

docs/source/recipes/how-to-use-color-picker-widget.md
```diff
@@ -0,0 +1,139 @@
...
+They will be injected by the StyleWrapper as style definitions and you will be able to use them in your CSS rules.
```
stevepiercy, line 111:
```suggestion
They will be injected by the `StyleWrapper` as style definitions, so you can use them in your CSS rules.
```

docs/source/recipes/how-to-use-color-picker-widget.md
```diff
@@ -0,0 +1,139 @@
...
+
+```html
+<div class="block teaser" style="--background-color': 'red'">
```
stevepiercy, line 114: Is `'red'` actually quoted in the output? And there's an extra `'`.

```suggestion
<div class="block teaser" style="--background-color: red">
```

docs/source/recipes/how-to-use-color-picker-widget.md
```diff
@@ -0,0 +1,139 @@
...
+```
+
+```css
+.block.teaser {
+  background-color: var(--background-color, transparent);
+}
+```
+
+The `name` key is mandatory in order to generate proper markup in the resultant HTML in both forms.
+It is also included by the StyleWrapper as a injected class name, in case you need it too in your CSS selectors.
```
stevepiercy, line 126:
```suggestion
It is also included by the `StyleWrapper` as an injected class name, in case you need it in your CSS selectors.
```

docs/source/recipes/how-to-use-color-picker-widget.md
```diff
@@ -0,0 +1,139 @@
...
+
+You can also use this selector in case it comes in handy:
```
stevepiercy, line 128:
```suggestion
You can also use this selector, where an element with class names `block` and `teaser` with a child element whose HTML attribute `style` contains the value of `--background-color`:
```

docs/source/recipes/how-to-use-color-picker-widget.md
```diff
@@ -0,0 +1,139 @@
...
+
+```css
+.block.teaser {
+  &[style*='--background-color'] {
+     padding: 20px 0;
+ }
+ ```
+
```
stevepiercy, line 136:
````suggestion

```{seealso}
See the MDN CSS Reference for selectors.

- [Attribute selectors](https://developer.mozilla.org/en-US/docs/Web/CSS/Attribute_selectors)
- [`&` nesting selector](https://developer.mozilla.org/en-US/docs/Web/CSS/Nesting_selector)
```


````

File: docs/source/recipes/how-to-use-color-picker-widget.md [hunk 10]
```diff
@@ -0,0 +1,139 @@
...
+## Use both basic and custom CSS properties definitions
+
+You can combine both basic and style definitions, it's up to you how to mix and match them.
```
stevepiercy, line 139:
```suggestion
You can combine both basic and custom CSS properties definitions.
It's up to you how to mix and match them.
```

packages/volto/news/5585.feature
```diff
@@ -0,0 +1 @@
+Enhance `ColorPickerWidget` with additional color definitions and save an object instead of a string @sneridagh
```
stevepiercy, line 1:
```suggestion
Enhanced `ColorPickerWidget` with additional color definitions, saving it as an object instead of a string. @sneridagh
```

docs/source/recipes/how-to-use-color-picker-widget.md
```diff
@@ -0,0 +1,139 @@
...
+  };
+
+  return schema;
+};
+```
+
+The widget name the color picker widget uses is `color_picker`.
+
+## Color definitions
+
```
stevepiercy, line 68: I'm not sure what version this enhancement will go in, so adjust accordingly.

````suggestion
```{versionchanged} 17.x.x
Enhanced `ColorPickerWidget` with additional color definitions, saving it as an object instead of a string.
```
````

File: docs/source/recipes/how-to-use-color-picker-widget.md [hunk 13]
```diff
@@ -0,0 +1,139 @@
...
+myst:
+  html_meta:
+    "description": "How to use the color picker widget in blocks settings and forms"
+    "property=og:description": "How to use the color picker widget in blocks settings and forms"
+    "property=og:title": "How to use the color picker widget"
+    "keywords": "Volto, Plone, frontend, React, blocks, forms, widget, color, picker"
+---
+
+# How to use the color picker widget
+
```
stevepiercy, line 11: I am guessing on the version in which the color picker was added. I think it is 16.0.0 based on this PR: https://github.com/plone/volto/pull/3269

````suggestion

```{versionadded} 16.0.0
```

````

**Review by stevepiercy — CHANGES_REQUESTED** [review]
Mostly grammar and MyST syntax fixes, but added some structure and some questions.

**Comment by sneridagh** [comment]
> In [this comment](https://github.com/plone/volto/pull/5585#issuecomment-1873950677) is the link for the Deploy Preview correct?
> 
> https://deploy-preview-5585--plone-components.netlify.app/
> 
> I expected to see something like this:
> 
> https://6.docs.plone.org/storybook/?path=/story/edit-widgets-colorpicker--default
> 
> If it is not correct, then is this a configuration item in GitHub Workflows or Netlify?

No, it's not correct. Good catch. I configured the `@plone/components` package in the monorepo to have also a build. Github should be confusing both. I guess it has to do with the monorepo setup in Netlify. I will take a look.

File: docs/source/recipes/how-to-use-color-picker-widget.md (hunk 13 above)
Comment (sneridagh) [reply docs/source/recipes/how-to-use-color-picker-widget.md:11]: IMHO, adding a versionadded is fine for new items, doing it for everything is useless, imagine we had to add in for every single existing feature... Here, this info is not relevant, since for me it's been there forever.

File: docs/source/recipes/how-to-use-color-picker-widget.md [hunk 14]
```diff
@@ -0,0 +1,139 @@
...
+Volto comes with a color picker widget that can be used in any Volto form.
```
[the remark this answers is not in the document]
sneridagh: No, they are not. They are the same. What did lead you to think that it's different?
I will add the reference.

docs/source/recipes/how-to-use-color-picker-widget.md
```diff
@@ -0,0 +1,139 @@
...
+    title: intl.formatMessage(messages.backgroundColor),
+    colors,
+    default: defaultBGColor,
+  };
+
+  return schema;
+};
+```
+
+The widget name the color picker widget uses is `color_picker`.
```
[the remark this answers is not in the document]
sneridagh: Not really, I want to express that the discriminator name is `color_picker`, and not anything else. If you don't use it, you don't get it.

docs/source/recipes/how-to-use-color-picker-widget.md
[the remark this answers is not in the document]
stevepiercy: The only place I found this as a title in Plone 6 docs is this: https://6.docs.plone.org/volto/recipes/environment-variables.html. Even then, that filename lacks "how-to-". Almost everything in docs is a "how to", and it does not really add anything.

"How to" in the `html_meta` stuff, in the narrative text, and in sub-headings (but not titles) is fine.

Are there other places in Plone 6 docs? I don't know about Plone 5 and earlier.

Anyway let's not get ahead of ourselves with docs reorganization stuff.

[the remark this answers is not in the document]
sneridagh: image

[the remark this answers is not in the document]
sneridagh: ok, let's rename this then.

docs/source/recipes/how-to-use-color-picker-widget.md
stevepiercy: I agree that adding a `version*` admonition for anything prior to 16.0.0 and its alphas would be silly at this point. I don't know when the color picker was added, but you make it sound like it's been around long before 16.x.x.

For 16.x.x, only because we still maintain it, I think we should make an effort to include it, but I'll leave it to you as a judgment call.

For 17.x.x, we should definitely make an effort.

docs/source/recipes/how-to-use-color-picker-widget.md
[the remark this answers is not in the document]
sneridagh: Renamed and added to the TOC.

sneridagh:
@stevepiercy the `emphasize-lines` does not work :( Something we can do?

stevepiercy:
@sneridagh I updated this branch with `main` in GitHub, in case you try to push something and it opens a can of nope-ass on you.

docs/source/recipes/color-picker-widget.md
```diff
@@ -0,0 +1,150 @@
...
+    "keywords": "Volto, Plone, frontend, React, blocks, forms, widget, color, picker"
+---
+
+# Color picker widget
+
+Volto comes with a color picker widget that can be used in any Volto form.
+You can combine the color picker widget with the {doc}`../blocks/block-style-wrapper` to have a powerful, yet simple way to manage color properties in your blocks.
+You can use it either in your custom block's setting's schema or enhance an existing block as follows:
+
+```js
```
stevepiercy, line 16: We need to use reStructuredText fallback syntax, instead of CommonMark shorthand, to get emphasize lines. In a future version of the theme, we will be able to use the shorthand.

````suggestion
```{code-block} js
````

File: docs/source/recipes/color-picker-widget.md [hunk 17]
```diff
@@ -0,0 +1,150 @@
...
+  html_meta:
+    "description": "How to use the color picker widget in blocks settings and forms"
+    "property=og:description": "How to use the color picker widget in blocks settings and forms"
+    "property=og:title": "How to use the color picker widget"
+    "keywords": "Volto, Plone, frontend, React, blocks, forms, widget, color, picker"
+---
+
+# Color picker widget
+
+Volto comes with a color picker widget that can be used in any Volto form.
```
stevepiercy, line 12: Insert a sentence with a link to the Storybook color picker demo.

```suggestion
Volto comes with a color picker widget that can be used in any Volto form.
You can [try a demo of the default color picker](https://6.docs.plone.org/storybook/?path=/story/edit-widgets-colorpicker--default).
```

docs/source/recipes/how-to-use-color-picker-widget.md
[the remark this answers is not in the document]
stevepiercy: Added a comment in a new review.

stevepiercy:
> @stevepiercy the `emphasize-lines` does not work :( Something we can do?

Yeah, see my review, once I finish it. I'll commit it as a fix.

docs/source/recipes/color-picker-widget.md
```diff
@@ -0,0 +1,150 @@
...
+    title: intl.formatMessage(messages.backgroundColor),
+    colors,
+    default: defaultBGColor,
+  };
+
+  return schema;
+};
+```
+
+The widget name the color picker widget uses is `color_picker`.
```
stevepiercy, line 65: Is this better? When you use "uses", it makes it sound like its name can be set to any value, but I don't think that is your intent.

```suggestion
The color picker widget's discriminator is `color_picker`.
```

docs/source/recipes/color-picker-widget.md
```diff
@@ -0,0 +1,150 @@
...
+
+## Color definitions
+```{versionchanged} 17.9.0
```
stevepiercy, line 68: Extra blank line for author legibility.

````suggestion

```{versionchanged} 17.9.0
````

File: docs/source/recipes/color-picker-widget.md [hunk 20]
```diff
@@ -0,0 +1,150 @@
...
+Enhanced `ColorPickerWidget` with additional color definitions, saving it as an object instead of a string.
+```
```
stevepiercy, line 70: Extra blank line for author legibility.

````suggestion
```

````

**Review by stevepiercy — CHANGES_REQUESTED** [review]
This has nitpicky things, and another stab at that discriminator sentence. Then it should be good to go.

**Comment by sneridagh** [comment]
@stevepiercy I was looking into the Netlify config, I haven't found anything related to having several builds in the same repo and the Netlify GH bot... it seems it only posts the last one...

**Comment by stevepiercy** [comment]
@sneridagh I will poke around inside Netlify and GitHub actions and Makefile. There must be something in there that triggers it.

**Comment by stevepiercy** [comment]
@sneridagh needs a change log entry.

**Comment by stevepiercy** [comment]
@sneridagh also unit tests fail in CI.

File: docs/source/recipes/color-picker-widget.md (hunk 17 above)
Comment (sneridagh) [reply docs/source/recipes/color-picker-widget.md:12]: Ah sorry I forgot.

**Comment by stevepiercy** [comment]
@sneridagh I found it. There is another site in Netlify, https://app.netlify.com/sites/plone-components/overview, that gets built on every pull request in Volto. I think we want it to build only when something in its package builds, not within the entire volto repo, correct?

**Comment by sneridagh** [comment]
@stevepiercy I will take a look at the unit tests I still want to polish the code a bit.

**Comment by sneridagh** [comment]
@stevepiercy is there a way to filter when to build it?

**Comment by stevepiercy** [comment]
> @stevepiercy is there a way to filter when to build it?

Yup, similar to what we do for Volto docs, we can do a diff. Let's do this in a separate PR, because I don't want this one to become a victim of feature creep. I also want to add the Storybook build for docs to be reachable at `volto-docs-preview-url/storybook` so we can begin to use arbitrary relative links to it. I'll do both in a separate PR.

File: docs/source/recipes/how-to-use-color-picker-widget.md (hunk 1 above)
Comment (sneridagh) [reply docs/source/recipes/how-to-use-color-picker-widget.md]: @davisagli I've moved the `name` key to the top, it makes more sense as you said, and the benefits are not that much in the other form. Please, take a look!

Comment (sneridagh) [reply docs/source/recipes/how-to-use-color-picker-widget.md]: @giuliaghisini I've modified a bit the definitions as per David's request, please take another quick look!

File: docs/source/recipes/color-picker-widget.md [hunk 21]
```diff
@@ -0,0 +1,161 @@
...
+  html_meta:
+    "description": "How to use the color picker widget in blocks settings and forms"
+    "property=og:description": "How to use the color picker widget in blocks settings and forms"
+    "property=og:title": "How to use the color picker widget"
+    "keywords": "Volto, Plone, frontend, React, blocks, forms, widget, color, picker"
+---
+
+# Color picker widget
+
+Volto comes with a color picker widget that can be used in any Volto form.
```
davisagli, line 12: People might be expecting a color widget that allows picking any color, so we might want to clarify that this one picks from a preset list of colors.

docs/source/recipes/color-picker-widget.md
```diff
@@ -0,0 +1,161 @@
...
+You can [try a demo of the default color picker](https://6.docs.plone.org/storybook/?path=/story/edit-widgets-colorpicker--default).
+You can combine the color picker widget with the {doc}`../blocks/block-style-wrapper` to have a powerful, yet simple way to manage color properties in your blocks.
+You can use it either in your custom block's setting's schema or enhance an existing block as follows:
```
davisagli, line 15:
```suggestion
You can use it either in your custom block's styles schema or enhance an existing block as follows:
```

docs/source/recipes/color-picker-widget.md
```diff
@@ -0,0 +1,161 @@
...
+The color picker widget's discriminator is `color_picker`.
+
+## Color definitions
+
+```{versionchanged} 17.9.0
+Enhanced `ColorPickerWidget` with additional color definitions, saving it as an object instead of a string.
+```
+
+The `colors` property of the widget controls which colors are available to choose in the widget.
+This is the signature of the object:
```
davisagli, line 75: The example below includes both the type signature and an example (which is great, but makes this sentence inaccurate)

docs/source/recipes/color-picker-widget.md
```diff
@@ -0,0 +1,161 @@
...
+  {
+    name: 'green',
+    label: 'green'
+  },
+]
+```
+
+### Basic color definition
+
+The basic color definition is the one that saves a string label as the widget value.
```
davisagli, line 109: Does this come from the `label` or the `name` above?

packages/volto/src/components/manage/Widgets/ColorPickerWidget.tsx
```diff
@@ -44,21 +69,29 @@ const ColorPickerWidget = (props) => {
...
+                  if (color.style !== undefined) {
+                    colorValue = color.style;
+                  } else {
+                    colorValue = color.name;
+                  }
                   return (
                     <Button
-                      key={id + color.name}
-                      className={color.name}
+                      key={id + colorName}
```
davisagli, line 81: I think we could just use the index in `colors` here and avoid requiring `name`. It seems unlikely that they will be reordered frequently.

docs/source/recipes/color-picker-widget.md
sneridagh: I improved it in the next sentence.

packages/volto/src/components/manage/Widgets/ColorPickerWidget.tsx
sneridagh: @davisagli so you'd change the `style` use case branch? as in:

```ts
type Color =
  | {
      name: undefined;
      label: string;
      style: Record<`--${string}`, string>;
    }
  | {
      name: string;
      label: string;
      style: undefined;
    };
```

davisagli: @sneridagh yeah, that's the suggestion, but I don't feel strongly about it

packages/scripts/news/5585.feature
```diff
@@ -0,0 +1 @@
+Support for TS/TSX files in i18n machinery @sneridagh
```
stevepiercy, line 1:
```suggestion
Added support for TS/TSX files in i18n machinery. @sneridagh
```

stevepiercy (changes requested):
Looking good: https://deploy-preview-5585--volto.netlify.app/recipes/color-picker-widget.html

Just a minor suggestion to the news item, and it's GTG.

packages/volto/src/components/manage/Widgets/ColorPickerWidget.tsx
sneridagh: @davisagli I don't know, I feel like this is more consistent, when we deal with choices, we always have a "display name" (label) and a "token" (value). The token is invariable, but the display name could be i18n'd. If at some point you need to identify the color entry, you will have to equal the style property (which you could still do...) Answering myself while writing :) 

Anyways, I don't have a strong opinion either at this point. 

/cc @tiberiuichim @ichim-david @giuliaghisini @pnicolli any opinions about this?

tiberiuichim (approved):

stevepiercy (approved):

davisagli (approved):

PR #5592, opened 2024-01-02
Clarify how CSS properties work

### LINKED ISSUE #5591: Clarify how CSS properties work
I read the above two sentences a half-dozen times, and I am not sure why both are needed.

```suggestion
As you can see, the custom CSS property applies only within the `div` in which it is defined.
```

_Originally posted by @stevepiercy in https://github.com/plone/volto/pull/5581#discussion_r1438402042_

stevepiercy:
Fixes #5591.

### DIFF
--- a/docs/source/blocks/block-style-wrapper.md
+++ b/docs/source/blocks/block-style-wrapper.md
@@ -218,8 +218,11 @@ Finally, assuming that you select the default value for the {guilabel}`Aspect Ra
 </div>
 ```
 
-The custom CSS property definition will only apply within the div that it's defined.
-As you can see, the custom CSS property applies only within the `div` in which it is defined.
+As you can see, the custom CSS declaration applies only in the `div` where you inject the property.
+
+```{seealso}
+[CSS basics](https://developer.mozilla.org/en-US/docs/Learn/Getting_started_with_the_web/CSS_basics)
+```
 
 If you want to use it in your custom components, you need to apply it in the root of your block's view component as follows:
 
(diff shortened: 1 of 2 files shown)


docs/source/blocks/block-style-wrapper.md
```diff
@@ -218,8 +218,11 @@ Finally, assuming that you select the default value for the {guilabel}`Aspect Ra
 </div>
 ```
 
-The custom CSS property definition will only apply within the div that it's defined.
-As you can see, the custom CSS property applies only within the `div` in which it is defined.
+As you can see, the custom CSS declaration applies only in the `div` where you inject the property.
+
+```{seealso}
+[CSS basics](https://developer.mozilla.org/en-US/docs/Learn/Getting_started_with_the_web/CSS_basics)
```
ichim-david, line 224:
```suggestion
[CSS custom properties basics](https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties)
```

sneridagh:
Thanks @ichim-david !

@stevepiercy Now better?

docs/source/blocks/block-style-wrapper.md
```diff
@@ -218,8 +218,11 @@ Finally, assuming that you select the default value for the {guilabel}`Aspect Ra
 </div>
 ```
 
-The custom CSS property definition will only apply within the div that it's defined.
-As you can see, the custom CSS property applies only within the `div` in which it is defined.
+As you can see, the custom CSS declaration applies only in the `div` where you inject the property.
```
ichim-david, line 221: @stevepiercy the reason it was mentioned as CSS property definition instead of the custom CSS declaration is due to how this feature is called in CSS.

I would write as 
The CSS property will be applied on the `div` where you inject the property.

This is because this property could affect also its children, you don't know where inside that section someone has a var() call that makes use of it

stevepiercy: Thanks for the clarification and the assist. Let's merge this!

PR #5595, opened 2024-01-03
Attempt to build the volto/components Storybook only when something c…

stevepiercy:
…hanges in its directory

See https://github.com/plone/volto/pull/5585#issuecomment-1874457512 and ensuing discussion.

See also https://docs.netlify.com/configure-builds/ignore-builds/#app

### DIFF
--- a/packages/components/netlify.toml
+++ b/packages/components/netlify.toml
@@ -1,0 +1,2 @@
+[build]
+  ignore = "git diff --quiet $CACHED_COMMIT_REF $COMMIT_REF ."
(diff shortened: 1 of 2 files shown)


stevepiercy:
@sneridagh I don't know whether this will actually work until it is merged to `main` and it has a cached commit ref, but this is the basic idea. We can look at the Netlify logs (for example, https://app.netlify.com/sites/plone-components/deploys/65953f2fe56175000813fb55#L11) for this site to see whether it does what we want, and fiddle around with it. Sometimes we have to use lots of `echo` commands to troubleshoot.

sneridagh (approved):

PR #5608, opened 2024-01-10
Fix autopopulated value of facet when settings the value for another (#5541)

sneridagh:


iFlameing (approved):
Looks perfectly fine. But checking it locally.

iFlameing:
@sneridagh checked locally everything is fine and working as expected :)

PR #5615, opened 2024-01-11
Fix wrong params url when removing the search text from search block

### LINKED ISSUE #5614: Wrong params url when removing the search text from search block.

iFlameing:
Fixes #5614 .

@sneridagh This is the same fix for the auto-populated issue. We just missed one case where we typed in the search box of searchblock and removed it.


ichim-david:
I'll test this tomorrow morning, alongside https://github.com/plone/volto/pull/5262
It would be great to get both fixes merged soon if we are at search block fixes.

ichim-david (commented):
@iFlameing @sneridagh 
I tested locally this branch and I tried a demo on the plone demo.
Take into consideration that my experience with the search block is limited so I don't know all of the patches that went into it throughout time.

I tried to replicate the cypress test and I have the following
things to say:

https://demo.plone.org/search-block-test?query=%5B%7B%22i%22%3A%22portal_type%22%2C%22o%22%3A%22paqo.selection.any%22%2C%22v%22%3A%5B%22Document%22%2C%22Event%22%2C%22Folder%22%5D%7D%2C%7B%22i%22%3A%22portal_type%22%2C%22o%22%3A%22paqo.list.contains%22%2C%22v%22%3A%5B%22Document%22%2C%22Event%22%2C%22Folder%22%5D%7D%5D&sort_order=ascending

To me the URL parameters are modified just fine when clicking on the
clear input even before this code.
However even though the URL does not contain the searchText to Event
there is no loading of results afterwards.

[loading](https://github.com/plone/volto/assets/152852/ec7dbbda-6c4c-4823-8087-d514ed313de4)

Locally this doesn't happen with your branch but neither does it happen with the main branch.

As such I'm still a bit confused on what this change fixes. 
I am curious as to why you get the loading hickup on plone demo
vs locally with 18.

I will try a test also with Volto 17 locally and see if I can replicate
the demo issue sometime today but my initial review is:
doesn't seem to harm but doesn't seem to fix either :)

fredvd:
@iFlameing Status?  (working on a search block implementation and checking for bugs, found this one)

PR #5621, opened 2024-01-14
App extras exceptions

dobri1408:
In situations where we do not want the appExtras rule to apply universally to all routes matched by a certain rule, the addition of an exclude property could be beneficial. This property would provide a way to specify exceptions to the rule. For instance:
`{
   match: '*',
   exclude: '/**/diff',
   component: RemoveSchema,
}`
To address this scenario, where the goal is to apply a rule to all pages except for the "history diff" page, incorporating an exclude property into the rule configuration is an effective solution.

### DIFF
--- a/packages/volto/src/components/theme/AppExtras/AppExtras.jsx
+++ b/packages/volto/src/components/theme/AppExtras/AppExtras.jsx
@@ -8,6 +8,8 @@ const AppExtras = (props) => {
   const { pathname } = props;
   const active = appExtras
     .map((reg) => {
+      const excluded = matchPath(pathname, reg.exclude || "");
+      if (excluded) return null;
       const match = matchPath(pathname, reg.match);
       return match ? { reg, match } : null;
     })


dobri1408:
Also, if this is merged, can we backport it to version 16.x?

stevepiercy:
This needs narrative documentation. See https://6.docs.plone.org/volto/recipes/appextras.html.

I would like to see an exclude code example, modeled on the default code example.

Additionally, we need to include a placeholder for this admonition, especially if it gets backported.

````
```{versionchanged}17.x.x and 16.x.x
Added the `exclude` property.
```
````

Side note, that page has some syntax and grammar issues that could be addressed at the same time.

**Review by ichim-david — CHANGES_REQUESTED** [review]
https://github.com/plone/volto/pull/5621/files#diff-76050b5a1b7f3811d861cba8076fb99be2f8162e0649b91432f7793702071078R3

The matching rule isn't correct, look at the test output, now every test returns null which means that your condition always renders true.

**Comment by dobri1408** [comment]
@ichim-david, you're right, the default value was incorrect.

**Review by ichim-david — CHANGES_REQUESTED** [review]
@dobri1408 you need to add a test for this feature

**Review by ichim-david — CHANGES_REQUESTED** [review]
@

File: packages/volto/src/components/theme/AppExtras/AppExtras.test.jsx [hunk 1]
```diff
@@ -46,6 +46,13 @@ beforeAll(() => {
         <div className="something">{JSON.stringify(props.match)}</div>
       )),
     },
+    {
+      match: '/test-exclude',
```
ichim-david, line 50:
```suggestion
      match: '/frontpage',
      exclude: '/frontpage/images',
      component: jest.fn((props) => (
        <div className="frontpage-content">{JSON.stringify(props.match)}</div>
      )),
    },
```
@dobri1408 you need to redo the test to make it more clear on what your feature does.
You want to show that you can load the content from https://github.com/plone/volto/pull/5621/files#diff-ab6967b7b0acf3c2cb23aea6a5252475fd6c058af52d53b6d1c8fc7a00b1a1b0R53
when the pathname is /test-exclude but when you are inside /text-exclude/edit
then you will get only the content of the match '' everything since that section excludes
the /test-exclude match.

I added a suggestion for you to take into consideration.

ichim-david (changes requested):

ichim-david:
> This needs narrative documentation. See https://6.docs.plone.org/volto/recipes/appextras.html.
> 
> I would like to see an exclude code example, modeled on the default code example.
> 
> Additionally, we need to include a placeholder for this admonition, especially if it gets backported.
> 
> ```
> ```{versionchanged}17.x.x and 16.x.x
> Added the `exclude` property.
> ```
> ```
> 
> Side note, that page has some syntax and grammar issues that could be addressed at the same time.

@dobri1408 don't forget to also do this step that @stevepiercy mentioned that you need todo

sneridagh:
out of curiosity, the react-router-dom `matchPath` support regex, right? Can't we build a regex that does this? (I am not a regex nerd, but still...)

dobri1408:
Hi @sneridagh! This was my first thought, but unfortunately no, it doesn't support regex so we have to introduce this new property.

sneridagh:
@dobri1408 please perform the suggestions from @ichim-david and @tiberiuichim I think they are valuable.

dobri1408:
> This needs narrative documentation. See https://6.docs.plone.org/volto/recipes/appextras.html.
> 
> I would like to see an exclude code example, modeled on the default code example.
> 
> Additionally, we need to include a placeholder for this admonition, especially if it gets backported.
> 
> ```
> ```{versionchanged}17.x.x and 16.x.x
> Added the `exclude` property.
> ```
> ```
> 
> Side note, that page has some syntax and grammar issues that could be addressed at the same time.

Hi @stevepiercy, how do I write on this https://6.docs.plone.org/volto/recipes/appextras.html. I mean where do I have to make the change?

docs/source/recipes/appextras.md
```diff
@@ -1,14 +1,18 @@
...
   html_meta:
-    "description": "The `AppExtras` component is a general use insertion point for general purpose code spanning the whole application or for third party services code."
-    "property=og:description": "The `AppExtras` component is a general use insertion point for general purpose code spanning the whole application or for third party services code."
-    "property=og:title": "AppExtras component"
-    "keywords": "Volto, Plone, frontend, React, app extra component"
+    'description': 'The `AppExtras` component is a general use insertion point for general purpose code spanning the whole application or for third party services code.'
+    'property=og:description': 'The `AppExtras` component is a general use insertion point for general purpose code spanning the whole application or for third party services code.'
+    'property=og:title': 'AppExtras component'
+    'keywords': 'Volto, Plone, frontend, React, app extra component'
 ---
```
stevepiercy, line 8: Please configure your edit not to rewrite " to ', and revert the above changes.

docs/source/recipes/appextras.md
```diff
@@ -1,14 +1,18 @@
...
 
 # AppExtras component
 
+```{versionchanged}17.x.x and 16.x.x
```
stevepiercy, line 12: We need explicit version numbers. Let's use 17.11.6 and 16.30.3, based on https://www.npmjs.com/package/@plone/volto?activeTab=versions. @sneridagh will need to verify upon release.

````suggestion
```{versionchanged} 17.11.6 and 16.30.3
````

File: docs/source/recipes/appextras.md [hunk 4]
```diff
@@ -23,14 +27,16 @@ Or you can use the new key of `config.settings`, the `appExtras`. This is
 a list of registrations, each registration is an object with:
 
 - `match`: The `match` key is for objects compatible with [react-router's
```
stevepiercy, line 29: Let's combine lines, and comply with Plone's one sentence per line for narrative documentation style.

```suggestion
- `match`: The `match` key is for objects compatible with [react-router's `matchPath`](https://v5.reactrouter.com/web/api/matchPath). It can be either a string or a [`match`](https://v5.reactrouter.com/web/api/match) object.
```

docs/source/recipes/appextras.md
```diff
@@ -23,14 +27,16 @@ Or you can use the new key of `config.settings`, the `appExtras`. This is
...
-matchPath](https://v5.reactrouter.com/web/api/matchPath), so it can be either
-a simple string or a match object.
+  matchPath](https://v5.reactrouter.com/web/api/matchPath), so it can be either
```
stevepiercy, line 30: Merged with above line.

```suggestion
```

docs/source/recipes/appextras.md
```diff
@@ -23,14 +27,16 @@ Or you can use the new key of `config.settings`, the `appExtras`. This is
...
+  a simple string or a match object.
```
stevepiercy, line 31: Merged with above line.

```suggestion
```

packages/volto/news/5621.feature
```diff
@@ -0,0 +1 @@
+Added exclude Property to Rule Configuration: Enhanced flexibility in rule application by introducing an exclude property. This new feature allows for specifying exceptions to rules that are applied to all routes. It's particularly useful for cases where a rule should apply universally, except for specific pages. @dobri1408
```
stevepiercy, line 1: Revise to correct property name, and make concise.

```suggestion
Added the `ignore` property to allow exceptions to rules that are applied to all routes. @dobri1408
```

docs/source/recipes/appextras.md
```diff
@@ -23,14 +27,16 @@ Or you can use the new key of `config.settings`, the `appExtras`. This is
...
+- `ignore`: The `ignore` key is for routes compatible with the match property from above and
```
stevepiercy, line 32: I'm not certain about the wording being 100% compliant with what the code actually does, so someone should double-check both the `ignore` and `match` properties rewording.

```suggestion
- `ignore`: The `ignore` key is for routes compatible with the [react-router's `matchPath`](https://v5.reactrouter.com/web/api/matchPath) property, and ignores all the results in the `ignore` rule.
  It can be either a string or a [`match`](https://v5.reactrouter.com/web/api/match) object.
```

docs/source/recipes/appextras.md
```diff
@@ -54,10 +60,17 @@ export settings = {
...
+       match: '/frontpage',
+       ignore: '/frontpage/images',
+      },
+      component: FrontPage
+    },
   ]
 }
 ```
 
 You can use this to render, for example, `<Portal>` components to be inserted
```
stevepiercy, line 75:
```suggestion
The above example will apply the rules to their respective routes.

-   Insert a Google Analytics tag everywhere.
-   Insert the `WordClouds` component to the `/blogs` route.
-   Insert the `ExtraToolbarButton` component when editing objects under the path `/blog/**/edit`.
-   Insert the `FrontPage` component to the `frontpage` route, except under `/frontpage/images`.
```

docs/source/recipes/appextras.md
```diff
@@ -54,10 +60,17 @@ export settings = {
...
-somewhere in the website.
+somewhere on the website.
```
stevepiercy, line 76: Replaced with an explanation of what the above example actually does.

```suggestion
```

docs/source/recipes/appextras.md
```diff
@@ -23,14 +27,16 @@ Or you can use the new key of `config.settings`, the `appExtras`. This is
 a list of registrations, each registration is an object with:
 
 - `match`: The `match` key is for objects compatible with [react-router's
-matchPath](https://v5.reactrouter.com/web/api/matchPath), so it can be either
-a simple string or a match object.
+  matchPath](https://v5.reactrouter.com/web/api/matchPath), so it can be either
+  a simple string or a match object.
+- `ignore`: The `ignore` key is for routes compatible with the match property from above and
+  ignores all the results that are also compatible with the ignore rule.
```
stevepiercy, line 33: Merged with above line.

```suggestion
```

stevepiercy (changes requested):
These suggestions might need more revision, and should be reviewed by someone who can make sure the rewording I suggested aligns with the code functionality. Thank you!

stevepiercy (changes requested):
One more change, after checking the preview on Netlify.

stevepiercy (approved):
Thank you!

sneridagh (approved):

ichim-david (approved):

ichim-david:
@sneridagh this is ready to merge for some time and everyone gave their approval, can we merge this or do you want to merge something else before?

PR #5647, opened 2024-01-17
Fix blocks chooser index and add a bit of breath to the left in the s…

sneridagh:
…earch input.


ichim-david (approved):

ichim-david:
I've reverted my changes to the block chooser search input made in 
https://github.com/plone/volto/pull/5647/commits/34e77db6ab94942b13d9ba54d87dd15383e9c621
 since I want to modify the component to get rid of the inline styles which modify the test snapshots
and thus is deemed a breaking change while these fixes are meant to be backported to 17.x.x branch as well.

I will open a separate pull request with my changes which are meant then for the main branch

sneridagh:
@ichim-david ok! Merging this then!

PR #5649, opened 2024-01-17
Fix order of preference in addons-registry for the theme

sneridagh:


stevepiercy:
@sneridagh Does this change require a change in the documentation? https://6.docs.plone.org/volto/configuration/how-to.html#the-configuration-registry

> Both the main project and individual add-ons can extend Volto's configuration registry. First the add-ons configuration is applied, in the order they are defined in package.json, then finally the project configuration is applied. Visualized like a pipe would be:
>
> > Default Volto configuration -> Add-on 1 -> Add-on 2 -> ... -> Add-on n -> Project

sneridagh:
Nope it's not about this, I checked and there's no explicit mention about the order, although the feature is covered in how to setup a theme.

PR #5658, opened 2024-01-18
Add image preview in object browser widget.

robgietema:


sneridagh:
@robgietema nice! Can we improve the popover so you can actually click on it?

https://github.com/kitconcept/kitconcept.com/issues/53#issuecomment-1870333371 and the other suggestions from comments?

Thanks!

robgietema:
The tooltip we have now is there so you can see what the path of the object is you are linking to. When this is clickable you leave the page which will add more UX problems then it solves. I don't see the usecase of going to the object from here.

davisagli:
@robgietema I want a way to open the linked item in a new tab all the time.

robgietema:
I still don't get the usecase, please explain.

stevepiercy:
@robgietema I agree with @davisagli and @sneridagh, where I would right-click or CTRL-click on a link to open it in a new browser window. This would be for convenience.

One implementation would be to use the `target="_blank"` attribute in the HTML. I hate those because I want control of what opens a new window, yet I understand that some users don't know how to right- or CTRL-click on links and it might be appropriate in this context so users don't lose where they are.

Does that make sense?

sneridagh:
@robgietema the use case is, I'd like to check something from the linked image, so I want to open it in a new window. This use case is quite common. Maybe you want to modify metadata or even replace the image (without having to create and link a new one).

The tooltip now it's fine because it's informative. It was supposed to solve this use case (because it's a link), but you are not able to reach it, because it closes before. Like David, I don't care if the link is in the tooltip (which I think it's still useful) but elsewhere (maybe in another icon besides or below the (x)?

robgietema:
I understand what you are all saying but I still don't see this as a wanted feature. Feel free to create another issue for this since it is out of scope for this issue.

sneridagh (approved):
Ok, then LGTM. I will create another ticket for the other request.

PR #5667, opened 2024-01-19
Fix `/links-to-item` should be a protected route

### LINKED ISSUE #5666: `links-to-item` should be a protected route.

iFlameing:
Fixes #5666


sneridagh (approved):

davisagli (commented):
@sneridagh I don't really like this solution. It could be a valid use case to want to enable access to view the links for a user who is not able to edit. Doing it this way puts a handicap on Plone's flexible permissions system. Not urgent, but let's discuss the right way to do it.

Issue #5688, opened 2024-01-23
Listing Block doesn't render initial results in SSR


ichim-david:
the getAsyncData calls to retrieve values with id while the values for the query search results is made this way
const subrequestID = content?.UID ? `${content?.UID}-${id}` : id;

ichim-david:
I have a fix for this already, working on adding also some cypress tests

PR #5689, opened 2024-01-23
Listing Block render of initial results in SSR

ichim-david:
Fixes https://github.com/plone/volto/issues/5688 by ensuring that the call to getQueryStringResults is done using the subrequestID construct made possible by passing in content to getAsyncData from App.jsx


ichim-david:
@sneridagh you can easily test the fix by disabling javascript and looking at the page contents of a page with a listing block before this branch and after using this branch

packages/volto/news/5688.bugfix
```diff
@@ -0,0 +1 @@
+Fixed listing SSR rendering by sending subrequestId instead of id only within getAsyncData the same way it happens when calling getQueryStringResults directly. @ichim-david
```
stevepiercy, line 1:
```suggestion
Fixed listing SSR rendering by sending `subrequestId` instead of `id` only within `getAsyncData`, similar to calling `getQueryStringResults` directly. @ichim-david 
```

stevepiercy (changes requested):
Tweak grammar and sprinkle a little MyST syntax.

tiberiuichim:
The listing block works in 2 scenarios:

- as folder contents (when it's just dropped into a page, without any settings done to it)
- normal, query based

I think it's meaningful to test both scenarios.

ichim-david:
> The listing block works in 2 scenarios:
> 
>     * as folder contents (when it's just dropped into a page, without any settings done to it)
> 
>     * normal, query based
> 
> 
> I think it's meaningful to test both scenarios.

@tiberiuichim added extra test in https://github.com/plone/volto/pull/5689/commits/d4e568fa85e86f6e8d164e9d28190730e85b54e5

ichim-david:
Hmm, the default listing works now with SSR but I want to test further with the other templates as well because on Volto 16 I've discovered that our Carousel variation from volto-listing-block is still seen as loading only.

As such this is not yet ready for merge as I might need to tweak extra code after I test every variation possible.

ichim-david:
> Hmm, the default listing works now with SSR but I want to test further with the other templates as well because on Volto 16 I've discovered that our Carousel variation from volto-listing-block is still seen as loading only.
> 
> As such this is not yet ready for merge as I might need to tweak extra code after I test every variation possible.

I tried with plain Volto every available variation and it work fine.
I will add SSR test for each template not just the default template tomorrow.

One thing that I will be curious to check is with a Volto project using listing block enhancers such as the ones from kitconcept and eea.

packages/volto/src/components/manage/Blocks/Listing/getAsyncData.js
```diff
@@ -2,12 +2,15 @@ import { getQueryStringResults } from '@plone/volto/actions';
...
 export default function getListingBlockAsyncData(props) {
-  const { data, path, id, dispatch, blocksConfig, content } = props;
+  const { data, path, location, id, dispatch, blocksConfig, content } = props;
 
   const { resolvedExtensions } = resolveBlockExtensions(data, blocksConfig);
 
   const subrequestID = content?.UID ? `${content?.UID}-${id}` : id;
 
+  // retrieve the current page from the location querystring
+  const currentPage = location?.query?.split(/page=(\d+)/)[1] || 1;
```
tiberiuichim: Maybe use querystring to parse the query? This feels dirty, we already import the qs package.

ichim-david: > Maybe use querystring to parse the query? This feels dirty, we already import the qs package.

Fake news, don't review a draft pr or your comments might end up stale as can be seen in this commit 
https://github.com/plone/volto/pull/5689/commits/ad68293b65870772bd38fd008ef01103f197e59f

ichim-david:
@tiberiuichim @sneridagh ready for testing.
I've fixed not only the initial page that wasn't rendering when the listing contained a querystring but also the
results when we have pagination using the page parameter.

I also cleaned up the testing by using commands that were defined and added the inInHTML helper which
tests if the markup is in the page render for each test we have for listing block.

tiberiuichim (approved):

ichim-david:
Regarding this commit https://github.com/plone/volto/pull/5689/commits/70beca094c4efca87d4153ab5ff3b058da291323
Once every 7-10 tries the test failed as seen in these 2 screenshots
listing-block-failure

path-is-null

I couldn't reproduce the error after I made that change, maybe it's a fluke and we might need to slightly adjust again the cypress command, what I know is that I had todo several types on other pull requests the run failed jobs because of this failure from the listing block tests

sneridagh (approved):

PR #5695, opened 2024-01-24
Backport i18n polishing (#5542) and update locales

sneridagh:


sneridagh:
@erral is it safe to backport this to 17?

sneridagh:
@erral merging because if not, with latest @plone/scripts the sites are borked out. :(

erral:
it should be safe

PR #5701, opened 2024-01-26
Fixed creation of both a slate block and a (proper) image block on pasting image directly from clipboard

### LINKED ISSUE #5697: Pasting an image directly from the clipboard in a slate block triggers the creation of both a slate block and a (proper) image block.
**Describe the bug**
Pasting an image directly from the clipboard in a slate block triggers the creation of both a slate block and a (proper) image block.

**To Reproduce**

1. Edit any content with blocks
2. Click on an empty slate block
3. Copy an image
4. Paste directly from the clipboard into a slate block
5. Two blocks, one pure slate and an image block are created.

**Expected behavior**
Only the image block is created.

AnshulKahar2729:
Fixes #5697 

Description : Earlier on the pasting the images in the blocks, rendering two images on the same time.


packages/volto-slate/src/editor/plugins/Image/render.jsx
```diff
@@ -1,22 +1,11 @@
...
-    maxWidth: '100%',
-    maxHeight: '20em',
-    boxShadow: selected && focused ? '0 0 0 2px blue' : 'none',
-  };
+  const { attributes, children } = props;
 
   return (
     <span {...attributes} style={{ display: 'inline-block' }}>
       {children}
-      <img alt="" src={element.url} style={style} />
```
sneridagh, line 19: @AnshulKahar2729 We can't do this. This is not the way to fix this issue. We have to dig deeper and find why it's emitting the slate image (which we don't want) and make that it does not emmit it. By removing the markup for the slate image is not an acceptable "fix".

 Do you agree with me, right?

sneridagh (changes requested):

saloni0419:
is this issue still open #5701 @sneridagh

stevepiercy:
@saloni0419 please read and follow [Contributing to Plone](https://6.docs.plone.org/contributing/index.html) and [First-time contributors](https://6.docs.plone.org/contributing/first-time.html), especially [Things not to do](https://6.docs.plone.org/contributing/first-time.html#things-not-to-do).

Using your super power of observation, you may see that the green "Open" icon reflects its status, and that by virtue of its URL and the "Pull requests" tab being selected, this is a pull request, not an issue. See [GitHub docs](https://docs.github.com/).

packages/volto-slate/src/editor/plugins/Image/render.jsx
AnshulKahar2729: Ok, I got the problem, we need to fix the function which creates extra block rather than removing it manually.

ichim-david:
@AnshulKahar2729 I am closing this pull request, as Victor mentioned this is not as easy as removing the image insertion from Slate. I suggest you study the problem more deeply and raise a pull request only if you've solved it and your work passes the code quality and testing commands.

Issue #5693, opened 2024-01-24
Use openExternalLinkInNewTab config setting for all instances of links with target blank 


ichim-david:
In pull request https://github.com/plone/volto/pull/4379 Rob added the **openExternalLinkInNewTab** config option.

We have a pull request now to document this option https://github.com/plone/volto/pull/5678

What is missing is for this option to be used everywhere in Volto where we have UniversalLinks or normal links
that would open with target="_blank". 
This way you can configure per project if you want to open the external links in a new tab or not

singhcoder694:
@ichim-david as per my understanding we need to incorporate same functionality of choice of **openExternalLinkInNewTab**
to all the links which are set to target="_blank" in Volto code.... Am I getting it right?

PR #5708, opened 2024-01-28
Open link in new tab

singhcoder694:
Changed all the links in Repo from only target='_blank' to target={settings.openExternalLinkInnewTab ? '_blank' : '_self'} as per issue #5693

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,19 +10,8 @@ SHELL:=bash
 MAKEFLAGS+=--warn-undefined-variables
 MAKEFLAGS+=--no-builtin-rules
 
-# Project settings - Update also `packages/volto/Makefile` to keep both in sync
-
-INSTANCE_PORT=8080
-DOCKER_IMAGE=plone/server-dev:6.0.9
-DOCKER_IMAGE_ACCEPTANCE=plone/server-acceptance:6.0.9
-KGS=
-NODEBIN = ./node_modules/.bin
-SCRIPTSPACKAGE = ./packages/scripts
-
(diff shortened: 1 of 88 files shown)


stevepiercy:
@singhcoder694 all required statuses must pass. See [Change log entry](https://6.docs.plone.org/contributing/index.html#change-log-entry) and [Code quality](https://6.docs.plone.org/volto/contributing/index.html#code-quality). We don't review failing pull requests.

ichim-david:
@singhcoder694 you don't modify the *.test.jsx.snap files, that happens by doing the following:
1. run the jest tests locally by running in the root of the volto repo:
   pnpm --filter @plone/volto test
2. once you get failures you will get the option to hit u to update the tests
3. look at every test or code analysis failure and run the required commands to test and fix
https://github.com/plone/volto/blob/main/packages/volto/package.json#L55
Maybe you need eslint in which case you run:
pnpm --filter @plone/volto lint-fix
or prettier or stylelint depending on the changes that you've made.

davisagli (changes requested):
Instead of making this change in many places, it would be better to update those places to use the UniversalLink component, which already pays attention to this setting. UniversalLink is supposed to be the consistent way to make links in volto; if we don't use it, then we have inconsistencies like this.

davisagli:
Thanks for your efforts, but we'd like a solution which centralizes the check for this setting in the UniversalLink component, rather than checking it in many places in the code. Also the PR is now outdated and incomplete, so I will close it.

PR #5715, opened 2024-01-30
Remove turbo from monorepo commands until it's really necessary

### LINKED ISSUE #5714: Volto main lint fails when running from packages/volto instead of project root
When I run:
pnpm lint:volto
turbo run lint --filter @plone/volto
No failures but when I cd into packages/volto and run there pnpm lint I get 2 errors
You can see the error in the following gist:
https://gist.github.com/ichim-david/1214e562abdf0fe9c6e5ee6544eca095

sneridagh:
Fixes #5714


ichim-david:
@sneridagh this fixes the lint command inside package/volto but it doesn't fix the false positive test run from the root of the monorepo and the testing infrastructure is testing from the root by calling pnpm lint

PR #5718, opened 2024-01-31
Support for passing whole data object in the initial block config

sneridagh:
Initial implementation, let's talk about your idea yesterday.

### DIFF
--- a/docs/source/blocks/settings.md
+++ b/docs/source/blocks/settings.md
@@ -205,6 +205,18 @@ const initialBlocks = {
 };
 ```
 
+you can also pass the full config for the block, using an object:
+
+```js
+const initialBlocks = {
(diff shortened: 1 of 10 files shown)


docs/source/blocks/settings.md
```diff
@@ -205,6 +205,18 @@ const initialBlocks = {
 };
 ```
 
+you can also pass the full config for the block, using an object:
```
stevepiercy, line 208:
```suggestion
You can also pass the full configuration for the block using an object:
```

stevepiercy (dismissed):

ichim-david (approved):
@sneridagh I like the idea, being able to pass the options to the default blocks that you want added per content type is appealing

PR #5722, opened 2024-01-31
fixed a11y Logo Issue and added Translations

### LINKED ISSUE #5720: Logo alt-Title not a11y/Screenreader compliant
When looking at the alt-Title of the Logo-Image, it displays the Site-Title of the Site-Root.

**To Reproduce**
Hover the Page-Logo or Inspect the Logo with the Browser-Inspector

**Expected behavior**
It should be displaying something like "Back to Homepage" to make it a11y-compliant and Screenreader-accessible

**Screenshots**
image
On Mouse-Over
image
In Inspector

This happens in any Browser

I am currently working on a fix

Molochem:
fixes #5720 

Fixed the Alt-Title of the Logo to say "Back to Homepage" in various Languages for a11y and Screenreader conformity

### DIFF
--- a/packages/volto/locales/ca/LC_MESSAGES/volto.po
+++ b/packages/volto/locales/ca/LC_MESSAGES/volto.po
@@ -462,6 +462,11 @@ msgstr ""
 msgid "Back"
 msgstr "Enrere"
 
+#. Default: "Back to homepage"
+#: components/theme/Logo/Logo
+msgid "Back to homepage"
+msgstr "Torna a la pàgina d'inici"
+
 #. Default: "Base"
 #: components/manage/Diff/Diff
 msgid "Base"
--- a/packages/volto/locales/de/LC_MESSAGES/volto.po
+++ b/packages/volto/locales/de/LC_MESSAGES/volto.po
@@ -461,6 +461,11 @@ msgstr "Verfügbare Inhaltsregeln:"
 msgid "Back"
 msgstr "Zurück"
 
+#. Default: "Back to homepage"
+#: components/theme/Logo/Logo
+msgid "Back to homepage"
+msgstr "Zurück zur Startseite"
+
 #. Default: "Base"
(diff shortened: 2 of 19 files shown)


davisagli (approved):
Looks good to me.

In the future, you don't have to provide all translations for a new message; you can leave that for native speakers of the languages to fill in later.

erral:
I would say that "back to home page" is not a good alternative text for the logo right? 

I understand that perhaps we need to signal that, but perhaps it is better to signal that in an `aria-whatever` attribute? Or perhaps add the said alternative text to the link instead of to the logo?

stevepiercy:
@erral is correct. A logo represents a brand or identity, not an action.

Also it is "home page", not "homepage".

And what if the user enters a site not on the home page? They would not "go back to home page", but would "visit home page".

ichim-david:
> @erral is correct. A logo represents a brand or identity, not an action.
> 
> Also it is "home page", not "homepage".
> 
> And what if the user enters a site not on the home page? They would not "go back to home page", but would "visit home page".
@sneridagh  Maybe it should simply be "Go to homepage" this way there is no visit or back involved.
@davisagli @sneridagh this is a breaking change as it modifies the markup, if my change to remove inline style from was deemed as breaking  From BlockChooserSearch.jsx
https://github.com/plone/volto/commit/34e77db6ab94942b13d9ba54d87dd15383e9c621
 which I've reverted so that the changes could have been merged to 17.x.x than so should this change be deemed as breaking.
Given this feedback from others like @erral and @stevepiercy  maybe we need to further discuss in the original or another issue if we need to change it to something else.

davisagli:
@erral @stevepiercy @ichim-david @sneridagh Sorry, it seems I was too quick to merge this. We can revert it and discuss further. I'll do that later today.

davisagli:
@erral @stevepiercy @ichim-david @sneridagh @Molochem Reverted.

PR #5725, opened 2024-02-01
Block search engines from indexing content on Netlify preview builds

stevepiercy:
See https://github.com/plone/documentation/issues/1619

### DIFF
--- a/docs/source/robots.txt
+++ b/docs/source/robots.txt
@@ -1,7 +1,3 @@
-# Disallow all user agents from the following directories and files
+# Disallow all user agents from indexing pages and links in the site on Netlify only.
 User-agent: *
-Disallow: /_sources/
-Disallow: /.doctrees/
-Disallow: /*.txt$
-Disallow: /.buildinfo$
-Disallow: /objects.inv$
+Disallow: /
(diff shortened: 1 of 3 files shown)


stevepiercy:
https://deploy-preview-5725--volto.netlify.app/robots.txt looks good.

No headers got set. I'm looking into why.

stevepiercy:
Actually the header does get set on Netlify only. I looked in the wrong place:

Screenshot 2024-02-01 at 2 47 50 AM

ksuess:
I wonder if it would be enough in every Netlify "site" but plone/documentation to set the option "branch deploy" to none.

grafik

It remains the preview, which is per Netlify default prevented to be indexed by `X-Robots-Tag: noindex`.

stevepiercy:
There is always at least one production deploy, which is either `6.0` for docs or `main` for volto. Setting it to `None` would be the default branch. I set both docs and volto to do just that. These PRs are still needed.

stevepiercy:
See Netlify docs: https://docs.netlify.com/site-deploys/overview/#branches-and-deploys

ksuess:
Yes, one published deploy always seems to be provided by Netlify. I also do not see a way to prevent every deploy but the preview, which is indeed where handy.
Approving.

ksuess (approved):

PR #5732, opened 2024-02-05
Fixed A11y Video Block Issue

Molochem:
Fixed the missing aira-label on Video-Blocks

It seems a bit hacky, and also im not sure about the Message Content. Suggestions are welcome.

### DIFF
--- a/packages/volto/locales/ca/LC_MESSAGES/volto.po
+++ b/packages/volto/locales/ca/LC_MESSAGES/volto.po
@@ -2558,6 +2558,11 @@ msgstr "Eines personals"
 msgid "Persons responsible for creating the content of this item. Please enter a list of user names, one per line. The principal creator should come first."
 msgstr "Responsables de la creació del contingut d'aquest article. Introduïu una llista de noms d'usuari, un per línia. El creador principal hauria de ser primer."
 
+#. Default: "Play Video"
+#: components/manage/Blocks/Video/Body
+msgid "Play Video"
+msgstr ""
+
 #. Default: "Please choose an existing content as source for this element"
 #: components/manage/Blocks/Teaser/DefaultBody
 msgid "Please choose an existing content as source for this element"
--- a/packages/volto/locales/de/LC_MESSAGES/volto.po
+++ b/packages/volto/locales/de/LC_MESSAGES/volto.po
@@ -2557,6 +2557,11 @@ msgstr "Persönliche Einstellungen"
 msgid "Persons responsible for creating the content of this item. Please enter a list of user names, one per line. The principal creator should come first."
 msgstr "Eine Liste von Personen, die an der Erstellung dieses Artikels beteiligt waren. Bitte geben Sie einen Benutzernamen pro Zeile ein. Der Hauptverantwortliche sollte zuerst genannt werden."
 
+#. Default: "Play Video"
+#: components/manage/Blocks/Video/Body
+msgid "Play Video"
+msgstr "Video abspielen"
+
 #. Default: "Please choose an existing content as source for this element"
 #: components/manage/Blocks/Teaser/DefaultBody
 msgid "Please choose an existing content as source for this element"
--- a/packages/volto/locales/en/LC_MESSAGES/volto.po
+++ b/packages/volto/locales/en/LC_MESSAGES/volto.po
@@ -2552,6 +2552,11 @@ msgstr ""
 msgid "Persons responsible for creating the content of this item. Please enter a list of user names, one per line. The principal creator should come first."
 msgstr ""
 
+#. Default: "Play Video"
+#: components/manage/Blocks/Video/Body
+msgid "Play Video"
+msgstr "Play Video"
+
 #. Default: "Please choose an existing content as source for this element"
 #: components/manage/Blocks/Teaser/DefaultBody
 msgid "Please choose an existing content as source for this element"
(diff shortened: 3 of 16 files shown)


packages/volto/src/components/manage/Blocks/Video/View.test.jsx
```diff
@@ -20,29 +24,42 @@ config.blocks.blocksConfig = {
...
-      }}
-    />,
+    <Provider store={store}>
+      <View
+        data={{
+          '@type': 'video',
+          url: 'https://youtu.be/KqjeO_ekW3g',
+        }}
+      />
+      ,
```
ichim-david, line 43: @Molochem by introducing this comma , you render it also in the test which results in a render of an array of a div and that comma instead of how it was previously, there is no need to have it as such and the test could be as it was with just the addition of the aria-label on the div

packages/volto/news/5732.bugfix
```diff
@@ -0,0 +1 @@
+Fixed the missing aira-label on Video-Blocks @Molochem
```
stevepiercy, line 1:
```suggestion
Added the missing `aria-label` in video blocks. @Molochem
```

stevepiercy (changes requested):

Molochem:
Just for the Record , some Tests are flaky. Idid nothing to the actual code, just changed the News fragment. Suddenly all Checks pass ?

sneridagh (approved):
LGTM, @ichim-david could you give your blessing here?

ichim-david:
@Molochem @sneridagh it's a small improvement from "Group main" to "Play Video".
My problem right now is that ok we say Play Video but if you hit enter or spacebar there is no video playback.
It's not a button to access and play the video.
You have to click on the placeholder image to then get the youtube embed as seen in this demo 
https://demo.plone.org/video-block

I would rather say something along the lines "Video embed placeholder image".
The video block needs improvement, you can't access it with the keyboard so I don't really understand why we
have the tabindex=0 on it.

All of this could be improvement with other tickets but I'm a bit -1 on this idea because you don't Play video
using the keyboard anyhow.
I'll try to research some accessible placeholder / embed video players and see what others have tried

ichim-david:
@Molochem @sneridagh 
https://adrianroselli.com/2024/01/embed-slides-youtube-videos-and-more.html#YouTube
If you search for:
The bulk of that code is the SVG play button and its focus & hover styles. You can make something simpler if you wish. 
Below it is a placeholder image with a button on top that says: "Play Video". 
As soon as you press it, the embed is loaded and video starts playback.
This is the correct pattern to me to signal "Play Video".

JeffersonBledsoe:
@ichim-david @Molochem I'm just taking a look over this PR. Unfortunately, I think there are a number of accessibility issues/ WCAG failures with the actual [Embed component](https://react.semantic-ui.com/modules/embed/). E.g. the lack of a suitable role (should be a button to load the video), the lack of announcement when you 'prepare' the video, the poor focus indicator, etc. So while I think this PR is an improvement, I think making the video block accessible is largely held back by these other issues. It would be nice to implement the workflow detailed in https://github.com/plone/volto/pull/5732#issuecomment-1979408736 , however I think that would be better suited for another PR. 

My suggestion for getting this improvement merged would be:
- set the `as` in the `embedSettings` to be a `button`
- fix any styling issues that might come up from the change of element (e.g. you might have to set `width: 100%`).
- have the label be something like "Prepare video for playback". This at least makes the initial screen reader pass over a bit clearer.
- have some text on the page saying something like "Pressing the video will load it. Press again to play" 
  - You can set a `aria-describedby` key in the `embedSettings` to link the button to this text then so it is read aloud.
  - If we need to, we can visually-hide the text so the current visual representation is not changed. I would argue this is a lesser experience (as those not using Assistive Technology need to know you must click twice too), but it is an option if there are any problems with changing the look of the block.

I look forward to hearing what you both think of the above proposal so we can get this merged!

tisto:
@JeffersonBledsoe FYI: we plan to look into this during the Salamina sprint this week.

JeffersonBledsoe:
> @JeffersonBledsoe FYI: we plan to look into this during the Salamina sprint this week.

See you there!

nileshgulia1:
@JeffersonBledsoe @ichim-david I was looking at this and I'm a bit torn on the approach. "Pressing the video will load it. Press again to play" For instance a user sees this, he press the button while on the placeholder (with role ="button"). This loads the iframe, pressing Enter again doesn't play the video, then we have to navigate all the way to the play button in Youtube video. So the above text makes no sense. Maybe I'm missing something.

ichim-david:
@plone/volto-accessibility @tisto @Molochem after discussions with Karin Maligoj which is our resident expert for this Sprint I proposed the aria-label "Load video player".
This is because this is the exact action that happens, when you press enter you get either the youtube or vimeo or whatever other provider there is.
Then after hitting tab you get to the video player controls where you enventually get also the "Play video" button.

Here is a demo page where you can see that as soon as you hit enter you don't play the video
https://volto.demo.plone.org/video-block

ichim-david:
The whole video player experience isn't at all accessible either. For instance, you have tabIndex on a div with an aria-label but there aren't any rotary actions to find this video (it isn't a heading, a form element, a link) as such you find out about it only if you tab around the page and find it eventually.
As part of a broader overhaul, this limitation should be addressed by those who desire to use the Volto Video block and ensure it's accessible.

davisagli:
Closing in favor of #6359

PR #5745, opened 2024-02-09
Initial work, we will have to wrap the table in view mode with a container for make this work.

sneridagh:
Which will be breaking. Waiting for the Block Model v3 for this.


davisagli (changes requested):
This has a bunch of whitespace changes to .md files 👎

davisagli:
Oh, this was old and replaced by #7916

PR #5747, opened 2024-02-09
Backport #5689 to Volto 16.x.x

ichim-david:


ichim-david:
16 branch has some failures regarding search block.
Will put this in draft until I get more information on what happens to search, if this was happening before or not.

ichim-david:
I've tried it locally with 16.x.x and the same search issue happens so it's not due to the changes from this pull request.
Honestly Cypress is flaky and I don't know how we will fix this.
testing-error

I did the same test with Volto selected only Event and refreshed the page and I get only the Event
result yet on the test with Electron I always get a failure

sneridagh (approved):
Not Cypress... The problem lies in the search block, it does more requests than needed, therefore, sometimes timing issues happen. If you look closely, there's one request that returns (good) results (1) then another that comes back to the original state (bad, 3 results), then another good with 1 result.

Any improvement is ok, so LGTM

PR #5748, opened 2024-02-11
Fix logging in test-acceptance-server commands

davisagli:
Start the acceptance image with a pseudo-tty, this fixing a buffering issue where the console output was not shown until exiting.


davisagli:
@sneridagh Why is the changelog check complaining? This is a change to the root level of the repository, not a package.

ichim-david (approved):
@davisagli
Thanks for your work! I felt bad if this was the only thing needed but seeing the extra
pull requests added for plone.app.robotframework made me happy you've tackled this issue!

sneridagh:
@davisagli @ichim-david It was like this because of something... I had to remove the tty. I barely recall it, but I think it was something in CI? well, we will find out...

davisagli:
@sneridagh Well, I wondered about that, but the acceptance tests passed, so... ¯\_(ツ)_/¯

Issue #5742, opened 2024-02-07
content id being showed vertically (distorted) if we add some more attributes then it is showing horizontally.


vanshdeepsingh-coder:
**Description:**
When we try to show the id of the content it is being displayed vertically which is taking a lot more vertical space, but if we try to add some more attributes such as 'expiration date' then it showing horizontally (in correct manner).

**To Reproduce:**

1. Go to home page
2. Click on content icon on left bar
3. On the home navigation bar click on the three dots
4. Select ID in the check box
5. See the display error

**Expected behavior**
The ID of the contents in content box should be shown horizontally

**Video**

[Screencast from 2024-02-07 22-27-04.webm](https://github.com/plone/volto/assets/82263082/355aa295-56d9-482e-8e53-356b32452ee1)

**Software:**

- OS: Ubuntu (23.10)
- Browser: Mozilla firefox (118.0.1)

snehalchaudhari98:
Hi @ichim-david, if this issue still exists then I would like to work on this. Can you please assign it to me?

vanshdeepsingh-coder:
@Snehal Chaudhari This issue is assigned to me and I’m currently working on
it.

stevepiercy:
@snehalchaudhari98 please read and follow [First-time contributors](https://6.docs.plone.org/contributing/first-time.html), especially [Things not to do](https://6.docs.plone.org/contributing/first-time.html#things-not-to-do), [Contributing to Plone](https://6.docs.plone.org/contributing/index.html), and [Contributing to Volto](https://6.docs.plone.org/contributing/volto.html). 

Using your super power of observation, you may see that the green "Open" icon reflects its status. See [GitHub docs](https://docs.github.com/).

PR #3245, opened 2022-04-12
fix: fixed the position of the dropdown menu and white-space of tbody

SaraBianchi:
folder contents bugfix
- fixed the position of the dropdown menu of each table row
- added min-height for the table to better show the dropdown menu even when there is only one item
- added style to wrap titles on desktop monitor

PR #5749, opened 2024-02-11
Removed word-break:break-word from contents table tbody

ichim-david:
- Due to having small table headers such as ID or UID which meant that there was no room to show those ids

This commit fixes https://github.com/plone/volto/issues/5742 by removing the change made in this pull request https://github.com/plone/volto/pull/3245/files#diff-741dc58bf11002ee21b76a53bd83a19041a2dc2e06fcd838346af8ed52c040e6R143

### DIFF
--- a/packages/volto/news/5742.bugfix
+++ b/packages/volto/news/5742.bugfix
@@ -1,0 +1,1 @@
+Removed css from `contents.less` that made the Contents table break words in tiny sections due to small table headers such as `ID` or `UID`.  @ichim.david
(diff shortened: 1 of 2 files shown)


ichim-david:
@pnicolli @sneridagh  one thing that we could do is perhaps to add sticky sections for the drag, select and actions for situation where you enabled many actions that will make you scroll until you get to the actions.
See these videos for what I mean.
The current behavior with non-sticky:
https://github.com/plone/volto/assets/152852/2ea97d61-34b0-4f25-a34b-cb144e47a21e

With sticky sections:
https://github.com/plone/volto/assets/152852/1dc012d6-7385-4e70-90a5-883c8e24f955

davisagli (approved):
LGTM. Now that the table scrolls properly, it makes more sense to not break words.

I also like the idea for a sticky actions column.

sneridagh:
+1 I like the idea too.

sneridagh (approved):

PR #5753, opened 2024-02-12
Add wait commands to flaky block-listing tests

ichim-david:
- Using only wait('@content') mean that every x times the assert command failed due to running too quick before proper page load


ichim-david:
@sneridagh I'm going to merge this pull request since it's meant to fix sporadic failures from the listing block testing that happens every now and then and they are annoying to have to re-run the failed test so that you have a green build even though it didn't crash because of your code.

PR #5756, opened 2024-02-13
Mention what version the 'links and references' view was added

davisagli:
~~@stevepiercy It's a bit more complicated than this... the feature was also backported to Volto 16, but there it is disabled by default. Should we mention this here?~~

The setting which controls it is documented here: https://6.docs.plone.org/volto/configuration/settings-reference.html#term-excludeLinksAndReferencesMenuItem ...should it also be mentioned on this page, or not since this is more for an editor audience?

### DIFF
--- a/docs/source/user-manual/links-to-item.md
+++ b/docs/source/user-manual/links-to-item.md
@@ -32,6 +32,15 @@ An item references another item by creating a hyperlink to this item, using the 
 In the toolbar on the left side, you can click on the item with the three horizontally aligned dots, called {guilabel}`More menu`.
 Click then on the item labeled {guilabel}`Links and references`.
 
+````{note}
+If you do not see {guilabel}`Links and references`, then you either have a version of Volto that does not support this feature, or this feature has not been enabled on your site.
+In either case, contact your site administrator for assistance.
+
+```{seealso}
+{term}`excludeLinksAndReferencesMenuItem`
+```
+````
+
 You can see now a table displaying all links to and references of the current item.
 
 ```{image} ../_static/user-manual/manage/link-to-items.png
--- a/docs/source/user-manual/links-to-item.md
+++ b/docs/source/user-manual/links-to-item.md
@@ -11,6 +11,14 @@ myst:
 
 # Finding links and references to the current page
 
+```{versionadded} Volto 16.28.0
+This feature was added in Volto 16.28.0, but it was disabled by default.
+```
+
+```{versionchanged} Volto 17.0.0-alpha.19
+This feature is now enabled by default.
+```
+
 Sometimes it can be hard to keep track from where a certain page is referenced.
 This is especially true whenever you want to reconstruct your website with a greater number of objects being moved or deleted.
 For this situation, you can visit the {guilabel}`Links and references` page, which is an overview of all content items that reference a certain content item.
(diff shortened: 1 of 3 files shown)


docs/source/user-manual/links-to-item.md
```diff
@@ -19,6 +19,9 @@ For this situation, you can visit the {guilabel}`Links and references` page, whi
 An item references another item by creating a hyperlink to this item, using the item in a block, referencing it in the {guilabel}`Related Items` field, or referencing it in a relation field.
 ```
 
+```{versionadded} Volto 17.0.0-alpha.19
```
stevepiercy, line 22: The version* admonitions should be below the headings to which they refer, because when people click anchors, the version* admonitions will scroll out of view.

````suggestion
```{versionchanged} Volto 17.0.0-alpha.19
This feature is now enabled by default.
```

```{versionadded} Volto 16.?.?
This feature was added in Volto 16.?.?, but it was disabled by default.
````

**Review by stevepiercy — CHANGES_REQUESTED** [review]
I think will address your question about 16 vs. 17, except I don't know which version in 16.x.x.

File: docs/source/user-manual/links-to-item.md (hunk 1 above)
Comment (stevepiercy) [reply docs/source/user-manual/links-to-item.md:22]: Oopsie! 16 admonition should come before 17 admonition!

Comment (stevepiercy) [reply docs/source/user-manual/links-to-item.md:22]: Edited suggestion.

Comment (davisagli) [reply docs/source/user-manual/links-to-item.md:22]: @stevepiercy This info is relevant to the feature covered by this entire page. Should it go just under the h1 heading?

Comment (davisagli) [reply docs/source/user-manual/links-to-item.md]: @stevepiercy I applied your suggestion manually so I could add the right version for 16. Thanks.

File: docs/source/user-manual/links-to-item.md [hunk 2]
```diff
@@ -1,10 +1,10 @@
...
 myst:
   html_meta:
-    "description": "User manual on how to find all links and relations to the current item."
-    "property=og:description": "User manual on how to find all links and relations to the current item."
-    "property=og:title": "Finding links and relations to the current item."
-    "keywords": "Volto, Plone, frontend, React, User manual, links, relations, references, related content"
+    'description': 'User manual on how to find all links and relations to the current item.'
+    'property=og:description': 'User manual on how to find all links and relations to the current item.'
+    'property=og:title': 'Finding links and relations to the current item.'
+    'keywords': 'Volto, Plone, frontend, React, User manual, links, relations, references, related content'
```
stevepiercy, line 7: Bad editor config, no cookie for you! Please revert above changes.

stevepiercy (changes requested):
Please revert editor s/"/'.

docs/source/user-manual/links-to-item.md
davisagli: @stevepiercy Sorry, will fix. But it's NOT simply bad editor config. As we've discussed before, I don't know how to make VSCode use double quotes here but single quotes in javascript code samples.

stevepiercy:
> The setting which controls it is documented here: https://6.docs.plone.org/volto/configuration/settings-reference.html#term-excludeLinksAndReferencesMenuItem ...should it also be mentioned on this page, or not since this is more for an editor audience?

Ah, I see. We need to modify both pages to reference each other with an appropriate context. I'll push something up for you to consider in little bit.

stevepiercy:
@davisagli okey dokey. Have a look at these changes to see if they sufficiently address the things you brought up.

davisagli:
@stevepiercy Thanks, looks good to me. Could you update your PR review?

stevepiercy (approved):
Thank you!

PR #5761, opened 2024-02-14
Exclude chromewebstore from linkcheck

stevepiercy:
chromewebstore recently changed its URL and has "too many redirects" during linkcheck, no matter what I tried. This might be fixed with a more recent version of Sphinx, so I put in a TODO to check when that happens.


stevepiercy:
I'm going to merge this, as it is trivial and doesn't need review, and it is required for its related PR in documentation.

PR #5762, opened 2024-02-14
Remove the isDisabled from all fields in the left side form of the babel view

sneridagh:
So they can be selected and copied over, for better UX.


giuliaghisini (approved):

davisagli:
@sneridagh Would it make sense to prefill all data in the translation with the data from the source language, so that it's not necessary to copy field by field in cases where the translation is unchanged?

Issue #4352, opened 2023-02-02
Extract basic Volto components to their own Volto core package: `@plone/components`


sneridagh:
## Abstract

Extract the basic Volto React components to their own Volto core add-on, and being able to release them separately from Volto core. 

## Motivation

Since long time ago that there is the need to render Plone content into other React-based frameworks (NextJS, Remix, Gatsby, etc...). Having a unified set of default basic components for reuse everywhere would help a lot in order to expand the influence of Plone as headless CMS.

One could thing about this library in a small SDK to show Plone content in React frameworks.

## Assumptions

No special assumptions.

## Proposal & Implementation

This PLIP has relation with PLIP #4347 (Modernize Data Fetching API). The components should be agnostic in data fetching and use the `@plone/actions` library proposed in that PLIP. Thus, they will be Redux-free.

The components of this library won't use any specific design system nor SemanticUI in order to be as much as possible unspecific and agnostic, to ease the reusability.

They should use the basic and structural components in `@plone/atoms` components (See #4332 Quanta PLIP).

A preliminary list of candidates:

- [ ] Header
- [ ] Footer
- [ ] Breadcrumbs
- [ ] Sections / Fat Menu
- [ ] AnonTools
- [ ] Comments
- [ ] LanguageSelector
- [ ] Pagination
- [ ] Login
- [ ] Volto Core Blocks View components

## Deliverables

- `@plone/components` or `@plone/react-sdk` (TBD) npm package, and as a Volto core package. 
- Documentation on how to use it
- StoryBook with detailed information about the components
- Cypress tests
- Upgrade guide

## Risks

At some point, when #4347 is ready, we have to start using them along with `@plone/actions` in Volto core. This will break all the existing shadows in projects. It will have to be proper documented.

## Participants

Víctor Fernández de Alba (Leading)
... anybody interested to help will be welcomed! Please contact @sneridagh

sneridagh:
> [!IMPORTANT]
> If you are not a member of the Volto Team or Developers Team in the Plone GitHub organization, then do not work on or comment on this issue.

pnicolli:
Update on this: CMSUI and Public Theme components will not be part of @plone/components, which will only be a React Components Library with optional Quanta styles and Tailwind flavor. Those will instead be in the @plone/layout package.
We are happy with the current state of the library, therefore I am closing this.
Of course the library will need to be kept up to date with new react-aria-components releases, but it's outside of the scope of this issue.

PR #5766, opened 2024-02-15
Move to only CSS-based approach - RAC completion - Removal of SCSS

sneridagh:


sneridagh:
@stevepiercy https://github.com/plone/volto/pull/5766/files#diff-4b24799689b090701f0bc36ea0a508a962d448d1feaec58750b8fcd61e150c21 Please a will appreciate a preliminary look. It's still moving, so don't take everything by the book. I'd like to merge it today.

sneridagh:
@stevepiercy also tell me what do you think about it and if you got the idea.

stevepiercy:
@sneridagh reviewing docs now, will push commits when done, and ask questions if I am not clear.

packages/components/package.json
```diff
@@ -91,13 +102,17 @@
...
     "globby": "^14.0.0",
     "history": "^5.3.0",
     "jest-axe": "^8.0.0",
     "jsdom": "^22.1.0",
-    "parcel": "^2.10.3",
+    "lightningcss": "^1.23.0",
+    "lightningcss-cli": "^1.23.0",
+    "parcel": "^2.11.0",
+    "parcel-optimizer-react-client": "workspace:^",
     "postcss-scss": "4.0.9",
```
pnicolli, line 116: Shall we remove this?

stevepiercy (approved):
Docs look good.

packages/components/src/styles/basiq/theme.css
```diff
@@ -0,0 +1,194 @@
...
+  font-family: system-ui;
+  font-size: 14px;
+  line-height: 1.5;
+}
+
+/* color themes for dark and light modes, generated with Leonardo.
+ * Light: https://leonardocolor.io/theme.html?name=Light&config=%7B%22baseScale%22%3A%22Gray%22%2C%22colorScales%22%3A%5B%7B%22name%22%3A%22Gray%22%2C%22colorKeys%22%3A%5B%22%23000000%22%5D%2C%22colorspace%22%3A%22RGB%22%2C%22ratios%22%3A%5B%22-1.12%22%2C%221.45%22%2C%222.05%22%2C%223.02%22%2C%224.54%22%2C%227%22%2C%2210.86%22%5D%2C%22smooth%22%3Afalse%7D%2C%7B%22name%22%3A%22Purple%22%2C%22colorKeys%22%3A%5B%22%235e30eb%22%5D%2C%22colorspace%22%3A%22RGB%22%2C%22ratios%22%3A%5B%22-1.12%22%2C%221.45%22%2C%222.05%22%2C%223.02%22%2C%224.54%22%2C%227%22%2C%2210.86%22%5D%2C%22smooth%22%3Afalse%7D%2C%7B%22name%22%3A%22Red%22%2C%22colorKeys%22%3A%5B%22%23e32400%22%5D%2C%22colorspace%22%3A%22RGB%22%2C%22ratios%22%3A%5B%22-1.12%22%2C%221.45%22%2C%222.05%22%2C%223.02%22%2C%224.54%22%2C%227%22%2C%2210.86%22%5D%2C%22smooth%22%3Afalse%7D%5D%2C%22lightness%22%3A98%2C%22contrast%22%3A1%2C%22saturation%22%3A100%2C%22formula%22%3A%22wcag2%22%7D */
+:root {
+  --background-color: #f8f8f8; /* Change to --basiq-slate-50 ? or to pure white? */
```
pnicolli, line 12: It's a strong move to set the main background color of the site to something different from white, I am not sure it fits the very basic theme idea that we want

packages/components/src/styles/basiq/theme.css
```diff
@@ -0,0 +1,194 @@
...
+    --highlight-foreground: HighlightText;
+    --highlight-foreground-pressed: HighlightText;
+    --color-invalid: LinkText;
+    --color-pressed-invalid: LinkText;
+  }
+}
+
+/* Dark: https://leonardocolor.io/theme.html?name=Dark&config=%7B%22baseScale%22%3A%22Gray%22%2C%22colorScales%22%3A%5B%7B%22name%22%3A%22Gray%22%2C%22colorKeys%22%3A%5B%22%23000000%22%5D%2C%22colorspace%22%3A%22RGB%22%2C%22ratios%22%3A%5B%22-1.12%22%2C%221.45%22%2C%222.05%22%2C%223.02%22%2C%224.54%22%2C%227%22%2C%2210.86%22%5D%2C%22smooth%22%3Afalse%7D%2C%7B%22name%22%3A%22Purple%22%2C%22colorKeys%22%3A%5B%22%235e30eb%22%5D%2C%22colorspace%22%3A%22RGB%22%2C%22ratios%22%3A%5B%22-1.12%22%2C%221.45%22%2C%222.05%22%2C%223.02%22%2C%224.54%22%2C%227%22%2C%2210.86%22%5D%2C%22smooth%22%3Afalse%7D%2C%7B%22name%22%3A%22Red%22%2C%22colorKeys%22%3A%5B%22%23e32400%22%5D%2C%22colorspace%22%3A%22RGB%22%2C%22ratios%22%3A%5B%22-1.12%22%2C%221.45%22%2C%222.05%22%2C%223.02%22%2C%224.54%22%2C%227%22%2C%2210.86%22%5D%2C%22smooth%22%3Afalse%7D%5D%2C%22lightness%22%3A11%2C%22contrast%22%3A1%2C%22saturation%22%3A100%2C%22formula%22%3A%22wcag2%22%7D */
+@media (prefers-color-scheme: dark) {
```
pnicolli, line 169: I have a personal opinion about this, that we should not include a dark mode theme by default. Most usecases for plone are public facing sites showing news items, events, blogs, products or whatever. These use cases usually don't have two separate themes or a theme switcher, it's something usually for dashboards, applications that are more focused on using some product or managing something.

packages/components/src/styles/basiq/theme.css
```diff
@@ -0,0 +1,194 @@
+/* Base styles */
+:root {
+  background: var(--background-color);
+  font-family: system-ui;
+  font-size: 14px;
+  line-height: 1.5;
+}
```
pnicolli, line 7: These are very strong assumptions that I would probably set not in basiq but in a more advanced theme.

packages/components/src/styles/basiq/theme.css
```diff
@@ -0,0 +1,194 @@
...
+:root {
+  --quanta-cobalt: #2597f4;
+  --quanta-sapphire: #0b78d0;
+  --quanta-royal: #085696;
+  --quanta-ballet: #fee9e7;
+  --quanta-rose: #f54e38;
+  --quanta-candy: #d0220b;
+  --quanta-neon: #3da72a;
+  --quanta-lemmon: #f6d355;
+}
```
pnicolli, line 82: These, and all other quanta colors (because they look nice :D ) would probably fit better in a separate theme file in the quanta folder.

packages/components/src/styles/typography.css
```diff
@@ -0,0 +1,252 @@
+body {
+  font-family: metropolis, sans-serif !important;
```
pnicolli, line 2: We should try to avoid at all cost to have `!important`s in the basic styles we provide. This can be left for later to change, it is just something that has a high risk of being forgotten here :)

stevepiercy:
See https://github.com/plone/volto/issues/4352

packages/components/README.md
```diff
@@ -7,48 +7,217 @@
...
+## Quanta icons
+
+This package provides an implementation of the Quanta Icon set in React components.
+They can be used directly in your components as:
+
+```tsx
+import { ChevronupIcon } from '../Icons/ChevronupIcon';
+
+const MyComponent = (props) => (
+  <Button>
```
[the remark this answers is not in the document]
sneridagh: I added `aria-label` on the icons. Question is if we want them to provide one by default or not. Is that what you meant?

packages/components/package.json
sneridagh: why? It's what bundles it all :)

packages/components/src/styles/basiq/theme.css
sneridagh: Agreed, moving to white.

packages/components/src/styles/basiq/theme.css
sneridagh: Let's remove it for now then, since for now it causes confusion.

packages/components/README.md
```diff
@@ -5,50 +5,226 @@
...
+Since we are using RAC as the base, the styling of this package components are done via the RAC styling options (https://react-spectrum.adobe.com/react-aria/styling.html).
+See the `react-aria-components` section below for more information about RAC.
+This package provides a simple and basic set of CSS rules to build upon.
+Alternatively, you can bring your own styles, removing the basic styling or complementing it, and build on top of it.
+You can even use RAC to use other CSS utility libraries, like TailwindCSS.
+
+## Styling
+
+This package provide a basic set of CSS rules.
+We call them "BasiQ" (the "Q" is for "Quanta").
```
ichim-david, line 32: @sneridagh @stevepiercy  No one asked my opinion but here I am giving it anyhow :) 
I am not so sure if it would be a good thing to have a name that is like an acronym.
I am thinking of how we are going to spell it when we give training about it.
Personally I wouldn't try to make clever names in trying to combine something
and giving it meaning then somehow.

I also think we should investigate if the naming doesn't clash with other  entities that use
this naming, performing a google search gave me a few results for basiq such as 
https://www.basiq.io/
I don't know how far are trademarks and copyright nowadays but it doesn't hurt for us todo
some searching and ensuring we don't use something already established and copyrighted

packages/components/src/styles/basiq/theme.css
sneridagh: But we need a default, right? Maybe make all use custom CSS Properties?

packages/components/src/styles/basiq/theme.css
sneridagh: Done! I was into it already.

packages/components/src/styles/typography.css
sneridagh: by all means, this is still copy/paste old Quanta thinggies that we should polish.

sneridagh:
@pnicolli pushed your suggestions and more.

packages/components/package.json
sneridagh: doh now I got it, sorry. Thanks @ichim-david

packages/components/README.md
stevepiercy: I had a similar moment of "hmmm", how do you even pronounce that? "Bay-sik" or "Bay-suh-que"?

"Basiq" is trademarked.

Let's avoid the attorney horde. I'll push a commit to replace it with "Quanta base CSS rules" and grammatically correct phrases.

stevepiercy: On second thought, I won't push a commit. It means renaming the package.

packages/components/README.md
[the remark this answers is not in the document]
pnicolli: I have to admit I did not notice the `aria-label` on the icons, sorry. I am not entirely sure it's enough though, for these reasons:

- `title` also provides a tooltip which is nicer for visual users that only see an icon instead (even if totally understandable like a chevron in this case)
- iirc the title or label should be on the button, because screen readers will read that as the button label and it is more straightforward to understand for the user. Also, the icon itself usually has `aria-hidden="true"` so I'm not sure if the label would be read, I haven't tested it.

To answer you question, I only meant to have a good example in the docs, I did not mean to change the components code.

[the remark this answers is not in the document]
sneridagh: OK! fortunately RAC will help with all these, since they already cover a lot, but these guidelines will be great to have sumarized somewhere (probably in the components docs too).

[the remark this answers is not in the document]
sneridagh: Btw, `title` is not in the RAC Button props... that is indicator that having the title is no longer the best practice or the browsers have now support for aria instead...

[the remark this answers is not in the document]
sneridagh: https://react-spectrum.adobe.com/react-spectrum/Button.html#accessibility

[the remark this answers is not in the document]
sneridagh: Also on that regard, I plan to make a test on A11y with AXE for all components.

packages/components/README.md
sneridagh: @stevepiercy 
I've already talked about this with @ichim-david 
I came up with the name in 5mins. It's an internal name for now, we won't use it anywhere else. 
We will sell `@plone/components` what's inside stays in Vegas.
Let's maintain it as internal name until someone has a better name. And yes, we need a name for them.
Quanta is also something else out there, nobody complained.

packages/components/src/styles/basiq/theme.css
pnicolli: Yeah we do need a default. I was thinking that if this is the basic css for the components only, maybe defaults for the website as a whole should not be here... 🤔

sneridagh: yep, let's have one for the storybook, then only customize the very basic.
I will take care this evening.

packages/components/README.md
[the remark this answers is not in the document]
pnicolli: > Btw, `title` is not in the RAC Button props... that is indicator that having the title is no longer the best practice or the browsers have now support for aria instead...

Yes, this is correct. Also, `aria-label` always takes precedence when set, it's the best way to go. What I kind of dislike of this new spec is that `title` also added the tooltip and I really don't get why that is not an important feature for people making decisions on these topics :D anyway, I'm happy if we stick to `aria-label` here

packages/components/README.md
sneridagh: @stevepiercy 
OTOH, I after giving it a second thought, change a Q for a c is easy anyways, but let's all think in the name of the basic/fundamental/foundational CSS rules.

stevepiercy: @sneridagh indeed, I think clever spelling in names end up being a footgun. I think it is better not to brand something that you do not intend to put much marketing effort toward. It's just confusing to developers.

sneridagh: Ready.

packages/components/README.md
```diff
@@ -29,66 +29,65 @@ You can even use RAC to use other CSS utility libraries, like TailwindCSS.
...
 ```
 
-Take a look at the implementation of the Quanta components, using the BasiQ ones as baseline in the `quanta` folders.
+Take a look at the implementation of the Quanta components, using the basic ones as baseline in the `quanta` folders.
 
 Alternatively, as RAC allows, you can also drop your own basic set of styles.
-You can take the BasiQ styles as reference.
-You can even bring your own CSS framework and make BasiQ get the styling in there using the utilities of your choice.
+You can take the basic styles as reference.
+You can even bring your own CSS framework and make basic get the styling in there using the utilities of your choice.
```
stevepiercy, line 85:
```suggestion
You can even bring your own CSS framework, and add your styles to the basic styles using the utilities of your choice.
```

packages/components/src/styles/quanta/Select.css
```diff
@@ -1,5 +1,5 @@
-/* Quanta no ha de redefinir colors, ha de tocar-los via redefinició dels originals
-que venen de basiq. Així si la gent toca basiq, no es redefiniran a Quanta */
+/* Quanta does not have to redefine colors, it has to override the originals coming
```
stevepiercy, line 1:
```suggestion
/* Quanta does not have to redefine colors, but it has to override the original ones coming
```

packages/components/src/styles/quanta/Select.css
```diff
@@ -1,5 +1,5 @@
...
+from basic. Then if people override the basic ones, then they won't get overriden in quanta */
```
stevepiercy, line 2:
```suggestion
from the basic styles. Then, if people override the basic ones, then they won't get overridden in Quanta. */
```

stevepiercy (changes requested):
Grammar fixes. Thanks for tweaking to "basic".

packages/components/README.md
sneridagh: Not really the intention... although valid, we can leave it as well. The initial intention is that you can remove them completely, and use your framework to style from scratch the default RAC classNames...

sneridagh:
I think it's ready as a first alpha, I'd like to merge it, so people can see the outcome in StoryBook.

We will continue improving from here.

packages/components/README.md
```diff
@@ -5,50 +5,223 @@
...
+
+This package provides an implementation of the Quanta Icon set in React components.
+They can be used directly in your components as:
+
+```tsx
+import { ChevronupIcon } from '../Icons/ChevronupIcon';
+
+const MyComponent = (props) => (
+  <Button aria-label="Unfold/Collapse">
+    {props.isOpen ? <ChevronupIcon /> : <ChevrondownIcon />}
```
ichim-david, line 144: @sneridagh I know this is just demo example code but there is no import of ChevrondownIcon,
or where Button comes from. We can keep examples to minimum but I still think they should be
as close to working as possible which to me means that this example is missing the Button and
the ChevrondownIcon import

packages/components/news/5766.breaking
```diff
@@ -0,0 +1,5 @@
+- Move to only CSS-based approach
+- Removal of SCSS in favor of the above
+- New BasiQ components, based in React Aria Components
```
ichim-david, line 3:
```suggestion
- New Basic components, based in React Aria Components
```

packages/components/src/stories/Basiq.colors.stories.mdx
```diff
@@ -0,0 +1,12 @@
+import { Meta, ColorPalette, ColorItem } from '@storybook/blocks';
```
ichim-david, line 1: @sneridagh this is still named Basiq

packages/components/README.md
```diff
@@ -5,50 +5,223 @@
...
+
+```tsx
+import { useFlattenToAppURL } from '@plone/components';
+
+const Link = (props: LinkProps, ref: ForwardedRef<HTMLAnchorElement>) => {
+  const { flattenToAppURL } = useFlattenToAppURL();
+  const flattenedURL = flattenToAppURL(props.href);
+
+  return (
+    <RACLink ref={ref} {...props} href={flattenedURL}>
```
ichim-david, line 177: @sneridagh I would add the import for RACLink in this example

packages/components/README.md
```diff
@@ -5,50 +5,223 @@
...
+const Link = (props: LinkProps, ref: ForwardedRef<HTMLAnchorElement>) => {
+  const { flattenToAppURL } = useFlattenToAppURL();
+  const flattenedURL = flattenToAppURL(props.href);
+
+  return (
+    <Link {...props} href={flattenedURL}>
+      {props.children}
+    </RACLink>
+  );
+};
```
ichim-david, line 181:
```suggestion
import { useFlattenToAppURL } from '@plone/components';
import { Link as RACLink } from 'react-aria-components';

const Link = (props: LinkProps, ref: ForwardedRef<HTMLAnchorElement>) => {
  const { flattenToAppURL } = useFlattenToAppURL();
  const flattenedURL = flattenToAppURL(props.href);

  return (
    <RACLink {...props} href={flattenedURL}>
      {props.children}
    </RACLink>
  );
};
```

sneridagh: ok, but both could work!

ichim-david (approved):
Approving this work it's awesome already. I still have to mention that we need to add some more info to the readme in the future about installation and how you expect ppl to import the most basic component if they have a simple create react app or a volto website

sneridagh:
@stevepiercy there's a link in main Volto README that fails because of expired cert. I assume that's temporary.

packages/components/README.md
sneridagh: See last commit.

stevepiercy:
@sneridagh correct. Someone who maintains [that site](https://www.talke.com/) needs to renew its cert, and that looks to be Kitconcept. Pinging @ericof and @tisto to address the expired cert. I'm not sure exactly who manages it.

sneridagh:
@stevepiercy 

A reason that is out of our hands can't break our CI. Can we disable such check for expired certs?

I think we are no in control of it, so we can't do nothing avout that situation.

stevepiercy:
@sneridagh this is why I don't think we should have external links for marketing purposes in the README. I still believe such marketing belongs in marketing materials, such as https://plone.org/why-plone/they-use-plone/they-use-plone-6 and https://github.com/collective/awesome-volto#websites-built-with-volto.

Until that happens, here are options:

1. Add the flag `--allow-ssl`, but then that defeats the purpose of checking links. A site with an invalid cert is broken and we shouldn't have a link to it.
2. Exclude the one site from checking in the configuration, but that is a bad precedent.
3. Remove the link from the README so that it does not break CI.
4. Wait for the next business day. It expired on the weekend. I assume that they will realize that no one can safely visit their site on Monday and renew it. This is not unusual. For personal blogs, because they lack corporation resources, I typically notify the owner, and they fix it and thank me for letting them know.

For options 1 and 2, I am a hard -1 because it is undue maintenance and violates the purpose of a linkchecker. I'm +1 for options 3 and 4.

sneridagh:
Point is, URL checking is fine, acknowledging that a link is broken is fine. Allow this process to break the build, is bot acceptable, and breaks momentum. 

Again a third party problem must not stop our process, nor making spend our precious time on this matters, instead of producing.

Let's assume they will fix it soon. If not, we remove the link.

PR #5769, opened 2024-02-18
Add include for use in main docs and Develop Volto core

stevepiercy:
@sneridagh do we need to add Yeoman or Yarn as requirements, aligning https://6.docs.plone.org/volto/contributing/developing-core.html with https://6.docs.plone.org/install/create-project.html?

See also https://github.com/plone/documentation/pull/1628


stevepiercy:
Failed linkcheck is due to an expired cert for https://www.talke.com/. We can ignore that for a while.

sneridagh (approved):

PR #5775, opened 2024-02-20
Slots in core

sneridagh:
Let's see if we finally make this happen, at least the "static" ones.

- [x] Finish docs
- [x] Add Cypress tests in coresandbox
- [x] Add a few slots here and there in Volto core?
- [x] Unregister slot component
- [x] Reorder slots components


packages/registry/src/index.ts
```diff
@@ -181,6 +182,84 @@ class Config {
...
+
+  getSlot<T>(name: string, args: T): SlotComponent['component'][] | undefined {
+    const { slots, data } = this._data.slots[name];
+    const slotComponents = [];
+    // For all enabled slots
+    for (const slotName of slots) {
+      // For all registered components for that slot, inversed, since the last one registered wins
+      // TODO: Cover ZCA use case, where if more predicates, more specificity wins if all true.
+      // Let's keep it simple here and stick to the registered order.
+      for (const slotComponent of data[slotName].toReversed()) {
```
tiberiuichim, line 194: You can use `find()` here, similar to https://github.com/plone/volto/blob/e7788b3358e56bc0cc13a650d56915ed6984e396/packages/volto-slate/src/blocks/Text/keyboard/index.js#L30

packages/types/src/config/index.d.ts
```diff
@@ -11,7 +11,17 @@ export type AddonRoutesConfig = {
   component: React.ComponentType;
 }[];
 
-export type SlotsConfig = Record<string, unknown>;
+export type SlotComponent = {
+  component: React.ComponentType<any>;
+  predicates: ((...args: any[]) => boolean)[];
```
tiberiuichim, line 16: this should be optional

tiberiuichim (approved):

packages/registry/src/registry.test.tsx
```diff
@@ -0,0 +1,374 @@
...
+  //   },
+  //   {
+  //     predicates: [RouteCondition('/de'), ContentTypeCondition(['News Item'])],
+  //     component: 'this is a Bar component',
+  //   },
+  //   {
+  //     predicates: [RouteCondition('/de/path/folder'), ContentTypeCondition(['News Item']), TrueCondition()],
+  //     component: 'this is a Bar component',
+  //   },
+  // ]
```
stevepiercy, line 135: Lots of commented code here. Is this necessary?

packages/registry/src/registry.test.tsx
```diff
@@ -0,0 +1,374 @@
...
+
+  // type Predicate = (predicateValues: unknown) = (predicateValues, args) => boolean
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const RouteConditionTrue = (route) => () => true;
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const RouteConditionFalse = (route) => () => false;
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const ContentTypeConditionTrue = (contentType) => () => true;
+  // eslint-disable-next-line @typescript-eslint/no-unused-vars
+  const ContentTypeConditionFalse = (contentType) => () => false;
```
stevepiercy, line 149: Are all those repeated comments needed?

docs/source/configuration/slots.md
```diff
@@ -0,0 +1,89 @@
+# Slots
```
stevepiercy, line 1:
```suggestion
---
myst:
  html_meta:
    "description": "Slots are insertion points in the Volto rendering tree structure."
    "property=og:description": "Slots are insertion points in the Volto rendering tree structure."
    "property=og:title": "Slots"
    "keywords": "Volto, Plone, frontend, React, configuration, slots, viewlets"
---

# Slots
```

docs/source/configuration/slots.md
```diff
@@ -0,0 +1,89 @@
...
+
+Slots are insertion points in the Volto rendering tree structure.
+This concept is inherited from Plone ClassicUI artifact `Viewlets`.
```
stevepiercy, line 4: I think this is the correct markup to use Intersphinx. We should check the Netlify preview build to verify it links to Plone 6 docs correctly.

```suggestion
This concept is inherited from the Plone Classic UI {doc}`plone:classic-ui/viewlets`.
```

docs/source/configuration/slots.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Anatomy
+
+Slots are named, and they can contain a list of different slot components.
+Slots components are also named, and they are registered in the configuration registry using a specific API for slots.
```
stevepiercy, line 9:
```suggestion
Slot components are also named, and they are registered in the configuration registry using a specific API for slots.
```

docs/source/configuration/slots.md
```diff
@@ -0,0 +1,89 @@
...
+The main trait of a slot component is that its render are controlled by a list of conditions (predicates).
```
stevepiercy, line 10:
```suggestion
The main trait of a slot component is that its renderer is controlled by a list of conditions called {term}`predicates`.
```

We should add the definition of `predicate` and `predicates` to the Plone 6 docs' Glossary. Let's steal from Pyramid: https://docs.pylonsproject.org/projects/pyramid/en/latest/glossary.html#term-predicate. What do you think?

docs/source/configuration/slots.md
```diff
@@ -0,0 +1,89 @@
...
+Multiple slot components can be registered under the same name, as long as they have different predicates.
+
+Slot (eg. `toolbar`)
+  - SlotComponent (eg. `edit`)
+    - predicates (eg. only appear in `/de/about`)
+    - predicates (eg. only appear if content type is one of `Document` or `News Item`)
+    - no predicates (eg. always appear)
+  - SlotComponent (eg. `contents`)
+  - SlotComponent (eg. `more`)
```
stevepiercy, line 19: I'm not clear of the concept that this markdown is supposed to represent. Is this accurate?

1. This is a tree structure, where the root is `Slot`. Assuming that is correct, `Slot` should be inside the nested list, not as a separate paragraph.
2. Then there are 3 `SlotComponent` thingies, each with their own name.
3. Within the `edit` `SlotComponent` thingie, there are 2 lists of predicates and a default condition for when none of the predicates return `true`.

Please let me know.

docs/source/configuration/slots.md
```diff
@@ -0,0 +1,89 @@
...
+
+The order in which the components should render is governed by the order that they were registered.
```
stevepiercy, line 21:
```suggestion
The order in which the components render is governed by the order in which they were registered.
```

docs/source/configuration/slots.md
```diff
@@ -0,0 +1,89 @@
...
+You can change the order of the defined slot components for a different slot using the API. (pending)
+You can even delete the rendering of a registered slot component using the API (pending)
```
stevepiercy, line 23: Let's wrap the above two lines in an admonition.

````
```{todo}
You can change the order of the defined slot components for a different slot using the API.
You can even delete the rendering of a registered slot component using the API.
```
````

File: docs/source/configuration/slots.md [hunk 12]
```diff
@@ -0,0 +1,89 @@
...
+
+Slot (eg. `toolbar`)
+  - `edit`
+  - `contents`
+  - `more`
```
stevepiercy, line 28: I don't understand why the above four lines are present. They lack context.

docs/source/configuration/slots.md
```diff
@@ -0,0 +1,89 @@
...
+
+Volto renders the slots using the `SlotRenderer` component.
+You can add more insertion points in your code as needed.
```
stevepiercy, line 31:
```suggestion
You can add insertion points in your code, as shown in the following example.
```

docs/source/configuration/slots.md
```diff
@@ -0,0 +1,89 @@
...
+
+```tsx
+<SlotRenderer name="toolbar" content={content} />
+```
+
+## Registering a slot component
```
stevepiercy, line 37:
```suggestion
## Register a slot component
```

docs/source/configuration/slots.md
```diff
@@ -0,0 +1,89 @@
...
+      name: 'save',
+      component: 'this is a toolbar save component with a true predicate',
+      predicates: [RouteConditionTrue('/de')],
+    });
+```
+
+`slot`: The name of the slot, where the slot components are stored
+`name`: The name of the slot component that we are registering
+`component`: The component that we want to render in the slot
+`predicates`: A list of functions that return a function with this signature:
```
stevepiercy, line 53: This should be in [glossary list syntax](https://6.docs.plone.org/contributing/documentation/myst-reference.html#glossary-terms).

Example:

````
```{glossary}
:sorted:

`inline literal term`
    definition

narrative term
    definition
```
````

The `:sorted:` option, is, well, optional. 😁

File: docs/source/configuration/slots.md [hunk 16]
```diff
@@ -0,0 +1,89 @@
...
+
+```ts
+export type SlotPredicate = (args: any) => boolean;
+```
+
+There are two predicate helpers available in the Volto helpers.
+
+### RouteCondition
```
stevepiercy, line 61:
```suggestion
### `RouteCondition`
```

docs/source/configuration/slots.md
```diff
@@ -0,0 +1,89 @@
...
+export function RouteCondition(path: string, exact?: boolean) {
+  return ({ pathname }: { pathname: string }) =>
+    Boolean(matchPath(pathname, { path, exact }));
+}
+```
+
+It allows to make a slot render if a certain route is matched.
+It takes the route and if the route match should be exact or not.
+
+### ContentTypeCondition
```
stevepiercy, line 73:
```suggestion
### `ContentTypeCondition`
```

docs/source/configuration/slots.md
```diff
@@ -0,0 +1,89 @@
...
+### RouteCondition
+
+```ts
+export function RouteCondition(path: string, exact?: boolean) {
+  return ({ pathname }: { pathname: string }) =>
+    Boolean(matchPath(pathname, { path, exact }));
+}
+```
+
+It allows to make a slot render if a certain route is matched.
```
stevepiercy, line 70:
```suggestion
It renders a slot if the specified route matches.
```

docs/source/configuration/slots.md
```diff
@@ -0,0 +1,89 @@
...
+It takes the route and if the route match should be exact or not.
```
stevepiercy, line 71: What is the opposite of exact match? begins with or contains? I guess begins with, but please correct me.

```suggestion
It accepts the route as the `path` argument, and whether matching should be exact or begins with the given `path` as the `exact` argument.
```

docs/source/configuration/slots.md
```diff
@@ -0,0 +1,89 @@
...
+### ContentTypeCondition
+
+```ts
+export function ContentTypeCondition(contentType: string[]) {
+  return ({ content }: { content: Content }) =>
+    contentType.includes(content['@type']);
+}
+```
+
+This helper predicate allows you to make a slot render if the current content type is match.
```
stevepiercy, line 82:
```suggestion
The `ContentTypeCondition` helper predicate allows you to render a slot when the given content type matches the current content type.
```

docs/source/configuration/slots.md
```diff
@@ -0,0 +1,89 @@
...
+It takes a list of possible content types.
```
stevepiercy, line 83: It might be helpful to refer to, or provide the list of, possible content types. Do we have that documented somewhere?

```suggestion
It accepts a list of possible content types.
```

docs/source/configuration/slots.md
```diff
@@ -0,0 +1,89 @@
...
+
+### Custom predicates
+
+You can provide your own predicate helpers to determine if your slot component has to render or not.
```
stevepiercy, line 87:
```suggestion
You can create your own predicate helpers to determine whether your slot component should render.
```

docs/source/configuration/slots.md
```diff
@@ -0,0 +1,89 @@
...
+The `SlotRenderer` will pass down the current `content` and the `pathname`.
```
stevepiercy, line 88:
```suggestion
The `SlotRenderer` will pass down the current `content` and the `pathname` into your custom predicate helper.
```

docs/source/configuration/slots.md
```diff
@@ -0,0 +1,89 @@
...
+If that is not enough you could tailor your own `SlotRenderer`s or shadow the original to match your requirements.
```
stevepiercy, line 89:
```suggestion
You can also tailor your own `SlotRenderer`s or shadow the original (original what? need a thingie) to satisfy your requirements.
```

stevepiercy (changes requested):
I asked a few questions, too, so these suggestions will need to be reviewed as onesies.

packages/registry/src/registry.test.tsx
sneridagh: Yes, unfortunately

docs/source/configuration/slots.md
sneridagh: Yeah! Great! @tiberiuichim and I we were doubting with the name. This is the confirmation that we were right!

docs/source/configuration/slots.md
sneridagh: Yes, it's like this.

docs/source/configuration/slots.md
sneridagh: I will continue working on this.

docs/source/configuration/slots.md
sneridagh: Yeah, it's ongoing still. They represent how the order is saved internally, and show that it's a list that can be reordered.

docs/source/configuration/slots.md
sneridagh: ok

packages/coresandbox/src/components/Slots/SlotTest.tsx
```diff
@@ -0,0 +1,12 @@
+import { Container } from 'semantic-ui-react';
```
ichim-david, line 1: @sneridagh This is just for test but given the ability to use a container component do we still want to have imports from semantic with new code that we want to merge?

docs/source/configuration/slots.md
sneridagh: Well, yeah, I think in terms of jargon "route match(es)" is what people uses for "begins with". So I would use it also here.

docs/source/configuration/slots.md
sneridagh: They will be the ones available in your site, could be that you added more or disabled Plone's defaults...

docs/source/configuration/slots.md
sneridagh: The original `SlotRenderer`, it's the direct complement of the sentence, right? We can repeat it.

packages/coresandbox/src/components/Slots/SlotTest.tsx
sneridagh: I know! :) I was reluctant too. What do we put, then? A custom style?

packages/coresandbox/src/index.ts
```diff
@@ -171,6 +174,13 @@ const applyConfig = (config: ConfigType) => {
   config.blocks.blocksConfig.listing = listing(config);
   config.views.contentTypesViews.Folder = NewsAndEvents;
 
+  config.registerSlotComponent({
+    slot: 'aboveContent',
+    name: 'hello.aboveFooter',
```
ichim-david, line 179: This doesn't sit well with me
slot: 'aboveContent',
name: 'hello.aboveFooter'
I'm thinking of aboveFooter as another slot, not a name part of the aboveContent slot

sneridagh: Yeah, leads to confusion. I will amend it.

packages/volto/src/components/theme/View/View.jsx
```diff
@@ -244,13 +245,15 @@ class View extends Component {
               : null
           }
         />
+        <SlotRenderer name="aboveContent" content={this.props.content} />
```
ichim-david, line 248: @plone/volto-team  would be interesting to hear what other slots you think should be added.
I am thinking of plone 4 viewlets managers that we had, if we need to add something to the header
and footer as well or we keep it simple for starters and add extra slots as needed.
So for sure I am hoping that others also give feedback on this work.

packages/registry/src/registry.test.tsx
stevepiercy: Oh, similar to `# noqa` in Python. TIL.

docs/source/configuration/slots.md
stevepiercy: Suggestion updated.

stevepiercy:
I think the single most common use case is a site-wide alert. For example, you are a school in America, and you need to announce this month's shooter on campus. Or you are a government agency in Oregon who needs to notify about this year's ice storm or wildfire.

packages/coresandbox/src/index.ts
sneridagh: pushed

docs/source/configuration/slots.md
stevepiercy: @sneridagh in this example, can you clarify what should happen with the `edit` slot component when both of its predicates return `false`? Originally you said "always appear", but then why would we need predicates for this slot component? Should it instead be "do not display"?

stevepiercy: See suggested revision in c4cfe86ce3aacfd829cde31ea868f7f09026d855

docs/source/configuration/slots.md
stevepiercy: Fixed in 520fc605c45145562c9f796eee3e05952c19aa95. I went with a definition list instead of a glossary to avoid conflict with https://github.com/plone/documentation/pull/1629.

docs/source/configuration/slots.md
stevepiercy: I revised this to be more API-ish, less narrative-ish, in a8f6de3a946d44bd2b6c8a047037a38e28c73b70.

sneridagh:
@stevepiercy awesome! Looks great! THX!

docs/source/configuration/slots.md
sneridagh: I just checked that you already pushed for this change. Resolving.

docs/source/configuration/slots.md
```diff
@@ -134,3 +134,91 @@ It accepts a list of possible content types.
...
+```
+
+`slot`
+:   The name of the slot, where the slot components are stored.
+
+`name`
+:   The name of the slot component inside it's the component we want to unregister.
+
+`position`
+:   The component position that we want to remove in the slot component registration. Use `getSlotComponent` to find out the position of the registered component that you want to remove.
```
stevepiercy: Is the position zero- or one-indexed? I assume zero, but want to verify.

docs/source/configuration/slots.md
```diff
@@ -134,3 +134,91 @@ It accepts a list of possible content types.
...
+config.getSlotComponent(slot: string, name: string): SlotComponent[]
+```
+
+`slot`
+:   The name of the slot, where the slot components are stored.
+
+`name`
+:   The name of the slot component we want to retrieve.
+
+### `unRegisterSlotComponent`
```
stevepiercy, line 213: "un" is not a word, so this is correct camelCase. Would this be a pain to recase throughout?

```suggestion
### `unregisterSlotComponent`
```

stevepiercy (changes requested):
Looks great! I pushed a couple of commits to standardize a pattern for the API, improve grammar, and clean up some nitpicky things.

I have a couple of questions for you to consider, most importantly zero/one-based indexing.

docs/source/configuration/slots.md
sneridagh: zero

JeffersonBledsoe:
Bring on the slots! 🚀 :shipit: 
We started working on a UI for 'slot' editing with blocks (https://github.com/pretagov/volto-slots-editor), hopefully we can update it to support real slots

packages/registry/src/registry.test.tsx
sneridagh: Removed

docs/source/configuration/slots.md
```diff
@@ -0,0 +1,260 @@
...
+  - `edit`
+  - `contents`
+  - `more`
+```
+
+Volto renders slots using the `SlotRenderer` component.
+You can add insertion points in your code, as shown in the following example.
+
+```ts
+<SlotRenderer name="toolbar" content={content} />
```
davisagli, line 59: Why include Renderer in the name and not just call it Slot? It's obviously a renderer, since it's a component.

docs/source/configuration/slots.md
```diff
@@ -0,0 +1,260 @@
...
+## Register a slot component
+
+You register a slot component using the configuration registry:
+
+```ts
+    config.registerSlotComponent({
+      slot: 'toolbar',
+      name: 'save',
+      component: 'this is a toolbar save component with a true predicate',
+      predicates: [RouteConditionTrue('/de')],
```
davisagli, line 72: Below mentions RouteCondition but not RouteConditionTrue. Is something out of date?

docs/source/configuration/slots.md
```diff
@@ -0,0 +1,260 @@
...
+`name`
+:   String.
+    Required.
+    The name of the slot component to reposition in the list of slot components.
+
+`position`
+:   Number.
+    Required.
+    The destination position in the registered list of slot components.
+    The position is zero-indexed.
```
davisagli, line 187: It'd be nice to provide an easy way to put a name before/after another name, or at the end of the list.

packages/volto/src/components/theme/SlotRenderer/SlotRenderer.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+  if (!slots) {
+    return null;
+  }
+
+  return (
+    <>
+      {slots.map((component: React.ComponentType<any>) => {
+        // Weird compilation issue, ^^ that forced to re-declare the type above
+        const id = uuid();
```
davisagli, line 32: Won't this force re-rendering all slots even if their props haven't changed? I think it would make more sense to use the name as the key.

docs/source/configuration/slots.md
```diff
@@ -0,0 +1,260 @@
...
+  html_meta:
+    "description": "Slots are insertion points in the Volto rendering tree structure."
+    "property=og:description": "Slots are insertion points in the Volto rendering tree structure."
+    "property=og:title": "Slots"
+    "keywords": "Volto, Plone, frontend, React, configuration, slots, viewlets"
+---
+
+# Slots
+
+Slots are insertion points in the Volto rendering tree structure.
```
davisagli, line 12: How about:

```suggestion
Slots provide a way for Volto addons to insert their own components at predefined locations in the rendered page.```

(i.e. mention what problem they solve, not just how they operate)

File: docs/source/configuration/slots.md [hunk 35]
```diff
@@ -0,0 +1,260 @@
...
+This concept is inherited from the Plone Classic UI {doc}`plone:classic-ui/viewlets`.
```
davisagli, line 13:
````suggestion
```{note}
This concept is inspired by the Plone Classic UI {doc}`plone:classic-ui/viewlets`.
```
````

File: docs/source/configuration/slots.md [hunk 36]
```diff
@@ -0,0 +1,260 @@
...
+
+
+## Anatomy
+
+Slots are named, and they can contain a list of different slot components.
+Slot components are also named, and they are registered in the configuration registry using a specific API for slots.
```
davisagli, line 19: Link to the specific API instead of just hand-waving at it.

docs/source/configuration/slots.md
```diff
@@ -0,0 +1,260 @@
...
+│   ├── predicates (only appear in `/de/about`)
+│   ├── predicates (only appear if the content type is either a `Document` or `News Item`)
+│   └── no predicates (default when all predicates return `false`)
+├── SlotComponent (`contents`)
+└── SlotComponent (`more`)
+```
+
+At the root of the tree, there is a slot named `toolbar`.
+It contains three slot components, named `edit`, `contents`, and `more`.
+`edit` contains two predicates and a default for when all of its predicates return `false`.
```
davisagli, line 38: I don't understand the default part. How can one slot component have both 2 predicates and no predicates? Doesn't the one with no predicates need to be a separate slot component that is also named "edit"?

docs/source/configuration/slots.md
```diff
@@ -0,0 +1,260 @@
...
+
+Thus, when either the route is `/de/about` or the content type is either a `Document` or `News Item`, then the `edit` slot component would appear in the slot `about`.
```
davisagli, line 40: There isn't a slot `about`. You mean `toolbar`, right?

```suggestion
Thus, when either the route is `/de/about` or the content type is either a `Document` or `News Item`, then the `edit` slot component would appear in the slot `toolbar`.
```

docs/source/configuration/slots.md
```diff
@@ -0,0 +1,260 @@
...
+It would not display elsewhere.
+
+The order in which the components render is governed by the order in which they were registered.
+
+```{todo}
+You can change the order of the defined slot components for a different slot using the API. (pending)
+You can even delete the rendering of a registered slot component using the API. (pending)
```
davisagli, line 47: I see some APIs below for ordering. Does this todo need to be updated?

docs/source/configuration/slots.md
davisagli: (I don't feel strongly about this. I suppose it helps distinguish it from other kinds of objects related to slots.)

packages/volto/src/components/theme/View/View.jsx
davisagli: I would for sure like to have built-in slots above and below the title in the title block.

Header and footer also makes sense conceptually...but then I guess we should move the existing rendering of these into slots, which is probably hard to do without it becoming a breaking change.

davisagli: We should include a list of the built-in slots in the documentation.

docs/source/configuration/slots.md
stevepiercy: @davisagli your suggestion got tweaked out. I think this version is correct:

```suggestion
Slots provide a way for Volto add-ons to insert their own components at predefined locations in the rendered page.
```

davisagli: @stevepiercy oops, thanks, updated

docs/source/configuration/slots.md
stevepiercy: Does the configuration registry API for slots exist yet? https://6.docs.plone.org/volto/configuration/how-to.html

davisagli: @stevepiercy It's currently discussed further down in this file: https://github.com/plone/volto/blob/33b96c9d3e6d1636bc7e6c0224b5b86df61bcf28/docs/source/configuration/slots.md?plain=1#L65

docs/source/configuration/slots.md
stevepiercy: Maybe if I remove the "no predicates", leaving the parent with a default, then that would help? I agree the "no predicates" appearing as a predicate in the tree is confusing.

davisagli: @stevepiercy I'm not confident I understand how it's intended to work, so I'd like @sneridagh's input here.

docs/source/configuration/slots.md
stevepiercy: Agreed.

docs/source/configuration/slots.md
stevepiercy: The wrapping `todo` markup can be removed, but the sentences without `(pending)` should remain. @sneridagh should verify, though.

docs/source/configuration/slots.md
stevepiercy: "add-ons" is hyphenated.

docs/source/configuration/slots.md
stevepiercy: Fixed in [049e16a](https://github.com/plone/volto/pull/5775/commits/049e16a846405392473e99ea08e245e407391443)

stevepiercy:
Again, 404s in the README:

1. [L134] 404 https://memorytwin.com/en  
2. [L141] 404 https://twincreator.com/en

@nzambello we will remove these two links from both the Volto README.md and https://github.com/collective/awesome-volto#websites-built-with-volto unless you update them. Thank you!

packages/volto/src/components/theme/View/__snapshots__/View.test.jsx.snap
```diff
@@ -7,9 +7,15 @@ exports[`View renders a document view 1`] = `
   <div
     id="ContentMetadataTags"
   />
+  <div
+    id="SlotRenderer"
+  />
```
JeffersonBledsoe, line 12: Is there anyway we can actually test things are rendered here rather than just seeing the render wrapper? Seems like a false sense of security with the tests seeing that things 'pass' but rendering could be completely broken.

docs/source/configuration/slots.md
sneridagh: Fixed.

packages/volto/src/components/theme/SlotRenderer/SlotRenderer.tsx
sneridagh: Good catch. I've enhanced what `getSlot` returns to be able to set the key as the name of the component.

docs/source/configuration/slots.md
sneridagh: For `edit` there's three slot components declared, one of them has no predicates defined. If the ones with predicates do not match, the one without predicates is returned. Not sure if call it "default". We can also modify the entry in the tree to be more precise.

docs/source/configuration/slots.md
sneridagh: Removed it.

packages/volto/src/components/theme/View/__snapshots__/View.test.jsx.snap
sneridagh: For the `View` it's better like this. We have a Cypress test for it, so we are "safe" I will add a unit test also for the `SlotRenderer`.

packages/volto/src/components/theme/View/View.jsx
sneridagh: @davisagli I've thought about the above and below title too. However, there's a couple of drawbacks, eg. How about the edit mode? Should they appear too there? How about if you move the title block around? Unlikely, but possible. But I agree that they will solve more problems than bring.

docs/source/configuration/slots.md
stevepiercy: Maybe this?

```
`edit` contains three slot components, two of which have predicates and a third without predicates.
When the slots with predicates both return `false`, then third slot is returned.
```

packages/volto/src/components/theme/View/__snapshots__/View.test.jsx.snap
sneridagh: Done.

docs/source/configuration/slots.md
sneridagh: This is next.

sneridagh: Done.

packages/registry/src/index.ts
```diff
@@ -181,6 +185,214 @@ class Config {
...
+  }) {
+    if (!position && !action) {
+      throw new Error(`At least a position or action is required as argument`);
+    }
+    if (position && target) {
+      throw new Error(
+        `You should provide only one of position or action as arguments`,
+      );
+    }
+    if (action && action !== 'first' && action !== 'last' && !target) {
```
stevepiercy, line 339: This is aligns better with the error message. Otherwise I have to check other values not in the enum. Also by changing the previous error condition, we can remove `action &&`.

```suggestion
    if ((action == 'after' || action == 'before') && !target) {
```

docs/source/configuration/slots.md
```diff
@@ -0,0 +1,295 @@
...
+│   ├── predicate (only appear in `/de/about`)
+│   ├── predicate (only appear if the content type is either a `Document` or `News Item`)
+│   └── no predicate (default when all predicates return `false`)
+├── SlotComponent (`contents`)
+└── SlotComponent (`more`)
+```
+
+At the root of the tree, there is a slot named `toolbar`.
+It contains three slot components, named `edit`, `contents`, and `more`.
+`edit` contains three slot components, two of which have a predicate and a third without a predicate.
```
davisagli, line 41: @sneridagh @stevepiercy I'm still confused by this tree and the description of a slot component that contains 3 slot components, since I don't see a notion of SlotComponents containing SlotComponents in the actual API. It's more that there are SlotComponents which are grouped by name before deciding which one wins for each name using the predicates, right?

Would it help and be accurate to show it like this instead?
```text
Slot (`toolbar`)
├── name=`edit`
│   ├── SlotComponent with predicate (only appear in `/de/about`)
│   ├── SlotComponent with predicate (only appear if the content type is either a `Document` or `News Item`)
│   └── SlotComponent with no predicate (default when all predicates return `false`)
├── name=`contents`
│   └── SlotComponent
└── name=`more`
    └── SlotComponent
```

docs/source/configuration/slots.md
```diff
@@ -0,0 +1,295 @@
...
+When the slots with predicates both return `false`, then the third slot is returned.
+
+Thus, when either the route is `/de/about` or the content type is either a `Document` or `News Item`, then the `edit` slot component would appear in the slot `toolbar`.
+It would not display elsewhere.
```
davisagli, line 45: I don't think this paragraph is accurate. As I understand it, one of 3 SlotComponents with the name `edit` would be picked, based on which predicates match.

packages/volto/src/components/manage/Form/Form.jsx
```diff
@@ -641,131 +642,147 @@ class Form extends Component {
...
-                this.props.setFormData(newFormData);
-              }
-            }}
-            onChangeField={this.onChangeField}
-            onSelectBlock={this.onSelectBlock}
-            properties={formData}
+        <>
+          <SlotRenderer
+            name="aboveContent"
+            content={this.props.content}
```
davisagli, line 648: @sneridagh Should it use the form state if it's an add/edit form? (I'm thinking for things like showing some field value above the title)

Do we need a way to register slot components that are only for view or only for add/edit?

sneridagh: I think it's not necessary, since you can access the form global state from the Slot components directly, add the conditions there, so no need for drill down anything.

davisagli: @sneridagh Hmm, ok. But if you want to use the same slot component for view and edit, how does it know whether to get its data from the form state or from the content?

docs/source/configuration/slots.md
sneridagh: I agree the naming is a bit confusing. First the Slot (toolbar) that contains (named) slot components.
So you can register as many slot components under the same named slot component ('edit'). These registrations can have different predicates or none at all.
Maybe the last two components should also expose a real case, so they are more consistent?

docs/source/configuration/slots.md
sneridagh: @davisagli I propose to read the whole thing together this evening, what do you think @stevepiercy ?

packages/volto/src/components/manage/Form/Form.jsx
sneridagh: We reset the store after edit, so it's empty.

davisagli: I guess if there's any complexity and we need to reuse it in multiple places, we can add a `useContent` hook.

docs/source/configuration/slots.md
sneridagh: @davisagli how ZCA name such things? When you register a named view under other interfaces?

davisagli: @sneridagh I would explain it like this: In each slot, multiple slot components can be registered. Each slot component has a name. If there are multiple slot components registered with the same name, then the predicates are evaluated to decide which slot component will be rendered. The first slot component with a matching predicate will be used.

sneridagh: Sounds great!

docs/source/configuration/slots.md
stevepiercy: Yup, maybe make it a topic of the Documentation Team meeting in Discord, 1:30 AM Pacific?

sneridagh: I think David's proposal below is good!

packages/registry/src/index.ts
sneridagh: Pushed, I run the tests to make sure.

packages/registry/src/index.ts
```diff
@@ -181,6 +185,214 @@ class Config {
...
+    slot: string;
+    name: string;
+    position?: number;
+    action?: 'after' | 'before' | 'first' | 'last';
+    target?: string;
+  }) {
+    if (!position && !action) {
+      throw new Error(`At least a position or action is required as argument`);
+    }
+    if (position && target) {
```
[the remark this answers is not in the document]
sneridagh: Pushed.

docs/source/configuration/slots.md
stevepiercy: Using our example, if what @davisagli wrote is true, then there would be a total of 5 registered SlotComponents. Is that correct? I tried [reading the code](https://github.com/plone/volto/pull/5775/files#diff-5e23fa7ab384750c59928974f281315a01f49d7e8751645b5a253fae32cf7890R236-R296), and it appears to be correct, but I'm not certain due to my lack of fluency in TypeScript.

If so, then our tree diagram should be flattened from 3 levels to 2, as well.

stevepiercy:
@sneridagh I significantly overhauled the Anatomy section, per our conversation yesterday. There are a couple of missing pieces, as noted.

docs/source/configuration/slots.md
```diff
@@ -0,0 +1,379 @@
...
+You can register multiple slot components with the same name under the same slot, as long as they have different predicates or components.
+
+To illustrate how slots are structured and work, let's register a slot component, where the component is `@sneridagh what goes here?`, and the predicate matches a route that begins with `/de/about`.
+
+```ts
+config.registerSlotComponent({
+  slot: 'toolbar',
+  name: 'save',
+  component: '@sneridagh what goes here?',
+  predicates: [RouteCondition('/de/about')],
```
davisagli, line 44: @stevepiercy The example of changing the save button in the toolbar for different predicates seems a bit contrived. How about adding a different header to the aboveContent slot? So the slot would be `aboveContent`, name would be `header`, component would be a hypothetical React component AboutHeader, and the predicates would remain the same.

docs/source/configuration/slots.md
```diff
@@ -0,0 +1,379 @@
...
+    └── predicate of "only appear under `/de/about`"
+```
+
+Next, let's register another slot component in the same slot, with the same name and component, but with a different predicate where the content type matches either `Document` or `News Item`.
+
+```ts
+config.registerSlotComponent({
+  slot: 'toolbar',
+  name: 'save',
+  component: '@sneridagh what goes here?',
```
davisagli, line 65: component: PageHeader

docs/source/configuration/slots.md
```diff
@@ -0,0 +1,379 @@
...
+    └── predicate of "only appear when the content type is either a Document or News Item"
+```
+
+Finally, let's register another slot component in the same slot, with the same name, but with a different component and without a predicate.
+
+```ts
+config.registerSlotComponent({
+  slot: 'toolbar',
+  name: 'save',
+  component: '@sneridagh what goes here as a different component?',
```
davisagli, line 92: component: DefaultHeader

davisagli (commented):
@stevepiercy Thanks, I think the rewritten Anatomy section is clearer.

docs/source/configuration/slots.md
stevepiercy: Agree. I had the idea of an alert that you would post sitewide on a university website for a campus shooter, wildfire, or ice storm, but that was depressingly too realistic and would not require predicates. Your example could apply for departments or other organizations in a university. I'll run with that.

stevepiercy: Done in [40c1dfd](https://github.com/plone/volto/pull/5775/commits/40c1dfd3125bb2ee1594bffcef8560948bbc4130)

docs/source/configuration/slots.md
stevepiercy: Done in [40c1dfd](https://github.com/plone/volto/pull/5775/commits/40c1dfd3125bb2ee1594bffcef8560948bbc4130)

docs/source/configuration/slots.md
stevepiercy: Done in [40c1dfd](https://github.com/plone/volto/pull/5775/commits/40c1dfd3125bb2ee1594bffcef8560948bbc4130)

stevepiercy: Whoops, I mean [275b2c5](https://github.com/plone/volto/pull/5775/commits/275b2c53ea92a21037da01afbe35a4cd12e7a1c7)

docs/source/configuration/slots.md
```diff
@@ -0,0 +1,379 @@
...
+The order in which the slot components render is governed by the order in which they were registered.
+
+You can change the order of the defined slot components for a different slot using the API.
+You can even delete the rendering of a registered slot component using the API.
+
+[@sneridagh, this section needs to be explicit.
+From our conversation yesterday, I think you said that the last slot component registered *per slot* wins.
+Is that correct?
+This has further implications, such as whether this rule applies per slot component's name, component, and predicates. --@stevepiercy]
+```
```
stevepiercy, line 138: This TODO section needs revision, and then I think the docs will be done. If anyone can provide details, I can write it up in narrative form.

stevepiercy (changes requested):

docs/source/configuration/slots.md
```diff
@@ -0,0 +1,407 @@
...
+│   ├── `component`=PageHeader
+│   └── predicate of "only appear when the content type is either a Document or News Item"
+└── SlotComponent
+    ├── `slot`=`aboveContent`
+    ├── `name`=`header`
+    └── `component`=`DefaultHeader`
+```
+
+When the slot components with the same name and component under a given slot have all of their predicates return `true`, then that component will render in the slot.
+Else, if there are slot components with the same name, but with a different component, and without predicates, then that different component will render.
```
sneridagh, line 117: This is not accurate.

sneridagh: The below example is cocrrect though.

docs/source/configuration/slots.md
```diff
@@ -0,0 +1,407 @@
...
+
+Thus the example slot renderer will have the following behavior.
+
+-   When both a user visits the route beginning with `/de/about`, and the content type is either a Document or News Item, then the component `PageHeader` will render in the `aboveContent` slot.
```
davisagli, line 121: No. PageHeader will be used if the route starts with `/de/about` OR if the content type is Document or News Item. Matching either condition is sufficient; they don't have to both be true.

sneridagh:
@stevepiercy let's go together this morning through all of it and let's fix the inconsistencies, I'd like to merge it asap.

sneridagh:
@davisagli waiting for your last input before merging!

davisagli (commented):
@stevepiercy @sneridagh Thanks, the docs look ok to me now

davisagli (approved):

sneridagh:
Thanks everybody for the effort on making this a reality, years in the making, I think it's one of the best additions in the last years to Volto (and not only just Volto, the feature it's on `@plone/registry` so it can be anywhere...

And it shouldn't end here, the persistent slots that @tiberiuichim worked on some years ago is still relevant and valuable, and it's there to resurrect when we are ready to do so.

PR #5794, opened 2024-02-24
Pin eslint-plugin-jsx-a11y to ^6.7.1 everywhere it's referenced

### LINKED ISSUE #5785: eslint-plugin-jsx-a11y pin from Volto  shouldn't  have exact version  pins
Running yarn why eslint-plugin-jsx-a11y returns
├─ @plone/volto@npm:17.15.0
│  └─ eslint-plugin-jsx-a11y@npm:6.7.1 [51660] (via npm:6.7.1 [51660])
│
├─ eslint-config-react-app@npm:7.0.1
│  └─ eslint-plugin-jsx-a11y@npm:6.8.0 (via npm:^6.5.1)

We get the following error which breaks eslint testing
Error: Plugin "jsx-a11y" was conflicted between ".eslintrc.js » ./node_modules/@plone/volto/.eslintrc" and ".eslintrc.js » ./node_modules/@plone/volto/.eslintrc » eslint-config-react-app".

ichim-david:
Fixes #5785 

- Also added the disabling "jsx-a11y/label-has-associated-control" in case it reaches version 6.8.0 as we have failures within Volto


packages/generator-volto/generators/addon/templates/.eslintrc
```diff
@@ -27,7 +27,8 @@
     ],
     "react-hooks/rules-of-hooks": "error",
     "react-hooks/exhaustive-deps": "warn",
-    "react/react-in-jsx-scope": "off"
+    "react/react-in-jsx-scope": "off",
```
sneridagh, line 30: For 18 alpha is not really required, we only need to add an entry to the upgrade guide.

sneridagh: (and fix the violations)

PR #5803, opened 2024-02-26
Removed memori websites that give 404 since they are no longer using Volto

ichim-david:


stevepiercy (changes requested):
https://twincreator.com/en also

stevepiercy:
Would you please also submit a PR in https://github.com/collective/awesome-volto#websites-built-with-volto? That triggers the Marketing Team to update plone.org as well.

ichim-david:
@sneridagh merging this change to the README, I had to remove the links from Nicola since they gave a 404 because those sites are no longer using Volto, they are now using Astro as you can see from the redirect https://memori.ai/en/

PR #5812, opened 2024-02-29
feat(search): show facets count and hide facets values that are not meeting the mandatory criterias in search block 

razvanMiu:


razvanMiu:
Requires https://github.com/plone/plone.restapi/pull/1637

Issue #5772, opened 2024-02-20
Navigation reducer should keep items extra-data sent from the navigation endpoint


tiberiuichim:
The fact that this reducer is so limiting induces using bad development patterns when it comes to custom Volto navigation components. Developers will avoid to enhance the `@navigation` endpoint, for example, to pass extra information from the items, via extra behavior fields.

https://github.com/plone/volto/blob/4e5d508d97f438315b3477d166e26d42797875af/packages/volto/src/reducers/navigation/navigation.js#L33-L37

tiberiuichim:
https://github.com/eea/volto-cca-policy/blob/1d49d2e3c18f37ed9ae642c013c3a94c1387bcd4/src/customizations/volto/reducers/navigation/navigation.js#L32-L40

Hrittik20:
Hi @tiberiuichim, I see that the earlier PR is closed. If this issue is still open then I can make a PR for it.

stevepiercy:
@Hrittik20 please see item 4 in [Things not to do](https://6.docs.plone.org/contributing/first-time.html#things-not-to-do). You are welcome to claim this issue.

PR #5817, opened 2024-03-01
Enhanced navigation reducer in Volto

Hrittik20:
Fixes: #5772


ichim-david:
@Hrittik20 when you add a pull request you need to check the CI tests and find out what fails.
unit-tests
You can see that the unit tests fail for plone/volto.
this means that you need to go inside your branch to packages/volto and run:
pnpm test

jest testing will run and eventually you will get failures.
hit f key and only the failed tests will run.
From there you can make changes to the code or tests and when saving the failed tests will re-run.
This way you will know if what you changed fixed the tests or not.

Hrittik20:
@ichim-david Yes, I'm working on it. I will soon make an updated PR

ichim-david:
> @ichim-david Yes, I'm working on it. I will soon make an updated PR

@Hrittik20 I think the navigation test can avoid being update as long as the id is filtered since we
have the url field.
 @sneridagh what do you think, are we ok with having the id of the item added or
should we remove it?

ichim-david:
> @ichim-david How about If I update `navigation.js` by adding a check within the reducer logic that compares the incoming navigation items with the existing ones based on their URLs?
> 
> ```
> function isSameItem(item1, item2) {
>   return item1.url === item2.url;
> }
> export default function navigation(state = initialState, action = {}) {
>   let hasExpander;
>   switch (action.type) {
>     case `${GET_NAVIGATION}_PENDING`:
>       return {
>         ...state,
>         error: null,
>         loaded: false,
>         loading: true,
>       };
>     case `${GET_CONTENT}_SUCCESS`:
>       hasExpander = hasApiExpander(
>         'navigation',
>         getBaseUrl(flattenToAppURL(action.result['@id'])),
>       );
>       if (hasExpander && !action.subrequest) {
>         const newItems = getRecursiveItems(
>           action.result['@components'].navigation.items,
>         );
>         const hasSameItem = newItems.some(incomingItem =>
>           state.items.some(existingItem => isSameItem(incomingItem, existingItem))
>         );
>         if (!hasSameItem) {
>           return {
>             ...state,
>             error: null,
>             items: newItems,
>             loaded: true,
>             loading: false,
>           };
>         }
>       }
>       return state;
> ```

@Hrittik20 I want to hear what @davisagli  or @sneridagh has to say about this change, we might be ok to have also the id present without having todo any filtering,  or we can simply filter it when destructuring it here https://github.com/plone/volto/pull/5817/files#diff-c1e66aca1aa305458203941b7e7d598522bbecc56ead1841087e2393b93a57b6R37

sneridagh (approved):
I'd like @tiberiuichim and @davisagli to take a final look as well.

packages/volto/src/reducers/navigation/navigation.js
```diff
@@ -34,6 +34,7 @@ function getRecursiveItems(items) {
     title: item.title,
     description: item.description,
```
davisagli, line 35: title and description can be removed here since all of `item` is now included.

```suggestion
```

packages/volto/src/reducers/navigation/navigation.js
```diff
@@ -34,6 +34,7 @@ function getRecursiveItems(items) {
...
     url: flattenToAppURL(item['@id']),
+    ...item,
```
davisagli, line 37: We'll end up with both `url` and `@id`. I guess that is no real problem.

davisagli (approved):
a minor suggestion, but this lgtm

Issue #5341, opened 2023-10-26
[docs] Review and update testing and code quality commands


stevepiercy:
I think Volto's documentation on what commands need to be run needs to be reviewed and updated.

- [Code quality](https://6.docs.plone.org/volto/contributing/index.html#code-quality), which has links to the three pages:
- [Linting](https://6.docs.plone.org/volto/contributing/linting.html)
- [Testing](https://6.docs.plone.org/volto/contributing/testing.html)
- [Acceptance testing](https://6.docs.plone.org/volto/contributing/acceptance-tests.html)

_Originally posted by @stevepiercy in https://github.com/plone/volto/issues/5331#issuecomment-1779955840_

kunal-511:
Please assign this issue to me i would like to work on this

stevepiercy:
@kunal-511 please read and follow [Contributing to Plone](https://6.docs.plone.org/contributing/index.html) and [First-time contributors](https://6.docs.plone.org/contributing/first-time.html).

ichim-david:
I'll try to work on this issue, I think we are too spread on all of these issues, and we need to have a page with A-Z steps on what is needed to get from bug report to pull request that is ready for review within he Contribute to Volto section

stevepiercy:
We should also update `Makefile` with eslint and prettier commands. I had to look them up in https://github.com/plone/volto/blob/main/.github/workflows/code-analysis.yml. Relying on one's editor configuration or https://github.com/plone/volto/blob/main/.husky/pre-commit is not a great option. All those dot files at the root! I have no clue what they do in regards to code quality and linting.

PR #5818, opened 2024-03-01
Fix issue of duplicated blocks upon pasting an image into the Slate E…

aryan7081:
Fixes [Issue](https://github.com/plone/volto/issues/5697) 

Inside the onImageLoad function in withDeserializers extension of Volto Slate.
uploadContent adds the block using Redux and FormData, but the .then promise callback of uploadContent also adds a new Node in the Slate Editor using `Transforms.insertNodes(editor, image);` which eventually leads to adding multiple images in the editor.


packages/volto-slate/news/5818.bugfix
```diff
@@ -0,0 +1 @@
+Fix issue of duplicated blocks upon pasting an image into the Slate Editor. @aryan7081
```
stevepiercy, line 1:
```suggestion
When pasting an image into a Slate block, now only an image block is created, instead of both a Slate block and an image block. @aryan7081
```

stevepiercy (changes requested):

stevepiercy (commented):
News looks good, PR needs review by a maintainer.

stevepiercy (commented):
Change log is OK. Code needs review by a maintainer. @plone/volto-team

ichim-david (changes requested):
@aryan7081 I wonder how did you test that this code change fixes the issue?
I will attach a video where I record testing your fix and the issue is still present
exactly as it was flagged into the issue so I don't understand what made you belive that this issue would be fixed by your change.

https://github.com/plone/volto/assets/152852/bd3b8d7a-4b50-4c9b-b0ea-9932b1e991d9

If you still want to try to work on this issue, your pull request needs to contain
a cypress test that uploads an image, you copy the image to the clipboard and you paste it in a page as you can see from this video:

https://github.com/plone/volto/assets/152852/e192dec8-bbff-42e0-b84b-6609f509f85a

ichim-david:
@aryan7081 A couple of extra context as upon further testing I noticed the following:
1. copy and pasting an image from the website or from the internet results in the correct behavior regardless of your change
2. The code that you disabled indeed prevents 2 images from being uploaded when pasting from the os (MAC OS Sonoma in my case)
3. That is not enough to prevent the slate empty block from being added or from the image being inserted in the first slate block

ichim-david:
> @ichim-david Sorry for the delay. Working on this and adding tests as well. Can I test this on linux and windows vm?

@aryan7081 do the minimum test first on whatever platform you're working on and if you can reproduce the issue there then try to fix it, there is no point in doing cross-testing until necessary.

aryan7081:
The` createImageBlock` function within `packages/volto-slate/src/utils/volto-blocks.js` currently utilizes `addBlock `to insert a new block at the index position plus one. However, the `addBlock` function adds multiple blocks: one for the block being added and another empty block if the type is not 'slate'.
To address this, I've attempted to resolve the issue by modifying the` createImageBlock` function as follows:

```
const currBlockId = properties.blocks_layout.items[index];
const currBlockHasValue = blockHasValue(properties.blocks[currBlockId]);
let id, newFormData;
if (currBlockHasValue) {
  [id, newFormData] = addBlock(properties, 'image', index + 1);
  newFormData = changeBlock(newFormData, id, { '@type': 'image', url });
} else {
  [id, newFormData] = insertBlock(properties, currBlockId, { '@type': 'image', url });
}
```
This change ensures that a block is inserted at the (index - 1) position if the current(index) block is empty. Otherwise, it proceeds to create another block as usual.
Please let me know your thoughts or any further adjustments needed.

https://github.com/plone/volto/assets/71966131/cb836cb1-c490-4525-b06a-015b2e1442da

stevepiercy:
@aryan7081 please run code quality checks locally before pushing commits. See [Linting](https://6.docs.plone.org/volto/contributing/linting.html).

However our documentation lacks the specific commands that need to be run, so I don't blame you for not running undocumented commands. See https://github.com/plone/volto/issues/5341

The CI jobs call https://github.com/plone/volto/blob/main/.github/workflows/code-analysis.yml. You can look at it for the commands that do not pass the checks, and try running them locally.

aryan7081:
@stevepiercy Working on cypress tests.

sneridagh (changes requested):
@aryan7081 I tried it out, and it does work well indeed.

We would need the tests if possible. We improved the documentation lately:
https://6.docs.plone.org/volto/contributing/acceptance-tests.html

Also, the code analysis checks are failing.

Could you imagine in work on this?

sneridagh:
@plone/volto-team This works fine for me, pity is that we don't have any Cypress test for it. I don't even know if it's possible to model that interaction I've investigated it a bit this morning, we have clipboard mocks for pasting text but I have no clue how to mock the image data. I've tried though, if someone wants to take over:

https://github.com/plone/volto/tree/testingpasteimages

For the rest I'd merge it.

@ichim-david I'd appreciate another review from you!

sneridagh (approved):

ichim-david:
@sneridagh will have a look at this tonight.

ichim-david (approved):

ichim-david:
@sneridagh Tonight ended up this morning :)
Tested it again and there is no more duplication of images but I still find some strange behavior
or some differences in behavior at least.

Pasting from  os clipboard will upload and use an image block.
Pasting from the same website or other websites will add an img tag with the link to that source.
Ex:
image-block
img-tag-external
same-web-image

Not saying that it's the fault of this ticket. It just seems to me less useful this img tag linking since you don't have
control over size or alignment like you have when it's uploaded through the image block.

ichim-david:
> @ichim-david paste from a website you mean copy the html with along text or only the image? Is this different than before?

@sneridagh right click copy image only. It's not different from before. 
I guess it would be useful to copy paste alongside text otherwise it's better to drag and image from desktop to get
proper image block usage alongside upload and image options.

Ready for merging from my part.

PR #5823, opened 2024-03-02
[components] Pass down the Popover context, if any, in Select

sneridagh:


packages/components/src/components/Select/Select.tsx
```diff
@@ -63,7 +68,7 @@ export function Select<T extends object>({
           </Button>
           {description && <Text slot="description">{description}</Text>}
           <FieldError>{errorMessage}</FieldError>
-          <Popover>
+          <Popover {...popOverProps}>
```
stevepiercy, line 71: Is the weird casing deliberate? I think it should be either  `PopoverProps` or `popoverProps` as ["popover" is a single word](https://developer.mozilla.org/en-US/docs/Web/API/Popover_API).

```suggestion
          <Popover {...popoverProps}>
```

stevepiercy (commented):
Please check strange casing.

sneridagh:
@stevepiercy thanks!

stevepiercy (approved):
I approve this commit message. 🤣

PR #5840, opened 2024-03-05
fix reordering of slots function

steffenri:


steffenri:
@sneridagh Added more tests for all the use cases to show what the problem is. Also fixed the cause of the problem. There might be a better way to do this though.

sneridagh (approved):

sneridagh:
@davisagli do you know if we can simplify this?

PR #5848, opened 2024-03-06
[components] Provide superpowers to the `Container` component

sneridagh:


packages/components/src/components/Container/Container.tsx
```diff
@@ -1,25 +1,36 @@
...
+) => {
+  const {
+    as: Component = 'div',
+    children,
+    className,
+    layout,
+    narrow,
+    ...rest
+  } = props;
+  const classes = cx('q', 'container', className, { layout, narrow });
```
davisagli, line 32: q for quanta?

I would spell it out, tbh. Too much risk of colliding with some random library that decided to use single-letter class names.

PR #5622, opened 2024-01-14
Fixed #5566 dropdown popup positioning in contents page

kanhaiya04:
Fixed #5566,

Implemented the dropdown with **PopperJS**.

https://github.com/plone/volto/assets/93936630/b91fd706-4536-47dd-9f19-cd4c0bef2712

PR #5571, opened 2023-12-24
fix: dropdown popup positioning in contents page

HarshitVashisht11:
fix the issue #5566 

image

Issue #5566, opened 2023-12-23
UI Issue : dropdown popup positioning in contents page


Mandy767:
**Describe the bug**
https://demo.plone.org/contents on this page when we click on 3 dots dropdown a dropdown popup appears . Due to that dropdown popup unnecessary extra space is there below footer due to the position of dropdown popup .

1. Go to 'https://demo.plone.org/contents'
2. Click on '3 dots dropdown in actions'
3. scroll bar appears at right scroll it
4. at bottom you will find extra space

Video for better understanding:
https://github.com/plone/volto/assets/92856810/e6287494-f50e-4c19-98f5-2d622bdecd86

nimishmedatwal:
Hello, I think I can solve this issue. I am new to Plone and want to start contributing by resolving this.

Mandy767:
Cool 👍

stevepiercy:
The video does not demonstrate the issue, but this screenshot does.

<img width="1840" alt="Screenshot 2023-12-23 at 5 45 53 PM" src="https://github.com/plone/volto/assets/102112/af9c03ec-8a37-4725-999c-7188fb21dd78">

I think that when anything extends the height of the canvas, then either the height of the doormat or the main content container should be extended.

## Extend height of doormat

Note that the padding leaves an unattractive empty space below the grey.

<img width="1840" alt="Screenshot 2023-12-23 at 5 55 39 PM" src="https://github.com/plone/volto/assets/102112/66ab2867-4692-455e-b8ba-6c3c1f6cc81c">

## Extend the height of the page

This example looks cleaner.

<img width="1840" alt="Screenshot 2023-12-23 at 5 57 59 PM" src="https://github.com/plone/volto/assets/102112/68dd23f4-1147-4833-aeae-9d22c5ab4a6b">

The following CSS might not work for non-webkit browsers, so a more comprehensive solution may be needed.

```css
#main > div.ui.basic.segment.content-area {
  height: -webkit-fill-available;
}
```

whoisgautxm:
is this issue open

stevepiercy:
@whoisgautxm what about the GitHub interface is not clear?

<img width="1111" alt="Screenshot 2023-12-24 at 9 02 35 AM" src="https://github.com/plone/volto/assets/102112/d2d56700-cacf-469d-8b91-f67a8aa8d335">

[...]

PR #5863, opened 2024-03-10
Fix Size of SelectMenu in SelectWidget.

### LINKED ISSUE #4058: Add Field modal when editing a content type schema needs resizing.
**Describe the bug**
The "Add Field" modal when editing a content type schema is cut off - Select box is too large for window.

**To Reproduce**
From the control panel -> content types (localhost:3000/controlpanel/dexterity-types ) 

Edit the schema of any content type (three dots at the end of the row, select 'schema')

Click "select" in the type drop-down.

Note that the list is cut off, and the user must scroll the main window to see the last few choices.

**Expected behavior**
The main modal shouldn't scroll.  Only the dropdown should scroll.  You shouldn't have to use 2 scroll bars to get to the "yes / no" field choice.

**Screenshots**
select_too_long

- OS: Ubuntu 20.04
- Browser  Chrome Version 107.0.5304.121 (Official Build) (64-bit)
Volto 16.3.0
Plone 6.0.0rc2
plone.restapi 8.32.4
CMF 2.6.0
Zope 5.7
Python 3.8.10 (default, Jun 22 2022, 20:18:18) [GCC 9.4.0]
PIL 9.3.0 (Pillow)
Add-ons
@plone/volto-slate 16.3.0
@eeacms/volto-accordion-block 6.0.0
@kitconcept/volto-blocks-grid 5.2.0
@kitconcept/volto-slider-block 4.0.0
@eeacms/volto-matomo 3.0.1

victorchrollo14:
Fixes #4058.

Fixes the Size of Select Menu in SelectWidget, whose height was greater than the height of parent modal as a result users had to scroll both the Select menu as well the main modal to view all the field choices. 

The Issue is fixed by passing a menuPortalTarget prop into the Select Menu, that renders the select menu directly in document.body rather than the parent Modal as a result when the height of select menu is greater than the parent modal the users won't have to scroll the select menu as well as the parent modal to view all the field choices.

[Screencast from 08-03-24 05:51:57 PM IST.webm](https://github.com/plone/volto/assets/89346667/db07d246-4017-4804-b8e8-01f18a84b4e8)


packages/volto/news/4058.bugfix
```diff
@@ -0,0 +1 @@
+Fix the size of Select Menu in SelectWidget to prevent users from having to scroll both the main modal and dropdown menu to view the field choices by adding a menuPortalTarget prop. @victorchrollo14
```
stevepiercy, line 1: Change logs should avoid implementation details, as the code covers that.

```suggestion
In a modal with a `SelectWidget`, the parent modal no longer requires scrolling to view the choices in the select's options list. @victorchrollo14
```

stevepiercy (changes requested):
This PR needs review from a maintainer. I only reviewed the change log entry.

packages/volto/news/4058.bugfix
victorchrollo14: Ok. Thank you. 
Shall I wait for the maintainers to review and make all the suggested changes in a single commit.

ichim-david (changes requested):
@victorchrollo14 this is not a good fix, it looks bad to show the select options below the modal which has the grey overlay to bring focus only to the modal content.

I would do the following:
1. set overflow: hidden instead of auto
2. give a max height for the react-select__menu added in this modal to avoid
going over the content area as just having overflow hidden means that URL 
is hidden as seen in my screenshot

overflow-hidden

EDIT:
You could try also without any overflow set on that content parent as hidden might hide something down the line or in responsive mode

victorchrollo14:
@ichim-david. 
Adding ` overflow: hidden ` to the parent component disable the user from seeing all the content in the modal.

Initially I tried adding a maxMenuHeight prop to select component which fixes the issue in `/controlpanel/dexterity-types/`, but it doesn't work in all the places for instances you can check out `/controlpanel/user` where the issue persists. 

[Screencast from 10-03-24 06:39:56 PM IST.webm](https://github.com/plone/volto/assets/89346667/549c6b91-08ff-42c8-a5c9-81fe690aa64a)

I do agree that showing the select options below the modal looks bad. 
The select option is shown below the modal in several places like `/controlpanel/date-and-time/ `, `/controlpanel/site` and ` /controlpanel/language`  etc. 

What do you think would be better to use, should we go with the `menuTargetPortal or use menuMaxHeight ` prop.

image

ichim-david:
> @ichim-david. Adding `overflow: hidden` to the parent component disable the user from seeing all the content in the modal.
> 
> Initially I tried adding a maxMenuHeight prop to select component which fixes the issue in `/controlpanel/dexterity-types/`, but it doesn't work in all the places for instances you can check out `/controlpanel/user` where the issue persists.
> Screencast.from.10-03-24.06.39.56.PM.IST.webm
> 
> I do agree that showing the select options below the modal looks bad. The select option is shown below the modal in several places like `/controlpanel/date-and-time/ `, `/controlpanel/site` and ` /controlpanel/language` etc.
> 
> What do you think would be better to use, should we go with the `menuTargetPortal or use menuMaxHeight ` prop.
> 
> image

@victorchrollo14 I appreciate the fact that you've done research and mentioned other places where we have these react-select input and how it would look, well done, it shows a good level of commitment to thinking of a fix that would be acceptable for more than one place.

That menu does not look good if it goes outside of the content area as it is and is for sure no no in a modal.
popup
What is acceptable to go over the content area is the popup component that is used in the Contents section
where it has some shadows and the icon that triggers it is telling you that you are about to open a menu.

I would rather have the extra scrolling rather than this look as at least that is all enclosed in the section that contains these elements and in the case of add user I find it acceptable to use scrolling to get to the options I need.

We had some changes on the contents panel where a designer added some bottom breathing area for situations
where you would open the popup and it was later removed as a regression because it didn't look right for the
situation where you didn't open the menu.

You can continue to experiment if you like but take into consideration that again it has to look good in order
for the change to be accepted.

victorchrollo14:
hey @ichim-david . I went through most of the pages where the SelectWidget is being used. I added a `maxHeightMenu` prop that would keep the Select menu within the modal. This is what I found out.
1) It fixes the issue in  routes `/controlpanel/dexterity-types/*/schema` along with a similar issue in the homepage, when we click on personal tools -> preferences -> and then try selecting a language.
2) Doesn't cause issues in other places where SelectWidget is used. at least in all the places I looked into. ( I went through all the routes, not sure If I missed any)
3) There are a few exceptions, where although we add a `maxMenuHeight` prop, The select menu height doesn't change, example `/controlpanel/date-and-time`, `/controlpanel/language` etc.

What do you think? Would this be fine or should we try some other alternative solutions for this problem.

image

[Screencast from 11-03-24 10:09:34 PM IST.webm](https://github.com/plone/volto/assets/89346667/72565cf5-13fe-4d79-a89e-1b07f5b802fa)

[Screencast from 11-03-24 10:02:37 PM IST.webm](https://github.com/plone/volto/assets/89346667/1cc60eb9-83e6-4769-96d6-9c1008fc0b58)

ichim-david:
> hey @ichim-david . I went through most of the pages where the SelectWidget is being used. I added a `maxHeightMenu` prop that would keep the Select menu within the modal. This is what I found out.
> 
>     1. It fixes the issue in  routes `/controlpanel/dexterity-types/*/schema` along with a similar issue in the homepage, when we click on personal tools -> preferences -> and then try selecting a language.
> 
>     2. Doesn't cause issues in other places where SelectWidget is used. at least in all the places I looked into. ( I went through all the routes, not sure If I missed any)
> 
>     3. There are a few exceptions, where although we add a `maxMenuHeight` prop, The select menu height doesn't change, example `/controlpanel/date-and-time`, `/controlpanel/language` etc.
> 
> 
> What do you think? Would this be fine or should we try some other alternative solutions for this problem.
> 
> image
> Screencast.from.11-03-24.10.09.34.PM.IST.webm
> Screencast.from.11-03-24.10.02.37.PM.IST.webm

@victorchrollo14 I don't see the maxHeightMenu prop after your latest commits and force-push. look at the changes
from https://github.com/plone/volto/pull/5863/files and tell me if this is the latest status

victorchrollo14:
@ichim-david. I just tried it locally earlier, I've pushed it now.

ichim-david:
> @ichim-david. I just tried it locally earlier, I've pushed it now.

React select has the default max-height of 300px set. You would set it to 13em which I think would be calculated from
14px as the base font size so 182px. It might be too low of a number and I'm thinking of there aren't ways to call the widget for different components with a number that makes sense for it. You also have the option of using CSS selectors for specific sections and modifying the max-height.

I'm not comfortable with thinking this is a good solution to basically say that anytime you use the select widget it has a max height of 182px instead of the default 300. Maybe 200, 250px might be a better default for us but again I am always in favor of not messing a lot with the defaults and modifying those defaults as needed per context.

If we do set the maxMenuHeight  and we decide on a default that we think is more sensible than 300 we still need to ensure that someone can pass a prop  and modify the value.

stevepiercy:
> Initially I tried adding a maxMenuHeight prop to select component which fixes the issue in `/controlpanel/dexterity-types/`, but it doesn't work in all the places for instances you can check out `/controlpanel/user` where the issue persists.

For this scenario, instead of the select menu popping downward, and assuming it uses popper.js, it should automatically detect in which direction it should pop up. Does this select menu not use popper.js? Here are docs for reference.

https://floating-ui.com/docs/react

See also related issue and pull requests:

- https://github.com/plone/volto/issues/5566
- https://github.com/plone/volto/pull/5571 (abandoned, did not pass tests)
- https://github.com/plone/volto/pull/5622 (awaiting review by maintainer)

I agree with @ichim-david that changing defaults would be a last resort because of its impacts throughout the UI.

Finally, what happened to using [`menuPortalTarget={document.body}`](https://github.com/plone/volto/issues/4058#issuecomment-1985609821)?

stevepiercy:
@ichim-david would this example work? https://react-select.com/advanced#portaling. I asked in https://github.com/plone/volto/issues/4058#issuecomment-1985486422, but that seems to have been lost in this PR. @victorchrollo14?

victorchrollo14:
> > @ichim-david. I just tried it locally earlier, I've pushed it now.
> 
> React select has the default max-height of 300px set. You would set it to 13em which I think would be calculated from 14px as the base font size so 182px. It might be too low of a number and I'm thinking of there aren't ways to call the widget for different components with a number that makes sense for it. You also have the option of using CSS selectors for specific sections and modifying the max-height.
> 
> I'm not comfortable with thinking this is a good solution to basically say that anytime you use the select widget it has a max height of 182px instead of the default 300. Maybe 200, 250px might be a better default for us but again I am always in favor of not messing a lot with the defaults and modifying those defaults as needed per context.
> 
> If we do set the maxMenuHeight and we decide on a default that we think is more sensible than 300 we still need to ensure that someone can pass a prop and modify this value.

@ichim-david  yeah, I agree. putting a default value would affect different parts of the UI. if we are not aware of all the places it is being used, we might break some ui as well. 
react-select allows us to add custom css styling by passing in a styles prop. which we can edit in `Widgets/SelectStyling.jsx`. I added a max-height in there, it doesn't work as expected. The only way to add maxHeight properly is to pass the `maxMenuHeight` Prop. 

adding maxHeight in css makes it look like below.
image

In the route `http://localhost:3000/controlpanel/date-and-time`. Although the default max-height of menu in react-select is 300. 
here it exceeds that number to 501px. 
The default height of 300 works until the size of the options prop is 25, but when the options size is greater than 25 then the size of menu goes to 501px. What should we do about this issue? 
image

In order to fix the issue with schema modal. We could go with a default of 215px, which is the maximum  maxMenuHeight we can put for the modal in `http://localhost:3000/controlpanel/dexterity-types/*/schema `  such that the menu stays within the modal. Along with that we can make sure to add a prop inside SelectWidget to modify the maxMenuHeight.

victorchrollo14:
> @ichim-david would this example work? https://react-select.com/advanced#portaling. I asked in [#4058 (comment)](https://github.com/plone/volto/issues/4058#issuecomment-1985486422), but that seems to have been lost in this PR. @victorchrollo14?
I used it in the initial commits for this PR, then removed it as it wasn't really that good of a solution.

PR #5865, opened 2024-03-10
Sort facet values if they are numbers

### LINKED ISSUE #5864: Search block configuration breaks when selecting a Catalog index with integer values
**Describe the bug**

I have a catalog index which contains int values (altitude of some mountains). 

If I want to select this index as a source of a facet in a Search block, the configuration screen breaks.

**Expected behavior**
It should allow selecting the index correctly.

**Additional context**

The point is that [the facet is sorting the values](https://github.com/plone/volto/blob/main/packages/volto/src/components/manage/Blocks/Search/components/Facets.jsx#L84-L86) to present them, and assumes that everything is a string.

erral:
Fixes #5864


davisagli (commented):
Would it make sense to factor this out into a named function that can be easily tested?

erral:
I have extracted the sorting operation to a function and added some basic tests to it.

davisagli (approved):
LGTM

PR #5867, opened 2024-03-11
Fix block chooser search is not focusable when clicked on add button

### LINKED ISSUE #5866: Block chooser search is not focusable when clicked on add button.
Screenshot 2024-03-11 at 8 29 08 PM

iRohitSingh:
Fix #5866 

https://github.com/plone/volto/assets/61353484/a27b4519-5cb0-4f3c-aebb-5b1a89699ddb

### DIFF
--- a/packages/volto/news/5866.bugfix
+++ b/packages/volto/news/5866.bugfix
@@ -1,0 +1,1 @@
+Fix block chooser search is not focusable when clicked on add button @iRohitSingh
--- a/packages/volto/src/components/manage/BlockChooser/BlockChooserSearch.jsx
+++ b/packages/volto/src/components/manage/BlockChooser/BlockChooserSearch.jsx
@@ -35,6 +35,7 @@ const BlockChooserSearch = ({ onChange, searchValue }) => {
           placeholder={intl.formatMessage(messages.search)}
           title={intl.formatMessage(messages.search)}
           ref={searchInput}
+          autoFocus
         />
         {searchValue && (
           <Button


iRohitSingh:
@sneridagh I debugged this issue and in initial implementation there was autofocus but someone removed it. I am not able to pinpoint the person in gitblame. I readded it.

stevepiercy:
@iRohitSingh it was part of this commit by @sneridagh. He did some other things with autofocus. I'm not clear what one has to do with the other.

sneridagh:
@stevepiercy which commit?

autofocus is meh for a11y things... don't know if that had to be with the removal.

/cc @ichim-david

ichim-david:
> @stevepiercy which commit?
> 
> autofocus is meh for a11y things... don't know if that had to be with the removal.
> 
> /cc @ichim-david

@sneridagh I am in favor of adding the autofocus back as the short version.

You've probably read something similar to this mention https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autofocus#accessibility_concerns.
I have seen examples where once you open something you focus to the first  form element that is focusable such as the first input look at react-aria dialog example https://react-spectrum.adobe.com/react-aria/Dialog.html.

For us since it focuses it on the search input if you hit tab you will get to the other block items from the popup which
is not possible right now so not only are you solving the convenience of the sighted users to have the search focused but you can also now use the keyboard to select a block to add.

ichim-david (approved):

stevepiercy:
@sneridagh I can't find it now. I may have been mistaken.

sneridagh:
ok! merging.

sneridagh:
@iRohitSingh @ichim-david I'm reverting this PR because it caused a regression:https://github.com/plone/volto/issues/6389

Let's find an alternative whenever possible.

PR #5875, opened 2024-03-14
(Fix): Sorting Order and Sorting On are not working in Search Block edit mode (#5262)

ichim-david:


dobri1408:
@ichim-david I have fixed the tests!

dobri1408:
The tests didn't pass because I mistakenly duplicated two lines which triggered the search button. This setup functioned in Volto 17 due to its compatibility with those lines, but with minor UI differences in Volto 16, it failed. I've entirely eliminated those two lines since the search button was already tested in a previous test.

sneridagh (approved):

PR #5881, opened 2024-03-14
Add additional parameters to ContentsUploadModal to be reusable in different scenarios

erral:
In a recent project I have found `ContentsUploadModal` very useful, and I have copied it over to provide the end-user with file upload capabilities.

But as I wanted to restrict its use (max file size, only allow images, ...) and I have seen that react-dropzone's `<Dropzone />` allows extra parameters, I have added the option to ContentsUploadModal to pass additional parameters to Dropzone.

This way, one can simply reuse the modal component, which is great.


packages/volto/src/components/manage/Contents/ContentsUploadModal.jsx
```diff
@@ -205,6 +212,32 @@ class ContentsUploadModal extends Component {
...
+    }
+    if (maxSize) {
+      dropzoneOptions['maxSize'] = maxSize;
+    }
+    if (accept) {
+      dropzoneOptions['accept'] = accept;
+    }
+    if (disabled) {
+      dropzoneOptions['disabled'] = disabled;
+    }
```
ichim-david, line 239:
```suggestion
   const dropzoneOptions = {
      multiple,
      minSize,
      maxSize,
      accept,
     disabled,
};
```

@erral you could simplify this code as such and you wouldn't need to have all of these if statements.
I assume that the drag and drop library would know to handle even the options that have null values

erral: Thanks @ichin-david surely it will. I am still a react novice for such structures and still go for if-like thingies. 😅 I will check if the library handles the null values correctly.

erral: I have checked this and indeed the plugin handless the null values correctly, so I have updated the PR. Thanks again for the review.

ichim-david (approved):

packages/volto/news/5881.feature
```diff
@@ -0,0 +1 @@
+Add additional parameters to ContentsUploadModal to be reusable in different scenarios @erral
```
stevepiercy, line 1:
```suggestion
Add parameters to `ContentsUploadModal` to be reusable in different scenarios. @erral
```

stevepiercy (commented):
I reviewed only the change log entry.

stevepiercy (approved):
Approve of news item. Still needs approval from @plone/volto-team.

PR #5885, opened 2024-03-16
Notify in documentation useLinkedHeadings setting change after changes from #5225

ichim-david:


docs/source/configuration/volto-slate/configuration-settings.md
```diff
@@ -245,12 +245,12 @@ slate.runtimeDecorators = [([node, path], ranges) => ranges];
 
 The setting `slate.useLinkedHeadings` controls whether `volto-slate` creates anchors for headings, such as `h1` and `h2`, in the editor.
 
-The default setting is `true`.
+The default setting is `false`.
```
davisagli, line 248: This doesn't match what I see in the code.

It creates anchor links unless the setting is explicitly equal to `false`:
https://github.com/plone/volto/blob/0deb876b34bcd698dc46282884ce5417aeae0fce/packages/volto-slate/src/editor/render.jsx#L179

So if useLinkedHeadings isn't set, it has the same effect as setting it to true.

davisagli (changes requested):

docs/source/configuration/volto-slate/configuration-settings.md
ichim-david: > This doesn't match what I see in the code.
> 
> It creates anchor links unless the setting is explicitly equal to `false`:
> 
> https://github.com/plone/volto/blob/0deb876b34bcd698dc46282884ce5417aeae0fce/packages/volto-slate/src/editor/render.jsx#L179
> 
> So if useLinkedHeadings isn't set, it has the same effect as setting it to true.

@davisagli you are correct the setting is missing, and therefore it is opt-out.
I would add a note that mentions in the documentation for this setting that it will add the anchors only
for logged in users after whatever version made those changes.
What do you think about my idea, would you add this info somewhere else?

davisagli: @ichim-david yeah, it makes sense to clarify that the feature is only for editors

ichim-david:
@davisagli I know that we had these changes https://github.com/plone/volto/pull/5225 due to https://github.com/plone/volto/issues/5174 but it rubs me the wrong way that we make an assumption about this feature being useful only for logged-in users. If you want to enable for everyone you need to customize the render file just to remove the token condition.

What do you think about another option *skipLinkedHeadlinesForAnon* set to true by default?

docs/source/configuration/volto-slate/configuration-settings.md
```diff
@@ -253,6 +253,10 @@ You can opt out of this feature by setting its value to `false`.
 slate.useLinkedHeadings = false
 ```
 
+```{versionchanged} Volto 17.0.0-alpha.28
+This feature is still enabled by default but has a condition to show only for logged-in users
```
stevepiercy, line 257:
```suggestion
This feature is still enabled by default, but has a condition to show only for authenticated users.
```

packages/volto/news/5885.documentation
```diff
@@ -0,0 +1 @@
+Modified `slate.useLinkedHeadings` documentation to mention feature is disabled for logged-in users after #5225 changes. @ichim-david
```
stevepiercy, line 1:
```suggestion
Modified `slate.useLinkedHeadings` documentation to mention feature is disabled for authenticated users after #5225 changes. @ichim-david
```

stevepiercy (changes requested):
Minor grammar and vocabulary suggestions.

stevepiercy (approved):
LGTM, pending resolution of @davisagli's comment.

davisagli (approved):

PR #5186, opened 2023-09-14
Upgrade slate libraries; fix empty link element left hanging when hit enter at end of link

tiberiuichim:
I had to upgrade the slate library, as it introduces an `editor.normalize()` function that can cleanup the links.

There's one styling fix that needs to be done, the new slate lib triggers a `:focus-visible` and needs to be disabled with `outline: 0px`.

image

PR #5898, opened 2024-03-19
Does not show borders in addon block inputs

### LINKED ISSUE #5894: Addon blocks with text input have a black border when editing
**Describe the bug**
Addon blocks with text input have a black border when editing.

**To Reproduce**

1. In an addon, create a block, like Volto's title block.
2. When editing content, add the created block.
3. Place focus on the field to enter text.
4. See the black border around the field

**Expected behavior**
The black border should not appear.

**Screenshots**
download (12)

- OS: Ubuntu 22
- Browser Chrome
- Volto Version 18.0.0-alpha.21
- Plone Version 6.0.10
- Plone REST API Version 9.5.0

**Additional context**
This does not occur in Volto blocks because it is removed via CSS:

https://github.com/plone/volto/blob/accb714bb86f47ef6407e6ae7acf809fab1bd3db/packages/volto/theme/themes/pastanaga/extras/blocks.less#L45-L52

Ref: https://github.com/plone/volto/pull/5186

wesleybl:
We made the css selector more generic, so that it can also be applied to addon blocks.

Fixes #5894

### DIFF
--- a/packages/volto/cypress/tests/core/blocks/blocks-image.js
+++ b/packages/volto/cypress/tests/core/blocks/blocks-image.js
@@ -31,6 +31,10 @@ describe('Blocks Tests', () => {
     cy.getSlate().click();
     cy.get('.ui.basic.icon.button.block-add-button').click();
     cy.get('.ui.basic.icon.button.image').contains('Image').click();
+    cy.get('.block-editor-image [tabindex="0"]')
+      .last()
+      .focus()
+      .should('have.css', 'outline', 'rgb(16, 16, 16) auto 1px');
     cy.get('.block.image .ui.input input[type="text"]').type(
       `https://github.com/plone/volto/raw/main/logos/volto-colorful.png{enter}`,
     );
--- a/packages/volto/cypress/tests/core/blocks/blocks-slate.js
+++ b/packages/volto/cypress/tests/core/blocks/blocks-slate.js
@@ -1,0 +1,25 @@
+describe('Slate Block Tests', () => {
+  beforeEach(() => {
+    cy.intercept('GET', `/**/*?expand*`).as('content');
+    cy.intercept('GET', '/**/Document').as('schema');
+
+    // given a logged in editor and a page in edit mode
+    cy.autologin();
+    cy.createContent({
+      contentType: 'Document',
+      contentId: 'my-page',
+      contentTitle: 'My Page',
+    });
+    cy.visit('/my-page');
+    cy.wait('@content');
+
+    cy.navigate('/my-page/edit');
+    cy.wait('@schema');
+  });
+
+  it('No border in input', () => {
+    cy.get('.block-editor-slate [role=textbox]')
+      .click()
+      .should('have.css', 'outline', 'rgba(0, 0, 0, 0.87) none 0px');
+  });
+});
--- a/packages/volto/cypress/tests/core/blocks/blocks-table.js
+++ b/packages/volto/cypress/tests/core/blocks/blocks-table.js
@@ -23,6 +23,14 @@ describe('Table Block Tests', () => {
     // Edit
     cy.addNewBlock('table');
     cy.wait(2000);
+
+    // No border in input
+    cy.get('.block-editor-slateTable [role=textbox]').should('be.visible');
+    cy.get('.block-editor-slateTable [role=textbox]')
+      .first()
+      .click()
+      .should('have.css', 'outline', 'rgb(135, 143, 147) none 0px');
+
     cy.get(
       '.celled.fixed.table thead tr th:first-child() [contenteditable="true"]',
     )
--- a/packages/volto/cypress/tests/core/blocks/blocks-title.js
+++ b/packages/volto/cypress/tests/core/blocks/blocks-title.js
@@ -1,0 +1,25 @@
+describe('Title Block Tests', () => {
+  beforeEach(() => {
+    cy.intercept('GET', `/**/*?expand*`).as('content');
+    cy.intercept('GET', '/**/Document').as('schema');
+
+    // given a logged in editor and a page in edit mode
+    cy.autologin();
+    cy.createContent({
+      contentType: 'Document',
+      contentId: 'my-page',
+      contentTitle: 'My Page',
+    });
+    cy.visit('/my-page');
+    cy.wait('@content');
+
+    cy.navigate('/my-page/edit');
+    cy.wait('@schema');
+  });
+
+  it('No border in input', () => {
+    cy.get('.block-editor-title [role=textbox]')
+      .click()
+      .should('have.css', 'outline', 'rgba(0, 0, 0, 0.87) none 0px');
+  });
+});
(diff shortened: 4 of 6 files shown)


ichim-david:
@wesleybl you wrote a cypress test to show some css changes, man that is dedication :) 
I salute you.

tiberiuichim (approved):

davisagli:
> @wesleybl you wrote a cypress test to show some css changes, man that is dedication :) I salute you.

I appreciate the effort but I would be careful about doing a lot of this. Each new cypress test adds to the amount of time it takes to run the acceptance tests on every pull request. IMO It would make more sense to add the assertion at the end of an existing test that already creates a table block, instead of adding a new test that does the steps to create the block again.

wesleybl:
@davisagli testing times are already high. Having a few more seconds won't make much difference to the total time. We often let the PR tests run and go have a drink of water. I think this is better than "contaminating" existing tests with something that is not related to that test.

ichim-david:
> @davisagli testing times are already high. Having a few more seconds won't make much difference to the total time. We often let the PR tests run and go have a drink of water. I think this is better than "contaminating" existing tests with something that is not related to that test.

@wesleybl I know I praised your effort and although I got @davisagli again somehow to disagree with my judgment :), I actually tend to think he is right regarding checking if an existing test and adding another assert statement.
If we look at cypress best practices documentation we have
https://docs.cypress.io/guides/references/best-practices#Creating-Tiny-Tests-With-A-Single-Assertion
Here you have the following statement
"It is common for tests in Cypress to issue 30+ commands. Because nearly every command has an implicit assertion (and can therefore fail), even by limiting your assertions you're not saving yourself anything because any single command could implicitly fail."

wesleybl:
@davisagli @ichim-david Okay I'll change that. In any case, some blocks did not have tests. So the new test will have to be small in these cases.

wesleybl:
@davisagli @ichim-david done.

davisagli (commented):
My other doubt here: maybe there are blocks that expect the default focus outline to remain visible?

tiberiuichim:
@davisagli the blocks already have their own border indicator for "active/focused block"

davisagli:
@tiberiuichim then it makes sense when the slate editor covers the entire area of the block, but that's not always the case.

ichim-david:
> @davisagli the blocks already have their own border indicator for "active/focused block"

@wesleybl @tiberiuichim Actually @davisagli is right, it seems I am becoming his cheerleader again :).
Although @wesleybl your intention is good the problem is that
[class^="block-editor-"] :focus-visible mean that any children inside the block-editor-* divs should have 0 outline in focus-visible.

I have made a video that showcases the effect of this rule on the focus indicator when trying to tab around.
You will see that we lose the outline for the drag and drop icon and inside the image block to know that we are
inside it.
If we have problems with input and outline maybe we could restrict the removal of the outline only to them
instead of having a generic removal since you will be affected more than you realize.


https://github.com/plone/volto/assets/152852/2cfc9e05-ed17-4caa-96cc-0e4ba5f0c703

ichim-david (changes requested):
@wesleybl see my comment https://github.com/plone/volto/pull/5898#issuecomment-2011367498

sneridagh:
I am +1 with @ichim-david we have to be careful about a11y in these ones. Not an expert here, I can't help much. When in doubt, I always rely on taking a look how the Adobe guys do it in RAC.

wesleybl:
@davisagli @ichim-david Ok. Can I create a new class "block-editor-no-outline" and add this class to blocks that cannot have an outline? So I can use this class in addos.

wesleybl:
> Can I create a new class "block-editor-no-outline" and add this class to blocks that cannot have an outline?

This could be done with a new `classesInEdit` block setting. What do you think?

ichim-david:
> > Can I create a new class "block-editor-no-outline" and add this class to blocks that cannot have an outline?
> 
> This could be done with a new `classesInEdit` block setting. What do you think?

@wesleybl I am going to defer to the decision of @davisagli, @sneridagh and @tiberiuichim, naming is hard and if we are to introduce new block settings I am not comfortable saying yay or nay.

I am looking better at the issue you've added and the outline problem that was flagged by @tiberiuichim  in the slate update
https://github.com/plone/volto/pull/5186/files#r1358197824
it would make sense to have the ability to have a generic slate class like tibi asked here and if that is present on that one you remove the outline.

At the same time you run into issues with an add-on input and that can easily be fixed by having a simple css rule in your blocks.less equivalent and I don't know if it's worth trying todo something about it in core.

As we get into comments about pros on cons we might run into situations where we deem that it's better not to introduce some change such as the pull request with changing the toolbar icon from user to the cog.

I hope that we can avoid this scenario, I would think that it would be useful to see how to add a generic class to slate fields but again maybe the other core developers have a better suggestion than what I write here, so let's wait and see what they will say.

wesleybl:
> I would think that it would be useful to see how to add a generic class to slate fields

@ichim-david the Slate doesn't have a class, but I can pass the `style` property to it:

https://github.com/ianstormtaylor/slate/blob/c2ae1eda91d0aae1cd63bd46af759c542c292a8a/packages/slate-react/src/components/editable.tsx#L118

I think I'll do it like this.

wesleybl:
> the Slate doesn't have a class, but I can pass the style property to it

humm but that wouldn't solve my problem that the all Slate in Volto doesn't have border.

wesleybl:
@ichim-david @davisagli @tiberiuichim the Slate has no class but has a div with the `data-slate-editor` attribute. So I used this attribute in the CSS selector to identify the Slate. I also added a test to ensure the border appears in the image block.

Can you take a look please?

ichim-david:
> @ichim-david @davisagli @tiberiuichim the Slate has no class but has a div with the `data-slate-editor` attribute. So I used this attribute in the CSS selector to identify the Slate. I also added a test to ensure the border appears in the image block.
> 
> Can you take a look please?

@wesleybl it's Friday night 21:15PM so no test for me tonight we'll test it tomorrow morning.
Using the has selector to make something generic is clever, I don't know if we need to bother anymore about support for has selector in browsers, maybe the existing selective skin be kept for the older browsers who might not have support for has.

wesleybl:
> I don't know if we need to bother anymore about support for has selector in browsers, maybe the existing selectors can be kept for the older browsers who might not have support for has.

@ichim-david I changed the code only to:

```CSS
[data-slate-editor='true'] {
  outline: none;
}
```

This works and we don't need to use has.

ichim-david:
> > I don't know if we need to bother anymore about support for has selector in browsers, maybe the existing selectors can be kept for the older browsers who might not have support for has.
> 
> @ichim-david I changed the code only to:
> 
> ```css
> [data-slate-editor='true'] {
>   outline: none;
> }
> ```
> 
> This works and we don't need to use has.

@wesleybl indeed it does, I tested locally with volto-18 and every block available and I had focus-visible present where needed as such this is the cleanest possible fix without crazy selectors and messing with the focus-visible.
Accepting the work from my part, to be determined the judgment of the other reviewers.

ichim-david (approved):

wesleybl:
@ichim-david thanks for your help!

sneridagh (approved):
LGTM, but asking the @davisagli for a last review.

davisagli (approved):
OK with me.

PR #5900, opened 2024-03-21
Update middleware.ts

aliomarmohamed:
I want to commit this changes


davisagli:
@aliomarmohamed Please don't submit multiple pull requests with the same change. This is not helpful.

PR #5924, opened 2024-03-26
Update volto.po "yes" translation si -> sì

yurj:
minor fix


pnicolli (approved):

packages/volto/news/5924.bugfix
```diff
@@ -0,0 +1 @@
+Update volto.po "yes" translation si -> sì. @yurj
```
ichim-david, line 1:
```suggestion
Update italian locales volto.po "yes" translation si -> sì. @yurj
```

ichim-david: I think it's better for us to mention which locales volto.po updates the yes translation as it might not be obvious for someone looking simply at the changelog which yes that belongs to

ichim-david:
@yurj thank you  for your contribution.

PR #5926, opened 2024-03-26
fix(slate): fix insert/remove element edgecase bug in slate

### LINKED ISSUE #5928: SlateEditor error when inserting/removing custom inline element in very specific ways
**Describe the bug**
In a slate block you can add text and you can wrap that text with block or inline elements. An element in slate is a type of node that has a specific renderer. We have default  [elements](https://github.com/plone/volto/blob/main/packages/volto-slate/src/editor/config.jsx#L234-L268) but we can also add custom elements. As an example [volto-slate-footnote](https://github.com/eea/volto-slate-footnote) is a custom inline element that allows wrapping some text with a footnote.
For understanding the issue let's say that the slate block render the following text: `this is a text` and the custom inline element applied over a portion of the text will result in: `this is a <CustomInlineElement>text</CustomInlineElement>`

**Bug 1**: when the cursor is at the end of a custom inline element, but not in the element itself, I see the unwrap button in the toolbar. If I want to remove (unwrap) it, pressing the unwrap button, I get the following error:
```
Uncaught TypeError: link is not iterable
    at utils.js:76:1
```
The error comes from the [_unwrapElement](https://github.com/plone/volto/blob/main/packages/volto-slate/src/elementEditor/utils.js#L76) function because indeed there is no [match](https://github.com/plone/volto/blob/main/packages/volto-slate/src/elementEditor/utils.js#L74) for that element when the cursor is at the end however I still see the unwrap button in the toolbar because if the cursor is at the end of a custom inline element SlateEditor behave as if that element is [active](https://github.com/plone/volto/blob/main/packages/volto-slate/src/elementEditor/utils.js#L97-L135).

_bug_remove_element

**Bug 2**: In a slate table when trying to wrap some text with a custom inline element I get the following error:
```
Uncaught Error: When setting the selection and the current selection is `null` you must provide at least an `anchor` and `focus`, but you passed: {"anchor":null,"focus":null}
    at utils.js:53:1
```
The error comes from [_insertElement](https://github.com/plone/volto/blob/main/packages/volto-slate/src/elementEditor/utils.js#L53) function. From my testing it seems like after wrapping the text with the element the new [selection](https://github.com/plone/volto/blob/main/packages/volto-slate/src/elementEditor/utils.js#L52) is ok but then when [applying](https://github.com/plone/volto/blob/main/packages/volto-slate/src/elementEditor/utils.js#L53) it to the editor it sees it as if being null so maybe a solution for this would be to set the selection asynchronous, in a setTimeout.

_bug_insert_element

**To Reproduce**

**Bug 1:**

1. Add some text in a slate block
2. Wrap the last word with a custom inline element
3. Move the cursor at the end of the word that was wrapped
4. You will see the context toolbar of that custom inline element allowing you to unwrap the element
5. Press the unwrap button

**Bug 2:**

1. Add some text in a cell in a slate table block
2. Wrap that text with a custom inline element

razvanMiu:
Closes #5928


stevepiercy:
@razvanMiu which issue does this PR fix? See Item 2, second bullet point in [Create a pull request from your fork](https://6.docs.plone.org/contributing/first-time.html#create-a-pull-request-from-your-fork).

packages/volto-slate/news/5926.bugfix
```diff
@@ -0,0 +1 @@
+Fix insert/remove element edgecase bug in slate. @razvanMiu
```
stevepiercy, line 1: What edge case does this fix? Please describe the edge case, and how does what you did fix it.

stevepiercy (changes requested):

stevepiercy:
@razvanMiu at least copy and paste the error stack into the PR, or better yet, create an issue with the error stack after searching the issue tracker for the error message. Someone else might have reported it or is already working on it, which could save you some time and effort.

I can write up a description, once you paste the error stack, so I don't have to transcribe the videos.

Sometimes the English language can get in the way of doing good work, and if so, if it's easier to write it in your preferred language, we can use a translation service to translate it to English.

razvanMiu:
> @razvanMiu at least copy and paste the error stack into the PR, or better yet, create an issue with the error stack after searching the issue tracker for the error message. Someone else might have reported it or is already working on it, which could save you some time and effort.
> 
> I can write up a description, once you paste the error stack, so I don't have to transcribe the videos.
> 
> Sometimes the English language can get in the way of doing good work, and if so, if it's easier to write it in your preferred language, we can use a translation service to translate it to English.

I created an [issue](https://github.com/plone/volto/issues/5928)

stevepiercy:
@razvanMiu thank you! That's an excellent write-up.

@plone/volto-team all green and ready for review!

tiberiuichim (approved):

Issue #5677, opened 2024-01-21
Pagination component in file upload modal 


Ayush-2301:
**Is your feature request related to a problem? Please describe.**

When uploading too many files to a folder content of a page, the modal height increases, affecting the user experience. The 'Save' or 'Cancel' button may also be pushed down.

**To Reproduce**

As admin
Go to [demo.plone.org](demo.plone.org)
Navigate to folder_contents.
Drag and drop or add multiple files.

**Describe the solution you'd like**

A pagination component is need to be added in the file upload modal 

**Additional context**

https://github.com/plone/volto/assets/114101923/2b9946b6-709a-4ac8-b964-4e628e45f57d

Ayush-2301:
working on it

stevepiercy:
> working on it

@Ayush-2301 clearly you did not read and follow item 2 in https://6.docs.plone.org/contributing/first-time.html#work-with-github-issues.

AKA717:
users has to scroll all the way down to hit the upload button.

stevepiercy:
@ichim-david @Ayush-2301 sorry, I meant to say item 3, not 2, in [Work with GitHub issues](https://6.docs.plone.org/contributing/first-time.html#work-with-github-issues).

Discussing what the correct implementation would be, including the possibility of reusing existing components, instead of reinventing the wheel, or worse, adding a new dependency, could save time for both contributors and reviewers.

As far as racing in, "Fools rush in where angels fear to tread".

stevepiercy:
@ichim-david the Contributors Team (read role) lacks a lot of permissions that the Developers Team (write role) have, and assignment of issues is one of those. Here's the complete list of permissions for all roles. https://docs.github.com/en/organizations/managing-user-access-to-your-organizations-repositories/managing-repository-roles/repository-roles-for-an-organization#permissions-for-each-role. We wish GitHub would grant customization of these roles for open source organizations, but they reserve it for paid Enterprise accounts only.

ichim-david:
@Ayush-2301 before rushing in on working on this issue, as @stevepiercy mentioned I think you should wait for an extra comment on the validity of this enhancement as I know there were some pull requests for folder_contents with enhancements plus other issues so I suggest you wait a bit longer before you start working on this.

Ayush-2301:
[...]

Issue #5423, opened 2023-11-22
PLIP: TUS reliable large uploads with simpler file upload UI in volto


djay:
## Abstract

This PLIP has three aims (and could be separated if desired)
- Implement TUS in volto for "contents > upload" with a UI similar to quanta design (inc progress bars)
    - pick a js tus client to use in core
    - brings parity with classic
- Implement TUS uploads when editing files, ie TUS support for any dexterrity content type with file fields and for images/videos fields in blocks
   - otherwise its hard to explain that large uploads can only be done via content > uploads
- Simplify new file object creation UI by replacing "Add > File" with "Add > Upload files" (as per quanta design)
   - because File objects are not an obvious way to add files

Context
- Classic had only support for TUS in the contents file upload. This seems to be currently broken?

## Motivation

- As an installer I can enable optional TUS support to enable large uploads and choose chunk size and temp folder to suit deployment but UX for the editors mostly the same
- As an editor it should be obvious and quick to bulk upload files
- As an editor, if the connection is slow or intermittent I should have clear indications upload is working, and options to resume if upload was not successful and this should apply to all places in the UI I can upload a file including editing a file object or in blocks.
- As Ops, I can enable POST size limits and still allow larger uploads.

## Assumptions

- TUS requires more complex installation so will be turned off unless installer knows what they are doing
- ```Folder contents > upload``` is a currently “hidden feature”
- ```Add > File``` is too many clicks for most people and it’s not obvious the title is optional

### TUS for contents upload
- enabled by env variables
   - frontend TUS support will be enabled with TUS_CHUNK_SIZE environment variable
      - otherwise file will be sent as a whole and not via TUS since there is no guarantee the tmp dir will be the same between requests. so backend might also need TUS_TMP_FILE_DIR to be set in addition.
- propose to use [tus-js-client](https://github.com/tus/tus-js-client) and semantic UI.
   - lighter. no UI. reference implementation
   - [60kb minified?](https://bundlephobia.com/package/tus-js-client@3.1.1)
- code will be borrowed for two existing attempts. Both branches replace the only contents upload page
    - using [Uploady](https://github.com/rpldy/react-uploady) - Pretagov
        - Unreleased 
        - uses semantic UI progress bars
        - minified 50kb? TODO: actual bundle size change
        - latest version incompatible with volto
        - MIT License
    - Uppy - Carlos de la Guardia, Franco
        - https://github.com/plone/volto/tree/contents-uppy-upload
        - https://uppy.io/
[...]

Issue #4332, opened 2023-01-28
Quanta


sneridagh:
## Abstract

Implement in Plone Volto the Plone's Design System (aka Quanta). Improve the Theming story by providing separation of concerns between Public Theme and CMSUI.

<img width="762" alt="image" src="https://user-images.githubusercontent.com/486927/215257302-a7a6cf85-3904-4366-bbb2-ad96fd001735.png">

Quanta mainly represents these common look and feel and UX that the CMSUI should have.

### Figma mockups

https://www.figma.com/file/lWRIiKmcy1kMRl1DgLkCSL/Quanta?t=pyh04lHU2vmBer8g-6
https://www.figma.com/file/YpaRSKjFcFaAkVYnTSVGuz/Volto-UI-%E2%80%93-Quanta?t=pyh04lHU2vmBer8g-6

![image](https://user-images.githubusercontent.com/486927/215260144-5ed844d9-2c16-4953-bf92-cbbcb875076b.png)
![image](https://user-images.githubusercontent.com/486927/215260233-e7a468d1-282d-47f5-85fa-bb695e928849.png)

### Plone Conference 2021 Talk
https://www.youtube.com/watch?v=fvTMw36YT6A

### Slides

https://www.slideshare.net/sneridagh/meet-quanta-plones-style-guide

## Motivation

It's been a while since Albert Casado created Quanta specifically for Volto. Some small parts and ideas where already implemented, but the main bulk of the design is yet to be done. It has been decided since then that Quanta will be next Plone's design system and it should be implemented. 

At the same time, there is a long requested feature that is the re-think of the whole Plone Volto theming system. It is desired to have complete separation of concerns between the public theme and the Volto CMSUI itself. Also, do not rely on any specific design system in the public part. By doing this, we will enable the possibility to use any design system or CSS framework of the election of the integrator/themer independent from the one used in the CMSUI.

## Assumptions

During these years, we have worked in providing all the bits and pieces that will make this happen in a sane and pluggable way. If possible, to ease the transition, themes still developed with the current Volto design system (Pastanaga over SemanticUI) should be still possible. The CMSUI will have to migrate to Quanta once in place.

## Proposal & Implementation

Provided as a Volto core package, see the PoC in:

https://github.com/plone/volto/pull/3877

Making heavy use of the Volto Component Registry, one can implement Quanta in the top of the current Volto. Could be that at some point, and specially when introducing the separation public theme/CMSUI could be that this will be no longer possible.

[...]

Issue #4206, opened 2023-01-01
Implement Quanta Toolbar


sneridagh:
## Abstract

The Quanta Design System for Volto features a common interface to interact with blocks in the block engine, called the Quanta Toolbar. It's a Volto CMSUI artifact that acts as a placeholder of buttons, handlers and menus that allows the user to interact with a block's settings.

## Motivation

The current implementation of the "move" and "delete" handler is showing its age, and limitations. The move to a thing that can hold more CMSUI real estate is due since a very long time. Also, the Quanta toolbar implementation will solve other long standing issues, like the placement of additional controls for the blocks, and the block container use case UX.

## Assumptions

There is no special precondition, the Quanta toolbar implementation will happen on the top of the existing one.

## Proposal & Implementation

There is a PoC initial implementation led by @tiberiuichim in this PR: #3161. The proposal is to include the feature as experimental in Volto 16, and make it default once the implementation and initial feedback is reviewed and initial issues fixed. The implementation should provide standard components so a block can interact with its own Quanta Toolbar, add more settings or modify the common ones.

<img width="519" alt="image" src="https://user-images.githubusercontent.com/486927/164911168-52fa8be4-2df1-4308-9506-5f1a9efac8a2.png">

This implementation could be tied to the one happening in #4204 , and should be adjusted accordingly, depending on which one is happening first.

Please observe that the Quanta Toolbar is coupled with the last (+) icon in the engine, and we could also experiment to include a visual handler on hover for the insert inline a block between blocks (a la WP).

If possible, the CSS should not be entangled or tightly coupled to SemanticUI components/classnames. (see next section)

### Mockups

https://www.figma.com/file/YpaRSKjFcFaAkVYnTSVGuz/Volto-UI-%E2%80%93-Quanta?node-id=1782%3A470&t=ojoCCLXSYoeVkmdw-4
https://www.figma.com/file/YpaRSKjFcFaAkVYnTSVGuz/Volto-UI-%E2%80%93-Quanta?node-id=1782%3A579&t=ojoCCLXSYoeVkmdw-4
https://www.figma.com/file/YpaRSKjFcFaAkVYnTSVGuz/Volto-UI-%E2%80%93-Quanta?node-id=1790%3A868&t=ojoCCLXSYoeVkmdw-4

### Optional

[...]

PR #5929, opened 2024-03-27
Replace spinner with Progress bar and fix it's position . (#5620)

### LINKED ISSUE #5620: Loading indicator (spinner) badly positioned when uploading hundreds of images via folder contents. Replace spinner with upload progress indicator (x of y files uploaded).
**Describe the bug**

When a user uploads hundreds of images (e.g. the Pastanaga icons set https://github.com/plone/pastanaga-icons/tree/main/Icons) she does not see any indication that the site is uploading images in the background. For the user it seems like the site is frozen.

**To Reproduce**

1. As admin
2. Go to demo.plone.org
3. Go to folder_contents (https://demo.plone.org/icons/contents) and drag and drop the Pastanaga UI icon set
4. Click the save button (right arrow)
5. -> there is no spinner visible, the page looks frozen

**Expected behavior**

Show a spinner that shows that the site is still loading. In an ideal world, a progress indicator would show how many files have been uploaded already (e.g. 4 of 345 files have been uploaded).

**Screenshots**

Inhalte (18)

victorchrollo14:
Fixes #5620.

* Replaced the spinner in contents -> upload, With a progress bar that indicates the number of files uploaded. ( Uploaded x out of y files ).
* Fixed the bad positioning of the spinner when a large number of files were uploaded by add a `max-height` and `overflow: auto` which fixes issue ( #5677 ).

### To Implement the progress bar.
* Added a new property of `uploadedFiles` to the contents redux state.
* Added a 'UPDATE_UPLOADED_FILES' type to ActionTypes.js, an action creator and a reducer function for the same.
* Added a local totalFiles state inside `ContetsUploadModal`.
* To update the number of files uploaded we dispatch the `updateUploadedFile`  action inside the redux api middleware, after each of the upload requests are fullfilled. and finally set the uploadedFiles to 0 after all the requests are fulfilled. 
* Added unit tests for the `updateUploadedFile` action creator and the reducer as well.

[Screencast from 26-03-24 06:57:09 PM IST.webm](https://github.com/plone/volto/assets/89346667/28a0f296-0dad-4968-b8a5-b3190f0d2dc9)

### DIFF
--- a/packages/volto/locales/de/LC_MESSAGES/volto.po
+++ b/packages/volto/locales/de/LC_MESSAGES/volto.po
@@ -3986,7 +3986,7 @@ msgstr "Laden Sie ein Lead-Bild im Feld 'Lead-Bild' hoch."
 #. Default: "Upload files"
 #: components/manage/Contents/ContentsUploadModal
 msgid "Upload files"
-msgstr "Datei hochladen"
+msgstr "Dateien hochladen"
 
 #. Default: "Uploading image"
 #: components/manage/Blocks/Image/Edit


stevepiercy:
I restarted the failed CI check, and we'll [see if it passes](https://github.com/plone/volto/actions/runs/8454968566/job/23173728410?pr=5929) this time.

stevepiercy (commented):
That's great that you took the extra effort to provide translations to all those languages. I'll request review from the suggested folks, but anyone who has an interest in getting this merged is welcome to add a review. Thank you!

stevepiercy:
CI is all green. Let's get a review from @plone/volto-team.

victorchrollo14:
> That's great that you took the extra effort to provide translations to all those languages. I'll request review from the suggested folks, but anyone who has an interest in getting this merged is welcome to add a review. Thank you!

Hey @stevepiercy. I didn't add the translations, the earlier message of 'uploading files' got automatically removed after I made the changes, I thought this is normal and committed those changes. 

I'll add the translations and push it up then.

giuliaghisini (approved):

packages/volto/src/components/manage/Contents/ContentsUploadModal.jsx
```diff
@@ -211,13 +215,13 @@ class ContentsUploadModal extends Component {
...
-              <FormattedMessage
-                id="Uploading files"
-                defaultMessage="Uploading files"
-              />
-            </Loader>
+            <Progress
+              className="progress-bar"
+              value={this.props.uploadedFiles}
+              total={this.state.totalFiles}
+            >{`Uploading: ${this.props.uploadedFiles} Files Out Of ${this.state.totalFiles}`}</Progress>
```
ichim-david, line 224: @victorchrollo14 you need to add the message with formatMessage using several attributes.
Here is an example of how it's done within Sharing.jsx
https://github.com/plone/volto/blob/main/packages/volto/src/components/manage/Sharing/Sharing.jsx#L90
https://github.com/plone/volto/blob/main/packages/volto/src/components/manage/Sharing/Sharing.jsx#L434
and how it looks when the message is extracted
https://github.com/plone/volto/blob/main/packages/volto/locales/en/LC_MESSAGES/volto.po#L407

We do not want the message to be hardcoded to english since Volto is a multi language framework.

ichim-david (changes requested):

packages/volto/src/components/manage/Contents/ContentsUploadModal.jsx
victorchrollo14: yeah @stevepiercy mentioned that, I'll add that up.

packages/volto/locales/ca/LC_MESSAGES/volto.po
```diff
@@ -3979,16 +3979,16 @@ msgstr "Pengeu una imatge principal al camp de contingut 'Imatge principal'."
...
-#: components/manage/Contents/ContentsUploadModal
-msgid "Uploading files"
-msgstr "Carregant fitxers"
-
 #. Default: "Uploading image"
 #: components/manage/Blocks/Image/Edit
 msgid "Uploading image"
 msgstr "S'està carregant la imatge"
 
+#. Default: "Uploading {uploadedFiles} Files Out of {totalFiles}"
```
stevepiercy, line 3987: For all instances of this phrase throughout the pull request, not just this one suggestion and file, please use sentence casing.

```suggestion
#. Default: "Uploading {uploadedFiles} files out of {totalFiles}"
```

packages/volto/locales/de/LC_MESSAGES/volto.po
```diff
@@ -3978,16 +3978,16 @@ msgstr "Laden Sie ein Lead-Bild im Feld 'Lead-Bild' hoch."
...
-
 #. Default: "Uploading image"
 #: components/manage/Blocks/Image/Edit
 msgid "Uploading image"
 msgstr "Bild hochladen"
 
+#. Default: "Uploading {uploadedFiles} Files Out of {totalFiles}"
+#: components/manage/Contents/ContentsUploadModal
+msgid "Uploading {uploadedFiles} Files Out of {totalFiles}"
+msgstr "Hochladen von {uploadedFiles} Dateien von insgesamt {totalFiles}"
```
stevepiercy, line 3989:
```suggestion
msgstr "Hochladen von {uploadedFiles} dateien von insgesamt {totalFiles}"
```

packages/volto/news/5620.bugfix
```diff
@@ -0,0 +1 @@
+Replaced the spinner with a progress bar that shows the number of files uploaded (Uploaded: x out of y files) and positioned the progress bar properly while uploading a large number of files. @victorchrollo14
```
stevepiercy, line 1:
```suggestion
Replaced the spinner with a progress bar that shows the number of files being uploaded (Uploading x files out of y), and positioned the progress bar properly, while uploading a large number of files. @victorchrollo14
```

stevepiercy (changes requested):

packages/volto/locales/ca/LC_MESSAGES/volto.po
victorchrollo14: okay. I'll change this up.

stevepiercy (approved):
Messages look good now. Let's get one more review from a @plone/volto-team member.

ichim-david:
@victorchrollo14 the feature looks like it works based on your recording. I wouldn't be surprised if you won't receive feedback to tweak the styles or for us todo so after merging.
I would still like to test it and see how it performs and it might take a while until this is merged but I have a good feeling about this one, congratulations.

victorchrollo14:
> @victorchrollo14 the feature looks like it works based on your recording. I wouldn't be surprised if you won't receive feedback to tweak the styles or for us todo so after merging.
> I would still like to test it and see how it performs and it might take a while until this is merged but I have a good feeling about this one, congratulations.

I just came up with a random implementation and was expecting someone would tell me to implement it in a better way or tweak it a bit.

I too am surprised that I was not asked to make changes in code. I'll fix it up if there are any issues.

avoinea (approved):
The Romanian translation looks good.

sneridagh (changes requested):
@victorchrollo14 please look at the comment. BTW, very nice to the eye.

victorchrollo14:
hey @sneridagh. does this look alright?

packages/volto/src/middleware/api.js
```diff
@@ -188,7 +187,15 @@ const apiMiddlewareFactory =
...
                   },
                 ).then((reqres) => {
-                  dispatch(updateUploadedFiles(uploadedFiles++));
+                  if (item.op === 'post') {
+                    let uploadedFiles =
+                      getState().content.subrequests?.['batch-upload']
+                        ?.uploadedFiles;
+
+                    dispatch(
+                      updateUploadedFiles(++uploadedFiles, 'batch-upload'),
```
sneridagh, line 196: @victorchrollo14 the one getting the SUBREQUEST='batch-upload' is createContent action, when it's uploading files.

https://github.com/plone/volto/blob/8aa167a6d5ffb94f329dec42c0d9c96700ea27d6/packages/volto/src/components/manage/Contents/ContentsUploadModal.jsx#L187-L204

Why the `updateUploadedFiles` needs it to be passed down at all?

victorchrollo14: > @victorchrollo14 the one getting the `SUBREQUEST='batch-upload'` is `createContent` action, when it's uploading files.
> 
> https://github.com/plone/volto/blob/8aa167a6d5ffb94f329dec42c0d9c96700ea27d6/packages/volto/src/components/manage/Contents/ContentsUploadModal.jsx#L187-L204
> 
> Why the `updateUploadedFiles` needs it to be passed down at all?

createContent does not handle updating the uploadedFiles count, So I created updateUploadedFiles action that accepts `subrequest, uploadedFiles` properties and a  reducer that takes care of updating the uploadedFiles count  which I have added into subrequests['batch-upload'].
What do you think? should we use `createContent` itself for updating  the uploadedFiles count.

sneridagh: No no, the question is, why do you need to pass down the `subrequest` to `updateUploadedFiles`?

packages/volto/src/actions/content/content.js
```diff
@@ -25,9 +25,10 @@ import config from '@plone/volto/registry';
  * @param {number} number of files uploaded
  * @returns {Object} update uploaded files action
  */
-export function updateUploadedFiles(uploadedFiles) {
+export function updateUploadedFiles(uploadedFiles, subrequest) {
   return {
     type: UPDATE_UPLOADED_FILES,
+    subrequest,
```
sneridagh, line 31: Here, why?

packages/volto/src/reducers/content/content.js
```diff
@@ -104,20 +103,36 @@ export default function content(state = initialState, action = {}) {
...
-            subrequests: {
-              ...state.subrequests,
-              [action.subrequest]: {
-                ...(state.subrequests[action.subrequest] || {
-                  data: null,
-                }),
-                loaded: false,
-                loading: true,
-                error: null,
+        ? action.subrequest === 'batch-upload'
```
sneridagh, line 106: Why did you have to add this?

packages/volto/src/middleware/api.js
sneridagh, line 196: I was thinking more in something more simple like 

```js
if (action.subrequest === 'batched-upload') {
    dispatch(updateUploadedFiles(++uploadedFiles))
}
```
and that's it. Without anything that you did in the last commit. Is it possible?

packages/volto/src/actions/content/content.js
victorchrollo14: I thought I'll pass 'batch-upload' as a subrequest to `updateUploadedFiles` action, then the reducer would use it to update the `subrequests['batch-upload']` to something like this 
```
subrequests: {
'batch-upload': {
   data: null,
   error: null,
   loading: true,
   loaded: false,
   uploadedFiles: 1
}
```

packages/volto/src/middleware/api.js
victorchrollo14: oh! that's what you meant.
Lol. I unnecessarily came up with something too complicated. 

I Tried out the above example. It works fine. 

[Screencast from 07-04-24 02:21:48 PM IST.webm](https://github.com/plone/volto/assets/89346667/7cd308e9-5e84-4907-9522-1d2ed248aff9)

sneridagh:
@victorchrollo14 if it works well, please then simplify it and we are good to go.

victorchrollo14:
> @victorchrollo14 if it works well, please then simplify it and we are good to go.

alright. I'll revert to the previous commit and push this change.

packages/volto/src/components/manage/Contents/ContentsUploadModal.jsx
```diff
@@ -388,7 +401,8 @@ export default compose(
   connect(
     (state) => ({
       request: state.content.subrequests?.[SUBREQUEST] || {},
+      uploadedFiles: state.content.uploadedFiles,
     }),
-    { createContent },
+    { createContent, updateUploadedFiles },
```
nileshgulia1, line 406: @victorchrollo14 Where are we dispatching this action in this file?

nileshgulia1 (commented):
I like the UI part. Just left few questions and its good to go.

packages/volto/src/components/manage/Contents/ContentsUploadModal.jsx
victorchrollo14: Thanks for pointing it out, I initially imported it thinking I'll be using in content upload modal, but then I didn't use it there and forgot to remove it.

packages/volto/src/middleware/api.js
```diff
@@ -133,6 +133,7 @@ const apiMiddlewareFactory =
     const { settings } = config;
 
     const token = getState().userSession.token;
+    let uploadedFiles = getState().content.uploadedFiles;
```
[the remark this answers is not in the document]
victorchrollo14: We could have created a new redux state for it. But then again it wouldn't make sense for a single property. So I added it to the contents state as Uploaded Files is related to it.

sneridagh:
@plone/volto-team I need a final ack on this one, @nileshgulia1 are all your questions answered?

ichim-david (changes requested):
@victorchrollo14 @plone/volto-team I don't like the way the the text is 
so janky as the counter increments and there is very poor contrast of the
text due to the gray overlay with the blue text.
I propose the following changes found in my gist
https://gist.github.com/ichim-david/0756e28aa1999d5d29ff7ebeede93413

And my changes would make the upload look and function as such.

https://github.com/plone/volto/assets/152852/59ef9812-ba0a-463b-be62-cb4c29025d66

If someone has a better solution of different color combination now's the time to speak or forever hold your tongue :)

EDIT:
In my video there is a small or close to small overlap when we have extra 0 such as 200. I've bumped up the width to 2.25 and now the issue is gone. Some of you guys might complain about having too much space to take into account the potential 3 digit file upload. 
Either someone comes up with a better css styling or maybe we need to pad the numbers with zeroes so if we have 50 it starts 01 if we upload 100+ it starts with 001 as a solution to avoid large spacing between the words and the incrementer.

stevepiercy:
I dislike the spacing around the counter. I think it is OK to have the text jump as this happens in every progress bar I have ever watched, but if not, then I would left-align the text and replace it with two lines:

```
Files uploaded: <count>
Total files to upload: <total>
```

I agree that the contrast is too low between the counter text and its grey overlay background. However I think that a background color on the entire progress indicator container would be the better option. The rounded corners of both the counter text container and the progress bar stacked on top of each other is inconsistent with Pastanaga's principles of no rounded corners.

Finally, I think the progress bar itself should be rectangular without rounded corners.

ichim-david:
> I dislike the spacing around the counter. I think it is OK to have the text jump as this happens in every progress bar I have ever watched, but if not, then I would left-align the text and replace it with two lines:
> 
> ```
> Files uploaded: <count>
> Total files to upload: <total>
> ```
> 
> I agree that the contrast is too low between the counter text and its grey overlay background. However I think that a background color on the entire progress indicator container would be the better option. The rounded corners of both the counter text container and the progress bar stacked on top of each other is inconsistent with Pastanaga's principles of no rounded corners.
> 
> Finally, I think the progress bar itself should be rectangular without rounded corners.

I like the 2 line solution and it would get rid of the spacing and counter padding needed and since it's the last text the previous text would probably remain in place.

I am fine with having a bg color for that section and getting rid of the rounded option if this is desired. 
@stevepiercy you might need to make a screenshot with your proposal so that we know exactly what you want for bg color

stevepiercy:
> @stevepiercy you might need to make a screenshot with your proposal so that we know exactly what you want for bg color

As long as there is sufficient contrast, I'm OK with whatever is chosen. White appears to be the default background color, so I would go with that. However there might be an existing style or class that can be reused instead of creating a new one. `blocks-chooser` may be a good class to reuse.

djay:
I'm not saying there is a need to change anything UI wise but quanta does have a design for how this should look.

If there are changes to the UI then it would be nice to make them in the direction of the quanta design?

image

Qunata doesn't say much about the progress bar works but we did an implementation of upload screen that worked with TUS that which means that it gives progress on individual files that are large. Not sure that can be done here as you might not get progress on progress on single files?

- image

The details for how we proposed the final upload UI would look are all in
#5423 

In any case, just a heads up this UI may change again in the future as the proposal was to switch to TUS for all uploads

stevepiercy:
There are two other Quanta related PLIPs:

- https://github.com/plone/volto/issues/4332
- https://github.com/plone/volto/issues/4206

They are targeted for Plone 7 and depend on other parts that are beyond the scope of fixing this bug. I'd prefer to fix the bug now, reusing styles that already exist.

ichim-david:
@stevepiercy @djay we have to take into consideration also the scope of this ticket which is to simply provide a progress bar with some info on how many files.

I took Steve's suggestion and tweaked the rounded corners of the progress and added the info on 2 lines.
I think this is a good solution but let's hear your opinion.

https://github.com/plone/volto/assets/152852/3df17c41-53a1-4ec0-879f-5786ca7d1c29

stevepiercy:
@ichim-david the contrast is sufficient, and there's no more jumpy text. I think it is good. Would you please push a commit with your work?

I think we're waiting for input from @nileshgulia1.

djay:
To be clear I wasn't referring to quanta styles but the layout and functionality.

nileshgulia1 (approved):

packages/volto/src/components/manage/Contents/ContentsUploadModal.jsx
```diff
@@ -40,6 +40,10 @@ const messages = defineMessages({
     defaultMessage:
       '{count, plural, one {Upload {count} file} other {Upload {count} files}}',
   },
+  uploadingFiles: {
+    id: 'Uploading {uploadedFiles} files out of {totalFiles}',
+    defaultMessage: 'Uploading {uploadedFiles} files out of {totalFiles}',
+  },
```
ichim-david, line 46:
```suggestion
   filesUploaded: {
    id: 'Files uploaded: ',
    defaultMessage: 'Files uploaded: ',
  },
  totalFilesToUpload: {
    id: 'Total files to upload: {totalFiles}',
    defaultMessage: 'Total files to upload: {totalFiles}',
  },
```

packages/volto/src/components/manage/Contents/ContentsUploadModal.jsx
```diff
@@ -234,13 +242,18 @@ class ContentsUploadModal extends Component {
...
-            </Loader>
+            <Progress
+              className="progress-bar"
+              value={this.props.uploadedFiles}
+              total={this.state.totalFiles}
+            >
+              {this.props.intl.formatMessage(messages.uploadingFiles, {
+                uploadedFiles: this.props.uploadedFiles,
+                totalFiles: this.state.totalFiles,
+              })}
```
ichim-david, line 255:
```suggestion
              {this.props.intl.formatMessage(messages.filesUploaded)}
              <span className="uploaded-files">{this.props.uploadedFiles}</span>
              <br />
              {this.props.intl.formatMessage(messages.totalFilesToUpload, {
                totalFiles: this.state.totalFiles,
              })}
```

ichim-david (changes requested):
@victorchrollo14 I added as suggestions that you can batch and commit in a single commit the changes I suggest to the styling and the text content.
You will need to run pnpm y18n command to update the translations po files and use the latest translations messages.

packages/volto/locales/de/LC_MESSAGES/volto.po
```diff
@@ -3978,16 +3978,16 @@ msgstr "Laden Sie ein Lead-Bild im Feld 'Lead-Bild' hoch."
 msgid "Upload files"
 msgstr "Datei hochladen"
```
ksuess, line 3979:
```suggestion
msgstr "Dateien hochladen"
```
Plural

ksuess (changes requested):

victorchrollo14:
> @stevepiercy @djay we have to take into consideration also the scope of this ticket which is to simply provide a progress bar with some info on how many files.
> 
> I took Steve's suggestion and tweaked the rounded corners of the progress and added the info on 2 lines. I think this is a good solution but let's hear your opinion.
> 
>  flex-two-lines.mp4

@ichim-david @stevepiercy @djay @sneridagh @nileshgulia1. Can we try something like this.
if this doesn't feel fine, then I'll make the changes @ichim-david suggested.

[Screencast from 01-05-24 03:10:15 PM IST.webm](https://github.com/plone/volto/assets/89346667/d61f09b7-4b10-4822-bd38-d620a7473e0d)

stevepiercy:
@victorchrollo14 absolutely, positively, 100% YES! That is exactly what I imagined, but could not put into words. It looks consistent, accessible, and well-integrated.

victorchrollo14:
> @victorchrollo14 absolutely, positively, 100% YES! That is exactly what I imagined, but could not put into words. It looks consistent, accessible, and well-integrated.

I'll push these changes then.

ichim-david (approved):
@victorchrollo14 I'm fine now with the changes, thank you for your work and the willingness to go through the necessary round of changes in order to make something even better than when it started as this is the strength of the open source community where you can receive feedback from several members and each can contribute to the successful outcome.

sneridagh (approved):

PR #5933, opened 2024-03-28
Fix cypress test "As editor, I can unlock a locked page"

wesleybl:
Sometimes this test fails because it does not find the alert with the message that the content is locked. What may be happening is that the test runs too quickly, and Plone does not consider that the content has been locked. Then we wait a bit for Plone to consider it locked.

This fix: https://github.com/plone/volto/actions/runs/8403447599/job/23013937367#step:10:649

### DIFF
--- a/packages/volto/news/5933.internal
+++ b/packages/volto/news/5933.internal
@@ -1,0 +1,1 @@
+Fix cyprees test "As editor, I can unlock a locked page". @wesleybl
--- a/packages/volto/news/5933.internal
+++ b/packages/volto/news/5933.internal
@@ -1,1 +1,1 @@
-Fix cyprees test "As editor, I can unlock a locked page". @wesleybl
+Fix cypress test "As editor, I can unlock a locked page". @wesleybl
--- a/packages/volto/cypress/tests/core/basic/locking.js
+++ b/packages/volto/cypress/tests/core/basic/locking.js
@@ -91,6 +91,9 @@ describe('Document locking', () => {
     cy.url().should('eq', Cypress.config().baseUrl + '/document/edit');
     cy.wait('@schema');
 
+    // Wait Plone to consider the content locked.
+    cy.wait(2000);
+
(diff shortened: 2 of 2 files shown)


packages/volto/news/5933.internal
```diff
@@ -0,0 +1 @@
+Fix cyprees test "As editor, I can unlock a locked page". @wesleybl
```
ichim-david, line 1:
```suggestion
Fix cypress test "As editor, I can unlock a locked page". @wesleybl
```

ichim-david (approved):

ichim-david:
@wesleybl thx, there is also a flaky one on the listing block tests which I will have to look at and add a pull request myself. It's annoying for me that we have to add these wait timeouts, I feel that we are not using Cypress to the best of its abilities but I guess it is what it is and we can always improve in the future.

ichim-david:
@wesleybl unfortunately the problem still occurs and I have a branch where I had to get rid of 
the check for the toast with lock info since every so often the toast didn't show up.
https://github.com/plone/volto/commit/984757505f17c00b3f32c481a0e88526427f7a9a#diff-fcada51db602a0197ba4aa1b737618004ca91e9d00e8fbdca4c0bff93b2b8eafL104

I myself have made pull requests with test fixes only to discover that they still crashed every so often
due to a flaky test scenario.
I have started to look into this topic and have found this user on youtube which has a series of tutorials
about flaky tests scenarios and how to fix them.
This video shows how you can wrap the test and duplicate it so that you see how it behaves after several runs.
https://www.youtube.com/watch?v=bWR6zFGywMI

In my own experience you need to run the test 20 times and several times to see if the test passes or not.


https://github.com/plone/volto/assets/152852/3ceb2af8-6276-4e63-b18b-b42102e34067

wesleybl:
@ichim-david great! I often cannot simulate an intermittent error locally. With your tip it seems possible. Thank you for sharing it!

PR #5937, opened 2024-04-01
Fix Image block field to wrap the entire image across left and right alignments.

### LINKED ISSUE #5922: Image block field fail to wrap the entire image across various alignments during editing.
**Describe the bug**
Image block field fail to wrap the entire image across right, left, and full alignments during editing.

**To Reproduce**
1-In a page create a block.
2-Upload an image.
3-Go to block section on sidebar and change the alignment to anything other than center.
4-Notice how the block field does not wrap the entire image.

**Expected behavior**
The field should wrap the entire image in any alignment view.

**Screenshots**
Issue

**Software:**

- OS: [Ubuntu 22]
- Browser [Microsoft edge]
- Volto Version [18.0.0-alpha.25]
- Plone Version [6.0.10]
- Plone REST API Version [9.5.0]

MostafaMagdyy:
Fixes #5922 for the image field container to wrap the entire image in right and left alignment.

For full-width alignment, I think it's another issue itself. Disregard the field to wrap the entire image because the image is going outside the whole container and above the sidebar on the right if you try to drag its block. So, I do not think this PR should fix it, because it is not related to the failure of the wrapper to fill the whole image. Let me know if I should fix it in this PR, and I will do so. Also, please clarify how the full width should be, **because I do not see any difference between it and when the image is aligned center with size=large, as both take size=100vw in getImageBlockSizes function**.

### DIFF
--- a/packages/volto/news/5922.bugfix
+++ b/packages/volto/news/5922.bugfix
@@ -1,1 +1,1 @@
-Fix image block field to wrap the entire image across various alignments.@mostafamagdyy
+Fix image block field to wrap the entire image across various alignments.@MostafaMagdyy
--- a/packages/volto/news/5922.bugfix
+++ b/packages/volto/news/5922.bugfix
@@ -1,1 +1,1 @@
-Fix image block field to wrap the entire image across various alignments.@MostafaMagdyy
+Fix image block field to wrap the entire image across various alignments. @MostafaMagdyy
--- a/packages/volto/news/5922.bugfix
+++ b/packages/volto/news/5922.bugfix
@@ -1,0 +1,1 @@
+Fix image block field to wrap the entire image across various alignments.@mostafamagdyy
--- a/packages/volto/theme/themes/pastanaga/extras/blocks.less
+++ b/packages/volto/theme/themes/pastanaga/extras/blocks.less
@@ -140,6 +140,7 @@
 }
 
 .block.align.left {
+  display: flex;
   z-index: 2;
 
   &.video .video-inner,
@@ -174,6 +175,8 @@
 }
 
 .block.align.right {
+  display: flex;
+  flex-direction: row-reverse;
   z-index: 2;
 
   &.maps iframe,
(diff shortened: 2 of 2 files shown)


MostafaMagdyy:
Please review this pull request. @stevepiercy ,@polyester, could you please take a look and provide your feedback? Thank you!

stevepiercy:
@MostafaMagdyy the lint CI check fails. You must fix it before review will happen. See https://6.docs.plone.org/volto/contributing/linting.html

packages/volto/news/5922.bugfix
```diff
@@ -0,0 +1 @@
+Fix image block field to wrap the entire image across various alignments.@MostafaMagdyy
```
stevepiercy, line 1:
```suggestion
Fix image block field to wrap the entire image across various alignments. @MostafaMagdyy
```

stevepiercy (changes requested):
Just missing a space in the change log.

MostafaMagdyy:
@stevepiercy Thank you for your feedback. I have addressed all your comments and fixed all the issues. Additionally, here are the screenshots displaying the new behavior for left and right alignment.
oie_PHzGyA7xqc6d
oie_cLX8eZVwphge

stevepiercy (approved):
I approve, but this still needs review from a Volto Team member.

ichim-david (changes requested):
@MostafaMagdyy you don't fully understand why the select area with the border around the image is as small as it is.
It's because of a float property.
That is useful when you want to have text to the right or left of your image.
Here is a demo on the current feature in action:
before-flex

and how it would function with your changes:
after-flex

You need some css that will affect the editing selected area alongside
extra code to have the text float to the left or right.

MostafaMagdyy:
> @MostafaMagdyy you don't fully understand why the select area with the border around the image is as small as it is. It's because of a float property. That is useful when you want to have text to the right or left of your image. Here is a demo on the current feature in action: before-flex
> 
> and how it would function with your changes: after-flex
> 
> You need some css that will affect the editing selected area alongside extra code to have the text float to the left or right.

Thank you for your feedback. 
Now i understand the importance of it and i will try to make the same behaviour as it was.

davisagli:
It looks like this PR did not solve the problem in a way that supported floating images next to text, and has been abandoned, so I will close it.

PR #5939, opened 2024-04-01
Improve the `ContentMetadataTags` component, add missing `preview_image_link` use case. Add exhaustive unit tests.

sneridagh:
@tiberiuichim finally I've found a a perfect use case to sneak `ts-pattern` in there :)


packages/volto/src/components/theme/ContentMetadataTags/ContentMetadataTags.tsx
```diff
@@ -0,0 +1,182 @@
...
+  }, [dispatch, pathname]);
+
+  const contentMetadataTagsImageField = config.settings
+    .contentMetadataTagsImageField as keyof PossibleImageFields;
+  const image = props.content[contentMetadataTagsImageField];
+  const { opengraph_image } = props.content;
+
+  const getContentImageInfo = (image: PossibleImageFieldsUnion) =>
+    match(image)
+      .with({ scales: { large: { download: P.string } } }, ({ scales }) => ({
```
davisagli, line 94: Note, an image can have scales but not have a `large` scale, if the full-size image is smaller than the `large` scale size.

sneridagh: I forgot completely about it. If I keep forgetting about this, imagine others.

It's one of my biggest complaints in the last years. Can't we just make PIL generate them anyways with the highest resolution image possible? Having unpredictable API responses is not good.

davisagli: We're definitely not going to waste time and space generating and storing duplicate images.

In general, there's not much you can depend on about what image scales exist. The configuration can be changed per site in the registry. I think client code needs to be written to specify a target size (not a scale name), and use a function to pick the best scale for that size (i.e. either the smallest one largest than the target, or if there isn't one, then the largest one available)

davisagli: @sneridagh I don't think it requires any sitewide info, just the list of available scales for the specific image you want to use.

sneridagh: Ok, so given the current info for the specific image you are handling, give me the nearest most appropriate one. Got it.

erral: I think that we should be indexing all scale information in the API, regardless they are bigger than the required scales. Right now we are doing one thing in the catalog results and another thing in the content serialization. See https://github.com/plone/plone.namedfile/issues/140 for reference.

sneridagh: That reminded me another one, why is the URL in some places context-less and in other there is the full URL? can we have just the full one everywhere (as in any other serialized URL in p.restapi)?

davisagli: @erral I discussed this with @sneridagh. We have to be careful about backwards compatibility but I think we can make some improvements.
1. We need to store a URL without the hostname in the catalog, because it can be served on multiple hostnames. But, we can convert it to a full URL when serializing the scales for the API. It's technically not backwards compatible, but code that handles image scales usually has to handle both cases already, since it can be passed scales from the catalog or not.
2. We can adjust the serializer for scales from the catalog to include all configured scale sizes (even if they are bigger than the original) but we should be careful to point to the same URLs as the original for those scales, to make sure we don't accidentally generate additional copies of the same image.

PR #5940, opened 2024-04-01
Add a new label `needs: triage` to new bug reports

stevepiercy:
We need to verify that a bug report is in fact a bug. Once the issue is confirmed, then this label can be replaced with `21 status: confirmed`. This label differs from `needs: confirmation` in that the latter is for old issues that have been hanging around for a long time.


stevepiercy:
See https://github.com/plone/volto/pull/5941 for readme-link-check failure

davisagli (approved):

PR #5941, opened 2024-04-02
Fix redirect of `https://sustainability.eionet.europa.eu` to `https:/…

stevepiercy:
…/www.eea.europa.eu/en/topics/at-a-glance/sustainability/`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,7 +155,7 @@ You should check the dependencies in their `package.json` for more details.
 - [Freshwater Information System for Europe](https://github.com/eea/freshwater-frontend) - Volto project for [Freshwater Information System for Europe website](https://water.europa.eu/freshwater)
 - [European Industrial Emissions Portal](https://github.com/eea/industry-frontend ) - Volto project for [European Industrial Emissions Portal website](https://industry.eea.europa.eu)
 - [Biodiversity Information System for Europe](https://github.com/eea/bise-frontend) - Volto project for [Biodiversity Information System for Europe website](https://biodiversity.europa.eu)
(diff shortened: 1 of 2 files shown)


davisagli (approved):

ichim-david:
@stevepiercy the better fix would of been to simply remove the link. There was indeed a separate website but it seems that the content from it was imported within the main eea website and that volto website was shut down.
https://github.com/eea/sustainability-frontend - was set as archive yesterday so indeed  it's better not to have this link anymore.
I'll add another pull request later.

stevepiercy:
@ichim-david thanks for the information. There is a reluctance to remove links from the README, unless they are 404s. I tend to just mindlessly update the 301s. I noticed that there are now three former separate sites that have been Borg-ed up. See https://github.com/search?q=repo%3Aplone%2Fvolto%20www.eea.europa.eu&type=code. Do these three links need to be consolidated into one? We also need to update https://github.com/collective/awesome-volto/pull/20 accordingly.

ichim-david:
> @ichim-david thanks for the information. There is a reluctance to remove links from the README, unless they are 404s. I tend to just mindlessly update the 301s. I noticed that there are now three former separate sites that have been Borg-ed up. See https://github.com/search?q=repo%3Aplone%2Fvolto%20www.eea.europa.eu&type=code. Do these three links need to be consolidated into one? We also need to update [collective/awesome-volto#20](https://github.com/collective/awesome-volto/pull/20) accordingly.

@stevepiercy yeah I understand you had no idea about the status of the links and there was no way for me to give feedback at 1AM when I was sleeping :)
I'll check the collective.awesome and then see about volto core

PR #5944, opened 2024-04-02
Add open/copy target in object browser widget.

robgietema:


ichim-david:
link-over
Looking at this screenshot that I took of the feature,
@robgietema I suggest the following:

- [ ]  remove the title before the link, it's duplicated and doesn't serve a better purpose than the linked text that follows
  it especially when you have an external link where you get the svg as external
- [ ] I think a button would be better for the copy link action, right now I feel it needs at minimum the css pointer cursor added
- [ ] You don't really have any indicator of popup functionality and it's not keyboard accessible but maybe you envision this only as a helper for editors who know that they need this

PR #5953, opened 2024-04-09
Fix pt_BR translation of invalid email message

wesleybl:


wesleybl:
I'm wondering if the "email" key shouldn't be changed to something with "invalid email".

stevepiercy:
I also would use "Email" instead of the hyphenated and archaic "e-mail". This comment from SO from 2010 is interesting: https://english.stackexchange.com/a/1928

wesleybl:
@stevepiercy In Brazil, the use of the hyphen is still more common. So I think it's better to leave it like that. In English it is already without a hyphen.

davisagli (approved):

davisagli:
> I'm wondering if the "email" key shouldn't be changed to something with "invalid email".

That would make sense to me -- but then we should update all the existing translations.

I'll merge this as is, and that can be done in a separate PR.

Issue #5961, opened 2024-04-16
Remove navigation settings that are not used by Volto


wesleybl:
**Is your feature request related to a problem? Please describe.**
Some navigation controlpanel settings are not used by Volto:

- Navigation depth
- Automatically generate tabs
- Generate tabs for items other than folders
- Sort tabs on
- Reversed sort order for tabs
- Sitemap depth

**Describe the solution you'd like**
These settings should be removed from the controlpanel.

**Describe alternatives you've considered**
There is already an option to remove fields from the control panel:

https://github.com/plone/volto/blob/81d0369e1628b3072f9e47d14831a2618d2633b6/packages/volto/src/config/ControlPanels.js#L60

Just add the fields there.

PR #5964, opened 2024-04-17
Removes navigation settings that are not used by Volto

wesleybl:
fixes: #5961

### DIFF
--- a/packages/volto/src/config/ControlPanels.js
+++ b/packages/volto/src/config/ControlPanels.js
@@ -90,6 +90,13 @@ export const filterControlPanelsSchema = (controlpanel) => {
       'picture_variants',
       'image_captioning',
     ],
+    navigation: [
+      'generate_tabs',
+      'navigation_depth',
+      'sort_tabs_on',
+      'sort_tabs_reversed',
+      'sitemap_depth',
+    ],
   };
 
   // Creates modified version of properties object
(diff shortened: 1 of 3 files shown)


packages/volto/src/config/ControlPanels.js
```diff
@@ -90,6 +90,14 @@ export const filterControlPanelsSchema = (controlpanel) => {
       'picture_variants',
       'image_captioning',
     ],
+    navigation: [
+      'generate_tabs',
+      'navigation_depth',
+      'nonfolderish_tabs',
```
davisagli, line 96: This one is used here: https://github.com/plone/plone.restapi/blob/main/src/plone/restapi/services/navigation/get.py#L116

wesleybl: I removed it.

davisagli (approved):

packages/volto/src/config/ControlPanels.js
wesleybl: @davisagli even though `nonfolderish_tabs` is checked in restapi, its help text says:

```By default, any content item in the root of the portal will appear as a tab. If you turn this option off, only folders will be shown. This only has an effect if 'automatically generate tabs' is enabled.```

First, the term "tab" doesn't make sense in Volto. Second, the `automatically generate tabs` field has been removed. So, I think we should remove the nonfolderish_tabs field. Do you agree?

davisagli: @wesleybl I see, so the help text doesn't match how it is actually used in plone.restapi. Yeah, I think we should probably remove it from the control panel.

PR #5966, opened 2024-04-18
Fix flaky test 'As editor I can add links' by using getSlateEditorAndType

ksuess:
This should fix the failing test in https://github.com/plone/volto/actions/runs/8726201362/job/23941446900?pr=5295
Locally I could reproduce the failure with Electron and fix it like this by using cy.getSlateEditorAndType.


pnicolli (approved):
Awesome thanks! I did not know this existed.

PR #5974, opened 2024-04-23
Add headers to avoid redirects being cached

sneridagh:
(eg. the one coming from the MultilingualRedirector).

### DIFF
--- a/src/server.jsx
+++ b/src/server.jsx
@@ -273,6 +273,10 @@ server.get('/*', (req, res) => {
       }
 
       if (context.url) {
+        // If there's a redirection, do not cache it
+        res.set({
+          'Cache-Control': 'no-cache',
+        });
         res.redirect(flattenToAppURL(context.url));
       } else if (context.error_code) {
         res.set({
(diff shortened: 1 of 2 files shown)


src/server.jsx
```diff
@@ -273,6 +273,10 @@ server.get('/*', (req, res) => {
       }
 
       if (context.url) {
+        // If there's a redirection, do not cache it
+        res.set({
+          'Cache-Control': 'no-cache',
+        });
```
davisagli, line 279: @sneridagh

1. I don't think we want to prevent caching of all redirects. It is specifically the multilingual redirect for `/` that is a problem to cache because it relies on the I18N_LANGUAGE cookie, but isn't served with a `Vary: Cookie` header.
2. This should be `no-store` rather than `no-cache`, so that it isn't stored in the user's browser cache either.

davisagli: @sneridagh Maybe you're right that it's better to handle this in the varnish config.

sneridagh:
closing, we don't want to cache all the redirects

PR #5976, opened 2024-04-23
fix: pass down locale to IntlProvider

tomschall:
Add possibility to pass down `locale`, `messages` and `defaultLocale` properties inside the `customStore` object to `IntlProvider`. With this change we can control react-intl language provider from inside storybook and switch for example from english to german with storybook args.

### DIFF
--- a/packages/volto/src/storybook.jsx
+++ b/packages/volto/src/storybook.jsx
@@ -1387,14 +1387,15 @@ export default class Wrapper extends Component {
   render() {
     const mockStore = configureStore();
     const store = mockStore(this.customState());
+    const state = store.getState();
 
     return (
       <Provider store={store}>
         <PluggablesProvider>
           <IntlProvider
-            locale={store.getState().intl.locale}
-            messages={store.getState().intl.messages}
-            defaultLocale={store.getState().intl.defaultLocale ?? 'en'}
+            locale={state.intl.locale}
+            messages={state.intl.messages}
+            defaultLocale={state.intl.defaultLocale ?? 'en'}
           >
             <StaticRouter location={this.props.location}>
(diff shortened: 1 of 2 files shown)


packages/volto/src/storybook.jsx
```diff
@@ -1438,7 +1442,11 @@ export class RealStoreWrapper extends Component {
     return (
       <Provider store={store}>
         <PluggablesProvider>
-          <IntlProvider locale="en">
+          <IntlProvider
+            locale={this.props.customStore.intl.locale}
+            messages={this.props.customStore.intl.messages}
+            defaultLocale={this.props.customStore.intl.defaultLocale ?? 'en'}
+          >
```
sneridagh, line 1449:
```suggestion
          <IntlProvider
            locale={store.getState().intl.locale}
            messages={store.getState().intl.messages}
            defaultLocale={store.getState().intl.defaultLocale ?? 'en'}
          >
```

sneridagh (changes requested):
@tomschall I tried it out, and it breaks in the stories that do not have a `customStore` passed down, take a look at my suggestion, I think we have to look into the generated store instead. So it takes the defaults, and the incoming `customStore`, if any. I already tried it out and works as expected.

tomschall:
> @tomschall I tried it out, and it breaks in the stories that do not have a `customStore` passed down, take a look at my suggestion, I think we have to look into the generated store instead. So it takes the defaults, and the incoming `customStore`, if any. I already tried it out and works as expected.

What i am not so happy with is, that we call store.getState() 3 times in each Wrapper. I made another suggestion, that we use a const to fetch the state and then use that for the props.

sneridagh:
Thanks @tomschall !!

PR #5978, opened 2024-04-23
Fix initialValue block setting

### LINKED ISSUE #5971: initialValue block setting doesn't work
**Describe the bug**
[initialValue](https://github.com/plone/volto/blob/2add6cd79242b28c2a54ecaf391bf4a5745b093b/docs/source/blocks/settings.md?plain=1#L157) block setting doesn't work.

**To Reproduce**

1. Create a block with the `initialValue` setting. For example:
```javascript
function initialValue({ id, value, formData, intl }) {
  return {
    ...formData.blocks[id],
    myField: 'test',
  };
}
const applyConfig = (config) => {
  config.blocks.blocksConfig.myBlock = {
    id: 'myBlock',
    initialValue: initialValue,
    ...
  };
  return config;
};

```
2. Add the block to a page
3. The `myField` field is not filled with the value 'test'.

**Expected behavior**
The myField field should be filled with the value 'test'.

Volto 18.0.0-alpha.27
Plone 6.0.10
plone.restapi 9.5.0

**Additional context**
The configuration does not work because the block id that is passed to the `_applyBlockInitialValue` function is the id of the previous block, and not the id of the newly created block. See:

https://github.com/plone/volto/blob/0edcd8083afee783a6463a14b18d0f712cf7c25b/packages/volto/src/helpers/Blocks/Blocks.js#L307

The id passed there should be `newBlockId` and not `id`.

Even making this fix, one thing that still happens is that when the previous one is a block of text, the setting is not applied. The [insertBlock](https://github.com/plone/volto/blob/0edcd8083afee783a6463a14b18d0f712cf7c25b/packages/volto/src/helpers/Blocks/Blocks.js#L288) method is not called in this situation. I wouldn't know where to call `_applyBlockInitialValue` in this situation. I know that the [changeBlock](https://github.com/plone/volto/blob/0edcd8083afee783a6463a14b18d0f712cf7c25b/packages/volto/src/helpers/Blocks/Blocks.js#L341) method is called in this situation. But I don't know if it would be the right place, since it is called several times.

wesleybl:
This fixes `initialValue` block setting in two situations:

- When the previous block is an empty block:

In this situation, only the `onChangeBlock` method was called, which does not call the `applyBlockInitialValue` method. Then we call the `applyBlockInitialValue` method before calling `onChangeBlock`.

- When the previous block is not an empty block:

In this situation, the `insertBlock` method is called. But it passed the id of the previous block to the `applyBlockInitialValue` method, instead of the id of the new created block.

fixes #5971


packages/volto/src/helpers/index.js
```diff
@@ -39,6 +39,7 @@ export {
   getLanguageIndependentFields,
 } from '@plone/volto/helpers/Content/Content';
 export {
+  _applyBlockInitialValue,
```
ichim-david, line 42: @wesleybl Usually when you have a method that starts with _ it's implied that it is private.
I don't think it should thus be exported for helpers, if that is the case that it would be useful
for add-ons to call this helper then I would remove the _.
Ex info
https://css-tricks.com/implementing-private-variables-in-javascript/#aa-using-an-underscore
but if you google js private convention you will see that this was very popular before true private
fields

wesleybl: @ichim-david done.

sneridagh (approved):
@ichim-david can I have your final review?

ichim-david (approved):

PR #5982, opened 2024-04-25
Ensure that sidebar field will not steal focus when metadata is edited 17.x.x backport

dobri1408:
When editing metadata externally, as in the case of our EEA metadata block, the current functionality causes the sidebar to automatically gain focus. However, this behavior is unnecessary in this context and results in the sidebar capturing focus inappropriately. To address this, I propose a solution where the sidebar will only receive focus if the user is actively working within it. This can be achieved by checking the activeElement to determine if the sidebar should be focused.

[screen-capture.webm](https://github.com/plone/volto/assets/50819975/6a771fa6-dabf-49c4-82a8-2ea579f9f5bf)

### DIFF
--- a/cypress/tests/core/basic/metadata.js
+++ b/cypress/tests/core/basic/metadata.js
@@ -40,7 +40,7 @@ describe('Add Content Tests', () => {
     cy.get('.sidebar-container .tabs-wrapper a.active.item').contains('Page');
   });
 
-   it('After removing value of widget the focus should be removed from the field', () => {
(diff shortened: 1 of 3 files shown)


ichim-david:
Closing this pull request for now, the backport is always done after the main work went through the review process and the work was merged.

ichim-david (approved):

ichim-david:
@sneridagh I'm going to merge this since this is a bugfix 17.x.x backport and we need this fix in the 17 branch.

PR #5983, opened 2024-04-25
Ensure that sidebar field will not steal focus when metadata is edited

dobri1408:
When editing metadata externally, as in the case of our EEA metadata block, the current functionality causes the sidebar to automatically gain focus. However, this behavior is unnecessary in this context and results in the sidebar capturing focus inappropriately. To address this, I propose a solution where the sidebar will only receive focus if the user is actively working within it. This can be achieved by checking the activeElement to determine if the sidebar should be focused.

[screen-capture.webm](https://github.com/plone/volto/assets/50819975/64494d61-863d-4052-9e71-621d29fdf70c)


sneridagh:
@dobri1408 are you sure that this does not break the original intention of the code (https://github.com/plone/volto/issues/4230). Apparently, seems it does not, but still.

dobri1408:
@sneridagh, I believe there might be a misunderstanding. Consider this perspective: the field in question is displayed in the sidebar, as indicated by the className being sidebar-metadata. Therefore, if we are not operating within the sidebar context, it seems unnecessary to focus on these fields. My condition ensures we check whether we are in the sidebar or not.

sneridagh (approved):
LGTM, although I'd like to have feedback from David and Tiberiu.

tiberiuichim:
Seems strange passing react props derived from the state of the DOM elements in the browser

ichim-david:
@sneridagh I'm putting the pull request in draft until we have a better look at the problem and consider if we need to handle it from Volto core. 
This is a problem of our doing within our add-on  https://github.com/eea/volto-metadata-block/ that allows you to mirror the content type fields from the sidebar to the content area. Because it renders with the same id and widget when there is a field with a react-select and given the field id it focuses on the sidebar which is bad if you interact with the widget in the content area.

Given Tiberiu's comment even if we do propose something for the core we might need todo it differently so for now we treat it as review given and we still have some work todo.

dobri1408:
After conducting research and consulting with my colleagues, we have concluded that the focus management approach used by the Select Widget is incorrect. As demonstrated in the attached video, the issue does not originate from our metadata-section-block.

The problem originates from this particular pull request: https://github.com/plone/volto/pull/4003/files#diff-e739dae43ff2e2a3a433411ca92aa574aa0502d440e3734905df9e1b96522ebfR318
 
My solution, as outlined in this pull request, addresses all these issues. I suggest we reconsider this approach and possibly enhance it if it doesn't meet your expectations.

https://github.com/plone/volto/assets/50819975/d5cb58a3-4c39-4335-858e-97cbeac38e58



We've had this bug for quite some time. As you can see here, we were forced to implement customizations because of it: https://github.com/eea/volto-cca-policy/tree/master/src/customizations/volto/components/manage/Widgets

dobri1408:
Hello @ichim-david, @tiberiuichim, @sneridagh! I wanted to highlight an issue that appears to be present in every Volto project, including demo.plone.org. This problem is specific to Volto and is not related to the EEA addons. I've added a cypress test and uploaded a video to clearly demonstrate the issue. Could you please review it when you have a moment? Thank you! 

https://github.com/plone/volto/assets/50819975/cfadb5b8-b29e-4f4e-a474-485f61d51954

ichim-david:
@dobri1408 simply add the test as a new it test here: https://github.com/plone/volto/blob/main/packages/volto/cypress/tests/core/basic/metadata.js there is no need to have a separate file which isn't a .js file and which isn't testing the FileWidget.

sneridagh:
@ichim-david @dobri1408 is it ready to merge?

ichim-david (approved):
@sneridagh yes it's ready for merging

dobri1408:
Hello @sneridagh, could you, also, backport this to Volto 17?

Issue #5985, opened 2024-04-26
Wrong parameters in the asyncConnect function


wesleybl:
**Describe the bug**
The `asyncConnect` function is declared with 5 required parameters:

https://github.com/plone/volto/blob/d4d7d58eee9d8b59ba84dd1884d5c3729a412da4/packages/volto/types/helpers/AsyncConnect/index.d.ts#L17

But only the first parameter is required. In many places it is called with only one parameter. For example:

https://github.com/plone/volto/blob/e7d6f1c0be483c6560a552bae615b3f80ec21c1e/packages/volto/src/components/manage/Controlpanels/index.tsx#L14-L25

When I run the tests for an addon, I receive the error:

```bash
● Test suite failed to run

    node_modules/@plone/volto/src/components/manage/Controlpanels/index.tsx:14:30 - error TS2554: Expected 5 arguments, but got 1.

    14 export const Controlpanels = asyncConnect([
                                    ~~~~~~~~~~~~

      node_modules/@plone/volto/types/helpers/AsyncConnect/index.d.ts:17:47
        17 export function asyncConnect(asyncItems: any, mapStateToProps: any, mapDispatchToProps: any, mergeProps: any, options: any): (Component: any) => any;
                                                         ~~~~~~~~~~~~~~~~~~~~
        An argument for 'mapStateToProps' was not provided.

```

**To Reproduce**

1. I ran tests on one of my addons.

**Expected behavior**
The tests must run without error. When defining parameters, only the first one must be required.

**Software (please complete the following information):**

Volto 18.0.0-alpha.28
Plone 6.0.10
plone.restapi 9.5.0
CMF 3.3
Zope 5.9
Python 3.11.1 (main, Dec 7 2022, 01:11:34) [GCC 11.3.0]
PIL 9.5.0 (Pillow)

**Additional context**

PR #5986, opened 2024-04-26
Defines the last 4 parameters of the `asyncConnect` function with optional

wesleybl:
The function works without these parameters. Avoid error:

```
TS2554: Expected 5 arguments, but got 1.
```

fixes: #5985


davisagli (approved):

sneridagh:
@wesleybl unfortunately, we cannot do this :( The types in `packages/volto/types` are autogenerated, given the information `tsc` can infer from the code and JSDoc annotations in the `.js` code. See https://github.com/plone/volto/blob/e7d6f1c0be483c6560a552bae615b3f80ec21c1e/packages/volto/package.json#L42

This was done for convenience, because the prior situation was far worse, when coming to TS support and VSCode autocompletion and "go to reference" matters. Problem is that any change will be overwritten afterwards.

What we can do is to improve the JSDocs and hints in the code itself, so the generated types are improved. @wesleybl could you try this approach instead, then run the `build:types` command and see if that improves? 

Unfortunately, we will have to bear with this kind of things until the vast majority of our code base is not TS-based. :(

sneridagh:
@wesleybl could you please share the test that cause the failure? Maybe we can add them to the test batteries in CI.

sneridagh:
@tiberiuichim could you please check the typings? I want to make sure that we are not messing this up.

wesleybl:
> @wesleybl could you please share the test that cause the failure? Maybe we can add them to the test batteries in CI.

@sneridagh It is not a specific test. I think any test that import:

https://github.com/plone/volto/blob/e7d6f1c0be483c6560a552bae615b3f80ec21c1e/packages/volto/src/components/manage/Controlpanels/index.tsx

will give the error. I commented on the test that failed and another test started to fail.

>  The types in packages/volto/types are autogenerated, given the information tsc can infer from the code and JSDoc annotations in the .js code

After your commit, the parameters became optional, even after running `build:types`. So I think it's ok now. Thanks!

PR #5997, opened 2024-05-01
Allow Makefile to be modified by Makefile.local if present

ichim-david:
This way we can modify any of the makefile commands or add extra without having to touch the
Volto core file.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -401,3 +401,4 @@ stop-test-acceptance-server-detached: ## Stop Test Acceptance Server Main Fixtur
 
 # include local overrides if present
 -include Makefile.local
+-include ../../../Makefile.local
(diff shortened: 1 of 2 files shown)


sneridagh:
@ichim-david if it's not present it does not complain? If so, I'd add also `../../Makefile.local`, for make it work with the projectless setups.

ichim-david:
> @ichim-david if it's not present it does not complain? If so, I'd add also `../../../Makefile.local`, for make it work with the projectless setups.

@sneridagh yes it doesn't complain if it's not present, I've tested the make with and without the file and it works as usual.

There is a warning locally but this is correct as I am overriding the setup command with my own and it just tells me about it.
Makefile.local:2: warning: overriding commands for target `setup'
Makefile:101: warning: ignoring old commands for target `setup'

sneridagh (approved):

PR #5998, opened 2024-05-01
Cypress replaced wait with assertions

ichim-david:
After looking at the videos from this playlist https://www.youtube.com/playlist?list=PLP9o9QNnQuAYfFNgCWIxzRVdomGINKhH9
I learned that it's better to use assertions in order to wait for items to show up rather than use
wait timeouts in seconds as the assertions can wait up to the default timeout.

I modified some of the flaky tests from block-listing and locking for the editors that @wesleybl 
was also flagged as being bothersome.

- Introduced typeWithDelay as I've noticed that if we add the delay of 20ms per typing it avoids selecting other things rather then the slate input and when typing it's less likely to turn an error.


sneridagh (approved):

wesleybl:
Thanks!

PR #5999, opened 2024-05-03
Fix 301 and 302 redirects.

robgietema:


sneridagh (approved):

JeffersonBledsoe:
I saw this was reverted. We started a PR which solved this along with the ++api++ URL in the past (https://github.com/plone/volto/pull/4854), but it needs a little love to finish off the work (mostly tests adding for all the URL cases; such as URLs with query string params, separate internal and external paths, etc.) if anybody wants

davisagli:
@JeffersonBledsoe Yep, thanks -- I think we'll come back to this.

PR #6008, opened 2024-05-08
Revert "Fix 301 and 302 redirects. (#5999)"

davisagli:
This reverts commit 0fbeea4ec9e21dfca80a61d1c2e73055caf4a381.

It is causing redirects to API URLs when there is a plone.app.redirector alias, instead of to the correct frontend URL.


davisagli:
Ignoring the changelog check since this is removing a changelog entry that is no longer needed.

PR #6010, opened 2024-05-09
Improve scoping of separator class

JeffersonBledsoe:
The `.separator` class, used in the blocks toolbar like shown below, has a really low specificity, causing it to leak (try creating a block named `separator` and seeing what happens to the blocks list 😉 ). This PR scopes it to `.block .toolbar .separator`, which is really it's intended location


stevepiercy (approved):

packages/volto/theme/themes/pastanaga/extras/blocks.less
```diff
@@ -318,6 +318,16 @@ body.has-toolbar.has-sidebar-collapsed .ui.wrapper > .ui.inner.block.full {
       0 2px 4px rgba(0, 0, 0, 0.05);
     transform: translate(-50%, 0);
 
+    .separator {
```
ichim-david, line 321: @JeffersonBledsoe  not quite, the html block on the edit page also uses this class
https://github.com/plone/volto/blob/09239bfde86c7dff628c7aec3df291d439dab33c/packages/volto/src/components/manage/Blocks/HTML/Edit.jsx#L295 
html-block

I am ok  with changing modifying the separator scope and/or prefixing it with other names so that it's not simply .separator.
I'm a bit surprised that on the demo site it's not prefixed .block .separator but even then it might conflict with a separator block

ichim-david (changes requested):

packages/volto/theme/themes/pastanaga/extras/blocks.less
JeffersonBledsoe: @ichim-david This is the toolbar that is mentioned and a screenshot taken of in the description. There's probably addons that use this CSS class too, but this is the only place I could find it in core

ichim-david: @JeffersonBledsoe you are correct, I did a better search and there are other separators such as within slate or the plone components but they have their own extra identifiers so this work is good.

ichim-david (approved):

PR #6018, opened 2024-05-15
Pass class into breadcrumbs RAC

Gomez:


packages/components/src/components/Breadcrumbs/Breadcrumbs.tsx
```diff
@@ -42,9 +43,9 @@ export function Breadcrumbs({
 
   return (
     <nav aria-label="breadcrumbs" role="navigation">
-      <RACBreadcrumbs items={itemsWithRoot || items}>
+      <RACBreadcrumbs className={className} items={itemsWithRoot || items}>
         {(item) => (
-          <RACBreadcrumb id={item['@id']}>
+          <RACBreadcrumb className={className} id={item['@id']}>
```
sneridagh, line 48: @Gomez not sure to use the same class in both, this could lead to style leaks. :/

packages/volto/news/6018.bugfix
```diff
@@ -0,0 +1 @@
+Fix ignored classname in breadcrumbs RAC @gomez
```
pnicolli, line 1: This would go in the news folder of the `components` package, not in volto

sneridagh:
@Gomez applied suggestions

PR #6028, opened 2024-05-17
Add controlpanel for plone.volto index (wip)

jonaspiterek:
# ToDo

- [ ] Resolve `Error: Hydration failed because the initial UI does not match what was rendered on the server.` which occurs when going to /controlpanel/blocktypes and reloading the page


jnptk:
Closing this in favor of #7418, which continues this work.

Issue #5783, opened 2024-02-22
`pnpm test` fails on updating snapshots


ksuess:
Change or add a new script
"test": "pnpm --filter @plone/volto run test:ci" 
->
"test": "pnpm --filter @plone/volto run test",

**`pnpm test` exits without the chance to update a snapshot.**

its-VinayKumar:
@ksuess Please assign me this issue. I will resolve it as soon as possible.

stevepiercy:
@its-VinayKumar please read and follow [First-time contributors](https://6.docs.plone.org/contributing/first-time.html), especially [Things not to do](https://6.docs.plone.org/contributing/first-time.html#things-not-to-do), [Contributing to Plone](https://6.docs.plone.org/contributing/index.html), and [Contributing to Volto](https://6.docs.plone.org/contributing/volto.html).

Hrittik20:
Wouldn't updating `"test": "jest"` to `"test": "jest --updateSnapshot"` fix this issue?

sneridagh:
@ksuess @Hrittik20 you can use:

```
pnpm test -- -u
```

ichim-david:
@sneridagh I think it's a mistake to close this issue given the fact that we have this difference from normal commands to :ci commands
Even in packages/volto we have test and test:ci
https://github.com/plone/volto/blob/main/packages/volto/package.json#L44
For consistency, I think the action from the monorepo package should also have the same parity  and not rely on knowledge on how to  get around a broken assumption given the difference in pnpm test in the monorepo vs pnpm test in packages/volto

sneridagh:
@ichim-david but which is the issue here? is it about the about the lack of parity, or is about issuing update command? or both?

sneridagh:
@ichim-david https://github.com/plone/volto/pull/6051

PR #6031, opened 2024-05-20
Fix updating snapshots on pnpm test

Hrittik20:
Fixes: #5783


ichim-david:
@Hrittik20 adding the -u flag means you update the snapshots regardless if you want todo that or not. You want to run the test, see that it fails because most of the time you made a mistake and hence tests fail.
There are instances where yes you make changes that require testing snapshot update but then you can mention that you want to update the test.

You didn't understand the issue @ksuess added and what she proposed.

If you look there are :ci commands
https://github.com/plone/volto/blob/main/package.json#L18
and non :ci commands. 
She proposes to modify the test command use from packages/volto
https://github.com/plone/volto/blob/main/packages/volto/package.json#L43
and add a new test:ci command and modify the scripts where test is used to be test:ci

This way if someone runs the tests from the root of the monorepo they can update the
tests since the test run allows to select the action todo
run-mode

ichim-david (changes requested):
See previous comment

Hrittik20:
@ichim-david Apologies for the misunderstanding. So will the issue be solved if I modified the [ package](https://github.com/plone/volto/blob/main/package.json) like this :
`"test": "pnpm --filter @plone/volto run test",
  "test:ci": "pnpm --filter @plone/volto run test:ci", `

sneridagh:
@Hrittik20 you can use:

```
pnpm test -- -u
```

It does the trick, by passing args to pnpm scripts (as it did always, it's not a pnpm trick).
If you do not mind, I'd close this one.

PR #6033, opened 2024-05-20
Server-side redirects for plone.app.redirector aliases (#4854)

davisagli:


davisagli:
@stevepiercy Is the readthedocs build supposed to work for the 16.x.x branch of volto, or do we need to find a way to disable that check?

stevepiercy:
@davisagli I need to either configure it to work against the `16.x.x` branch (and also `17.x.x) by backporting the relevant changes from `main`, or remove or disable the check. IIRC older branches previously used Netlify, but I'd have to check. I think we want the former, correct?

davisagli:
@stevepiercy Are the 16.x docs published somewhere? If so we should fix it; otherwise it's probably ok to just disable it.

stevepiercy:
> @stevepiercy Are the 16.x docs published somewhere?

Nope.

> If so we should fix it; otherwise it's probably ok to just disable it.

I'll see what I can do. I'm not sure if the configuration is via the RTD web UI or can be done in the CI.

stevepiercy:
RTD uses a webhook, which fires on a pull request or push but cannot be scoped to a specific branch. That's OK, we can cancel the build, which is what Netlify did.

https://docs.readthedocs.io/en/stable/build-customization.html#cancel-build-based-on-a-condition

I think that we would need to put a `readthedocs.yaml` file at the root of each branch we don't want to build in RTD.

```yaml
version: 2
build:
  os: "ubuntu-22.04"
  tools:
    python: "3.12"
  jobs:
    post_checkout:
      exit 183;
```

I'll try it on this PR to see if it works, and if not, it can be reverted.

stevepiercy:
Yup, that did it.

davisagli:
@stevepiercy Thanks!

sneridagh (approved):

PR #6038, opened 2024-05-21
Use pnpm corepack instead of installing it in CI

sneridagh:


stevepiercy (changes requested):
These changes LGTM. However:

- it needs a change log
- these steps need to align with the changes, too: https://github.com/plone/volto/blob/main/.readthedocs.yaml#L19-L20

sneridagh:
@stevepiercy done!

stevepiercy:
@sneridagh RTD still no like: https://readthedocs.org/projects/volto/builds/24443126/. I'm not sure how to resolve this one. I was using https://6.docs.plone.org/volto/contributing/developing-core.html#pnpm as a guide.

Maybe we need to install Node.js first, even though it is declared in https://github.com/plone/volto/pull/6038/files#diff-03efc769b870804394632e45d7885272b44c16939517fb31c9d7c614d2ffae57R13, so don't need to use nvm? Sorry my JS install knowledge is novice.

sneridagh:
@stevepiercy depends on how node is installed in that image... :(  do we know more about the image they are using? maybe we can install a package.

stevepiercy:
@sneridagh here's what I found, which is not very helpful. https://dev.readthedocs.io/en/latest/design/build-images.html

stevepiercy:
Still failing: https://readthedocs.org/projects/volto/builds/24443350/

sneridagh:
@stevepiercy it should work now. However, it seems that GHA have a hickup now

PR #6039, opened 2024-05-21
Use search operation instead of contains in search block

instification:


davisagli:
Closing as abandoned. The proposal to add this operator in plone.app.querystring was closed (https://github.com/plone/plone.app.querystring/pull/148)

PR #6043, opened 2024-05-22
Allow X-Robots-Tag header in images/files middleware

sneridagh:
----
📚 Documentation preview 📚: https://volto--6043.org.readthedocs.build/

### DIFF
--- a/packages/volto/news/6043.bugfix
+++ b/packages/volto/news/6043.bugfix
@@ -1,0 +1,1 @@
+Allow X-Robots-Tag header in images/files middleware @sneridagh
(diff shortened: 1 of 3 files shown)


davisagli (commented):
@sneridagh Okay, but this isn't all we need. We want to pass X-Robots-Tag through when downloading the image data via the middleware, yes, but we also want to include it in the Volto SSR response for an image view.

(Maybe it's easier to add the [robots meta tag](https://developers.google.com/search/docs/crawling-indexing/robots-meta-tag), using react-helmet?)

davisagli (approved):
@sneridagh reminded me the meta tag is already added in ContentMetadataTags.jsx. So this looks good to me.

PR #6051, opened 2024-05-27
Fix test script in monorepo root

### LINKED ISSUE #5783: `pnpm test` fails on updating snapshots

sneridagh:
Closes #5783

----
📚 Documentation preview 📚: https://volto--6051.org.readthedocs.build/


ichim-david (approved):
@sneridagh as I was telling to @Hrittik20 that he should check where pnpm test is used and to modify it with pnpm test:ci I noticed that the testing is done in https://github.com/plone/volto/blob/main/.github/workflows/unit.yml#L47 and there either way it's using directly the pnpm test from packages/volto and not from the  monorepo root.
I also wonder how come it's not using test:ci there?
In fact only i18n:ci is used in an automated CI https://github.com/plone/volto/blob/e87d129c8387bf509f2b92294f39a34b2db8e58f/.github/workflows/code-analysis.yml#L148
Oh well, this restores the parity of commands from monorepo root to packages/volto so yes I think this is a good change.

sneridagh:
@ichim-david that is a typo probably, but jest detects the envvar probably anyways... so the outcome is the same.

PR #6055, opened 2024-05-28
Listing Block template for Page/News Item/Event based on content type.

### LINKED ISSUE #4765: Listing/Search Block Variants for Page/News Item/Event/File
## Abstract

Create new listing block variants that show the following content objects:

- page
- news item
- event
- file

in a different way (tbd).

Listing Block Variants:

- Standard
- Standard with Image
- Grid?

iFlameing:
Still long road to go. Fixes #4765 

1. Added the variation for listing.
2. Needed thought for migration.

Replicate same things for search.

----
📚 Documentation preview 📚: https://volto--6055.org.readthedocs.build/


packages/volto/src/components/manage/Blocks/Listing/RenderAccordingToContentType.jsx
```diff
@@ -0,0 +1,42 @@
...
+  image,
+  grid,
+}) => {
+  const type = item && item['@type'];
+  const Template =
+    (type &&
+      config?.getComponent &&
+      config.getComponent({ name: 'ListingTemplate', dependencies: [type] })
+        .component) ||
+    contentTemplate[type];
```
davisagli, line 25: @iFlameing Using the registry here is good, but I think the way RenderAccordingToContentType is currently implemented, it needs to know too much about the specific listing variations that exist (i.e. the boolean props for image and grid).

Also the contentTemplate mapping should not be hardcoded here. That is what the registry lookup is for.

Instead I would suggest to get rid of RenderAccordingToContentType, and do the registry lookup directly in the variation templates (i.e. in DefaultTemplate, SummaryTemplate, and Grid).

Each variation should use a different component name in the registry lookup:
- default variation -> ListingTemplate component type
- summary variation -> ListingWithImageTemplate component type
- grid variation -> ListingGridTemplate component type

If you have questions about how to use the registry while I am asleep, @sneridagh can probably help.

davisagli:
@iFlameing For migration / backwards compatibility, we have 2 options:

1. Tell users how to run a script to update all existing blocks content with the new names
2. Add a mapping in the volto settings from old listing variation name to new listing variation name, so that the block can automatically convert old names when it is rendering.

I like option 2 better. It's a lot simpler for users, and still pretty easy to implement.

We'll need a section in the upgrade guide to explain how to configure this mapping if the user wants something other than the default one.

iFlameing:
@davisagli pleaser review it once again. I will rename the variable once it is done. Same for default content type. I will register it wit config.register component.

tisto:
@iFlameing this PLIP/PR is mentioned in the Plone roadmap that is about to being published next week. We should try to put some effort into this. Maybe some folks at the Salamina sprints can help. cc @davisagli @sneridagh @fredvd @jackahl

davisagli:
We are trying a different solution in Volto Light Theme: https://github.com/kitconcept/volto-light-theme/pull/462 -- that one introduces a Summary component which makes it easier to add a new content type's summary that is used in several different blocks, and also avoids the need to migrate existing block variation names. If it works well then we can consider handling it that way in Volto or Seven in the future.

PR #6057, opened 2024-05-29
 Fix js events association on error pages

### LINKED ISSUE #6048: Error pages do not associate events with elements
**Describe the bug**
In error pages, in production mode and rendered on the server, events are not associated with elements. This can be seen when we click on the search lupe, and it goes to a page not found:

http://localhost:3000/undefined/search?SearchableText=

In vanilla Plone this is not a problem, as we don't have many clickable elements on the page. But on sites that have more links this is a problem, since no Universal Link works.

**To Reproduce**

1. Start Volto in production mode:
```
pnpm build && pnpm --filter @plone/volto start:prod
```
2. In the Browser's address bar, type a non-existent URL. For example: http://localhost:3000/xxxxx
3. Press Enter.
4. Click on the search lupe.

**Expected behavior**
The Browser should be redirected to the search page:

http://localhost:3000/search?SearchableText= 

- OS: Ubuntu
- Browser: Chrome
- Volto Version: 18
- Plone Version: 6.0.11

**Additional context**
In the situation described above, javascripts are not inserted into the page, causing the problem to occur.

This is because the `extractScripts` parameter is passed as `undefined` here:

https://github.com/plone/volto/blob/082e919d372a96af0ab40938375a65c6bdeb0c34/packages/volto/src/server.jsx#L289-L292

This causes this code not to be executed:

https://github.com/plone/volto/blob/0edcd8083afee783a6463a14b18d0f712cf7c25b/packages/volto/src/helpers/Html/Html.jsx#L189-L197

This has been implemented by @tiberiuichim in:

https://github.com/plone/volto/pull/1948

I think this code should be executed, regardless of whether it is in development or production mode. Is there any reason not to load scripts on error pages?

wesleybl:
Now the scripts are added to SSR error pages, in production mode. As a result, the `extractScripts` property of the `Html` component is no longer necessary.

The setting `config.settings.serverConfig.extractScripts.errorPages` It has also been removed as it is not used anywhere else.

fixes #6048 

----
📚 Documentation preview 📚: https://volto--6057.org.readthedocs.build/


stevepiercy (changes requested):

tiberiuichim (approved):
Should be marked as a breaking change.

sneridagh (changes requested):
@wesleybl LGTM, though I'd like also @davisagli approval.

Also, please mark it as breaking, and add some breaking notes too. Thanks!

stevepiercy:
@sneridagh with breaking, it's not just notes, but an entry in the [Volto upgrade guide](https://6.docs.plone.org/volto/upgrade-guide/index.html). I'll help with syntax and grammar, but content needs to come from @wesleybl. This is both a removal and a change, so this would be the pattern to use.

````

```{versionremoved} 18.0.0-alpha.22 [or whatever is the next alpha]
Thing that was removed.
```

```{versionchanged} 18.0.0-alpha.22 [or whatever is the next alpha]
Thing that changed.
```

Explanation of things that were removed and changed.

````

sneridagh:
@stevepiercy sure, I meant the upgrade guide.

stevepiercy (changes requested):

stevepiercy (approved):

stevepiercy:
Docs look good, back to you @sneridagh.

PR #6065, opened 2024-05-31
Make css layer more specific

pnicolli:
This also allows `@plone/components` to play nicely with tailwindcss. See https://github.com/tailwindlabs/tailwindcss/discussions/6694#discussioncomment-3489265

They say they will allow a more flexible use of layers in v4 in order to overcome this issue, but who knows when it's gonna be there.

----
📚 Documentation preview 📚: https://volto--6065.org.readthedocs.build/

### DIFF
--- a/packages/components/news/6065.bugfix
+++ b/packages/components/news/6065.bugfix
@@ -1,0 +1,1 @@
+Make css layer more specific @pnicolli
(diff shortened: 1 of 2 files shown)


sneridagh (approved):
LGTM! This was a trial anyways, we can continue further exploring this

PR #6067, opened 2024-06-02
[slots] `@plone/slots` first developments - Login form

sneridagh:
## Third level strategic package experiments
Experiments on the `@plone/slots` package. First trials on the shape as a not transpiled, not Volto add-on package to make it work-able as a `node_modules` package, and compatible with Next.js, vite-based frameworks and ultimately, Volto-Razzle powered, just in case they are necessary there too. In order to make it work, this kind of packages must not be under `src` directory. Let's not fight the ecosystem anymore in this matter. If the package is not being transpiled, the direct resolution must work OOTB, so the bundlers resolution is direct, and the TypeScript resolution too.

## Login Component
First iteration of a real life component in `@plone/slots` using the other 1-2 levels `@plone/*` packages. Separation of concerns with a driver HOC, to improve shadow-ability and simplify custom-ability and improve extensibility story.

It's still rough, though, we should continue polishing the ideas.


davisagli:
This was superceded by https://github.com/plone/volto/pull/6409

Issue #6071, opened 2024-06-05
Login status is not properly determined by the application


tiberiuichim:
How to reproduce:

- open demo.plone.org in two separate tabs
- login with admin in one of the tabs
- in the other tab, navigate to another page using an internal link (nav menu)

Expected: I should be able to see the toolbar, because the actions are properly sent together with the main content, thanks to the expand=actions

I think the token props is not properly reconstructed based on succesful backend interaction.

https://github.com/plone/volto/blob/a2f8d5997030dc8becc0c403c03f327edb5d0e36/packages/volto/src/components/manage/Toolbar/Toolbar.jsx#L373C6-L373C23

ichim-david:
Confirming the testing of this behavior within https://volto.demo.plone.org/

tiberiuichim:
Another one:

* Have some published pages available from the global nav menu
* With two open tabs, login, refresh them both
* Logout in one of the tabs
* In the other tab, click on one of the published public pages
* You get Unauthorised

This is because the api sends the store token, which is invalidated and properly fails in the acl_users somewhere.

Ni1esh-Yadav:
@tiberiuichim I am not able to reproduce the bug can you help me out??

stevepiercy:
@Ni1esh-Yadav please read and follow [First-time contributors](https://6.docs.plone.org/contributing/first-time.html), especially Item 2 in [Things not to do](https://6.docs.plone.org/contributing/first-time.html#things-not-to-do), [Contributing to Plone](https://6.docs.plone.org/contributing/index.html), and [Contributing to Volto](https://6.docs.plone.org/contributing/volto.html).

Shyam-Raghuwanshi:
anything pending in this?

PR #6079, opened 2024-06-07
When user changes location, set the userSession.token value based on cookie

tiberiuichim:
A fix for #6071

The problem is like this: the userSession.token is written in 2 cases:

- normal login/logout
- [in the SSR rendered HTML from the server](https://github.com/plone/volto/blob/02fb83c812012e62cd2478de73735159c43e703d/packages/volto/src/server.jsx#L194)

With this change, when the user performs a `GET_CONTENT` (perhaps it should be a route change?), we trigger a fake login/logout success action, to mutate the `userSession.token` in the store.

----
📚 Documentation preview 📚: https://volto--6079.org.readthedocs.build/


sneridagh:
@tiberiuichim I assume this is WIP still, right? Ping me when you are done with it.

stevepiercy:
@tiberiuichim is this ready to merge? CI is all green.

packages/volto/src/middleware/api.js
```diff
@@ -222,17 +222,40 @@ const apiMiddlewareFactory =
...
+            //   } else if (state.userSession?.token && !token) {
+            //     dispatch({
+            //       type: 'LOGOUT_SUCCESS',
+            //       result: {
+            //         token,
+            //       },
+            //     });
+            //   }
+            // }
+
```
stevepiercy, line 254: It's not clear why you would add commented code. Suggest removing all the above.

stevepiercy (changes requested):

davisagli (approved):
Looks good and works as expected.

PR #6080, opened 2024-06-07
Fix deselecting lists.

robgietema:
----
📚 Documentation preview 📚: https://volto--6080.org.readthedocs.build/

### DIFF
--- a/packages/volto-slate/news/6080.bugfix
+++ b/packages/volto-slate/news/6080.bugfix
@@ -1,1 +1,1 @@
-Fix deselecting lists. @robgietema
+Fix removal of slate formatting applied to text when toggling the list buttons @robgietema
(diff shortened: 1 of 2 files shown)


ichim-david (approved):
@robgietema I tested your fix and now when you go from list to non list you are no longer loosing the formatting.
Here is the behavior before your change:

https://github.com/plone/volto/assets/152852/5fc99213-fb39-4730-b895-7105452e3a47

and here is after your fix:

https://github.com/plone/volto/assets/152852/0e3f5e2f-2a53-41da-b04e-64345d61df15

My only suggestion would be to add a bit more description to the bugfix
entry, something along the lines of:
Fix removal of slate formatting applied to text when toggling the list buttons.

PR #5809, opened 2024-02-28
Improve wayfinding for various Volto audiences

stevepiercy:
Closes #5730

See also https://github.com/plone/documentation/pull/1630

PR #6085, opened 2024-06-09
Fix link in pop-up RelationsMatrix.jsx

stevepiercy:
This link is outdated after https://github.com/plone/volto/pull/5809, and its anchor never existed. This PR corrects both issues.

----
📚 Documentation preview 📚: https://volto--6085.org.readthedocs.build/

### DIFF
--- a/packages/volto/news/6085.bugfix
+++ b/packages/volto/news/6085.bugfix
@@ -1,0 +1,1 @@
+Fix link in pop-up in `RelationsMatrix.jsx`. @stevepiercy
--- a/packages/volto/src/components/manage/Controlpanels/Relations/RelationsMatrix.jsx
+++ b/packages/volto/src/components/manage/Controlpanels/Relations/RelationsMatrix.jsx
@@ -414,7 +414,7 @@ const RelationsMatrix = (props) => {
                           <Popup
                             trigger={
                               <a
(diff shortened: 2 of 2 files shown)


ichim-david:
@stevepiercy  the selection of the toc is a bit strange with your link
relation-fields
It shows in the toc the previous section selected when you open the link.
I suggest we simply use: https://6.docs.plone.org/volto/development/widget.html#relation-fields
then it shows up like this:
relation-fields-toc

stevepiercy:
@ichim-david I get different results. The `id` is for the `span` above the `h2`. I don't know how you get different results.

Screenshot 2024-06-09 at 7 17 56 AM

stevepiercy:
Also in Sphinx and MyST, it is preferred to use the label, not the heading text, as that heading text may change, but the label almost never does. When the label changes, it is much easier to do a find and replace.

ichim-david (approved):

stevepiercy:
@ksuess is the link with its anchor correct? I made an educated guess, but I'd like your feedback.

ksuess:
Yes, https://6.docs.plone.org/volto/development/widget.html#widget-relation-field-label where the info is located. Thank you for bringing this up and taking time to find a solution!

ksuess (approved):

PR #6093, opened 2024-06-11
Fixes #6046 pass proper defaults to align and size fields of Image block

### LINKED ISSUE #6046: Widgets should not fake present a default value
https://github.com/plone/volto/blob/b1797b4c256e18e7e300f5c50d2565d2cf44699b/packages/volto/src/components/manage/Widgets/ImageSizeWidget.jsx#L60

@ichim-david  please explain the problem

ichim-david:
Fixes #6046 pass proper defaults to align and size fields of Image block

----
📚 Documentation preview 📚: https://volto--6093.org.readthedocs.build/

### DIFF
--- a/packages/volto/news/6046.bugfix
+++ b/packages/volto/news/6046.bugfix
@@ -1,0 +1,2 @@
+Add default 'l' and 'center' values to size and align fields of `Image` block.
+This fixes data not having any value adding proper options to the `Image` block. @ichim-david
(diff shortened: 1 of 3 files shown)


davisagli (approved):

ichim-david:
@davisagli @sneridagh technically speaking this is a breaking change because now sizes="100vw" will be added by default.
This should happen since you see the "center" and "L" size selected by default due to the "false" defaults but before since it wasn't passed in data this code didn't run.
https://github.com/plone/volto/blob/c08e7a38bf89da4318bfc880ff3736ad8c4cda3f/packages/volto/src/components/manage/Blocks/Image/utils.js#L4

Take into consideration that if you were to click on a different align and size and then set it back to "center" and "L"
then this behavior did happen.

https://github.com/plone/volto/assets/152852/74a0c4ab-2e21-4156-b428-09b836714716

I have no problem modifying the issue to bugfix and adding a point in the upgrade guide that now we fix the fact that sizes weren't added even though "center" and "L" size looked like they were selected and as such sizes should have been added.

PR #6099, opened 2024-06-13
Improve container detection, `config.settings.containerBlockTypes` is no longer needed

sneridagh:
This fixes the order tab to detect all containers, including EEA's nested ones in `data` keys.

@ichim-david I updated also the pagination `findBlocks` helper too, with the new code, please take a look.

----
📚 Documentation preview 📚: https://volto--6099.org.readthedocs.build/


ichim-david (approved):
I like it because it tests now recursively for data instead of checking for a particular block type

sneridagh:
@ichim-david an edge case, fixed.

PR #6102, opened 2024-06-14
fix: nonContentRoutes diff path

giuliaghisini:
fixed "diff" path in nonContentRoutes.
If we had a path like "/test/test-a/difference" where the last part of url contains the word 'diff' it was considered as non-content route and it was wrong

----
📚 Documentation preview 📚: https://volto--6102.org.readthedocs.build/


packages/volto/news/6102.fix
pnicolli, line 1: Change filename to 6102.bugfix, otherwise towncrier won't get it

packages/volto/news/6102.fix
```diff
@@ -0,0 +1 @@
+fixed 'diff' path for nonContentRoutes. @giuliaghisini
```
pnicolli, line 1:
```suggestion
Fixed 'diff' path for nonContentRoutes. @giuliaghisini
```

Nice and clean with capitalized first letter 😄

pnicolli (changes requested):

sneridagh (approved):

sneridagh:
@giuliaghisini I fixed the changelog, although it seems that there is one test failing, could you please take a look? Thx!!

PR #6112, opened 2024-06-18
Prevent duplicated UUUIDs in inner blocks when copying container blocks

sneridagh:
----
📚 Documentation preview 📚: https://volto--6112.org.readthedocs.build/


packages/volto/src/helpers/Blocks/cloneBlocks.test.js
```diff
@@ -0,0 +1,54 @@
...
+      blocks: {
+        a: { '@type': 'slate', value: 1 },
+        b: { '@type': 'slate', value: 2 },
+      },
+      blocks_layout: { items: ['a', 'b'] },
+    };
+    const result = cloneBlocks(blocksData);
+
+    expect(result.blocks_layout.items[0]).not.toBe('a');
+    expect(result.blocks[result.blocks_layout.items[0]]).not.toBe({
```
davisagli, line 16: Isn't this the assertion we want here?

```suggestion
    expect(result.blocks[result.blocks_layout.items[0]]).toStrictEqual({
```

packages/volto/src/components/manage/Form/BlocksToolbar.jsx
```diff
@@ -97,8 +98,8 @@ export class BlocksToolbarComponent extends React.Component {
         return mode === 'copy'
           ? blockConfig.cloneData
             ? blockConfig.cloneData(blockData)
-            : [uuid(), blockData]
-          : [uuid(), blockData]; // if cut/pasting blocks, we don't clone
+            : [uuid(), cloneBlocks(blockData)]
+          : [uuid(), cloneBlocks(blockData)]; // if cut/pasting blocks, we don't clone
       })
       .filter((info) => !!info); // some blocks may refuse to be copied
```
davisagli, line 104: Can we avoid duplicating cloneWithIds between here and cloneBlocks.ts?

sneridagh: @davisagli do you mean unify all in a single helper? or improve the condition?

packages/volto/src/helpers/Blocks/cloneBlocks.test.js
sneridagh: Fixed

packages/volto/src/components/manage/Form/BlocksToolbar.jsx
davisagli: @sneridagh I mean use a helper function in both places

sneridagh: @davisagli let's talk before you leave, I think that the condition was wrong, from the beginning. If it's not copy, does the same as copy, maybe it's a copy/paste omission error? I'd say if it's move, we should not clone.

davisagli: @sneridagh Ah I see what you are saying.

You can ignore my comment here, I thought this was the same code as in cloneBlocks.ts but it isn't.

But you're right, cloneBlocks should not be used when moving

sneridagh: @davisagli okay, I implemented a way to preserve the original uuid if it's a cut operation.

packages/volto/src/components/manage/Form/BlocksToolbar.jsx
```diff
@@ -91,14 +92,14 @@ export class BlocksToolbarComponent extends React.Component {
...
+      .filter(([blockId, blockData]) => blockId && !!blockData['@type']) // Removes null blocks
+      .map(([blockId, blockData]) => {
         const blockConfig = config.blocks.blocksConfig[blockData['@type']];
         return mode === 'copy'
           ? blockConfig.cloneData
             ? blockConfig.cloneData(blockData)
-            : [uuid(), blockData]
-          : [uuid(), blockData]; // if cut/pasting blocks, we don't clone
+            : [uuid(), cloneBlocks(blockData)]
+          : [blockId, cloneBlocks(blockData)]; // if cut/pasting blocks, we don't clone
```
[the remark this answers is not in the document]
sneridagh: ummm correct!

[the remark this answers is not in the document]
sneridagh: Fixed.

davisagli (approved):
@sneridagh That looks better now

PR #6114, opened 2024-06-18
Rename files with wrong extension `js->jsx` when they contain JSX.

sneridagh:
This is in preparation for Vite, who does not allow you to mix and match, and it's better.
Along with #6113 it won't be breaking, the shadows will continue working as they are.

----
📚 Documentation preview 📚: https://volto--6114.org.readthedocs.build/


ichim-david:
@sneridagh tests fail
    TypeError: (0 , _Loadable.injectLazyLibs)(...) is not a function

      94 | };
      95 |
    > 96 | export default injectLazyLibs(['reactBeautifulDnd'])(
         |                                                     ^

sneridagh:
@ichim-david fixed!

PR #6124, opened 2024-06-25
Added three missing German translations.

jensens:
----
📚 Documentation preview 📚: https://volto--6124.org.readthedocs.build/


packages/volto/locales/de/LC_MESSAGES/volto.po
```diff
@@ -2495,12 +2495,12 @@ msgstr "Objekt-Browser öffnen"
...
 #: components/manage/Sidebar/Sidebar
 msgid "Order"
-msgstr ""
+msgstr "Reihung"
 
 #. Default: "Ordered"
 #: components/manage/Blocks/ToC/Schema
 msgid "Ordered"
-msgstr ""
+msgstr "Sortiert"
```
davisagli, line 2503: This sets whether to use an `<ol>` element -- i.e. to produce a numbered list in the ToC. The list is not sorted in any way, but still shown in the order of the headings from the page's blocks.

jensens: It is about using numbers vs dots? Then the term is  "Nummeriert".

davisagli (approved):

PR #6126, opened 2024-06-26
fix: state and styles for toolbar handler for hp

sabrina-bongiovanni:
Added a new class for the toolbar handler in the homepage based on the path.

- The homepage is the only page with path `''`.
- The homepage is always public, so it needs a style regardless of the review state of the page
- Since it's always public, I used the `@teal-blue` color used for public pages

----
📚 Documentation preview 📚: https://volto--6126.org.readthedocs.build/


sneridagh:
@sabrina-bongiovanni didn't we have already a className that identifies the root in the body class?

ichim-david:
@sneridagh @sabrina-bongiovanni there is indeed .contenttype-plone-site class on the body.
there could be a selected constructed from this class  adding the same styles as the publish
without adding extra code in the toolbar.

sabrina-bongiovanni:
> @sneridagh @sabrina-bongiovanni there is indeed .contenttype-plone-site class on the body. there could be a selected constructed from this class adding the same styles as the publish without adding extra code in the toolbar.

@ichim-david right! Just committed the redacted changes

packages/volto/theme/themes/pastanaga/extras/toolbar.less
```diff
@@ -752,3 +752,10 @@ body:not(.has-sidebar):not(.has-sidebar-collapsed) {
   }
 }
 // End Orphaned CSS
+
+// Toolbar handler color in homepage
+.contenttype-plone-site {
+  #toolbar .toolbar .toolbar-handler button:before {
```
ichim-david, line 758:
```suggestion
  #toolbar .toolbar-handler button:before {
```
@sabrina-bongiovanni you can keep the selector as specific as it was 3 levels plus the body override, there is no reason to have .toolbar in there if we look at how the background color is defined on the button
teal-override

Personally, it bothers me that we have a generic button selector which is less efficient than a class selector.
I would add ".toolbar-handler-button" here https://github.com/plone/volto/blob/main/packages/volto/src/components/manage/Toolbar/Toolbar.jsx#L627
so that we always have such a class and use it instead of the simple button.
For published or draft it can still be
 #toolbar .toolbar-handler .published:before
#toolbar .toolbar-handler .private:before
but toolbar.less contains a bunch of button selectors which is a pity :)

sabrina-bongiovanni: @ichim-david That makes sense. Changed the class and the selectors inside the .less file. Let me know if this is okay now :)

ichim-david:
@sabrina-bongiovanni you need to run pnpm test and update the test snapshot. This is why you have the unit tests failing.

ichim-david (approved):

packages/volto/theme/themes/pastanaga/extras/toolbar.less
ichim-david: > Personally, it bothers me that we have a generic button selector which is less efficient than a class selector. I would add ".toolbar-handler-button" here https://github.com/plone/volto/blob/main/packages/volto/src/components/manage/Toolbar/Toolbar.jsx#L627 so that we always have such a class and use it instead of the simple button. For published or draft it can still be #toolbar .toolbar-handler .published:before #toolbar .toolbar-handler .private:before but toolbar.less contains a bunch of button selectors which is a pity :)

@sabrina-bongiovanni I added this comment as my personal preference but I didn't make a suggestion for you to change this. I personally like this but technically speaking now since it modifies the testing structure it's a breaking change, although it simply adds a new class and not removing something.

The fix could have been fine with just this section that you've added https://github.com/plone/volto/pull/6126/files#diff-db29f0e3d44ba2993a71b0198daf2217864bf4318283240e32e3db9fcc25b32dR757

To be determined of @sneridagh opinion.

pnicolli (approved):

sneridagh (approved):

PR #6127, opened 2024-06-26
(feat): Add loading visual and succes message in controlpanel when deleting users and groups

dobri1408:
This pull request adds a loading indicator for deletion requests and provides a toast for successful deletion action for both user and group deletion operations.

See video for before and after effect:

https://github.com/plone/volto/assets/152852/e69b1bc5-ffb9-4f62-b437-cd0dc2aec3f6

----
📚 Documentation preview 📚: https://volto--6127.org.readthedocs.build/


ichim-david (approved):

ichim-david:
@sneridagh I have already pre-tested the feature to ensure that it works. We had a situation at EEA where on ldap users it was talking a long time to delete an user and there was no indication regarding the operation.
This change brings the same behavior from the Add user "loading" if necessary plus toast message also to the delete action.

Might be worth adding also onDeleteUserError and onDeleteGroupError in case we run into an error since there is such an option for onAddUserError and onAddGroupError. To be determined if it's worth it or not.

stevepiercy (approved):
I think adding the error messages, as @ichim-david, would be a good addition as well. Your call on those.

sneridagh (approved):
@ichim-david @dobri1408 @stevepiercy Agree that improve it would be great, but for now I'm merging this.

PR #6132, opened 2024-06-28
Cleanup Image widget and pass down `onSelectItem` prop if any

sneridagh:
@dobri1408 @ichim-david integrating it into VLT I found this.

----
📚 Documentation preview 📚: https://volto--6132.org.readthedocs.build/

### DIFF
--- a/packages/volto/src/components/manage/Blocks/Image/Edit.jsx
+++ b/packages/volto/src/components/manage/Blocks/Image/Edit.jsx
@@ -4,13 +4,8 @@
  */
 
 import React from 'react';
-import { connect } from 'react-redux';
-import { compose } from 'redux';
-
-import { injectIntl } from 'react-intl';
 import cx from 'classnames';
 import { ImageSidebar, SidebarPortal } from '@plone/volto/components';
-import { createContent } from '@plone/volto/actions';
 
 import {
   flattenToAppURL,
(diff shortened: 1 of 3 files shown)


packages/volto/src/components/manage/Blocks/Image/Edit.jsx
```diff
@@ -110,6 +93,7 @@ function Edit(props) {
             block={props.block}
             id={props.block}
             objectBrowserPickerType={'image'}
+            onSelectItem={props.onSelectItem}
```
ichim-david, line 96: @sneridagh I don't understand this passing of onSelectItem because at least in volto core there is no props.onSelectItem for Image block Edit.jsx. I do understand giving the ability to call this method  if it's passed in the props of ImageWidget.jsx but in the case of Image block it's simply undefined and the fallback is called.

It's nice that you found more things to cleanup such as the extra compose actions!

ichim-david (approved):

dobri1408 (approved):

packages/volto/src/components/manage/Blocks/Image/Edit.jsx
dobri1408: Yes, I agree, I think that it can be removed.

sneridagh: @ichim-david @dobri1408 Agreed! yeah, I'll remove it.

sneridagh: Done.

PR #6150, opened 2024-07-03
Fix aria-label in ObjectBrowserNav component

sneridagh:


ichim-david (approved):

packages/volto/news/6150.bugfix
```diff
@@ -0,0 +1 @@
+Fix aria-label in ObjectBrowserNav component @sneridagh
```
ichim-david, line 1:
```suggestion
Fix aria-label of items that are `folderish` in ObjectBrowserNav component when performing item search. Previously it said `Select item.title` now `Browser item.title`. This brings it in line with the aria-label when not performing an item search. @sneridagh
```

ichim-david: @sneridagh it might be too verbose, feel free to summarise further, Fix aria-label in ObjectBrowserNav is a bit too generic and encompassing IMOHO.

stevepiercy: Explicit > Implicit, even if it means verbosity.

sneridagh:
@ichim-david @stevepiercy Suggestion applied! Thanks a lot!

PR #6153, opened 2024-07-04
Debounced searching for users and groups in controlpanel

pnicolli:
Small performance improvement. This avoids spamming the backend with requests.

Not sure if the news item should be Internal or Feature, it does not really feel like a bugfix to me.


ksuess (approved):

ichim-david (approved):

packages/volto/news/6153.internal
```diff
@@ -0,0 +1 @@
+Debounced searching for users and groups in controlpanel @pnicolli
```
ichim-david, line 1:
```suggestion
Debounced searching for users and groups in the `User Group Membership` Control Panel. @pnicolli
```

PR #6147, opened 2024-07-03
Teaser Block Image Overide

Tishasoumya-02:
#6137

PR #6156, opened 2024-07-07
Allow Image Widget value to be an object in which case we use the '@id' for value of the Image Sidebar preview

ichim-david:
@sneridagh I am not proposing we change the teaser preview_image widget however I was trying as a test bench for the new image widget what would be needed in order to remove the object browser widget and use the new image widget.

Given how the ImageWidget assumes the value is only a string and Teaser/ObjectBrowser always assumes an array of value 1 with an object I think the ImageWidget should be more flexible in displaying the "image preview" by checking the value for the array first result.

Taking the work from https://github.com/plone/volto/pull/6147 these changes would display the external images set as preview_image overrides, any image selected or any image uploaded.

What I would propose is to make the ImageWidget a bit more flexible as seen in the changes made  here
[packages/volto/src/components/manage/Widgets/ImageWidget.jsx](https://github.com/plone/volto/compare/teaser_image_widget?expand=1#diff-0a16812deda14b522753664f63a68d086529f0c024a187dea1a62f97911b8dd0R81)

What do you think about my proposal?


ichim-david:
Removed the proof of content code in favor of the proposed fix to the ImageWidget. If anyone wants to try the Teaser block using the ImageWidget you can check out this commit https://github.com/plone/volto/pull/6156/commits/f68545097110dbec32932991d2907ae4d4af398d

sneridagh (approved):

PR #6161, opened 2024-07-10
Add foundations for extensible validation in forms

sneridagh:


ericof:
@wesleybl, @luxcas This is about getting validation also on blocks

docs/source/configuration/validation.md
```diff
@@ -0,0 +1,197 @@
...
+```ts
+config.registerComponent({
+  name: 'fieldValidator',
+  dependencies: ['default', 'minLength'],
+  component: minLengthValidator,
+});
+```
+
+It takes two `dependencies`.
+The first element should be the `default` identifier, and the second you can set it up to identify the validator.
```
tiberiuichim, line 34: It's not explained why `default` should be the first identifier.

docs/source/configuration/validation.md
```diff
@@ -0,0 +1,197 @@
...
+It can be any string.
+
+### Per field `type` validators
+
+These validators are applied depending on the specified `type` of the field.
+
+```ts
+config.registerComponent({
+  name: 'fieldValidator',
+  dependencies: ['integer', 'maximum'],
```
tiberiuichim, line 45: Here the `default` is missing.

docs/source/configuration/validation.md
```diff
@@ -0,0 +1,197 @@
...
+You should specify the validator in the JSON schema of the block (or as additional data in the content type definition).
+
+```ts
+let blockSchema = {
+  properties: {
+    ...schema.properties,
+    customField: {
+      title: 'Default field',
+      description: '',
+      validator: 'isURL',
```
tiberiuichim, line 150: Maybe allow passing directly a validator function as well? We're way past the point "schemas should be serializable" and for one-offs it's easier to declare them inline.

sneridagh: If the pattern we are teaching is go for "named" components and call them by name (partially because of the duality forms from the backend/content types and the client defined forms in blocks), I'd refrain to do so, because then we will have two use cases, one where you can do it, and another use case in which you can't. 

Also, in terms of explaining this in docs will became tricky and separate this is already confusing (I've been there when reviewing this with @ericof). Some other considerations is testability, because in that regard, modularization >> inlining things.

So, in principle, I would be -1 for this.

docs/source/configuration/validation.md
sneridagh: ummm, in that line, the `default` is not needed here, because it's not a default validator, it's a `type` one, so the first dependency is the name of the type.

docs/source/configuration/validation.md
```diff
@@ -0,0 +1,204 @@
...
+### Per field `widget` validators
+
+These validators are applied depending on the specified `widget` of the field.
+
+```ts
+config.registerComponent({
+  name: 'fieldValidator',
+  dependencies: ['phoneNumber', 'isValidPhone'],
+  component: phoneValidator,
+});
```
davisagli, line 79: What makes this a widget validator instead of a type validator?

docs/source/configuration/validation.md
```diff
@@ -0,0 +1,204 @@
...
+
+```ts
+config.registerComponent({
+  name: 'fieldValidator',
+  dependencies: ['slider', 'url', 'isURL'],
+  component: urlValidator,
+});
+```
+
+The first dependency should be the `id` of the block, and the second the `id` of the field.
```
davisagli, line 130: Isn't it the `@type` of the block?

docs/source/configuration/validation.md
```diff
@@ -0,0 +1,204 @@
...
+      description: '',
+      validator: 'isURL',
+    },
+  },
+  required: [],
+};
+```
+
+It does not need to be tied to any field `type` or `widget` definition.
+It runs in addition to all the above, so it complements the other validators if any apply.
```
davisagli, line 165: My biggest piece of feedback on this overall is using an array for the `dependencies` feels like trying to fit a square peg into a round hole. It's quite hard to keep track of all the different ways validators can be registered and which values need to go in the dependencies for each one. And it also seems to be like there might be some ambiguous cases, like if there is a field type and a widget type with the same name.

I don't know if the registry supports this but would it be possible to make things a bit more explicit by specifying the dependencies as an object instead of an array? So for example:
```
config.registerComponent({
  name: 'fieldValidator',
  dependencies: {
    blockType: 'slider',
    fieldName: 'url',
    validatorName: 'isURL',
  },
  component: urlValidator,
});
```

sneridagh: @davisagli TBH, I'm not fully satisfied with it... I thought that the reason was mainly because we can't get it to work as satisfying as the ZCA works :) I envisioned the resolution algorithm using a string and concating the deps if we have an array, so I've played with that concept, even in the new `getComponents` registry API method.

Using an object would be possible too, I guess. I can play with the idea, and leave the original API in place also.

sneridagh: @davisagli I did an initial implementation of using an object instead of an array but I have doubts on how to hash the resultant key to store that object, so then `getComponents` work in the same simple way, under the same concepts that we had before. It does work, but under some assumptions that I do not like or they are not satisfactory/safe :/ 

Maybe if you have some time, we can talk about it.

tiberiuichim:
@sneridagh Is it possible to pass parameters to validators? For example, if I have a validator "minmax", would I have to create and register a custom one for each use case?

docs/source/configuration/validation.md
```diff
@@ -0,0 +1,204 @@
...
+    "property=og:title": "Form fields validation"
+    "keywords": "Volto, Plone, frontend, React, configuration, form, fields, validation"
+---
+
+# Client side form field validation
+
+Volto provides a mechanism for delivering form field validation in an extensible way.
+This extensibility is based on the Volto component registry.
+
+## Registering a validator
```
stevepiercy, line 15:
```suggestion
## Register a validator
```

docs/source/configuration/validation.md
```diff
@@ -0,0 +1,204 @@
...
+  dependencies: ['default', 'minLength'],
+  component: minLengthValidator,
+});
+```
+
+It takes two `dependencies` being the first element a fixed `default` identifier, and the second you can set it up to identify the validator itself.
+In the case of the example, this other dependency is `minLength`.
+It can be any string.
+
+### Per field `type` validators
```
stevepiercy, line 37:
```suggestion
### Field `type` validators
```

docs/source/configuration/validation.md
```diff
@@ -0,0 +1,204 @@
...
+
+These validators are applied depending on the specified `type` of the field.
```
stevepiercy, line 39:
```suggestion
Field `type` validators are applied according to the specified `type` of the field.
```

docs/source/configuration/validation.md
```diff
@@ -0,0 +1,204 @@
...
+  dependencies: ['integer', 'maximum'],
+  component: maximumValidator,
+});
+```
+
+It takes two `dependencies` since we can potentially have several validators for the same `type`.
+The first element should be the field `type`, and the second you should set it up to identify the validator itself.
+You should specify the `type` in the JSON schema of the block (in a content type, it is included in the default serialization of the field).
+If a field does not specify type, it assumes a `string` type as validator.
+The next example is for the use case of JSON schema defined in a block:
```
stevepiercy, line 53:
```suggestion
The following example shows how to use a JSON schema defined in a block.
```

docs/source/configuration/validation.md
```diff
@@ -0,0 +1,204 @@
...
+      title: 'My custom field',
+      description: '',
+      type: 'integer',
+    },
+  },
+  required: [],
+};
+```
+
+### Per field `widget` validators
```
stevepiercy, line 70:
```suggestion
### Field `widget` validators
```

docs/source/configuration/validation.md
```diff
@@ -0,0 +1,204 @@
...
+
+These validators are applied depending on the specified `widget` of the field.
```
stevepiercy, line 72:
```suggestion
Field `widget` validators are applied according to the specified `widget` for the field.
```

docs/source/configuration/validation.md
```diff
@@ -0,0 +1,204 @@
...
+      title: 'Phone number',
+      description: '',
+      widget: 'phoneNumber',
+    },
+  },
+  required: [],
+};
+```
+
+### Per behavior and field name validator
```
stevepiercy, line 102:
```suggestion
### Behavior and field name validator
```

docs/source/configuration/validation.md
```diff
@@ -0,0 +1,204 @@
...
+
+These validators are applied depending on the behavior (usually coming from a content type definition) in combination with the name of the field.
```
stevepiercy, line 104:
```suggestion
Behavior and field name validators are applied according to the behavior combined with the name of the field.
The behavior usually comes from a content type definition.
```

docs/source/configuration/validation.md
```diff
@@ -0,0 +1,204 @@
...
+
+```ts
+config.registerComponent({
+  name: 'fieldValidator',
+  dependencies: ['plone.eventbasic', 'start', 'startValidator'],
+  component: startEventDateRangeValidator,
+});
+```
+
+The first dependency should be the name of the behavior, and the second the name (`id`) of the field.
```
stevepiercy, line 114:
```suggestion
The `dependencies` key accepts an array of two or three values.
The first element is a string that is the name of the behavior.
The second element is the name (`id`) of the field.
```

docs/source/configuration/validation.md
```diff
@@ -0,0 +1,204 @@
...
+It can get a third dependency in case you want to specify several validators for the same behavior - field id combination.
```
stevepiercy, line 115:
```suggestion
If you want to specify several validators for the same behavior, you can provide an optional third dependency as a field id combination.
```

docs/source/configuration/validation.md
```diff
@@ -0,0 +1,204 @@
...
+This type of validator only applies to content-type validators.
```
stevepiercy, line 116:
```suggestion
This type of validator applies only to content type validators.
```

docs/source/configuration/validation.md
```diff
@@ -0,0 +1,204 @@
...
+
+### Per block type and field name validator
```
stevepiercy, line 118:
```suggestion
### Block type and field name validator
```

docs/source/configuration/validation.md
```diff
@@ -0,0 +1,204 @@
...
+
+These validators are applied depending on the block type in combination with the name of the field in the block settings JSON schema.
```
stevepiercy, line 120:
```suggestion
Block type and field name validators are applied according to the block type combined with the name of the field in the block settings JSON schema.
```

docs/source/configuration/validation.md
```diff
@@ -0,0 +1,204 @@
...
+
+```ts
+config.registerComponent({
+  name: 'fieldValidator',
+  dependencies: ['slider', 'url', 'isURL'],
+  component: urlValidator,
+});
+```
```
stevepiercy, line 128: Move example after description.

docs/source/configuration/validation.md
stevepiercy, line 130:
```suggestion
The `dependencies` key accepts an array of two values.
The first element is a string that is the `type` of the block.
The second element is the `id` of the field.
```

docs/source/configuration/validation.md
```diff
@@ -0,0 +1,204 @@
...
+
+The first dependency should be the `id` of the block, and the second the `id` of the field.
+It can get a third dependency in case you want to specify several validators for the same block type - field id combination.
```
stevepiercy, line 131:
```suggestion
To specify several validators for the same block type, you can provide an optional third value as a field id combination.
```

docs/source/configuration/validation.md
```diff
@@ -0,0 +1,204 @@
...
+This type of validator only applies to blocks.
```
stevepiercy, line 132:
```suggestion
This type of validator applies only to blocks.
```

docs/source/configuration/validation.md
```diff
@@ -0,0 +1,204 @@
...
+
+### Specific validator using the `validator` key in the field
+
+A final type of validator is applied to the field if the `validator` key is present in the JSON schema definition of the form field.
```
stevepiercy, line 136:
```suggestion
You can apply a specific validator to a field by using the `validator` key in the JSON schema definition of the form field.
```

docs/source/configuration/validation.md
```diff
@@ -0,0 +1,204 @@
...
+
+```ts
+config.registerComponent({
+  name: 'fieldValidator',
+  dependencies: ['isURL'],
+  component: urlValidator,
+});
+```
```
stevepiercy, line 144: Move after description.

docs/source/configuration/validation.md
```diff
@@ -0,0 +1,204 @@
...
+
+The dependencies take one single name, in this case, the name of the validator.
```
stevepiercy, line 146:
```suggestion
The `dependencies` key accepts an array with a single value, the name of the validator.
```

docs/source/configuration/validation.md
```diff
@@ -0,0 +1,204 @@
...
+You should specify the validator in the JSON schema of the block (or as additional data in the content type definition).
```
stevepiercy, line 147:
```suggestion
You should specify the validator either in the JSON schema of the block or as additional data in the content type definition.
```

docs/source/configuration/validation.md
```diff
@@ -0,0 +1,204 @@
...
+    ...schema.properties,
+    customField: {
+      title: 'Default field',
+      description: '',
+      validator: 'isURL',
+    },
+  },
+  required: [],
+};
+```
```
stevepiercy, line 162: Move example after description.

docs/source/configuration/validation.md
```diff
@@ -0,0 +1,204 @@
...
+
+It does not need to be tied to any field `type` or `widget` definition.
+It runs in addition to all the above, so it complements the other validators if any apply.
+
+## Volto's default validators
+
+Volto provides a set of validators by default, you can find them in this module: `packages/volto/src/config/validators.ts`
```
stevepiercy, line 169:
````suggestion
Volto provides a set of validators.
You can find them in the module {file}`packages/volto/src/config/validators.ts`.

```{todo}
Document the API as a reference in another file, after the dust settles.
```
````

File: docs/source/configuration/validation.md [hunk 29]
```diff
@@ -0,0 +1,204 @@
...
+
+### How to override them
```
stevepiercy, line 171:
```suggestion
### Override Volto's default validators
```

docs/source/configuration/validation.md
```diff
@@ -0,0 +1,204 @@
...
+
+You can override them in your add-on as any other component defined in the registry, by redefining them using the same `dependencies`, and providing your own.
```
stevepiercy, line 173:
````suggestion
You can override Volto's default validators in your add-on as any other component defined in the registry.
To do so, use the `dependencies` key, and provide your own values.

```{todo}
Provide usage example.
```
````

File: docs/source/configuration/validation.md [hunk 31]
```diff
@@ -0,0 +1,204 @@
...
+  // The field schema definition object
+  field: Record<string, any>;
+  // The form data
+  formData?: any;
+  // The intl formatMessage function
+  formatMessage: Function;
+};
+```
+
+This is an example of an `isNumber` validator:
```
stevepiercy, line 192:
```suggestion
The following example shows an `isNumber` validator.
```

docs/source/configuration/validation.md
```diff
@@ -0,0 +1,204 @@
...
+```ts
+export const isNumber = ({ value, formatMessage }: Validator) => {
+  const floatRegex = /^[+-]?\d+(\.\d+)?$/;
+  const isValid =
+    typeof value === 'number' && !isNaN(value) && floatRegex.test(value);
+  return !isValid ? formatMessage(messages.isNumber) : null;
+};
+```
+
+Using the `formData` you can perform validation checks using other field data as source.
```
stevepiercy, line 203:
```suggestion
Using `formData`, you can perform validation checks between multiple fields' data.
```

docs/source/configuration/validation.md
```diff
@@ -0,0 +1,204 @@
...
+This is interesting in the case that two fields are related, like `start` and `end` dates.
```
stevepiercy, line 204:
````suggestion
This is useful when you need to compare the values of multiple fields, such as ensuring that the end date and time of an event is after its start.

```{todo}
Provide usage example.
```
````

stevepiercy (changes requested):

ichim-david:
Closing this pull request since we have the superseding https://github.com/plone/volto/pull/6181

PR #6162, opened 2024-07-10
Remove `react-share` library and `SocialSharing` component

sneridagh:
The builds start to break because of `react-share` today, dependencies problem because we are using an ancient version.
Let's remove it, since core is not using it anymore.

----
📚 Documentation preview 📚: https://volto--6162.org.readthedocs.build/


sneridagh:
@stevepiercy Merging, since this is breking builds right now.

pnicolli:
@sneridagh I just created a new project with cookieplone because I wanted to finally try it and maybe jump on board and bumped into this. Maybe releasing a new alpha would help people that create new projects like I did. I can easily change mrs-developer for now so no rush, but I think it's needed.

sneridagh:
@pnicolli Today I knew that I overlooked that we would need a release for this, and I missed to do so. I will release Volto asap, I'll try this evening.

PR #6173, opened 2024-07-19
feat: publicUi routes to not have cms-ui in public nonContentRoutes

giuliaghisini:
if we have nonContentRoutes that aren't cms-ui, 
for example '/search'
now are considered as cmsUi. 
This adds unnecessary things to the page like:
- cms-ui class to the body tag

defining this public routes in config we can avoid this and everithing works well :-D


packages/volto/src/config/PublicUiRoutes.jsx
```diff
@@ -0,0 +1,17 @@
+import config from '@plone/volto/registry';
```
pnicolli, line 1: Remove unused import

packages/volto/src/config/index.js
```diff
@@ -8,6 +8,7 @@ import {
   layoutViewsNamesMapping,
 } from './Views';
 import { nonContentRoutes } from './NonContentRoutes';
+import { publicUiRoutes } from './PublicUiRoutes';
```
pnicolli, line 11: Proposal: I would suggest a more specific name for this config key, like `publicNonContentRoutes`

pnicolli (changes requested):

packages/volto/news/6173.feature
```diff
@@ -0,0 +1 @@
+Added config.settings.publicNonContentRoutes to avoid cms-ui in this public routes. @giuliaghisini
```
pnicolli, line 1:
```suggestion
Added config.settings.publicNonContentRoutes to avoid cms-ui in these public routes. @giuliaghisini
```

pnicolli (approved):

packages/volto/src/config/PublicUiRoutes.jsx
```diff
@@ -0,0 +1,17 @@
...
+  '/login',
+  '/logout',
+  '/sitemap',
+  '/register',
+  '/search',
+  '/change-password',
+  '/contact-form',
+  '/register',
+  /\/passwordreset\/.*$/,
+  '/passwordreset',
```
mamico:
```suggestion
  '/login$',
  '/logout$',
  '/sitemap$',
  '/register$',
  '/search$',
  '/change-password$',
  '/contact-form$',
  /\/passwordreset($|\/)/,
```
make regexp more rigingorous to solve the problem of match something like `/searching-content` ...

packages/volto/src/helpers/Url/Url.js
```diff
@@ -139,7 +139,10 @@ export const isCmsUi = memoize((currentPathname) => {
   // because the regexp test does not take that into account
   // https://github.com/plone/volto/issues/870
   return settings.nonContentRoutes.reduce(
-    (acc, route) => acc || new RegExp(route).test(`/${fullPath}`),
+    (acc, route) =>
+      acc ||
+      (settings.publicUiRoutes.indexOf(route) < 0 &&
+        new RegExp(route).test(`/${fullPath}`)),
```
mamico, line 145:
```suggestion
        new RegExp(route).test(fullPath)),
```
wen need the additional `/` ?

packages/volto/src/config/PublicUiRoutes.jsx
giuliaghisini: with $ this means that you don't consider parameters.. i've just copied path from nonContentRoutes. 
I think for now they colud be remain the same

packages/volto/src/helpers/Url/Url.js
giuliaghisini: yes, fullPath isn't an entry of the list above, but the current path 
nothing changed from before. it was as is and worked well

packages/volto/news/6173.feature
```diff
@@ -0,0 +1 @@
+Added config.settings.publicNonContentRoutes to avoid cms-ui in these public routes. @giuliaghisini
```
stevepiercy, line 1:
```suggestion
Added `config.settings.publicNonContentRoutes` to avoid `isCmsUi` issues in these public routes. @giuliaghisini
```

packages/volto/src/config/PublicNonContentRoutes.jsx
```diff
@@ -0,0 +1,15 @@
+// PublicUi routes that are noContentRoutes, and we won't cms-ui
```
stevepiercy, line 1:
```suggestion
// PublicUi routes that are nonContentRoutes and are not members of isCmsUi
```

packages/volto/src/config/PublicUiRoutes.jsx
stevepiercy: `$` would not match `/login/`, for example, which would be bad assuming they were redirected from a page that requires authentication and want to redirect them back to that page after authentication. I think it is better to leave all these open-ended here.

I'm not sure why we have this one-off, though. Is there a good reason not to simplify it to just the string `/passwordreset`?

```js
/\/passwordreset($|\/)/
```

packages/volto/src/helpers/Url/Url.js
stevepiercy: Does `currentPathname` include or exclude a leading forward slash (`/`)?

If it includes a slash, then we don't need to use one in the regexp test.

If it excludes a slash, then we need a slash in the regexp test.

packages/volto/src/config/PublicNonContentRoutes.jsx
```diff
@@ -0,0 +1,15 @@
...
+// You can include either RegEx or a string representing the ending of the
```
stevepiercy, line 2:
```suggestion
// You can include either RegExp or a string representing the ending of the
```

stevepiercy (changes requested):
Clean up news and comments, and a couple of questions.

stevepiercy (approved):
@plone/volto-team this PR needs a technical review.

ichim-david:
@mamico @giuliaghisini This doesn't look ready for testing yet due to the unit test failure
unitest-failures

packages/volto/src/config/PublicUiRoutes.jsx
mamico: I've doublchecked code here. IMO, thinking of fixing the regexes here is useless. The real regexes used are in `nonContentRoutes`, the `publicNonContentRoutes` list is about the regexes in `nonContentRoutes` that should not be used for matching when it comes to evaluating `isCMSui`. For better readability, I moved the two lists in the same file and added an explicit comment.

stevepiercy: This makes sense to me.

stevepiercy:
The failing CI check for readme-link-check is due to https://www.cmscom.jp not responding. @terapyon would you please look into this?

Otherwise it is all green, but could use one more round of reviews. @giuliaghisini would you please add your review? Thank you!

ichim-david:
@mamico you've turned a bugfix into a breaking change. If you modify where the config is at it will need a breaking notice and an upgrade mention. I would wait for this action though to hear what @davisagli thinks about your proposal to move the config setting.

ichim-david:
> @mamico you've turned a bugfix into a breaking change. If you modify where the config is at it will need a breaking notice and an upgrade mention. I would wait for this action though to hear what @davisagli thinks about your proposal to move the config setting.

@mamico disregard this statement I should of paid more attention that you added something instead of modifying the name and import. There is no documentation about this new option though

stevepiercy:
For docs, see https://6.docs.plone.org/volto/contributing/documentation.html.

@ichim-david as far as what to document and where, I don't see anything for `nonContentRoutes` in [Routing](https://6.docs.plone.org/volto/contributing/routing.html) (so sad), [The configuration registry](https://6.docs.plone.org/volto/configuration/how-to.html), or [Settings reference guide](https://6.docs.plone.org/volto/configuration/settings-reference.html). Would one of these be appropriate locations?

ichim-david:
> For docs, see https://6.docs.plone.org/volto/contributing/documentation.html.
> 
> @ichim-david as far as what to document and where, I don't see anything for `nonContentRoutes` in [Routing](https://6.docs.plone.org/volto/contributing/routing.html) (so sad), [The configuration registry](https://6.docs.plone.org/volto/configuration/how-to.html), or [Settings reference guide](https://6.docs.plone.org/volto/configuration/settings-reference.html). Would one of these be appropriate locations?

It would be good to get the setting documented and in the routing docs  to link to the routing based settings

giuliaghisini:
> > For docs, see https://6.docs.plone.org/volto/contributing/documentation.html.
> > @ichim-david as far as what to document and where, I don't see anything for `nonContentRoutes` in [Routing](https://6.docs.plone.org/volto/contributing/routing.html) (so sad), [The configuration registry](https://6.docs.plone.org/volto/configuration/how-to.html), or [Settings reference guide](https://6.docs.plone.org/volto/configuration/settings-reference.html). Would one of these be appropriate locations?
> 
> It would be good to get the setting documented and in the routing docs to link to the routing based settings

updated how-to file

docs/source/configuration/how-to.md
```diff
@@ -150,3 +150,24 @@ export default function applyConfig(config) {
 
   return config;
 }
+```
+
+## nonContentRoutes and publicNonContentRoutes
```
stevepiercy, line 155:
```suggestion
## `nonContentRoutes` and `publicNonContentRoutes`
```

docs/source/configuration/how-to.md
```diff
@@ -150,3 +150,24 @@ export default function applyConfig(config) {
...
+
+The `nonContentRoutes` is a list of routes declaration for which is applied the body-class `cms-ui`. This are the routes usesed in the `backoffice` side of Volto.
```
stevepiercy, line 157:
```suggestion
`nonContentRoutes` is a list of routes reserved in Volto for its functionality as a content management system.
These functions include user authentication and registration, changing settings through control panels, generating a site map, and other functions.
Examples of these routes include `/login`, `/register`, and `/\/controlpanel\/.*$/`.
Editors can't use them for content.
The HTML attribute class value `cms-ui` is applied to members of `nonContentRoutes`.
```

docs/source/configuration/how-to.md
```diff
@@ -150,3 +150,24 @@ export default function applyConfig(config) {
...
+You can include either RegEx or a string representing the ending of the url.
```
stevepiercy, line 158:
```suggestion
You can configure `nonContentRoutes` with either a regular expression or a string representing the end of the URI.
```

docs/source/configuration/how-to.md
```diff
@@ -150,3 +150,24 @@ export default function applyConfig(config) {
...
   return config;
 }
+```
+
+## nonContentRoutes and publicNonContentRoutes
+
+The `nonContentRoutes` is a list of routes declaration for which is applied the body-class `cms-ui`. This are the routes usesed in the `backoffice` side of Volto.
+You can include either RegEx or a string representing the ending of the url. 
+
+The `publicNonContentRoutes` is a list of nonContentRoutes used in the `public` side of Volto. For example, some static public route such as '/search', or '/sitemap', ...
```
stevepiercy, line 160:
```suggestion
`publicNonContentRoutes` is a subset of `nonContentRoutes`.
These routes are used for public sections of a Volto site that do not require authentication.
This subset includes `/login`, `/search`, and `/sitemap`.
```

docs/source/configuration/how-to.md
```diff
@@ -150,3 +150,24 @@ export default function applyConfig(config) {
...
+
```
stevepiercy, line 161:
```suggestion

The following example shows how to configure settings for `nonContentRoutes` and `publicNonContentRoutes`.

```

stevepiercy (changes requested):
I expanded on @giuliaghisini's documentation. This should be reviewed by another contributor.

mamico (approved):

stevepiercy (approved):
Thanks for checking the docs @mamico.

I think we should do one more technical review, as @mamico changes made after @pnicolli. Maybe @davisagli can fill in, or @pnicolli with a second review? Other folks are on vacation for another week or two.

davisagli (approved):
I think this is okay for Volto 18. It's arguably a breaking change to backport it, since some existing sites may have styling that depends on the `cms-ui` class being there for these views.

giuliaghisini:
@sneridagh could we merge this in volto 18?

ichim-david:
@plone/volto-team I am +1 on having such a feature however I am not completely sold on the new setting name.
If the new option would share the same base it would be easier to find the extension.
I'm thinking of naming such as:
nonContentRoutesPublic
nonContentRoutesPublicUI
nonContentRoutesOfPublicUI

If we were to have good typescript integration I think that having the same base would make it easier to find every option used.
Maybe someone has an even better naming example or the majority decides that publicNonContentRoutes is good enough, I am not strongly against the current suggestion but again I am not 100% sold either :).

ichim-david (approved):

stevepiercy:
I tend to follow the same naming convention suggested by @ichim-david in my projects: `BaseNameOptionalModifier` or `base-name-optional-modifier`. We did something similar for almost all of the make commands.

If a renaming refactor is done, I prefer the shorter proposed version `nonContentRoutesPublic`, as these are routes. I don't think the `UI` segment helps, as it is an implementation detail better described in code comments.

packages/volto/src/config/NonContentRoutesPublic.jsx
```diff
@@ -0,0 +1,15 @@
+// PublicUi routes that are noContentRoutes, and we won't cms-ui
```
stevepiercy, line 1:
```suggestion
// PublicUi routes that are nonContentRoutes, and should not be members of isCmsUi
```

packages/volto/src/config/NonContentRoutesPublic.jsx
```diff
@@ -0,0 +1,15 @@
...
+// You can include either RegEx or a string representing the ending of the
+// nonContentRoute eg. '/add' will match '/foo/bar/add'
```
stevepiercy, line 3:
```suggestion
// nonContentRoute. For example, '/add' will match '/foo/bar/add'.
```

stevepiercy (changes requested):
Thanks for the refactor! This LGTM, with a few minor comment suggestions. I could not understand two of them, so I made a best guess.

docs/source/configuration/how-to.md
```diff
@@ -1,10 +1,10 @@
...
 myst:
   html_meta:
-    "description": "Volto has a centralized configuration registry used to parameterize Volto."
-    "property=og:description": "Volto has a centralized configuration registry used to parameterize Volto."
-    "property=og:title": "The configuration registry"
-    "keywords": "Volto, Plone, frontend, React, configuration, registry"
+    'description': 'Volto has a centralized configuration registry used to parameterize Volto.'
+    'property=og:description': 'Volto has a centralized configuration registry used to parameterize Volto.'
+    'property=og:title': 'The configuration registry'
+    'keywords': 'Volto, Plone, frontend, React, configuration, registry'
```
stevepiercy, line 7: Please revert the above changes, and configure your editor not to convert " to ' for .md files.

stevepiercy (changes requested):

stevepiercy (approved):
Let's merge!

PR #6181, opened 2024-07-22
Extensible validation support in Volto forms

sneridagh:
Supersedes #6161 
Refactored and re-document for clarity and simplicity from it.

@stevepiercy I think now it's the good one. I'm more satisfied with this one now. I will review myself again, but I think it's ready for check.

@davisagli I've refactored it a bit after talking to @robgietema in lieu of clarity and simplicity. Also reflected it in the docs too. Decisions I made, we can talk about it:

## Create an API for `registerUtility`

I think this is much more cleaner and fixes the thing you've mentioned in the other PR, for the "trying to push the square shaped piece in the circle shape hole". Also, I took it from a previous conversation last week with @ericof but that's another story. TL;DR: it was about the convenience of having an API that does and registers things for you, whatever it is, if it's a widget or a view, or other thing where you should "touch" and manually set things in different places. This makes much more sense to me now, with the current shape.

It will also make sense in other scenarios where I think we will need such "named utilities", eg. in other frameworks, having utility helpers, like flattenToAppUrl, globally, then ask for them in a named way.

## Using the `format` property from JSON schema specs

@robgietema pointed me out to this one, I'm using the default property for defining named validators as defined in the spec.

I still have to cleanup, will continue working polishing this a bit.

----
📚 Documentation preview 📚: https://volto--6181.org.readthedocs.build/


docs/source/configuration/validation.md
```diff
@@ -0,0 +1,273 @@
...
+But the validation engine will take care too:
+
+```json
+{
+  "properties": {
+    "customField": {
+      "title": "Custom URL field",
+      "widgetOptions": {
+        "frontendOptions": {
+          "format": "url"
```
robgietema, line 91: Can we change the api call so it will be "flattened" to the root of the field so we only have 1 way specifying the format?

docs/source/configuration/validation.md
```diff
@@ -0,0 +1,273 @@
...
+  method: maximumValidator,
+})
+```
+
+You should specify the `type` in the JSON schema of the block (in a content type, it is included in the default serialization of the field).
+If a field does not specify type, it assumes a `string` type as validator.
+
+#### Per field `widget` validators
+
+These validators are applied depending on the specified `widget` of the field.
```
robgietema, line 142: In the above examples you have to specify in the json schema if the validator is enabled or not. In the registerUtility method you only specify if the validator is available or not. It seems in this example registering the validator will have it enabled by default. I think for consistency we should also specify it in the schema and not enable it by default.

robgietema (approved):

docs/source/configuration/validation.md
```diff
@@ -0,0 +1,273 @@
...
+    },
+  },
+  required: [],
+};
+```
+
+```ts
+config.registerUtility({
+  type: 'validator',
+  name: 'maximum',
```
robgietema, line 129: This is indeed part of the validation spec of json schema (same for maxLength, minLength, maxItems, pattern etc). Should we make these validators customisable? I think it would make more sense to have these "build-in" as validators since the name already implies what it should do.

packages/volto/news/6161.breaking
```diff
@@ -0,0 +1,5 @@
+Add foundations for extensible validation in forms @sneridagh
```
stevepiercy, line 1:
```suggestion
Add foundations for extensible validation in forms. @sneridagh
```

packages/volto/news/6161.breaking
```diff
@@ -0,0 +1,5 @@
...
+
+Breaking:
+`packages/volto/src/helpers/FormValidation/FormValidation.jsx` has been heavily refactored.
+If you have shadowed this component in your project/add-on, please review and update your shadow.
```
stevepiercy, line 5:
```suggestion
If you shadowed this component in your project or add-on, you should review it and update it accordingly.
```

docs/source/configuration/validation.md
```diff
@@ -0,0 +1,273 @@
+---
+myst:
+  html_meta:
+    "description": "Client side form field validation"
+    "property=og:description": "Client side form field validation"
+    "property=og:title": "Form fields validation"
```
stevepiercy:
```suggestion
    "property=og:title": "Client side form field validation"
```

stevepiercy (changes requested):
For docs, it's easier for me to just edit it, than use GitHub's klunky review, when there are this many things to change. Thus you should double-check that I did not screw up anything.

docs/source/configuration/validation.md
sneridagh: I had this wish back in the day too, but it never happened, since we already had this "scape hatch" for all customizable things in Python schema hints. Maybe we can push again in the backend side for this to happen.

@davisagli do you think we could try to unify this also in the backend? Or do you remember the reason behind we never did it?

docs/source/configuration/validation.md
sneridagh: What we have today as built in:
- minLength
- maxLenght
- minimum
- maximum
- uniqueItem

We could implement the rest, for sure.

packages/volto/locales/volto.pot
```diff
@@ -1264,6 +1264,11 @@ msgstr ""
 msgid "End Date"
 msgstr ""
 
+#. Default: "Event end date happens before the event start date"
```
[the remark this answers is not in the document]
sneridagh: Leaving this because this has to be paired with code in the validator side.

packages/volto/locales/volto.pot
```diff
@@ -3433,6 +3438,11 @@ msgstr ""
 msgid "Start Date"
 msgstr ""
 
+#. Default: "Event start date happens later than the event end date"
```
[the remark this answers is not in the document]
sneridagh: Leaving this because this has to be paired with code in the validator side.

docs/source/configuration/validation.md
sneridagh: I will clarify which validators are included and enabled by default, and put it on the top, instead than at the bottom.

packages/volto/locales/volto.pot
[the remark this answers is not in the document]
sneridagh: @stevepiercy fixed

packages/volto/locales/volto.pot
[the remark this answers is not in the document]
sneridagh: @stevepiercy fixed

packages/volto/src/config/validation.ts
```diff
@@ -0,0 +1,124 @@
...
+  maxLengthValidator,
+  urlValidator,
+  emailValidator,
+  isNumber,
+  maximumValidator,
+  minimumValidator,
+  isInteger,
+  hasUniqueItems,
+  startEventDateRangeValidator,
+  endEventDateRangeValidator,
```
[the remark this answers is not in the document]
sneridagh: @stevepiercy Validator suffix added where they were missing.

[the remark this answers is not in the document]
sneridagh: Added the `pattern` validator.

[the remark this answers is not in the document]
sneridagh: Added minItems/maxItems validator.

packages/volto/src/helpers/MessageLabels/MessageLabels.js
stevepiercy, line 1: Do we need to add new strings to `packages/volto/locales/volto.pot` as well?

packages/volto/locales/volto.pot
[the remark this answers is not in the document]
stevepiercy: Shweeeeet!

stevepiercy (changes requested):
A couple of my goofs in docs fixed on second reading.

Also a couple of questions to address.

docs/source/configuration/validation.md
stevepiercy: Yeah, having that at the end of the docs, with a `seealso` where it was first mentioned, didn't feel right. This would be a good change.

docs/source/configuration/validation.md
```diff
@@ -0,0 +1,297 @@
...
+These usually come from a content type definition.
+This type of validator only applies to content type validators.
+It takes the `behaviorName` and the `fieldName` as dependencies.
+
+```ts
+config.registerUtility({
+  type: 'validator',
+  name: 'dateRange',
+  dependencies: {
+    behaviorName: 'plone.eventbasic',
```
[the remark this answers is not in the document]
davisagli: @stevepiercy This behavior has been defined in plone.app.event for years. It can't be changed without breaking backwards-compatibility.

docs/source/configuration/validation.md
```diff
@@ -0,0 +1,297 @@
...
+  html_meta:
+    "description": "Client side form field validation"
+    "property=og:description": "Client side form field validation"
+    "property=og:title": "Client side form field validation"
+    "keywords": "Volto, Plone, frontend, React, configuration, form, fields, validation"
+---
+
+# Client side form field validation
+
+Volto provides a mechanism for delivering form field validation in an extensible way.
```
davisagli, line 12:
```suggestion
Volto provides an extensible way to validate form field values.
```

docs/source/configuration/validation.md
davisagli: It depends on the property. For things that are valid JSON Schema, it would make sense to have plone.restapi serialize them at the top level. For other properties that are not part of JSON Schema, it makes sense to keep them inside the widgetOptions.

docs/source/configuration/validation.md
```diff
@@ -0,0 +1,297 @@
...
+      }
+    }
+  }
+}
+```
+
+
+### Advanced scenarios
+
+In case you need more granularity, and you don't have access to modify the existing implementation of the JSON schema definitions for existing content types, blocks, or forms, you can use the following advanced validator registrations, using `field`, `widget`, `behaviorName`, or `blockType` validator registrations.
```
davisagli, line 110: I don't think "more granularity" is the right way to describe this. Nothing is more granular (specific) than assigning one specific validator to a specific field using `format`. This is less granular: it's a generic approach to register validators that apply to multiple fields automatically based on their field/widget/behaviorName/blockType. Right?

docs/source/configuration/validation.md
```diff
@@ -0,0 +1,297 @@
...
+  required: [],
+};
+```
+
+```ts
+config.registerUtility({
+  type: 'validator',
+  name: 'phoneNumber',
+  dependencies: {
+    widgetName: 'phoneNumber',
```
davisagli, line 179: Why is it `widget` in the schema but `widgetName` here?

docs/source/configuration/validation.md
```diff
@@ -0,0 +1,297 @@
...
+})
+```
+
+
+(voltos-default-validators-label)=
+
+## Volto's default validators
+
+Volto provides a set of validators by default.
+You can find them in the module {file}`packages/volto/src/config/validators.ts`.
```
davisagli, line 251: We should ideally document them here rather than just pointing to source code.

packages/registry/news/6161.feature
```diff
@@ -0,0 +1 @@
+Add `getComponents` that match a partial set of dependencies, given a name. @sneridagh
```
davisagli, line 1: This is out of date

packages/registry/src/index.ts
```diff
@@ -406,6 +420,68 @@ class Config {
...
+  }) {
+    const { name, type, method, dependencies = {} } = options;
+    let depsString: string = '';
+    if (!method) {
+      throw new Error('No method provided');
+    } else {
+      depsString = Object.keys(dependencies)
+        .map((key) => `${key}:${dependencies[key]}`)
+        .join('+');
+    }
```
davisagli, line 438: I think you need to sort the keys both here and when the utility is looked up. Otherwise it will be different depending on the order in which the keys were added to the object.

packages/types/news/6161.bugfix
```diff
@@ -0,0 +1 @@
+Add `errors` shape to the `BlockEditProps`. @sneridagh
```
davisagli, line 1: Also added types for Utilities...

davisagli (commented):
I like this approach a lot better than #6161!

packages/types/news/6161.bugfix
sneridagh: Done

packages/registry/src/index.ts
sneridagh: Done

packages/registry/news/6161.feature
sneridagh: Done.

docs/source/configuration/validation.md
sneridagh: Done.

docs/source/configuration/validation.md
sneridagh: Moved to use `widget` only.

docs/source/configuration/validation.md
sneridagh: @davisagli I tried to improve it, take a look if now it's better.

packages/volto/src/helpers/MessageLabels/MessageLabels.js
sneridagh: Updated!

packages/volto/src/config/validation.ts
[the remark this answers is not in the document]
sneridagh: @stevepiercy I added this:

https://github.com/plone/volto/pull/6181/files#diff-0fb5051092a06c331a0607394901c0217dc85e7d2f3ce393fadab348e0832b56R321-R343

docs/source/configuration/validation.md
sneridagh: @stevepiercy moved to the top.

sneridagh:
@davisagli @stevepiercy I think I went through all the suggestions, let me know if I left something!

packages/volto/src/components/manage/Form/Form.jsx
```diff
@@ -730,6 +793,7 @@ class Form extends Component {
                 history={this.props.history}
                 location={this.props.location}
                 token={this.props.token}
+                errors={this.state.errors.blocks}
```
tiberiuichim, line 796: We should pass the whole error state to the BlocksForm (or introduce a `blocksError` prop to the BlocksForm.

I haven't looked in detail, but maybe also move the computing of blocks error to the BlocksForm? 

The idea is that some blocks need to have access to the whole form error state. For example, the volto-metadata-block exposes Dexterity content fields such as Title, etc and makes them editable from a block. If there's an error for that field, the error exists in the Form component, but the block doesn't have access to that whole state form error variable.

sneridagh: Then the content type field is the one who has to have the validator, right? What's the point in having the same validator in the proxied metadata field?

Also, if you want to make an extra one, in the validator you have access to the full `formData` to check the current values. 

The point of doing it this way (also in a nested way in blocks) is because it does work in the same way for content types, forms and blocks so we don't have to come up with fancy new way of doing things. The whole blocks use case was implemented, literaly, withing 10 minutes, passing down the `error` prop.

IMO this is better than nothing which is the current state.

tiberiuichim: Let's take this situation: I want to expose the content item title field as editable field in a block. I can initialize a Field for that field, but I also want to be able to pass the errors prop to that field. The errors is computed in the Form, and not passed down to the BlocksForm. All I'm saying is we should pass the whole state.errors to the BlocksForm, rather then just state.errors.blocks

docs/source/configuration/validation.md
```diff
@@ -9,13 +9,54 @@ myst:
...
+- maxItems
+- minItems
+- uniqueItems
+
+### Per widget
+- email
+- url
+
+### Event content type
+- start/end dates check
```
stevepiercy, line 55:
```suggestion
- event end date must be on or after its start date
```

docs/source/configuration/validation.md
```diff
@@ -107,8 +144,7 @@ The `urlValidator` method validator will be applied for the content type field `
 
 ### Advanced scenarios
 
-In case you need more granularity, and you don't have access to modify the existing implementation of the JSON schema definitions for existing content types, blocks, or forms, you can use the following advanced validator registrations, using `field`, `widget`, `behaviorName`, or `blockType` validator registrations.
-
+If, for some reason, you can't modify the existing implementation of the JSON schema definitions for existing content types, blocks, or forms, you can use the following advanced validator registrations, which allows you to register validators depending on the `field`, the `widget`, the `behaviorName` (in case of a content type), or the `blockType` (in case of a block).
```
stevepiercy, line 147:
```suggestion
If, for some reason, you can't modify the existing implementation of the JSON schema definitions for existing content types, blocks, or forms, you can use the following advanced validator registrations.
These allow you to register validators according to whether it is a `field`, `widget`, `behaviorName` (for content types), or `blockType` (for blocks).
```

docs/source/configuration/validation.md
```diff
@@ -291,7 +319,21 @@ export const isNumber = ({ value, formatMessage }: Validator) => {
 Using `formData`, you can perform validation checks using other field data as source.
 This is useful when you want to validate two related fields, such as ensuring the end date of an event is after its start date.
 You can create invariant validator types.
+In the following code snippet you can check how to create a validator method for checking if the Event content type `start`/`end` fields are valid:
```
stevepiercy, line 322: Note that this narrative text reflects that the start and end dates of an event content type are different. I think I read this code sample correctly, but another pair of non-jetlagged eyes would be good.

```suggestion
The following code snippet shows how to create a validator method that ensures the event content type's end date is after its start date.
```

packages/registry/news/6161.feature
```diff
@@ -1 +1 @@
-Add `getComponents` that match a partial set of dependencies, given a name. @sneridagh
+Added Utilities registry, `registerUtility`, `getUtility` and `getUtilities` @sneridagh
```
stevepiercy, line 1:
```suggestion
Added `Utilities` registry for `registerUtility`, `getUtility`, and `getUtilities`. @sneridagh
```

packages/types/news/6161.feature
```diff
@@ -0,0 +1,2 @@
+Added `errors` shape to the `BlockEditProps`.
+Added typings for Utilities registry @sneridagh
```
stevepiercy, line 2:
```suggestion
Added typings for `Utilities` registry. @sneridagh
```

stevepiercy (changes requested):
Looks pretty good. I am uncertain about one of my suggestions for the code sample.

stevepiercy (approved):

sneridagh:
@tiberiuichim I implemented your `blocksErrors` prop suggestion.

ichim-david (approved):
@sneridagh my only concern left is regarding the type validator as we've spoken if field shouldn't also be suffixed with ? to mark it as optional as both isNumber and startEventDateRangeValidator makes no direct use of the field
https://github.com/plone/volto/pull/6181/files#diff-0fb5051092a06c331a0607394901c0217dc85e7d2f3ce393fadab348e0832b56R298

sneridagh:
@ichim-david done.

packages/volto/src/components/manage/Form/Form.jsx
sneridagh: Done `blocksErrors` and `errors` props are passed down now.

sneridagh:
Merging! Thanks to all for the effort and for having this done!

PR #6186, opened 2024-07-23
improve url regex

mamico:
Improved the url regex to allow non-public urls, such as https://intranet/ or file://server/share, e.g., for intranet.


davisagli:
@mamico Makes sense. Needs a towncrier entry before it's ready to merge.

packages/volto/news/6186.feature
```diff
@@ -0,0 +1 @@
+Improved url regex to allow non-public urls, such as https://intranet/ or file://server/share, e.g., for intranet. @mamico
```
stevepiercy, line 1: We need to wrap URLs with backticks to make these inline literals, else linkcheckbroken attempts to visit the URL and fails.

```suggestion
Improved URL regex to allow non-public or intranet URLs, such as `https://intranet/` or `file://server/share`. @mamico
```

stevepiercy (changes requested):
Update news.

stevepiercy (approved):
@davisagli already approved in a comment in this PR, so I'm approving, too, so that it can be merged when CI is all green.

PR #6188, opened 2024-07-24
Improved i18n docs for customizations

### LINKED ISSUE #1846: Translations in customizations

pnicolli:
Fixes #1846

In addition to improving the customizations docs which I think might be clearer now, I would change the suggestion regarding overriding translations. I would in fact suggest to keep overridden translations in a dedicated file. The name is actually not important, but I would advide against cluttering the `src/config.js` file with translated messages and would suggest using a dedicated file. The name of the file is not important, but `src/messages.js` sounds like a sane default suggestion imho. Any thoughts?

----
📚 Documentation preview 📚: https://volto--6188.org.readthedocs.build/

### DIFF
--- a/packages/volto/news/6188.documentation
+++ b/packages/volto/news/6188.documentation
@@ -1,1 +1,1 @@
-Improved i18n docs for customizations @pnicolli
+Improved i18n docs regarding new translated messages not being picked up by the i18n translation mechanism when added in shadowed components. @pnicolli
--- a/packages/volto/news/6188.documentation
+++ b/packages/volto/news/6188.documentation
@@ -1,0 +1,1 @@
+Improved i18n docs for customizations @pnicolli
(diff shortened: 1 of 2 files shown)


packages/volto/news/6188.documentation
```diff
@@ -0,0 +1 @@
+Improved i18n docs for customizations @pnicolli
```
ichim-david, line 1:
```suggestion
Improved i18n docs regarding new translated messages not being picked up by the i18n translation mechanism when added in shadowed components. @pnicolli
```

ichim-david (approved):
@pnicolli always nice when we can improve the documentation, especially gotcha situations :)

PR #6189, opened 2024-07-24
`SchemaWidget` CSS and registration

sneridagh:
- Improve CSS for the `SchemaWidget` widget.
- Remove widget registration as `id`, move to `widget` type instead.

### DIFF
--- a/docs/source/upgrade-guide/index.md
+++ b/docs/source/upgrade-guide/index.md
@@ -376,6 +376,21 @@ It is unlikely that your code uses it, unless you heavily customized the Jest te
 The `react-share` library and `SocialSharing` component has not been used in the core since some time ago, and it is more suitable as an add-on and not in core.
 If you still use it, you can add it to your main add-on dependency, and extract the `SocialSharing` component from Volto 17 as a custom component in your add-on code.
 
+
+### `SchemaWidget` widget registration change
+
+Previously, it was registered as a widget `id`, but this definition could leak the widget and be applied to unwanted fields.
+Now it's registered as a `widget`, so if you are using it in your project/add-ons you should update the field definition and add the `widget` property.
+
+```ts
+// more form definition above...
+schema: {
+  title: 'Schema',
+  widget: 'schema'
+}
+// rest of the form definition...
+```
+
 (volto-upgrade-guide-17.x.x)=
 
 ## Upgrading to Volto 17.x.x
--- a/docs/source/upgrade-guide/index.md
+++ b/docs/source/upgrade-guide/index.md
@@ -379,7 +379,7 @@ If you still use it, you can add it to your main add-on dependency, and extract 
 
 ### `SchemaWidget` widget registration change
 
-Previously, it was registered as a widget `id`, but this definition could leak the widget and be applied to unwanted fields.
+Previously, it was registered as a widget `id` assigned to the `schema` key. Due to this common key name, this definition could leak the widget and be applied to unwanted fields.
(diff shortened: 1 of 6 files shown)


sneridagh:
@stevepiercy a bit of docs in this one.

packages/volto/news/6189.breaking
```diff
@@ -0,0 +1 @@
+Remove widget registration as `id`, move to `widget` type instead. @sneridagh
```
ichim-david, line 1:
```suggestion
Removed `SchemaWidget` widget registration as `id` widget mapping assigned to the `schema` key, moved to `widget` mapping registration with the same `schema` key.  @sneridagh
```

docs/source/upgrade-guide/index.md
```diff
@@ -376,6 +376,21 @@ It is unlikely that your code uses it, unless you heavily customized the Jest te
 The `react-share` library and `SocialSharing` component has not been used in the core since some time ago, and it is more suitable as an add-on and not in core.
 If you still use it, you can add it to your main add-on dependency, and extract the `SocialSharing` component from Volto 17 as a custom component in your add-on code.
 
+
+### `SchemaWidget` widget registration change
+
+Previously, it was registered as a widget `id`, but this definition could leak the widget and be applied to unwanted fields.
+Now it's registered as a `widget`, so if you are using it in your project/add-ons you should update the field definition and add the `widget` property.
```
ichim-david, line 383:
```suggestion
Previously, it was registered as a widget `id` assigned to the `schema` key. Due to this common key name, this definition could leak the widget and be applied to unwanted fields.
Now it's registered as a `widget`, so if you are using it in your project/add-ons you should update the field definition and add the `widget` property.
```

packages/volto/theme/themes/pastanaga/collections/form.overrides
```diff
@@ -123,16 +123,50 @@
...
+  background: transparent;
+  box-shadow: none;
+  transform: none;
 
   .item {
     cursor: pointer;
   }
 }
 
+.block.form {
```
ichim-david, line 140: @sneridagh @robgietema This pull requests mentions that it improves the css for the SchemaWidget
I see that it is assigned to ContentTypeSchema as such it should affect pages such as
https://volto.demo.plone.org/controlpanel/dexterity-types/Document/schema
However I don't see any .block.form.

Please give guidance on how to find the page or component that this css touches.

sneridagh: @ichim-david this will be useful in the case that the widget lives inside a block. Which is the next step that we are looking forward to.

image (1)

ichim-david: I am not a fan of the important on the segment > div backend and neither of the static height: 66px but I also don't have any content that would be affected by this style.
Perhaps the folks from Redturtle with products such as volto-form-block might be affected.
if that's the case then @pnicolli have a look at this css as well

ichim-david (approved):

packages/volto/theme/themes/pastanaga/collections/form.overrides
sneridagh: @ichim-david I guess that's because we have to overcome other of semanticUI definitions that already have an `!important`... @robgietema did these the other day, so I guess that's why. As you said, since they are for that use case, and it's touching semanticUI, I'd not bother much.

PR #6198, opened 2024-07-26
WIP: Self registration tests

JeffersonBledsoe:


JeffersonBledsoe:
@davisagli I'm trying to add a test to reflect the changes in https://github.com/plone/volto/pull/5935, but I'm getting the following error from the backend:

```json
{
  "error": {
    "errors": [
      {
        "field": "sendPasswordReset",
        "message": "Property 'Send a confirmation mail with a link to set the password' is not allowed."
      }
    ],
    "message": "Error in fields. Property 'Send a confirmation mail with a link to set the password' is not allowed.",
    "type": "WrongParameterError"
  }
}
```

This is with the following request:

`POST` `http://127.0.0.1:55001/plone/++api++/@users` `{"fullname":"Placeholder name","email":"test@example.com","sendPasswordReset":true}`. Toggling _Let users select their own passwords_ doesn't help. Any suggestions as to why this might be failing?

davisagli:
@JeffersonBledsoe Currently `sendPasswordReset` is only allowed in the REST API if the current user has permission to manage users. (https://github.com/plone/plone.restapi/blob/main/src/plone/restapi/services/users/add.py#L75)

I don't see a reason to restrict it in that way. It should be possible to request a password reset email regardless of whether the user is being added by an admin or self-registering. I'd accept a PR to plone.restapi to make that change.

PR #6221, opened 2024-07-31
Fix search block showing no option select in sort on property (#5055) (Backport 17.x.x)

ichim-david:


ichim-david:
> @ichim-david Thanks for preparing this backport

@davisagli yup you're welcomed, I was asking you if you wanted to backport also to 17 and then Victor said let's do it so I figured I would do my 10 min help part as well :)

I noticed that you've removed the extra translation, I guess we should clean up the message definition in the main branch since it's not used.
https://github.com/plone/volto/blob/4e5ecfde9f5c65a1b5f2d70dd2dd1bc7d5749f77/packages/volto/src/components/manage/Blocks/Search/components/SortOn.jsx#L25

davisagli:
@ichim-david Yeah -- I noticed the translation was unused after I merged the PR to main

PR #6227, opened 2024-08-02
Redirect to original URL when logging in with OIDC (volto-authomatic)

rstefko:


stevepiercy:
@rstefko see [Change log entry](https://6.docs.plone.org/contributing/index.html#change-log-entry).

davisagli (changes requested):
I think this should be handled in volto-authomatic's login component, instead of adding a parameter here that is only meaningful when volto-authomatic is installed.

ichim-david:
Closing this pull request given @davisagli comment, @rstefko thank you for your contribution however we have decided that your fix is best handled in your addon since Volto itself makes no use of next_url url parameter.

PR #6232, opened 2024-08-04
Revert rename of .gitkeep for consistency

### LINKED ISSUE #6225: Add towncrier files to ignore or rename `.keep`
From a release on July 29, 2024, we need to either rename all towncrier fragments that we want to ignore to `.keep` or add `.gitkeep` to towncrier's new ignore configuration.

Reference: https://github.com/twisted/towncrier/pull/622

@Tishasoumya-02 I know it's not your fault but the pull request isn't green
https://github.com/plone/volto/actions/runs/10196668962/job/28207886197?pr=6147
Due to
https://github.com/twisted/towncrier/blob/trunk/src/towncrier/_builder.py#L117

try to rename package/volto/news/.gitkeep to .keep as they recognize that as something to ignore.

_Originally posted by @ichim-david in https://github.com/plone/volto/issues/6147#issuecomment-2262823508_

stevepiercy:
Merge after the inclusion of the merged PR https://github.com/twisted/towncrier/pull/647, which closed https://github.com/twisted/towncrier/issues/643, in the next release of towncrier. 

Closes #6225.

### DIFF
--- a/packages/volto/news/6232.internal
+++ b/packages/volto/news/6232.internal
@@ -1,0 +1,1 @@
+Revert rename of `.gitkeep`. @stevepiercy


ichim-david:
I would argue this work isn't ready yet given the fact that there is no release yet for towncrier and as such the changelog run fails.

I also wonder if there shouldn't be a bit more info in the doc entry regarding the need for the rename to fix the towncrier which we can revert due to fix in the upstream project.

I also think this link should be in the issue description, https://github.com/twisted/towncrier/pull/647 it's your fix that we are waiting to be released.

davisagli:
@stevepiercy It looks like there's a release of towncrier with this change available now.

stevepiercy:
@davisagli I had the wrong link in the description. Updated. No release since 2024-07-31, so we have to wait a bit longer. I subscribed to Watch > Custom... > Releases so I will get a notification.

stevepiercy:
Towncrier 24.8.0 was released today, so we can merge the PR, after it passes checks.

stevepiercy:
Towncrier is happy now. Merging.

Issue #6171, opened 2024-07-18
Installing Component Library with css import


Nimo-19:
**Describe the bug**
Currently I am struggling to use a component library with a css import. 

I tried creating an addon and installing @adobe/react-spectrum
At first all worked fine, but after a `yarn install` in the base project I get the following error.

```
[...]/volto-test/node_modules/@react-spectrum/actiongroup/dist/vars.406b82f3.css:1
.Dniwja_i18nFontFamily {
^

SyntaxError: Unexpected token '.'
    at wrapSafe (node:internal/modules/cjs/loader:1281:20)
    at Module._compile (node:internal/modules/cjs/loader:1321:27)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1416:10)
    at Module.load (node:internal/modules/cjs/loader:1208:32)
    at Function.Module._load (node:internal/modules/cjs/loader:1024:12)
    at Module.require (node:internal/modules/cjs/loader:1233:19)
    at require (node:internal/modules/helpers:179:18)
    at Object.<anonymous> (/home/nimo/workspace/hrz/plone/frontend/volto-test/node_modules/@react-spectrum/actiongroup/dist/ActionGroup.main.js:1:1)
    at Module._compile (node:internal/modules/cjs/loader:1358:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1416:10)
sswp> !!! Script exited with code 1
```

**To Reproduce**

1. Bootstrap a new volto project `yo @plone/volto volto-test; cd volto-test;  ...`
2. Create a addon test-addon `yo @plone/volto:addon`
3. Add  @adobe/react-spectrum `yarn workspace test-addon add @adobe/react-spectrum`
4. Install with `yarn install`
5. In the addon create a simple Block like in the plone training with:
```js
import { ActionButton } from '@adobe/react-spectrum';

export default function TestView(props) {
  return (
    <div>
      <ActionButton>Hello World</ActionButton>
    </div>
  );
}
```
6. Start with `yarn start`
7. The aforementioned server side error

**Expected behavior**
The installed package should go through the bundler to be picked up by the css-loader

**Software (please complete the following information):**
```json

{
[...]
  "peerDependencies": {
    "@plone/volto": "^17.0.0"
  },
  "dependencies": {
    "@adobe/react-spectrum": "3.35.1"
  }
}

```

**Additional context**
I think the culprit is somewhere here:
[...]

PR #6238, opened 2024-08-06
Error using a Component Library fix

KUNDAN1334:
fix issue #6171

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,8 @@
     "lint-staged": "15.2.2",
     "tsconfig": "workspace:*",
     "typescript": "^5.4.5",
-    "vitest": "^1.5.0"
+    "vitest": "^1.5.0",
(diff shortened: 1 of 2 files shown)


stevepiercy:
@KUNDAN1334 please read and follow [First-time contributors](https://6.docs.plone.org/contributing/first-time.html), [Contributing to Plone](https://6.docs.plone.org/contributing/index.html), and [Contributing to Volto](https://6.docs.plone.org/contributing/volto.html).

ichim-david:
@KUNDAN1334 we are not using adobe/react-spectrum and we don't want to depend on it.
The issue signaled first needs to be confirmed and if it's valid then we might even propose a way to fix and only
then will any pull request be taken into consideration.
Please take the time to study the volto docs, work with the training and have experience with it before trying to contribute, you are not qualified as it stands to contribute to this project and will not be until you complete the required steps.

PR #6240, opened 2024-08-06
fix: handle form.ui.hovered

giuliaghisini:
fixed change of form.ui.hovered when editing blocks, because if you have a FormBlock component inside another one, 
onMouseOver fires for all stacked blocks and you cannot use the nested form. 

working example:

https://github.com/user-attachments/assets/b839a4d5-6459-491a-844c-a92dd5fb8780


stevepiercy:
@giuliaghisini did you mean to include files under the `api/` directory? I think that was a mistake.

stevepiercy:
@giuliaghisini btw, we recently updated documentation with the current best practice of using Docker on the backend. The `api` folder is outdated. See https://6.docs.plone.org/volto/contributing/developing-core.html#start-the-backend-and-volto

davisagli (approved):
The fix makes sense to me.

PR #6248, opened 2024-08-09
Allow enhancing the ContentsPropertiesModal schema

davisagli:
This has a couple improvements for the "properties" modal in the Contents view.

1. It makes it possible for an addon to add more properties with a schema enhancer, without needing to shadow the ContentsPropertiesModal component.
2. If all selected items have the same value for a field, the form is now prepopulated with that value.

### DIFF
--- a/packages/volto/news/6248.feature
+++ b/packages/volto/news/6248.feature
@@ -1,0 +1,2 @@
+The schema for the ContentsPropertiesModal can be enhanced using the `contentPropertiesSchemaEnhancer` setting.
+Also, the properties form is now prepopulated with values if all selected items share the same value. @davisagli
--- a/packages/volto/news/6248.feature
+++ b/packages/volto/news/6248.feature
@@ -1,2 +1,2 @@
-The schema for the ContentsPropertiesModal can be enhanced using the `contentPropertiesSchemaEnhancer` setting.
+The schema for the `ContentsPropertiesModal` can be enhanced using the `contentPropertiesSchemaEnhancer` setting.
 Also, the properties form is now prepopulated with values if all selected items share the same value. @davisagli
--- a/packages/volto/src/components/manage/Contents/Contents.jsx
+++ b/packages/volto/src/components/manage/Contents/Contents.jsx
@@ -1532,6 +1532,9 @@ class Contents extends Component {
                     onCancel={this.onPropertiesCancel}
                     onOk={this.onPropertiesOk}
                     items={this.state.selected}
+                    values={map(this.state.selected, (id) =>
+                      find(this.state.items, { '@id': id }),
+                    )}
                   />
                   {this.state.showWorkflow && (
                     <ContentsWorkflowModal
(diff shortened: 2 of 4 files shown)


packages/volto/news/6248.feature
```diff
@@ -0,0 +1,2 @@
+The schema for the ContentsPropertiesModal can be enhanced using the `contentPropertiesSchemaEnhancer` setting.
```
stevepiercy, line 1:
```suggestion
The schema for the `ContentsPropertiesModal` can be enhanced using the `contentPropertiesSchemaEnhancer` setting.
```

stevepiercy (changes requested):
Just some markdown on the news item.

stevepiercy (approved):
I only reviewed the news item, and it still needs technical review.

robgietema (approved):

ichim-david:
Will check in 30 min

ichim-david:
@pnicolli @erral I am checking this pull request from @davisagli which besides adding the schema enhancing support there is also this functionality where it prefills the options that are the same from the selected objects.
This means that if 2 objects have exclude from navigation set then it shows up in the content properties as checked.

One thing that might be a bit confusing for editors is the expiration date which right now is not visible as you can see from this screenshot
expiration-date

while with this branch it shows like this
expiration-set

This is technically the value set for the expiration but I doubt the editors are aware of this and how they would act if seeing this set if they would come to support and say but I didn't set an expiration date.

I would probably allow filtering, especially for the expiration date which could be seen as confusing.

Do you guys have any feelings about this or am I the only one feeling funky about it? :)

erral:
If the value for expiration date is not set it should not be visible in the form. Showing 2500-12-31 as a "default expiration date" is wrong in my opinion, in such case the widget should appear empty as it is right now.

ichim-david (changes requested):
See suggestion from @erral https://github.com/plone/volto/pull/6248#issuecomment-2288025503

davisagli:
@erral @ichim-david I'll take a look at the problem with the expiration date. I suspect this is a side effect of how an empty expiration date gets stored in the catalog. I need to figure out if we fix this in the indexer or the serializer or just do a workaround here.

davisagli:
Dexterity content returns this "ceiling date" in the `expires` method if the actual date set is None. (https://github.com/plone/plone.dexterity/blob/master/plone/dexterity/content.py#L611) I think we probably could change this to return None but it might have some unwanted side effects, like making it hard to sort by expiration date.

We should probably fix the serializer to return null for items that have this ceiling date in the catalog. But, then there will still be a problem if this change in volto is used with an old version of plone.restapi.

So I think I will also add a check here to ignore dates after 2499

davisagli:
@ichim-david @erral I added a check for the default effective & expires dates.

ichim-david (approved):
@davisagli I'm fine with this change.  Haven't retested but the fix looks fine too me at least at 10pm :)

PR #6249, opened 2024-08-12
Add missing for-attribute 

Gomez:
We had an a11y audit, these improvements were noted.


JeffersonBledsoe:
@plone/volto-accessibility Needs review

packages/components/news/6249.fix
```diff
@@ -0,0 +1 @@
+Add missing for-attribute to checkbox * label to input (a11y). - @gomez
```
stevepiercy, line 1:
```suggestion
Add missing `for` attribute to checkbox label and `aria-label` to search input to improve accessibility. @gomez
```

stevepiercy (changes requested):
Notes should describe all changes. Also added some MyST markup for inline literals.

stevepiercy:
The failing changelog check is a lie, and can be ignored. See https://github.com/plone/volto/pull/6232.

The failing tests, however, cannot be ignored and must pass. I restarted them in case it was flaky and temporary.

erral (approved):
It doesn't change the working of the widget, just adds a11y tags. Working as expected.

packages/volto/src/components/manage/Blocks/Search/components/SearchInput.jsx
```diff
@@ -22,15 +22,15 @@ const SearchInput = (props) => {
...
   return (
     <div className="search-input">
       <Input
         id={`${props.id}-searchtext`}
         value={searchText}
-        placeholder={
-          data.searchInputPrompt || intl.formatMessage(messages.search)
-        }
+        aria-label={label}
+        placeholder={label}
```
ichim-david, line 33: @Gomez there is no need to add the aria-label as well as it will end up duplicating the label of the placeholder and the screen reader will read both.

Although I didn't screen record your change we have a similar situation with the search input of the site.
Have a look at this video where I show how Android 14 talkback reads the input that has
<input type="text" aria-label="Search" name="SearchableText" value="" autocomplete="off" placeholder="Search Site" title="Search">

https://github.com/user-attachments/assets/37f86ff6-85c1-430f-8289-eea9534d0273

Here is how the same input is read by VoiceOver on a mac using Firefox, Edge and Safari

https://github.com/user-attachments/assets/38356ce6-9179-49ec-9408-d8fbd5fe0285

And here is how it sounds without the aria-label

https://github.com/user-attachments/assets/62689156-1841-4a20-a3a7-16a0fcb7e2fb
That sounds better not to have Search Search site which isn't even what shows up in
the Form rotary of VoiceOver

ichim-david (changes requested):
@Gomez have a look at my issue https://github.com/plone/volto/pull/6249/files#r1716735124

packages/volto/src/components/manage/Blocks/Search/components/SearchInput.jsx
ichim-david: > @Gomez there is no need to add the aria-label as well as it will end up duplicating the label of the placeholder and the screen reader will read both.
> 
> Although I didn't screen record your change we have a similar situation with the search input of the site. Have a look at this video where I show how Android 14 talkback reads the input that has
> 
>  talk.back.search.input.mp4 
> Here is how the same input is read by VoiceOver on a mac using Firefox, Edge and Safari
> 
>  input-with-aria-label-with-audio-optimized.mp4 
> And here is how it sounds without the aria-label
> 
> https://github.com/user-attachments/assets/62689156-1841-4a20-a3a7-16a0fcb7e2fb That sounds better not to have Search Search site which isn't even what shows up in the Form rotary of VoiceOver

I've made a demo with the search input from the search block. You can safely use only the placeholder and it will use that information. 

https://github.com/user-attachments/assets/0cd5b57b-3a5b-43bb-be11-b49fa30ef0cc

With your changes aria-label and placeholder has the same value. In theory for Firefox it doesn't repeat the placeholder and aria label if they are the same but it does if they are different.
Haven't tried with Talkback on Android will try some more tonight if I get the time but I remain -1 on the aria-label insertion

Gomez:
Thanks @ichim-david! This is a useful comment, i removed the duplication.  

Changelog updated with @stevepiercy comments.

stevepiercy (approved):
Changelog is good. @ichim-david will review upon his return, or someone else can do a technical review.

ichim-david:
@Gomez I'll work on this tomorrow as part of the Salamina sprint.

ichim-david (approved):
This can be merged now. The fix was good although I had to improve it due to how the checkbox component works 
https://github.com/Semantic-Org/Semantic-UI-React/blob/master/src/modules/Checkbox/Checkbox.js#L201
By passing the id the label gets the htmlFor from the component without us needing to add it manually

PR #6242, opened 2024-08-07
Complete upgrade to Cypress 13.13.2

stevepiercy:
See https://github.com/plone/volto/pull/6241

PR #6250, opened 2024-08-12
fix: disable save button upon submitting

sabrina-bongiovanni:
When submitting the Add form upon creating a new Content Type, if the connection is too slow, the button will start loading and the spinner will appear, but it will remain clickable, thus allowing to submit the form and trigger the query multiple times, which in turn results in creating more instances of the same CT.

Adding the disabled prop to the button when the query is still loading prevents this issue from happening.

### DIFF
--- a/cypress/tests/core/blocks/listing/blocks-listing-templates.js
+++ b/cypress/tests/core/blocks/listing/blocks-listing-templates.js
@@ -37,7 +37,7 @@ describe('Folder Contents Tests', () => {
     cy.getSlate().click();
     cy.get('button.block-add-button').click();
     cy.get(
-      '[style="transition: opacity 0.1s 0.2s;"] > :nth-child(2) > .ui',
+      '[style="transition: opacity 500ms ease 0ms;"] > :nth-child(2) > .ui',
     ).click();
     cy.get('#field-variation').click().type('summary{enter}');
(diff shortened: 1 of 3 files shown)


sabrina-bongiovanni:
It seems there are some broken acceptance tests. I tried to run them locally and it appears that the issues where the tests break are not related to the code I changed in this PR.

I tried to run the test locally and it seems that the only wrong thing with that test is that the breaking line should be:

`cy.get(
      '[style="transition: opacity 500ms;"] > :nth-child(2) > .ui',
    ).click();`

as the style tag actually only has that attribute.
However, I've seen all other PR not having any issues with that, so I don't know whether I should fix that and it's something limited to Volto 17 or if I should just leave it as it is.

ichim-david:
> @sabrina-bongiovanni some tests are known to be flaky. I triggered a rerun of failed jobs in CI. Let's see if the 5 failed ones pass this time.

@stevepiercy  Remember that you changed the upload action to v4 within https://github.com/plone/volto/pull/6242/files but that is only for main. That change should be backported also to 17

ichim-david (approved):

stevepiercy:
@sabrina-bongiovanni @ichim-david PR to bump actions/upload-artifact created in https://github.com/plone/volto/pull/6253.

PR #6258, opened 2024-08-16
Fix edge case in ContentsPropertiesModal

davisagli:
The item can be undefined if it's in this.state.selected but not this.state.items. This can happen after navigating to a different folder.

### DIFF
--- a/packages/volto/news/6258.bugfix
+++ b/packages/volto/news/6258.bugfix
@@ -1,0 +1,1 @@
+Fix `TypeError: values[0] is undefined` in Contents properties modal. @davisagli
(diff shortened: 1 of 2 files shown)


ichim-david:
@davisagli of course you found a bug after backporting to 16 and 17 :)). Talking about bad karma alert :). Maybe a cypress test wouldn't hurt for this functionality

davisagli:
@ichim-david Yes, I found the bug once I was testing it integrated with a project. I'm not sure a cypress test would have helped in this case, because it's a bit of an obscure one that I probably wouldn't have thought to test.

Are there any existing Cypress tests for the contents view? I looked but didn't find any. We should absolutely have some, but that's a bigger project than I anticipated when I started working on this change to the properties modal.

ichim-david (approved):

PR #6269, opened 2024-08-23
Move Tags component to the slot belowContent

wesleybl:
Ref: https://github.com/plone/volto/pull/4381#issuecomment-2276168262


wesleybl:
I removed showTags, but I don't know if it's the best option. It might make it easier to remove the tags.

CC @davisagli

stevepiercy (changes requested):
Add a little MyST markup to the news.

Otherwise this looks OK to me, but one more person should give a technical review.

davisagli (commented):
I think this makes sense for Volto 18. Removing the showTags setting and changing the parameters expected by the Tags component are breaking changes that can't be backported.

erral (changes requested):
Here you are changing the internal contract or the Tags component, which needs proper Upgrade Notes.

If it isn't completely necessary I wouldn't change the component itself, I would just do the slots thing.

davisagli:
> If it isn't completely necessary I wouldn't change the component itself, I would just do the slots thing.

@erral The change is necessary in order to accept the parameters that are passed to slot components. (They get `content`, not `tags`.)

wesleybl:
@erral @davisagli I changed the news to breaking and put a note in the update guide.

erral (approved):

davisagli:
@wesleybl On second thought, I think we should keep the `showTags` setting. It's very easy to check it (inside the Tags component) and makes it easier for someone who is upgrading a site that disabled the tags (so they don't need to figure out how to override the slot component).

wesleybl:
@davisagli I brought the `showTags` setting back.

packages/volto/news/6269.breaking
```diff
@@ -0,0 +1 @@
+Move Tags component to the slot `belowContent`. @wesleybl
```
davisagli, line 1:
```suggestion
Move `Tags` component to the slot `belowContent`. @wesleybl
```

davisagli (approved):

PR #6276, opened 2024-08-30
missing params fix in getQuerystringSearchQuery

### LINKED ISSUE #6266: @plone/client: getQuerystringSearchQuery discards all parameters except for the query
https://6.docs.plone.org/plone.restapi/docs/source/endpoints/querystringsearch.html and https://6.docs.plone.org/volto/client/actions/querystringsearch.html both document being able to use parameters such as `b_size` and `b_start` to adjust the query. However, the JS client only allows you to do this with the POST request, the [current query string GET code](https://github.com/plone/volto/blob/16b61471d33f84eb9b48f79bc66186bbcc2b01b1/packages/client/src/restapi/querystring-search/get.ts) only passes the query parameter onto the API request

Faakhir30:
Fixes #6266 

getQuerystringSearchQuery is discarding all params except `query` param, hopefully, I've fixed it in this repo.

However, I see in https://github.com/plone/plone.restapi/blob/main/src/plone/restapi/services/querystringsearch/get.py#L99 backend service is also only processing the `query` param, will create a separate PR in `plone/plone.restapi` to fix there as well once status of this PR is confirmed.


Faakhir30:
@JeffersonBledsoe  kindly you could confirm if these are the desired changes I'm heading in the right direction, thanks.

stevepiercy:
@Faakhir30 please update your branch with `main` and push to this pull request. See https://6.docs.plone.org/contributing/first-time.html#update-your-pull-request-from-your-fork

JeffersonBledsoe:
Hi @Faakhir30 . Thanks for looking into this!
If the REST API is also not expecting anything other than the `query` parameter,  then it could be that this is actually the intended behaviour and the documentation may need an update. I'll bring this up at the next Volto team meeting and report back here with any decisions.

JeffersonBledsoe:
> @Faakhir30 @JeffersonBledsoe When calling the /@querystring-search endpoint with the GET method, it expects `query` as a serialized JSON object which can contain multiple keys including `query`, `b_start`, `b_size`, etc.

Thanks for the clarification! I'd skimmed over the docs when re-visiting this and hadn't noticed that both the searchParam and the inner key are both referred to as `query` in the GET request.

Faakhir30:
> @Faakhir30 @JeffersonBledsoe When calling the /@querystring-search endpoint with the GET method, it expects `query` as a serialized JSON object which can contain multiple keys including `query`, `b_start`, `b_size`, etc.

yeah,  thanks for the clarification, I'm closing this PR as this isn't required then.

PR #6278, opened 2024-09-04
Remove "Generate tabs for items other than folders" field from navigation control panel

wesleybl:
The help for this field says:

`By default, any content item in the root of the portal will appear as a tab. If you turn this option off, only folders will be shown. This only has an effect if 'automatically generate tabs' is enabled.`

This doesn't make sense in Volto. Volto doesn't display item as tab.

### DIFF
--- a/packages/volto/news/6278.bugfix
+++ b/packages/volto/news/6278.bugfix
@@ -1,0 +1,1 @@
+Remove `Generate tabs for items other than folders` field from navigation control panel. @wesleybl
--- a/packages/volto/src/config/ControlPanels.js
+++ b/packages/volto/src/config/ControlPanels.js
@@ -90,6 +90,7 @@ export const unwantedControlPanelsFields = {
   navigation: [
     'generate_tabs',
     'navigation_depth',
+    'nonfolderish_tabs',
     'sort_tabs_on',
     'sort_tabs_reversed',
     'sitemap_depth',
--- a/packages/volto/src/config/config.test.js
+++ b/packages/volto/src/config/config.test.js
@@ -221,6 +221,7 @@ test('test navigation controlpanel fields', () => {
   const not_in_navigation = [
     'generate_tabs',
     'navigation_depth',
+    'nonfolderish_tabs',
     'sort_tabs_on',
     'sort_tabs_reversed',
     'sitemap_depth',


davisagli (approved):

fredvd:
@davisagli  @sneridagh 

I'm checking the release notes for Volto 18a43 and found this issue.  The change still might be good, but the motivation in the issue itself is strange.  With 'tabs'  the visual representation of the main navigation section is meant. This used to be a tab, but with Plone 5.2 we already got nested navigation.

These settings can still make sense both in Volto and Classic UI.   just replace 'tab' with 'main section'.  You could disable the automatic population of main section items based on the contents of the root folder.   (if you disable it, use actions to manually fill  the main section).  

And in case you do automatic population, you could still choose to ony show Folders, or in Volto's default case Folderish pages, or allow any item of a content type to be in the main section. 

So why was this option removed from the Volto control panel?

wesleybl:
> These settings can still make sense both in Volto and Classic UI. just replace 'tab' with 'main section'.

@fredvd what exactly are you calling the "main section" in Volto? Is it a navigation menu item? After your question, I did a test, and I saw that when this option is selected, the navigation menu also displays non-Folderish items, such as Link. So, it might make sense to have this option in Volto. But the term "tab" is really strange.

fredvd:
The term tabs is strange indeed.  but is just a help/field label text that hasn't been updated for the Volto frontend and Plone 5.2 dropdownmenu's.    with main section is the main 'horizontal' navigation section we have in a plain Plone site.

wesleybl:
I think it would be better:

`Generate navigation items for contents other than folders`

davisagli:
That still doesn't make much sense for Volto because Volto doesn't have folders, and there's a separate setting for which content types to show in navigation.

fredvd:
> And in case you do automatic population, you could still choose to ony show Folders, or in Volto's default case Folderish pages

The displayed_types you refer to not only work for the main section navigation, but also for the contextnavigation component and I suspect the site-index.   It's a different setting with different purpose.

wesleybl:
> and there's a separate setting for which content types to show in navigation.

If the content is not Folderish, such as Link, and this option is not selected, the item will not appear in the navigation menu, even if the Link type is selected to be displayed in the navigation menu.

wesleybl:
> That still doesn't make much sense for Volto because Volto doesn't have folders

It could be:

Generate navigation items for contents other than Folderish

wesleybl:
> if you disable it, use actions to manually fill the main section

@fredvd  in Plone Classic, is it possible to manually add items to the navigation menu? How to do this? Will this work on Volto?

PR #6280, opened 2024-09-08
Empty GridBlocks will not be rendered in 'View'

### LINKED ISSUE #6279: Saving empty 'gridBlock' breaks volto
**Describe the bug**
While editing if you add an empty gridBlock and click on 'save' it will break the site.

**To Reproduce**

1. Go to 'Edit mode'
2. Add Grid Block
3. Click on Save

**Expected behavior**

**Screenshots**

MAX-786:
Fixes #6279 

This can be a single line fix:
 https://github.com/plone/volto/blob/7690677589ad881d794a9484d9a8ee1aa005fbcf/packages/volto/src/components/manage/Blocks/Grid/View.jsx#L10

adding this check just before the above mentioned line will fix this
```js
  if (data.blocks_layout === undefined) {
    return null;
  }
  ```
  
then this won't render any Grid block since its empty so it makes sense too, but yeah it remains in the formData so going into edit again will open up the TemplateChooser again.

### DIFF
--- a/packages/volto/news/6279.bugfix
+++ b/packages/volto/news/6279.bugfix
@@ -1,0 +1,1 @@
+Empty GridBlocks will not be rendered in 'View' @MAX-786
--- a/packages/volto/news/6279.bugfix
+++ b/packages/volto/news/6279.bugfix
@@ -1,1 +1,1 @@
-Empty GridBlocks will not be rendered in 'View' @MAX-786
+Fix error rendering empty grid blocks. @MAX-786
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: packages/volto/news/6279.bugfix [hunk 1]
```diff
@@ -0,0 +1 @@
+Empty GridBlocks will not be rendered in 'View' @MAX-786
```
davisagli, line 1:
```suggestion
Fix error rendering empty grid blocks. @MAX-786
```

davisagli (approved):
Thanks for this fix!

Issue #6289, opened 2024-09-13
Update cookie* references to current Volto 18.0.0-alpha.43 release


stevepiercy:
With the release of Volto 18.0.0-alpha.43, we need to review and potentially update or supplement references to `cookiecutter-plone-starter`, `cookiecutter-zope-instance`, and 
[`plone/generator-volto`](https://github.com/search?q=repo%3Aplone%2Fvolto%20plone%2Fgenerator-volto&type=code) with [Cookieplone](https://github.com/plone/cookieplone).

- [ ] https://6.docs.plone.org/volto/contributing/developing-core.html

See also:
- https://github.com/plone/documentation/issues/1706
- https://github.com/plone/documentation/issues/1705
- https://github.com/plone/documentation/pull/1714

davisagli:
Cookieplone is designed for use with Volto 18.

For earlier Volto, it's still recommended to use plone/generator-volto (if you need just a frontend) or cookiecutter-plone-starter (it also uses plone/generator-volto, but generates a full project with frontend and backend).

cookiecutter-zope-instance is still relevant; it is used by both cookiecutter-plone-starter and cookieplone to create a Zope instance to run the backend.

stevepiercy:
@davisagli I updated the description and title of this issue with your feedback. Thank you!

stevepiercy:
There is only one file to manage in this issue. @sneridagh @davisagli should we update it to use Cookieplone or replace its entire content with a landing page with a link or redirect to the new [Create a project with Volto (development or pre-release)](https://plone6--1714.org.readthedocs.build/install/create-project-cookieplone.html) in the PR https://github.com/plone/documentation/pull/1714?

I am in favor of maintaining a single version of a file and adding a landing page and redirect. We've done this with contributing docs for volto, plone.api, and plone.restapi.

davisagli:
@stevepiercy Cookieplone isn't an appropriate tool for developing Volto core. It generates the scaffold for a new repository, but when you're developing Volto you don't need or want a new repository.

Most of this page should remain as is. There are a couple references to cookie* that need minor updates; I'll prepare a PR.

PR #6290, opened 2024-09-14
Upgrade Guide Cookieplone review

stevepiercy:
I updated the Upgrade Guide Cookieplone section. This also brings up new issues that must be resolved before the final Volto 18 release.

See also:

- https://github.com/plone/documentation/issues/1706
- https://github.com/plone/volto/issues/6289
- https://github.com/plone/documentation/pull/1707

----
📚 Documentation preview 📚: https://volto--6290.org.readthedocs.build/

### DIFF
--- a/docs/source/upgrade-guide/index.md
+++ b/docs/source/upgrade-guide/index.md
@@ -450,19 +450,28 @@ schema: {
 The `Tags` component has been moved to the `belowContent` slot.
 It now receives the `content` property instead of the `tags` property.
 
-### CookiePlone is now the recommended project and add-on generator for Volto 18
 
-```{versionadded} 18.0.0-alpha.43
+{upgrade-18-cookieplone-label}=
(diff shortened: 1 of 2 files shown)


sneridagh (approved):
LGTM, although I dislike the "Sponsored OSS Docker images" naming.
For me it's the recommended community published Docker Image, the fact that it's right now in Docker Hub is circumstantial and could change in the future, eg. hosted in GH.

stevepiercy:
Yeah, I'm not thrilled about the naming, either, but we have an "Official" Plone image (which IMO should be deprecated to avoid confusion for Plone 6) https://hub.docker.com/_/plone.

I forgot to ask, in the Upgrade Guide, we mention that developers need to adapt their projects to the new way of doing things, but we don't tell them how to do that. I think that should also go into the Upgrade Guide.

PR #6271, opened 2024-08-26
Passes the `user` to the `restricted` function of the block settings

wesleybl:
The `restricted` function is called inside a `useMemo`. So we can't call `useSelector` on it. To work around this, we get the user before `useMemo` and pass it to the `restricted` function.

fixes #6264

PR #6293, opened 2024-09-16
Fetch user before pass it to the restricted function

wesleybl:
It may not be in the Redux store yet when we try to pass it in. So we dispatch if it's not in the store.


ichim-david (approved):

packages/volto/src/hooks/user/useFetchUser.js
```diff
@@ -0,0 +1,22 @@
...
+  useEffect(() => {
+    if (!user?.id) {
+      dispatch(getUser(userId));
+    }
+  }, [dispatch, userId, user]);
+
+  return user;
+};
+
+export default useFetchUser;
```
davisagli, line 22: I'd be inclined to name it `useUser`. The point of refactoring this into a hook is that the code using the hook doesn't really need to be aware of how the user data is found.

packages/volto/src/hooks/user/useFetchUser.js
```diff
@@ -0,0 +1,22 @@
...
+  const userId = useSelector((state) =>
+    state.userSession.token ? jwtDecode(state.userSession.token).sub : '',
+  );
+  const dispatch = useDispatch();
+
+  useEffect(() => {
+    if (!user?.id) {
+      dispatch(getUser(userId));
+    }
+  }, [dispatch, userId, user]);
```
davisagli, line 17: If there are multiple components which use this hook, they might dispatch multiple copies of the getUser action at the same time.

Maybe there's a way to instead make sure that the user data is always fetched immediately after the user logs in.

packages/volto/src/hooks/user/useFetchUser.js
wesleybl: Done

packages/volto/src/hooks/user/useFetchUser.js
wesleybl: Done

packages/volto/src/components/theme/Login/Login.jsx
```diff
@@ -85,6 +86,9 @@ const Login = (props) => {
 
   const previousToken = usePrevious(token);
 
+  // Get user and place their data in the Redux store
+  useUser();
```
davisagli, line 90: This was my suggestion, but now I'm realizing it wasn't the best idea.

For one thing, this runs every time the login form renders, including before the user successfully logs in.

More importantly, it doesn't populate the redux store if the user reloads the page while already logged in.

I'll comment below with what I think we actually need to avoid sending multiple requests for the user at once.

packages/volto/src/hooks/user/useUser.js
```diff
@@ -0,0 +1,22 @@
...
+
+const useFetchUser = () => {
+  const user = useSelector((state) => state.users?.user);
+  const userId = useSelector((state) =>
+    state.userSession.token ? jwtDecode(state.userSession.token).sub : '',
+  );
+  const dispatch = useDispatch();
+
+  useEffect(() => {
+    if (!user?.id && userId) {
```
davisagli, line 14: Rather than checking user?.id, I think we should check state.users?.get which should exist as soon as getUser is dispatched the first time (including while the request is in progress)

davisagli (changes requested):

packages/volto/src/hooks/user/useUser.js
wesleybl: `state.users?.get` exists even when `user?.id` does not exist.

It looks like this:

```javascript
{
    "error": null,
    "loaded": false,
    "loading": false
}
```
I think this if would work:

```javascript
if (!user?.id && users?.get.loading === false) {
```

wesleybl: Done

davisagli (approved):
Thanks!

pnicolli:
I am not sure I understand what problem we are trying to solve here. Aren't the components getting an update (rerender triggered by useSelector) when the user is finally fetched, even if it's not there when these components are first rendered? In other words, does it break somehow, or what is the issue here?
I'm trying to understand if we are firing an extra query to the backend that is worth firing or if it can be avoided. Moreover, would it be possible, if we have a use case for fetching user data here, to just do it during SSR instead? This way we could potentially avoid any issue from the beginning.

davisagli:
@pnicolli Consider the case where the user is logged out, then they log in and edit the page. Then we have to dispatch getUser, because the user isn't in the store yet. In this scenario there is no SSR request at a point in time when the user is logged in.

sneridagh:
@pnicolli this is related to https://github.com/plone/volto/pull/6271, as a continuation, if the user info is not in the store, this PR makes sure it's in there, so we can use it in a `restrict` key in the block config.

It LGTM.

sneridagh (approved):

pnicolli:
> @pnicolli Consider the case where the user is logged out, then they log in and edit the page. Then we have to dispatch getUser, because the user isn't in the store yet. In this scenario there is no SSR request at a point in time when the user is logged in.

I thought the user should get into the store eventually when the user logs into the portal, but I trust you to have done more research than I did on the subject 😁  👍

sneridagh:
@pnicolli not the case, but it's a good point. @davisagli thoughts?

pnicolli:
Well I was thinking... maybe we should consider for Volto 19 to move the user info to an expander we get with the content? Would it be worth it? I don't want to clutter anybody's notifications though, so maybe I could give this a little more thought and discuss it at a Volto Team meeting

davisagli:
I also had the idea of fetching the user info when the user logs in, but it doesn't avoid the need for what we have in this PR, because sometimes the user loads the page from the server while already logged in.

I don't think user info belongs in an expander because it's information that does not depend on which content item is being fetched. It would also make it hard to cache content responses in a way that is shared between multiple users with the same roles.

PR #6294, opened 2024-09-17
fix: state and styles for toolbar handler for hp for volto17

sabrina-bongiovanni:
Porting of PR for Volto 18 to Volto 17. 
This is what the PR for Volto 18 entailed the following:

Added a new class for the toolbar handler in the homepage based on the path.

The homepage is the only page with path ''.
The homepage is always public, so it needs a style regardless of the review state of the page
Since it's always public, I used the @teal-blue color used for public pages

### DIFF
--- a/news/6294.bugfix
+++ b/news/6294.bugfix
@@ -1,1 +1,1 @@
-Fix the toolbar handler color for the homepage to match its "published" state for Volto17. @sabrina-bongiovanni
+Fix the toolbar handler color for the homepage to match its "published" state. @sabrina-bongiovanni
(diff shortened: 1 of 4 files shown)


pnicolli (approved):

news/6294.bugfix
```diff
@@ -0,0 +1 @@
+Fix the toolbar handler color for the homepage to match its "published" state for Volto17. @sabrina-bongiovanni
```
pnicolli, line 1:
```suggestion
Fix the toolbar handler color for the homepage to match its "published" state. @sabrina-bongiovanni
```

pnicolli (approved):

PR #6295, opened 2024-09-17
Join validation errors in one single toast and update api error resp

cekk:
With this changes error messages (and labels on single fields) are updated each time a POST/PATCH return a validation error (before this, error messages in fields weren't updated).

I've also joined error messages in one single toast because i've found that raising multiple toasts, only last will be shown.
For me it's also more readable a single toast with all the informations, but this can be reverted if you don't like it (i had to change also a wrapper into Toast component to allow insert an ul tag).

This is an example of the result:

Screenshot 2024-09-24 alle 15 20 58

### DIFF
--- a/packages/volto/news/6295.bugfix
+++ b/packages/volto/news/6295.bugfix
@@ -1,0 +1,2 @@
+- Join validation errors in one single toast and update errors from response. @cekk
+- Toast content now has a <div> wrapper instead of a <p>. @cekk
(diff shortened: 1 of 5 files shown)


davisagli (approved):
Minor: the list items look like they have too much indent on the left.

Otherwise LGTM.

giuliaghisini (approved):

pnicolli (approved):

PR #6299, opened 2024-09-20
fix: find blocks if fieldname is similiar to 'blocks'

giuliaghisini:
Fixed findBlocks because if we have a block with a property inside which name ends with 'blocks' the findBlocks function is called assuming that field contains something similar to volto's block, but it couldn't be. 

For example, if we have some blocks with the property 'subblocks' inside that is simply an array of object (not volto's blocks) an error is thrown


giuliaghisini:
@sneridagh @ichim-david do you think we could merge this one during Salamina sprint?

ichim-david (approved):
@giuliaghisini I'm ok with the fix and merging for 18. Victor next :)

PR #6309, opened 2024-09-24
improve DiffField.jsx with better support for displaying HTML elements such as images

dobri1408:
The HistoryDiff requires a redesign, as the current layout is broken on complex pages and isn't functioning as expected. My plan is to avoid treating the HTML as a simple string and layering classes on top, which could cause issues. Instead, I'll take a more structured approach by properly parsing the HTML.

Before:
Captură de ecran din 2024-09-24 la 16 45 40

After:
Captură de ecran din 2024-09-24 la 16 47 08


ichim-david:
@dobri1408 having the .snap test files changed means this is a breaking change to the code. I wonder which of your logic makes it so that the content is without the span tag.

I would also like to see a before and after of the actual volto stock demo instead of the EEA website diffing in action.

dobri1408:
I made a change by removing the DefaultView for the RenderBlocks (https://github.com/plone/volto/pull/6309/files#diff-1d656037a4db345e1154459194c480148b2b6a85acf74691f76a113f5df89199L81). The reason for this is that the DefaultView also renders elements like the footer and navigation bar, which is not ideal. We want to focus solely on the differences in the blocks. This is why one span tag is removed.         

Here are some screenshots :  

Before:
Captură de ecran din 2024-09-25 la 18 02 02


After
Captură de ecran din 2024-09-25 la 18 02 58

dobri1408:
@ichim-david The coloring is controlled by the CSS style found here: https://github.com/plone/volto/blob/a2d997b3e7a9746970f97c5cd7bec30a08a46daf/packages/volto/theme/themes/pastanaga/extras/main.less#L250. This style applies only to elements where the background can be modified, such as text, so it doesn't affect images. However, I don't believe it's necessary to extend it to images since changes to them are visually apparent. The coloring is mainly intended for text. That said, if you think adding something like a border for images would be better, I can implement that. Let me know what you prefer.

sneridagh (approved):
@dobri1408 @ichim-david I tried it out, and it works quite well!! I just tried a couple of things, using teasers etc, but LGTM.

@davisagli @ichim-david if you give your final blessing, we can merge.

packages/volto/news/6309.bugfix
```diff
@@ -0,0 +1 @@
+(fix+feat): improve DiffField.jsx to work with complex pages @dobri1408
```
ichim-david, line 1:
```suggestion
Improve DiffField.jsx to render only the blocks on the page with better support for displaying HTML elements such as images. @dobri1408
```

ichim-david (approved):
@sneridagh I am fine with these changes, it's what we had in our EEA projects for quite some time and we haven't received any bug reports as of late. 
@dobri1408 I think this should be a .feature instead of bugfix since it provides enhancements to the diff functionality.

PR #6310, opened 2024-09-24
Core js splitting - second part

giuliaghisini:


packages/volto/src/components/theme/Widgets/DatetimeWidget.test.js
```diff
@@ -6,6 +6,18 @@ import configureStore from 'redux-mock-store';
...
+  async () =>
+    await require('@plone/volto/helpers/Loadable/Loadable').__setLoadables(),
+);
+// jest.mock('./DatetimeWidget', () =>
+//   jest.fn(() => (
+//     <span className="metadata datetime widget">
+//       Tuesday, August 4, 2020 9:00 AM
+//     </span>
+//   )),
+// );
```
pnicolli, line 20: Cleanup

pnicolli (changes requested):

giuliaghisini:
@ichim-david @sneridagh pr is ready for review. 
Unfortunatly a cypress test for 'locking' fails, but i don't understand why... no code related to locking is touched by this pr.. do you have some idea?

ichim-david:
@giuliaghisini maybe it's a flaky test that passes after a re run or increase of wait timer. I will have a look at this next week when I'm back to work, right now I'm waiting for the flight delay to pass :)
Thank you for the work so far, I cannot wait to test it :)

pnicolli:
@ichim-david yeah I was looking at this right now and I don't think there will be time to polish this before thursday. The fact that most acceptance runs break makes me think something needs to be fixed and it's not just flakiness

davisagli:
There was a test stability problem that was never resolved, and by now this has a lot of conflicts. We can revisit this if someone is willing to bring it up to date.

PR #6313, opened 2024-09-24
Add outline to toolbar buttons

### LINKED ISSUE #6311: Toolbar buttons don't provide a focus outline [SC 2.4.7]
https://www.w3.org/TR/WCAG22/#focus-visible

Currently we don't have any focus indicator for the buttons in the toolbar. This is confusing when using a keyboard to navigate the site and is a failure of the WCAG 2.X success criterion noted in the title

JeffersonBledsoe:
Fixes #6311 

## Current progress

Screenshot 2024-09-24 at 16 56 59

Screenshot 2024-09-24 at 16 59 54

### DIFF
--- a/packages/volto/news/6313.bugfix
+++ b/packages/volto/news/6313.bugfix
@@ -1,0 +1,1 @@
+Fixed toolbar buttons not having a focus outline. @JeffersonBledsoe
--- a/packages/volto/theme/themes/pastanaga/extras/toolbar.less
+++ b/packages/volto/theme/themes/pastanaga/extras/toolbar.less
@@ -291,10 +291,11 @@ body:not(.has-sidebar):not(.has-sidebar-collapsed) {
       background: transparent;
       cursor: pointer;
       text-align: initial;
+      box-shadow: unset !important; // Some buttons have double outlines due to default styles. Safely disable those styles here.
 
       svg {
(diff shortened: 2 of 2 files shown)


JeffersonBledsoe:
@ichim-david Checks all green now. Not sure if this needs design input from anybody else but it's ready for merging from my perspective

giuliaghisini (approved):

Wagner3UB (approved):

pnicolli:
A behavior that I have recently seen in good component libraries is to only add the outline when navigating with the keyboard, while avoiding it when clicking things, because that's redundant. I haven't looked into how to do that, but I think it would be a nice improvement here.

ichim-david:
> A behavior that I have recently seen in good component libraries is to only add the outline when navigating with the keyboard, while avoiding it when clicking things, because that's redundant. I haven't looked into how to do that, but I think it would be a nice improvement here.

@JeffersonBledsoe this is why we had focus-visible selector instead of simply focus to ensure the outline shows up when using the keyboard navigation.  I would set it back as well to avoid outlines when mouse clicking

sneridagh (approved):
LGTM if someone promises me no outline will show when clicking 😂

pnicolli (approved):

PR #6315, opened 2024-09-24
fix(SideBarPopup): close sidebarPopup on ESC keypress

nileshgulia1:
As an extension of https://github.com/plone/volto/pull/2509#issuecomment-1016368522
The problem with block losing focus on ESC was originating from https://github.com/plone/volto/blob/355bbad4e61420c4ede481a1bcae21a3ed82148e/packages/volto/src/components/manage/Blocks/Block/BlocksForm.jsx#L77
The solution was to intercept the event by making `<aside>` focusable. Since non-interactive elements should not be made focusable, therefore added tabIndex={-1}.

### DIFF
--- a/packages/volto/src/components/manage/Sidebar/SidebarPopup.jsx
+++ b/packages/volto/src/components/manage/Sidebar/SidebarPopup.jsx
@@ -16,6 +16,13 @@ const SidebarPopup = (props) => {
     onClose();
   };
 
+  const handleEscapeKey = (e) => {
+    if (open && e.key === 'Escape') {
+      onClose();
+      e.stopPropagation();
+    }
+  };
+
   const [isClient, setIsClient] = React.useState(false);
   React.useEffect(() => {
     setIsClient(true);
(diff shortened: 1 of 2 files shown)


nileshgulia1:
Edit: I think introducing focus was a smell, I modified using intercepting even conditionally.

sneridagh:
@nileshgulia1 this has to be merged along with the other?
From the comments in the other PR I understand that it's solved already and no longer reproduceable.

sneridagh (approved):

packages/volto/src/components/manage/Sidebar/SidebarPopup.jsx
```diff
@@ -16,18 +16,26 @@ const SidebarPopup = (props) => {
...
+  };
+
   const [isClient, setIsClient] = React.useState(false);
   React.useEffect(() => {
     setIsClient(true);
   }, []);
 
   React.useEffect(() => {
     document.addEventListener('mousedown', handleClickOutside, false);
+    document.addEventListener('keyup', handleEscapeKey, false);
```
sneridagh, line 33: Just a comment why adding the `keyup`?

sneridagh: @nileshgulia1 ^^

nileshgulia1: @sneridagh I had to cut down the propagation to https://github.com/plone/volto/blob/1b8328cc0029f092525f6791d5458665b5008a4b/packages/volto/src/helpers/Utils/useDetectClickOutside.js#L52 that removes the focus on selected block.

PR #6317, opened 2024-09-24
feat!: new implementation for the proxy middleware. (express-http-middleware)

mamico:
This is an alternative implementation of #5299 which uses `express-http-middleware`. Compared to the other implementation, this one tries to implement the task ‘handle errors (not found, not authorised, ...) correctly’.

It is a PoC in which only the 404 error is implemented, and there is some code to correct (e.g. duplicate code).


wesleybl:
@mamico Is the idea of ​​the proxy to handle private content? I wonder if the request from anonymous users could not go directly to Plone.

mamico:
> @mamico Is the idea of ​​the proxy to handle private content? I wonder if the request from anonymous users could not go directly to Plone.

Not really. The file/image proxy implementation has been there on Volto from the beggining, this fix and its twin #5299, where you can find more details, is intended to improve performance in terms of memory and response time.

On the other hand, regarding whether to use the Volto proxy or set route on the reverse proxy to point directly to Plone. The points: handling authentication (as you said), but also, for example, handling error pages, such as 404. In fact pointing from reverse proxy directly to Plone, without any specific intervention on reverse proxy or on Plone, when faced with a missing resource, you would have the 404 page from Plone. Which is not so good ...

mamico:
implented as an addon https://github.com/collective/volto-middleware-binary

PR #6318, opened 2024-09-25
fix: Listing block default summary should use ul tag for its listing

ana-oprea:
This PR continues the work from [PR #4994](https://github.com/plone/volto/pull/4994), which is now outdated. To ensure progress on the development, I've opened this new PR to carry the changes forward and address any remaining tasks or updates.

### DIFF
--- a/packages/volto/src/components/manage/Blocks/Listing/SummaryTemplate.jsx
+++ b/packages/volto/src/components/manage/Blocks/Listing/SummaryTemplate.jsx
@@ -2,6 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import { ConditionalLink, Component } from '@plone/volto/components';
 import { flattenToAppURL } from '@plone/volto/helpers';
+import { List } from 'semantic-ui-react';
 
 import { isInternalURL } from '@plone/volto/helpers/Url/Url';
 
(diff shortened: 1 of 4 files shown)


ichim-david:
@JeffersonBledsoe never mind the review request there needs to be some changes done to this pull request, one of them being not reintroducing Semantic-ui component within the Summary Template

ichim-david:
@ana-oprea there are a couple of issues currently:
1. the listing gets empty p tag even when no content is found (no description added for item)
2. the listing also affects the coloring which might only be of importance for editors and otherwise
confusing for anon users and wonder why there are blue or black titles
3. it's using again semantic ui react which we want to avoid adding back into core and would be better
to use simple html elements


review-state
classic-summary-view

sneridagh:
@ana-oprea Thanks for removing SemanticUI fromt here! However, I won't remove the current `.ui.list` CSS, since you could break a lot of projects out there.
Just add the new one instead. Also, we need an upgrade guide entry since it's a breaking change in HTML of the component.

pnpm-lock.yaml
pnicolli, line 1: This change is very likely here by mistake for some git shenanigans I don't fully understand. Let's clean this up and push this forward.

pnicolli (changes requested):

pnicolli:
@plone/volto-accessibility If this could have another look now that SemanticUI is not in there anymore it would be great.

packages/volto/src/components/manage/Blocks/Listing/SummaryTemplate.jsx
```diff
@@ -22,17 +22,28 @@ const SummaryTemplate = ({ items, linkTitle, linkHref, isEditMode }) => {
...
-                <h3>{item.title || item.id}</h3>
-                <p>{item.description}</p>
+        <ul className="ui list summary-listing">
+          {items.map((item) => (
+            <li key={item['@id']}>
+              <div
+                className={`listing-item ${
+                  item.review_state ? `state-${item.review_state}` : ''
+                }`}
+                role="listitem"
```
JeffersonBledsoe, line 32: The `role="listitem"` isn't needed here as the containing `<li>` element already provides this role
```suggestion
```

packages/volto/src/components/manage/Blocks/Listing/SummaryTemplate.jsx
```diff
@@ -22,17 +22,28 @@ const SummaryTemplate = ({ items, linkTitle, linkHref, isEditMode }) => {
...
+              >
+                <ConditionalLink item={item} condition={!isEditMode}>
+                  <Component componentName="PreviewImage" item={item} alt="" />
+                  <div className="listing-body">
+                    <div className="listing-item-header">
+                      {item.title ? item.title : item.id}
+                    </div>
```
JeffersonBledsoe, line 39: I think we might actually want the option to set this as a header (and customise the heading level) as it's common practise to use heading levels in listings.
I'll comment on the original issue and see if we can come to an agreement

JeffersonBledsoe (changes requested):
One code change here but I'd like to discuss the removal of setting a heading level before this gets merged. See https://github.com/plone/volto/issues/4982#issuecomment-2910785072

Issue #2487, opened 2021-06-02
A11y: Table block Headline text has insufficient color contrast


ThomasKindermann:
The Headline text Color has insufficient color contrast (3.11:1) with the background of the Table Headline.
For better A11y it needs to have a color contrast of 4.5 or higher. 

**To Reproduce**
1. Add Table block to any page
2. Enter text in Headline Field
3. Open Axe-Dev and check site
4. Axe-Dev tool shows error

**Expected behavior**
Axe-Dev tool should show no color contrast violation for table headline
Color contrast should be above 4.5:1

**Screenshots**
![Screenshot 2021-06-02 at 15 54 18](https://user-images.githubusercontent.com/50659809/120499938-c0c1de00-c3c0-11eb-985c-74dfd38761fe.png)

![Screenshot 2021-06-02 at 16 39 52](https://user-images.githubusercontent.com/50659809/120500529-375edb80-c3c1-11eb-89af-86f51b667f21.png)

**Software:**
 - OS: OSX
 - Browser chrome
 - Volto Version "13.1.2"

ThomasKindermann:
https://github.com/plone/volto/pull/2488 i changed the text color to black but now its harder to distinguish the headline from the other fields, any suggestions?

@sneridagh @tisto

pnicolli:
Fixed by #6333

PR #6333, opened 2024-09-26
Increase contrast in table block for a11y reasons

jackahl:
fix for #2487 and #6334

### DIFF
--- a/packages/volto-slate/locales/pt_BR/LC_MESSAGES/volto.po
+++ b/packages/volto-slate/locales/pt_BR/LC_MESSAGES/volto.po
@@ -1,0 +1,205 @@
+msgid ""
+msgstr ""
+"Project-Id-Version: Plone\n"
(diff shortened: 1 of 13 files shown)


JeffersonBledsoe:
Some details on the change:
Current contrast ratio as tested by [WebAIM](https://webaim.org/resources/contrastchecker/) is 6.66:1 which surpasses the 4.5:1 contrast ratio required for 'normal' size text required by WCAG 2.2 SC 1.4.3 Contrast Minimum (AA). The body text doesn't meet SC 1.4.6 Contrast Enhanced, but I don't see this as an issue as we generally don't aim for AAA.

Screenshot of the PR in it's current state is below. 

Screenshot showing light blue text against a dark background.

JeffersonBledsoe (changes requested):

JeffersonBledsoe (approved):
Looks good!

pnicolli (approved):

PR #6338, opened 2024-09-26
Update Italian Italian translations.

gianniftp:
Added some translation to the italian locales.


stevepiercy:
@gianniftp @fredvd this is the same issue that @folix-01 experienced with @mister-roboto. I notice that their PRs are both indicated as their first to `plone/volto`. Could that be a factor, generating a false-positive?

gianniftp:
Sorry for my mistake. There is no email address in the commits. I will fix this asap.

Issue #5129, opened 2023-08-24
a11y - sharing page - Under table text description - cms ui 


Wagner3UB:
**Describe the bug**
The text refers to checkmark symbols that shows information through color, the font is too small, and are also hidden with 'aria-hidden

**Expected behavior**
Use symbols that distinguish checkmarks not only by color but also by shape (such as using an 'x' instead of a checkmark) and enable 'aria-label' to describe the symbols rather than hiding them.

**Screenshots**
<img width="1184" alt="Schermata 2023-08-24 alle 17 24 39" src="https://github.com/plone/volto/assets/60133113/f7047cd1-774d-488b-85fe-7ca03da849c1">

**Software (please complete the following information):**
Volto Version 17.0.0-alpha.19
Plone Version 6.0.4

PR #6347, opened 2024-09-27
Fix Sharing page icons for a11y

cekk:
This solves #5129 

I should update translations too because i moved the text from under the table to page description (because it's better for the user to read instructions before the table).


Wagner3UB (approved):

packages/volto/src/components/theme/Icon/Icon.jsx
```diff
@@ -56,8 +56,9 @@ const Icon = ({
     onClick={onClick}
     id={id}
     aria-hidden={ariaHidden}
+    aria-label={title || ''}
```
pnicolli, line 59: Do we really want aria-label as empty string in every current usage of this component, since we just fix the Sharing view here? What if we changed it to 
```suggestion
    aria-label={title}
```
so that aria-label is not rendered if title is not given? @plone/volto-accessibility

packages/volto/src/components/theme/Icon/Icon.jsx
```diff
@@ -56,8 +56,9 @@ const Icon = ({
...
     dangerouslySetInnerHTML={{
-      __html: title ? `<title>${title}</title>${name.content}` : name.content,
+      __html: title ? `<title >${title}</title>${name.content}` : name.content,
```
pnicolli, line 61:
```suggestion
      __html: title ? `<title>${title}</title>${name.content}` : name.content,
```
Seems like a useless change to me, is it not?

pnicolli (changes requested):

Wagner3UB:
@cekk I did the modifications suggested by @pnicolli, but there is some fail unit tests to re-do. Can you please run those tests and fix it if needed?

PR #6301, opened 2024-09-20
fix: getQuerystringSearchQuery doesn't accept path as an input

askadityapandey:
This Pull Request fixes #6254 , which is "Without being able to pass in a path to the querystring-search, all searches are performed from the root, breaking relative queries (e.g. ../::1)"
I've added all the important tests which are required, also made it optional so it does not break for the existing users. 
This is a cool project btw, will look forward to contribute more <3

PR #6349, opened 2024-09-27
Fixed client copy mutation, cleanup up move mutation for consistency

pnicolli:


ichim-david:
@pnicolli @sneridagh it seems that the path needs to be allowed to be set for these crud operations, I noticed another pull request added https://github.com/plone/volto/pull/6301/files#diff-ca875ccb3b13f54d06884d6a8d9a696d1d331322bcc0c06ef618e083ce147eb4R11 that does something similar to Pierro's pull request but for querystring-search.

sneridagh (approved):
LGTM

@ichim-david do you think that this collides with the PR you mention? Or you wanted to raise attention to the other as well?

ichim-david:
> LGTM
> 
> @ichim-david do you think that this collides with the PR you mention? Or you wanted to raise attention to the other as well?

Just raised awareness that it seems that the path needs to be added in several places

pnicolli:
> Just raised awareness that it seems that the path needs to be added in several places

Yeah that's true, these will need to be fixed as we start using the library.

PR #6356, opened 2024-09-30
Add visually-hidden class

JeffersonBledsoe:


packages/volto/theme/themes/pastanaga/extras/utils.less
```diff
@@ -61,3 +61,16 @@ body:not(.has-sidebar):not(.has-sidebar-collapsed) {
     }
   }
 }
+
```
ichim-david, line 64: there might be something from semantic already for visually hidden

sneridagh:
@JeffersonBledsoe @ichim-david if we do this, we need that also in `@plone/components` (for now) to have it also in the future theming.

sneridagh:
@JeffersonBledsoe it seems you pushed more changes than the needed in this PR. Can you take a look, please? you deleted things and files, that we cannot remove.

ichim-david:
@sneridagh @JeffersonBledsoe I've taken a deeper look and it seems that bootstrap 5 renamed sr-only to visually-hidden.
barceloneta is now based on bootstrap 5 as I see
https://getbootstrap.com/docs/5.3/migration/#helpers-2
https://github.com/plone/plonetheme.barceloneta/tree/master/scss
If we stick to the visually-hidden class, we have the same class name in both systems.

This is such a standard nowadays that I doubt that it could lead to class conflicts on account of modifying such class.
As such I don't think we need to rename it to sr-only unless @sneridagh you have another opinion.

Wagner3UB:
@plone/volto-accessibility I'll try to run some tests within our theme to see if I can find any conflicts.

ichim-david (approved):
I think this is fine from my point of view, waiting also for @Wagner3UB review if there would be any conflict with the bootstrap theme

packages/volto/theme/themes/pastanaga/extras/utils.less
Wagner3UB: Semantic doesn’t have anything for that purpose. I know Bootstrap uses .sr-only, but the structure is always the same.

Wagner3UB: Info: Bootstrap uses .visually-hidden starting from version 5, before that, it was .sr-only.

Wagner3UB:
All good here for me.

Wagner3UB (approved):

Wagner3UB:
@JeffersonBledsoe @ichim-david Can we merge this one?

JeffersonBledsoe:
@Wagner3UB I still need to add this to `@plone/components`. Should get that done this week

giuliaghisini:
@JeffersonBledsoe @ichim-david what about this? Is it ready to merge? we need this fixes..

docs/source/upgrade-guide/index.md
```diff
@@ -34,6 +34,16 @@ It is usually better and quicker to move your items into new locations and copy
 
 (19-removed-support-for-loading-configuration-from-project-label)=
 
+### Breaking Changes — `.visually-hidden` utility class
```
Wagner3UB: @stevepiercy Hey Steve, I could use your help here. It’s my first time handling a breaking change in Volto, so I’m not too familiar with this file. I put together a first draft, but I’d really appreciate your input to make sure it fits the right style.

stevepiercy: It's my understanding that all changes in the Upgrade Guide are breaking. Otherwise you wouldn't need to mention it. I'll wordsmith a suggestion.

docs/source/contributing/accessibility-guidelines.md
```diff
@@ -39,3 +39,7 @@ The reasoning is that the `<a></a>` HTML tag has specific behavior that screenre
 ## Make sure form elements have a label
 
 This is true for one-element forms as well, such as the "Search" form on the folder-contents component. Putting an icon on it does not convey any meaning to screenreader-users, you should clarify it with an aria-label.
+
+## Additional information for non-visual users
+
+Sometimes extra information needs to be provided to users who are not able to see the screen, The `.visually-hidden` class can be used to wrap elements which need to be exposed to assistive technologies such as screen readers without them being visible.
```
stevepiercy:
```suggestion
Sometimes extra information needs to be provided to users who are not able to see the screen.
The `.visually-hidden` class can be used to wrap elements which need to be exposed to assistive technologies, such as screen readers, without them being visible.
```

docs/source/upgrade-guide/index.md
```diff
@@ -34,6 +34,16 @@ It is usually better and quicker to move your items into new locations and copy
...
 
+### Breaking Changes — `.visually-hidden` utility class  
+PR: [Add visually-hidden class #6356](https://github.com/plone/volto/pull/6356)
+A new global CSS utility class called .visually-hidden has been introduced to Volto’s SCSS base.
+
+This class allows developers to visually hide elements while keeping them accessible to screen readers, improving accessibility for assistive technologies.
+
+Why this is breaking: If your project already defines a .visually-hidden class or uses similar accessibility helpers, the new global definition may override or conflict with existing custom styles.
+
+What to do: Check whether your project or custom theme defines a .visually-hidden class. If so, rename your custom implementation or override Volto’s default as needed. Review any components that depend on hidden accessibility elements to ensure visual and functional consistency.
```
stevepiercy: This was pretty good already, I only moved around the bits and pieces and added some glue.

````suggestion
### New utility class `visually-hidden`

```{versionadded} Volto 19.0.0-alpha.8
```

A new global CSS utility class called `visually-hidden` has been introduced to Volto's SCSS base.

This class allows developers to visually hide elements while keeping them accessible to screen readers, improving accessibility for assistive technologies.

If your project, add-on, or custom theme already defines a `visually-hidden` class, or uses similar accessibility helpers, the new global definition may override or conflict with existing custom styles.
In which case, you should rename your custom implementation, or override Volto's default, as needed.
Also review any components that depend on hidden accessibility elements to ensure visual and functional consistency.

```{seealso}
[Add visually-hidden class #6356](https://github.com/plone/volto/pull/6356)
```
````

**Comment by wesleybl** [comment]
> I think it requires a small note in the upgrade guide though, it can be seen as a breaking change.

@pnicolli @Wagner3UB I find it strange that this is considered a breaking change. Is this because the class name `visually-hidden` is common in CSS frameworks? Can't we have a name like `visually-hidden-volto`? That way we would avoid name conflicts. Otherwise, how can we have this functionality in Volto 18?

**Comment by nileshgulia1** [comment]
> > I think it requires a small note in the upgrade guide though, it can be seen as a breaking change.
> 
> @pnicolli @Wagner3UB I find it strange that this is considered a breaking change. Is this because the class name `visually-hidden` is common in CSS frameworks? Can't we have a name like `visually-hidden-volto`? That way we would avoid name conflicts. Otherwise, how can we have this functionality in Volto 18?

I guess yes, this a11y change would silently changes behavior of any custom class with the same name. I see it as a behaviour change not a developer error.

**Review by nileshgulia1 — APPROVED** [review]
Maybe we can add some code examples in the upgrade guide to show the exact class name. Otherwise lgtm as well.

**Comment by Wagner3UB** [comment]
> > I think it requires a small note in the upgrade guide though, it can be seen as a breaking change.
> 
> @pnicolli @Wagner3UB I find it strange that this is considered a breaking change. Is this because the class name `visually-hidden` is common in CSS frameworks? Can't we have a name like `visually-hidden-volto`? That way we would avoid name conflicts. Otherwise, how can we have this functionality in Volto 18?

@wesleybl @JeffersonBledsoe @pnicolli This can be discussed, but it creates room for duplicate classes to remain in the code. I’m not sure if that would be acceptable. If it is, we can certainly create a specific class to reference this kind of behavior. Keep in mind that there are other older classes from different libraries that might have been included, such as sr-only, an old class from Bootstrap 4.

**Comment by Wagner3UB** [comment]
> Maybe we can add some code examples in the upgrade guide to show the exact class name. Otherwise lgtm as well.

@nileshgulia1 Sure, add info is never a problem!

**Comment by pnicolli** [comment]
> I find it strange that this is considered a breaking change. Is this because the class name `visually-hidden` is common in CSS frameworks?

@wesleybl I personally usually consider anything that could change the end user website appearance as a breaking change. Adding a global css class with a name that is commonly already used elsewhere, even if it usually has the same goal, can fall into these breaking change category. Again, this is just my personal opinion, and I understand it might look a little too strict. @sneridagh you got any opinion about this matter?

> Can't we have a name like `visually-hidden-volto`? That way we would avoid name conflicts. Otherwise, how can we have this functionality in Volto 18?

I am not sure, Plone used to have these "branded" classes in Plone 5 and I found them confusing eventually. I would not be 100% against it but hoestly I would just put the class in your custom projects when you need it if those are Volto 18 still, it's a super quick addition.

File: docs/source/upgrade-guide/index.md [hunk 5]
```diff
@@ -50,6 +50,23 @@ If you can't upgrade immediately, you may continue to run Volto 19 on Node.js 20
 
 (19-removed-support-for-loading-configuration-from-project-label)=
 
+### New utility class `visually-hidden`
+
+```{versionadded} Volto 19.0.0-alpha.8
```
stevepiercy: We blinked.

````suggestion
```{versionadded} Volto 19.0.0-alpha.10
````

sneridagh:
@pnicolli @wesleybl luckily we have this written down: 

https://6.docs.plone.org/volto/release-management-notes/index.html#dom-structure-and-default-styling

Yes, it is a breaking change and should be documented as one properly. Also, this makes it non-backportable to 18. As @pnicolli said, a quick addition to any 18 project.

Regarding the `@plone/components`, yes it should be backported, since the "canonical" version that we use is in the seven branch.

Wagner3UB:
@JeffersonBledsoe I think we are done here! ~~Only thing to do is the backport to 18.x.x~~

@pnicolli @stevepiercy We have your blessing?

pnicolli:
I'm so sorry to bring up more problems, but I realized that we are adding this to `@plone/components` but actually react-aria has its own way of doing this: https://react-spectrum.adobe.com/react-aria/VisuallyHidden.html
We should probable then avoid adding this to `@plone/components`, or is there a reason that I don't see at the moment? @sneridagh 

Also since this is a breaking change, we cannot backport it to Volto 18.

wesleybl:
> but hoestly I would just put the class in your custom projects when you need it if those are Volto 18 still, it's a super quick addition.

@pnicolli @sneridagh @Wagner3UB What I'm trying to say is this: We have an accessibility issue in both Volto 19 and Volto 18. If the fix in Volto 19 is a breaking change, fine. But we need to find a way to make it work in Volto 18 as well. Volto 18 is the current version of Volto. We need to fix accessibility issues in it. This is a bug. We can't just say "fix it in your project."

pnicolli:
> What I'm trying to say is this: We have an accessibility issue in both Volto 19 and Volto 18. If the fix in Volto 19 is a breaking change, fine. But we need to find a way to make it work in Volto 18 as well. Volto 18 is the current version of Volto. We need to fix accessibility issues in it. This is a bug. We can't just say "fix it in your project."

Just want to make sure here that I explain my thoughts in a clear way: we are only talking about adding the `visually-hidden` class here in this PR, not fixing any bug. Adding this class as a global class is the breaking change, and I am willing to have this breaking change because I find this useful for the future and for some of the currently open bugs that we have.

If you are thinking about the bugfix in #7500 then yes, I agree we should have a fix for that in Volto 18 as well. That PR is out of scope here, even if it is what triggered the discussion on this PR, and thank you again for bringing this up over there.

We should try to stick to our policy about breaking changes at all times unless really unavoidable. In the specific case of #7500 we should finish and merge this PR and then use the class that we are adding here to fix that issue in Volto 19. Then we should fix the issue in Volto 18, absolutely, but I would simply change the fix to make it non-breaking. For example, for Volto 18, we should not use this class and instead apply these styles directly to the affected item. I would say let's continue the discussion about that other PR in its specific thread and not here, otherwise we risk mixing things up too much. Am i making sense here? Always happy to discuss!

sneridagh:
@pnicolli you are right, let's remove the addition to `@plone/components` for now. I didn't realise of this fact. 
Could someone take care of it?

Wagner3UB:
@pnicolli Ok, completely understood. Thank you.

Wagner3UB:
@pnicolli @sneridagh Done, I have removed the plone/components part. Now we are good to go.

PR #6366, opened 2024-10-03
Fix links to Vite website

stevepiercy:


stevepiercy:
I put a changelog entry in `packages/volto/news`, because our docs don't say where to put changes to apps. See https://6.docs.plone.org/contributing/index.html#change-log-entry.

I can update the docs in a separate PR, but I need to know the correct location.

Also one of the changelog checks passed because we don't check apps.

sneridagh:
@stevepiercy wut! they should have changed it yesterday, I've been cheking docs the last weeks in `vitejs.dev`. Today it's their conf, I guess they changed it to have it by then.

PR #6369, opened 2024-10-03
fix: default language for content if not set

giuliaghisini:
if language from retrived content is not set for some reasons, set default language from settings


davisagli (changes requested):
In a multilingual site, the default language should be inherited from the container, not from the site default language.

In my opinion this should be handled in the REST API, not here.

PR #6376, opened 2024-10-04
Remove pin on Vale, and upgrade to v3.x configuration

stevepiercy:
Upgrade to Vale 3.

See https://github.com/plone/documentation/issues/1602.

----
📚 Documentation preview 📚: https://volto--6376.org.readthedocs.build/


requirements-docs.txt
```diff
@@ -7,4 +7,4 @@ sphinx-copybutton
 sphinx-examples
 sphinxcontrib-video
 sphinxext-opengraph
-vale==2.30.0
+vale
```
davisagli, line 10: Let's not tackle this in this PR, but I would think we'd want all of these pinned so that the docs build doesn't break at some random time in the future just because a new version of something is released.

davisagli (approved):

requirements-docs.txt
stevepiercy: I created a new issue in plone-sphinx-theme to pin requirements, as that will soon become upstream for all docs.

https://github.com/plone/plone-sphinx-theme/issues/30

PR #6384, opened 2024-10-08
Backport #6309 to Volto 17

dobri1408:
This is a backport PR for [PR #6309](https://github.com/plone/volto/pull/6309).

**Why I believe this is not a breaking change:**

1) This can be viewed as a bug fix, as the History page appears broken when used on a complex page. This PR aims to resolve that issue.

2) Although the snapshot changes, it doesn't introduce a breaking change since there is no visual impact. The removal of the span occurs because this fix ensures that RenderBlocks, rather than DefaultView, is used to render the block changes, as it should be.

### DIFF
--- a/news/6384.feature
+++ b/news/6384.feature
@@ -1,0 +1,1 @@
+improve DiffField.jsx with better support for displaying HTML elements such as images @dobri1408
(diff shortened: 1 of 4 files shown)


sneridagh (approved):
@dobri1408 Thanks for pointing this. It is in fact, not a breaking change, as documented in the Release Management Notes:

https://6.docs.plone.org/volto/release-management-notes/index.html?highlight=breaking%20change#private-api

Let's go for it then!

PR #6385, opened 2024-10-09
Fix syntax of `html_meta` under `myst` field.

stevepiercy:


stevepiercy:
Merging as this is not worthy of a news item or testing. @MrTango FYI the error message when building the docs was:

```
/<path>/volto/theming/theming-a-base-theme.md:1: WARNING: top-level 'html_meta' key is deprecated, place under 'myst' key instead [myst.topmatter]
```

PR #6388, opened 2024-10-09
Upgraded nextjs deps and configuration

pnicolli:


sneridagh:
@pnicolli I am one of the late "ESM" all the things. How about declare the nextjs app `type: module`?

pnicolli:
@sneridagh I didn't dig deep into that world yet. What would be the change you are proposing? What I did here is basically to create an empty nextjs project with the suggested tooling in the nextjs docs and copy the config over here 😛

sneridagh:
@pnicolli yeah, basically then you don't have to do the `.mjs` dance anymore. I don't think there are other implications from the app point of view, unless you have to deal with importing .cjs files around.

pnicolli:
@sneridagh I gave it some thought and I think `type: module` is not a good fit for this app. I would like to keep it as close as possible to out-of-the-box nextjs.

I merged main and fixed a couple eslint warnings, the PR about providers would also benefit from this PR imho.

apps/nextjs/.gitignore
```diff
@@ -4,6 +4,7 @@
 /node_modules
 /.pnp
 .pnp.js
+.yarn/install-state.gz
```
sneridagh, line 7: ?

sneridagh (approved):
LGTM, this can go before the providers refactor

apps/nextjs/.gitignore
pnicolli: It's in the latest create-next-app basic template 😅

PR #6419, opened 2024-10-20
(fix): Enable Volto Login to Redirect Back to Original Route After Authentication 

dobri1408:
There are situations where you may want the Volto login to appear on a route other than `/login`, as demonstrated in this example: https://github.com/collective/volto-authomatic/blob/a78f354e2c8fc4a41cf9d71d76a93f1d1627d268/src/index.js#L18. This implies that the Volto login should be accessible from any route invoked, thus the redirect after you are logged in should work also.

To achieve this, I removed the hardcoded `/login` route and implemented a regex to strip the last segment of the route where the login was initiated.


packages/volto/news/6419.bugfix
```diff
@@ -0,0 +1 @@
+ Enable Volto Login to Redirect Back to Original Route After Authentication @dobri1408
```
stevepiercy, line 1:
```suggestion
Enable login to redirect back to original requested route after authentication. @dobri1408
```

stevepiercy (changes requested):
Will this change work in multi-language sites, too, where the route may have a language prefix?

stevepiercy:
I think this is not a "fix" but a "feature". `login` was not broken. It could also be "breaking".

I have wanted this feature in Volto for a long time. Thanks for bringing it forward.

dobri1408:
@stevepiercy, yes it will work also on a multilingual site, because the login route is still the last one

stevepiercy (approved):
News item looks good. This still needs technical review before merging.

nileshgulia1:
@dobri1408  Just thinking out loud, What about the cases where we have `http://localhost:3000/login/page-one/page-two`? (However its hard to imagine feasibility of these cases). As per current scenario we only replace paths ending with /login, given that we have Login component at that path. For custom paths its ideal to deal with it in its own component? 🤔

JeffersonBledsoe:
Perhaps I'm not understanding what this PR achieves, but can this not already be done by appending `/login` to any path? e.g. `/my/long/path/login` will return you to `/my/long/path` after you login.

This is works by the path being caught in [routes.js](https://github.com/plone/volto/blob/27120d94b21c12350f977531f85c822e99420b3e/packages/volto/src/routes.js#L139) and the `props.location` in [Login.jsx](https://github.com/plone/volto/blob/27120d94b21c12350f977531f85c822e99420b3e/packages/volto/src/components/theme/Login/Login.jsx#L83) that gets set by the router (from the wildcard catch in routes.js) is `return_url` which gets returned to after a successful login.

stevepiercy:
> Perhaps I'm not understanding what this PR achieves, but can this not already be done by appending `/login` to any path? e.g. `/my/long/path/login` will return you to `/my/long/path` after you login.

Confirmed!

https://demo.plone.org/vertical-spacing/grid-and-text/login

We should document this feature, as I never knew it existed. I imagine both end users and developers would want to know about it.

packages/volto/news/6419.feature
```diff
@@ -0,0 +1 @@
+Enable login to redirect back to original requested route after authentication. @dobri1408
```
davisagli, line 1:
```suggestion
Fix redirect after login if the `Login` component is used on a route other than `/login` or `/logout`. @dobri1408
```

packages/volto/news/6419.feature
davisagli, line 1: Rename to 6419.bugfix

davisagli (approved):
@JeffersonBledsoe @stevepiercy The release note was a bit misleading, so I suggested an edit. This is not about solving the general problem of taking you to where you were going after an interstitial login page. It's just a fix for a specific case where the login form was served at /fallback_login instead of /login, and didn't redirect anywhere after logging in.

PR #6423, opened 2024-10-22
Fix backend error when there is no query supplied

tedw87:
When you add a listing block with no criteria, it will give a backend error.

Fixes https://github.com/plone/volto/issues/3839


avoinea (approved):

davisagli:
I tried to reproduce the bug this is trying to solve, but I don't see the "No query supplied" exception. What am I missing?

tedw87:
Hi @davisagli, you can reproduce this by adding a listing-block and just save it with no criteria selected. Then refresh that page and you should see the error on the backend console. 
I will share here all the issues that are pointing to this problem:
https://github.com/plone/volto/issues/3839
https://github.com/plone/volto/pull/3866
https://github.com/plone/volto/issues/3474

davisagli:
> you can reproduce this by adding a listing-block and just save it with no criteria selected. Then refresh that page and you should see the error on the backend console.

Sorry if I'm being dumb, but this is exactly what I tried and I do not see the error. What version of the backend are you running and how did you start it? And are you running Volto from this branch with `make start`, or some other version?

tedw87:
So I tested in Volto 18 (main branch) and backend I started with `make backend-docker-start`, and in Volto I make a new page, add a Listing block with no criteria selected. I save the block and there is no error at this point, error apears when i refresh the page where i created the listing block.

davisagli (approved):
Looks good. I'm not sure why I couldn't reproduce the error the other day.

PR #6460, opened 2024-11-03
Fix display of nvm version when installing it for Contributing to Volto.

stevepiercy:
We previously consolidated these instructions into a single include file, so the replacement is no longer needed, and it stopped working when we moved it into a code block.

See https://github.com/plone/documentation/pull/1746/

----
📚 Documentation preview 📚: https://volto--6460.org.readthedocs.build/


stevepiercy:
Merging as I already tested docs locally, and that's all that's changing here.

PR #6461, opened 2024-11-03
Fix types, better exports and use tsup for build, remove parcel

sneridagh:
We are having problems in some setups in macos where `parcel` errors with a segmentation fault and we found also some weird errors related with it in `pnpm` lock, where the lock keeps updating because of it (apparently, their entries in the lock keep cycling infinitely). In addition, the resultant code has some weird names with a looong hash that end up in the dev tools. Then I also learned that `tsup` is more pure with their outputs and it was the recommended setup, see:

https://www.totaltypescript.com/how-to-create-an-npm-package

Totally worth it from top to bottom.

I also fixed the remaining types that forced us to `||true` in the build output. So overall this PR improves the current situation, in a non-breaking way.


packages/registry/package.json
```diff
@@ -70,10 +63,9 @@
   },
   "scripts": {
     "watch": "parcel watch",
```
davisagli, line 65: There's still a parcel command here.

davisagli (commented):
Could you explain more about the switch from parcel to tsup? What's the motivation and are there any tradeoffs?

sneridagh:
@davisagli just added some more background in the description.

packages/registry/news/6461.bugfix
```diff
@@ -0,0 +1 @@
+Fix types for addons TS, fix tsconfig for node side @sneridagh
```
stevepiercy, line 1:
```suggestion
Fix types for add-on's TypeScript. Fix `.tsconfig` for Node.js side. @sneridagh
```

packages/registry/news/6461.internal
```diff
@@ -0,0 +1 @@
+Replace `parcel` with `tsup` for build @sneridagh
```
stevepiercy, line 1:
```suggestion
Replace `parcel` with `tsup` for build. @sneridagh
```

packages/scripts/news/6461.internal
```diff
@@ -0,0 +1 @@
+Improve exports in @plone/registry, adapt the detection in i18n script @sneridagh
```
stevepiercy, line 1:
```suggestion
Improve exports in `@plone/registry`. Adapt its detection in i18n script. @sneridagh
```

stevepiercy (changes requested):
Was `parcel` something the developer directly called? If so, does its replacement merit more than a change log entry?

PR #6466, opened 2024-11-04
[client] Replace `parcel` with `tsup`

sneridagh:
Replace `parcel` with `tsup`.
Use a better tsconfig based on experience and best practices, fix typings imports.


davisagli (approved):

packages/client/package.json
pnicolli, line 1: Not seeing tsup in the devDependencies in package.json here, is that on purpose?

packages/client/package.json
```diff
@@ -22,17 +22,13 @@
     "src",
     "README.md"
   ],
-  "source": "./src/index.ts",
```
pnicolli, line 25: Should we also remove `src` from `files` above since we don't export sources anymore?

packages/client/package.json
sneridagh: I did some test in the past, and the devDep was left in there :/ So that's why it's not in the changes.

packages/client/package.json
sneridagh: Done!

PR #6467, opened 2024-11-04
Replace `parcel` with `tsup`. Better types, better tsconfig. Move to ESM.

sneridagh:


davisagli (approved):
@sneridagh What does adding `type` in the imports do?

sneridagh:
@davisagli type declares the import "Typescript only", so the interpreter knows that it does not belong to the code, so it strips it on build time when transpiling. It's a best practice nowadays. This is enforced by the `tsconfig` option "verbatimModuleSyntax".

I'm trying to come up with a golden `tsconfig` for all our browser libs, then one for the Node ones. I think we are near.

davisagli:
@sneridagh So the type system can't even tell which things are types without giving manual hints? Pssh.

pnicolli (approved):

sneridagh:
@davisagli it can, but sometimes it seems it has a hard time doing it, so it's better to explicit tell him :) It kind of makes sense in a bundler world!

PR #6472, opened 2024-11-05
React Router 7 experimental PoC

sneridagh:
----
📚 Documentation preview 📚: https://volto--6472.org.readthedocs.build/


sneridagh:
@stevepiercy @pnicolli I'd like to have this merged asap, I need it in the my slots PR to continue in there. Could you please take a look?

apps/rr7/app/routes/home.tsx
```diff
@@ -0,0 +1,79 @@
...
+    getContentQuery({ path: flattenToAppURL(request.url), expand }),
+  );
+
+  return { dehydratedState: dehydrate(queryClient) };
+}
+
+function Page() {
+  const { getContentQuery } = usePloneClient();
+  const pathname = useLocation().pathname;
+  const { data } = useQuery(getContentQuery({ path: pathname, expand }));
```
pnicolli, line 51: It's ok to go full steam ahead with this for a PoC, but as food for thought for our future discussions, I don't think we need useQuery here, imho we don't really want to revalidate data automatically or the other good features react-query offers when we are simply rendering content.

pnicolli (approved):

apps/rr7/app/routes/home.tsx
sneridagh: @pnicolli something we have to see with time. React Query excels in one thing, that it's when your data is stale, it asks for it again automatically, so the page updates.

Let's imagine you are in a page, you leave it open the whole night. Someone updates it, when you resume your work, the page autoreloads with the new data. Yes, the data is in SSR response already (like in Next.js), but how about stale data? This is also true for everything in your app that it's in "the cache" (old store). In Volto we didn't have this luxury. Now that we have it, let's use it. 

For clarification: there's no request happening, since it's getting the data from the hydration anyways. So, the full circle is complete.

sneridagh: Another clarification: The current Remix app has a bug, and it DOES the request anyways. I'll fix it after this is closed.

pnicolli: I would say "go with this" and then we can take some time to discuss it because I don't think we really need that shiny feature by default, but I understand the opportunity and power of it

PR #6496, opened 2024-11-21
Update instructions to install pipx in `RELEASING.md`

stevepiercy:


davisagli (approved):
Thanks. I tested this on my machine by removing my existing pipx binary and removing ~/.local/bin from my PATH, then followed the instructions in a new terminal.

PR #6497, opened 2024-11-21
Fixes ICS download in non-public event content 

sneridagh:


davisagli (commented):
@sneridagh Needs changelog

davisagli:
@sneridagh Is there also a fix for main?

PR #6513, opened 2024-12-05
Remove SemanticUI CSS declaration `overflow-x: hidden` in BODY tag.

sneridagh:
We had this behavior:

https://github.com/user-attachments/assets/f43aadc6-83af-4734-9269-79cb74698323

Turns out that SemanticUI is forcing `overflow-x: hidden` in the BODY class. That provokes the screen to jump when the popover is shown.

This PR resets this declaration.

We will test this in VLT 6.0.0a3: https://github.com/kitconcept/volto-light-theme/pull/440 and give feedback. Not sure if this could be potentially breaking.


pnicolli:
At RedTurtle we bumped into this a couple times already and didn't notice this PR.
Love this, but I think it's definitely a breaking change. On the other hand, this could be considered almost like a blocker for properly adopting RAC if one wants to. Maybe putting this in 19 and documenting it is enough? 🤔

pnicolli (approved):

packages/volto/news/6513.bugfix
```diff
@@ -0,0 +1 @@
+Removed SemanticUI CSS declaration `overflow-x: hidden` in BODY tag. @sneridagh
```
stevepiercy, line 1:
```suggestion
Removed Semantic UI CSS declaration `overflow-x: hidden` from the `body` tag. @sneridagh
```

packages/volto/theme/themes/pastanaga/extras/main.less
```diff
@@ -17,6 +17,9 @@
 
 body {
   display: flex;
+  // RAC Popovers have problems with the default declaration coming from Semantic UI
+  // that forces `overflow-x: hidden` on the BODY.
```
stevepiercy, line 21: I have no idea what RAC means, but this is my guess. Please correct me if it is wrong.

```suggestion
  // React Autocomplete Component popovers have problems with the default declaration
  // in Semantic UI that forces `overflow-x: hidden` on the body.
```

stevepiercy (changes requested):

packages/volto/theme/themes/pastanaga/extras/main.less
sneridagh: LOLOLOL

sneridagh: Fixed.

stevepiercy (approved):

PR #6509, opened 2024-12-02
Removed all imports from barrel files for components, actions, helpers, hooks and lodash

pnicolli:
Closes #6507

PR #6516, opened 2024-12-06
improve link integrity popup

giuliaghisini:
fixed and improved link integrity popup when deleting item/items from folder contents. 
Moved all code in a separate component 'ContentsDeleteModal'.

Added loader when loading references. If the call to @linkintegrity took a long time to load, it wasn't clear to the user that it was checking for broken references:
 
loading

If only one item is selected to delete, it will display all the possibile broken references in a table, instead in a list. In the last column could be more than one reference: 
1-selected-item

If more than one item is selected to delete, it will not display the possibile broken references, but only the number, because it  will be a very long list (It worked like that before too):
more-than-one-selected


packages/volto/locales/volto.pot
```diff
@@ -2355,6 +2356,11 @@ msgstr ""
 msgid "Navigate back"
 msgstr ""
 
+#. Default: "Navigate to this item"
+#: components/manage/Contents/ContentsDeleteModal
+msgid "Navigate to this items"
```
stevepiercy, line 2361: This should be either "Navigate to this item" or the "Navigate to these items", and align with the Default. I'm not sure whether it should be plural or singular. Please advise.

packages/volto/locales/volto.pot
```diff
@@ -4690,6 +4706,21 @@ msgstr ""
...
+#: components/manage/Contents/ContentsDeleteModal
+msgid "link-integrity: Delete"
+msgstr ""
+
+#. Default: "Delete item and break links"
+#: components/manage/Contents/ContentsDeleteModal
+msgid "link-integrity: Delete item and break links"
+msgstr ""
+
+#. Default: "Checking references..."
```
stevepiercy, line 4719: Let's not introduce punctuation that might not be translatable.

```suggestion
#. Default: "Checking references"
```

stevepiercy (changes requested):
I only reviewed the message strings and defaults. This still requires a technical review. But it looks pretty good.

stevepiercy (approved):
Messages only approved. Still requires technical review.

sneridagh:
@giuliaghisini we merged #6509 so now importing from barrel files in core is not allowed anymore, so expect some ESlint failures. Could you please update the PR accordingly? Thanks!

packages/volto/.env.local
pnicolli, line 1: Remove this file (maybe we should gitignore *.local files? @sneridagh )

davisagli (approved):
@giuliaghisini Nice improvements, thank you.

PR #6517, opened 2024-12-06
Improve link integrity popup volto17

giuliaghisini:
backport of https://github.com/plone/volto/pull/6516/

fixed and improved link integrity popup when deleting item/items from folder contents. 
Moved all code in a separate component 'ContentsDeleteModal'.

Added loader when loading references. If the call to @linkintegrity took a long time to load, it wasn't clear to the user that it was checking for broken references:
 
loading

If only one item is selected to delete, it will display all the possibile broken references in a table, instead in a list. In the last column could be more than one reference: 
1-selected-item

If more than one item is selected to delete, it will not display the possibile broken references, but only the number, because it  will be a very long list (It worked like that before too):
more-than-one-selected

----
📚 Documentation preview 📚: https://volto--6517.org.readthedocs.build/

### DIFF
--- a/locales/ca/LC_MESSAGES/volto.po
+++ b/locales/ca/LC_MESSAGES/volto.po
@@ -531,6 +531,7 @@ msgstr "No es pot editar el disseny per al tipus de contingut <strong>{type}</st
 
 #. Default: "Cancel"
 #: components/manage/Add/Add
+#: components/manage/Contents/ContentsDeleteModal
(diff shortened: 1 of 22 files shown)


sneridagh (approved):

sneridagh:
@plone/volto-team approved, since it does not break the policy, since it applies to the CMSUI.

pnicolli (approved):

PR #6523, opened 2024-12-08
update: skip scroll to top on inital SSR completion.

### LINKED ISSUE #6518: Page scrolls to top after initial load.
If I start to load a page (initial load via SSR) and start to scroll down, the page scrolls back to the top when it is done loading, which interrupts me. I guess this comes from the ScrollToTop component connected to the router. This is okay for navigating after the initial load (with some exceptions like #3662) but we should try to avoid it for the initial load.

Faakhir30:

-----

Closes #6518 , skipping scroll to top on first `componenetDidUpdate` after SSR, as that is being triggered by SSR completion and is causing that intruption


stevepiercy (approved):
News looks good to me.

Needs a technical review from @plone/volto-accessibility and @plone/volto-team.

Thank you!

packages/volto/news/6523.internal
stevepiercy, line 1: Whoops, reviewed too quickly. This file should be renamed to `6523.bugfix`.

stevepiercy (changes requested):

packages/volto/src/helpers/ScrollToTop/ScrollToTop.jsx
```diff
@@ -22,12 +28,30 @@ class ScrollToTop extends React.Component {
     children: PropTypes.node.isRequired,
   };
 
+  /**
+   * Used to indicate client-side rendering
+   * @memberof ScrollToTop
+   */
+  componentDidMount() {
+    this.isClientSide = true;
```
avoinea, line 36: @Faakhir30 I think the convention for this variable name is `isClient` and maybe you should use the component state for it :thinking: 

See for example: https://github.com/plone/volto/blob/main/packages/volto/src/components/theme/View/View.jsx#L114

avoinea (changes requested):

packages/volto/src/helpers/ScrollToTop/ScrollToTop.jsx
Faakhir30: @avoinea Thanks for sharing the sample reference. I did think, IMO, that `isClient` wasn't required in my case, so I removed that variable in my recent commit. The reason is that `componentDidUpdate` never gets triggered during SSR; it is only invoked on the client side.

packages/volto/src/helpers/ScrollToTop/ScrollToTop.jsx
```diff
@@ -22,12 +22,21 @@ class ScrollToTop extends React.Component {
...
    * @returns {null} Null
    * @memberof ScrollToTop
    */
   componentDidUpdate(prevProps) {
+    // Skip the first client-side update that happens right after hydration
+    if (this.state.isFirstClientUpdatePending) {
+      this.setState({ isFirstClientUpdatePending: false });
+      return;
+    }
+
```
davisagli, line 39: @Faakhir30 We can do this a bit more simply without using component state. Instead, add a check for `&& prevProps.location !== undefined` after line 41. Could you make this change please?

davisagli (changes requested):

davisagli (approved):
Thanks, looks good to me now.

PR #6527, opened 2024-12-10
Add missing styleWrapper style builder information in container/Grid

sneridagh:
This PR fixes the missing style information for inner container/grid blocks.
Adds a schemaEnhancer for disabling Teaser align handlers for inner container teaser elements.

### DIFF
--- a/packages/volto/src/components/manage/Blocks/Teaser/schema.js
+++ b/packages/volto/src/components/manage/Blocks/Teaser/schema.js
@@ -127,3 +127,15 @@ export const gridTeaserDisableStylingSchema = ({ schema, formData, intl }) => {
   schema.fieldsets = schema.fieldsets.filter((item) => item.id !== 'styling');
   return schema;
 };
+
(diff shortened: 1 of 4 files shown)


packages/volto/src/components/manage/Blocks/Teaser/schema.js
```diff
@@ -127,3 +127,15 @@ export const gridTeaserDisableStylingSchema = ({ schema, formData, intl }) => {
...
+  schema,
+  formData,
+  intl,
+}) => {
+  schema.properties.styles.schema.fieldsets[0].fields =
+    schema.properties.styles.schema.fieldsets[0].fields.filter(
+      (item) => !['align'].includes(item),
+    );
+  return schema;
+};
```
davisagli, line 141: @sneridagh Why is this here, but not used?

sneridagh: Do you think this should be only in VLT?

davisagli (approved):

PR #6530, opened 2024-12-11
Fix some NL translation typos.   

fredvd:
-  'lijt' typo stands out too much on the 404 page translation.  
-  Use werkkopie consistently.


fredvd:
That was strange, were did my previous PR commits come from..  :-S.  another try.

PR #6536, opened 2024-12-12
`tsconfig` centralization + Update RAC everywhere

sneridagh:


packages/blocks/news/6536.internal
```diff
@@ -0,0 +1 @@
+Update RAC to latest. @sneridagh
```
pnicolli, line 1: This is not what changed in `blocks` :D it's the tsconfig

sneridagh: Fixed!

PR #6551, opened 2024-12-19
Show preview image in folder content on hover

### LINKED ISSUE #6550: Show Preview Image in folder contents view
There are no preview images for images in the folder contents, which makes it very difficult to decide which image is the right image.
There must be preview images in the folder contents.

Screenshot 2024-12-19 at 7 35 18 PM

iRohitSingh:
Fix #6550 

https://github.com/user-attachments/assets/5f3d0443-0693-459e-88dc-031669a0e664

### DIFF
--- a/packages/volto/src/components/manage/Contents/ContentsItem.jsx
+++ b/packages/volto/src/components/manage/Contents/ContentsItem.jsx
@@ -4,7 +4,13 @@
  */
 
 import React from 'react';
-import { Button, Table, Menu, Divider } from 'semantic-ui-react';
+import {
+  Button,
+  Table,
+  Menu,
+  Divider,
+  Popup as SemanticUiPopup,
+} from 'semantic-ui-react';
 import { Link } from 'react-router-dom';
 import PropTypes from 'prop-types';
 import map from 'lodash/map';
@@ -97,6 +103,8 @@ export const ContentsItemComponent = ({
   order,
 }) => {
   const intl = useIntl();
+  const previewImageUrl = `${item['@id']}/${item.image_scales?.image?.[0]?.scales?.teaser?.download}`;
+  const previewIconUrl = `${item['@id']}/${item.image_scales?.image?.[0]?.scales?.thumb?.download}`;
 
   return connectDropTarget(
     connectDragPreview(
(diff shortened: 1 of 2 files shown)


tisto:
@iRohitSingh the implementation looks good on the video. I have a few questions regarding the implementation:

- did you test this for a folder contents view with 100 large images to see how the current implementation scales? I recall that @robgietema had put quite some effort into making the "image grid" option for the right content sidebar scale properly for our DLR project. Maybe he can point you to the implementation.
- image on hover: we should lazy load those images or load the images only on hover to avoid long loading times for the folder contents view
- image without hover: there was a PR some time ago from EEA (@avoinea, @tiberiuichim, @ichim-david?) for this feature. We did not enable or merge it because of loading performance concerns. We should try to look up that PR/discussion. In any case, if we load all images, we should at least use a smaller image size.

packages/volto/src/components/manage/Contents/ContentsItem.jsx
```diff
@@ -97,6 +103,8 @@ export const ContentsItemComponent = ({
   order,
 }) => {
   const intl = useIntl();
+  const previewImageUrl = `${item['@id']}/${item.image_scales?.image?.[0]?.scales?.teaser?.download}`;
```
ichim-david, line 106: @iRohitSingh this lookup should be added in a new component that can be placed in this file
which is called if the item type is image. Right now you would attempt to check this for everything
even if there is no need for it.

packages/volto/src/components/manage/Contents/ContentsItem.jsx
```diff
@@ -156,14 +164,41 @@ export const ContentsItemComponent = ({
...
             <div className="expire-align">
-              <Icon
-                name={getContentIcon(item['@type'], item.is_folderish)}
-                size="20px"
-                className="icon-margin"
-                color="#878f93"
-                title={item['Type'] || item['@type']}
-              />{' '}
-              <span title={item.title}> {item.title}</span>
+              {item['@type'] === 'Image' ? (
```
ichim-david, line 167: All of @tisto advice is quite valid, lazy loaded when needed, you don't want to load 100 images if you never even interact with them.

ichim-david (changes requested):
@iRohitSingh I left some concerns as comments in the code.

DuaaScripts (approved):

packages/volto/src/components/manage/Contents/ContentsItem.jsx
iRohitSingh: done.

PR #6556, opened 2024-12-24
Newfeat 17 hide show blocks

SaraBianchi:
New feature already explained here #5157 (I'm opening a PLIP about this).

### Some improvements made compared to the old version:
- Accessibility of opacity text, now the color contrast passes
Screenshot 2024-12-23 alle 11 50 00

- The "show/hide" button has been positioned at the top center of the block, when the block is hidden, the focus on the block has a dotted border to better identify that it has been hidden
Screenshot 2024-12-24 alle 15 05 55

- checked layout also on mobile
Screenshot 2024-12-23 alle 12 32 39

### New feature added
- an on/off switch in the left side toolbar to hide or show blocks preview from the page, hidden ones are collapsed, if there are no hidden blocks the switch is disabled.
https://github.com/user-attachments/assets/8027ed8e-e6d3-47be-80cc-74208feb0010

### Only sore point
- When the switch is activated, when a block is hidden and the next one is deleted by the keyboard, the focus is positioned on the collapsed one, if it is a slate text block it is possible to edit/delete it even while the content is hidden, it is necessary to expand it or exit with another key combination if you are not using the mouse.
A subsequent improvement could be to tell the Slate text field to bypass hidden blocks if you use the keyboard only when they are collapsed.


SaraBianchi:
I had to change the position of the eye icon because we noticed that in some blocks with a central toolbar it was not possible to click the button below.

Screenshot 2025-01-08 alle 23 06 24

I standardized the block toolbars (html, grid, table) to always have them centered and with information tooltips.

Screenshot 2025-01-09 alle 00 22 51

I also changed the eye icon in the preview of the HTML block because it could be misleading with the show/hide button.

Screenshot 2025-01-08 alle 23 59 40

Also added a configuration in the config to activate or not the functionality, by default it is `true` but we can set it to false.

`experimental: {
    hideShowBlocks: {
      enabled: true,
    },
  },`

PR #6562, opened 2025-01-02
Stop generating sitemap.xml.gz (#6561)

### LINKED ISSUE #6561: Do not deliver the old sitemap.xml.gz
We generate sitemap-index.xml which is also put correctly into robots.txt. However we still provide the old sitemap.xml.gz.

But, while this serves no purpose in addition to the index, it can cause problems if Google somehow parses it (for example it's submitted to the Google Search console). For this reason, we should stop providing sitemap.xml.gz.

reebalazs:
We generate sitemap-index.xml which is also put correctly into robots.txt. However we still provide the old sitemap.xml.gz.

But, while this serves no purpose in addition to the index, it can cause problems if Google somehow parses it (for example it's submitted to the Google Search console). For this reason, we  stop providing sitemap.xml.gz.

- [X] I succesfully ran [code linting checks](https://6.docs.plone.org/volto/contributing/linting.html) on my changes locally.
- [X] I succesfully ran [unit tests](https://6.docs.plone.org/volto/contributing/testing.html) on my changes locally.
- [X] I succesfully ran [acceptance tests](https://6.docs.plone.org/volto/contributing/acceptance-tests.html) on my changes locally.

-----

Closes #6561

### DIFF
--- a/packages/volto/news/6561.bugfix
+++ b/packages/volto/news/6561.bugfix
@@ -1,1 +1,1 @@
-Stop generating sitemap.xml.gz @reebalazs
+Serve `sitemap.xml.gz` as an alias for `sitemap-index.xml`. @reebalazs
--- a/packages/volto/news/6561.bugfix
+++ b/packages/volto/news/6561.bugfix
@@ -1,0 +1,1 @@
+Stop generating sitemap.xml.gz @reebalazs
--- a/packages/volto/src/express-middleware/sitemap.js
+++ b/packages/volto/src/express-middleware/sitemap.js
@@ -47,11 +47,24 @@ export const sitemapIndex = function (req, res, next) {
   });
 };
 
+export const sitemapIndexCompatibility = function (req, res, next) {
+  generateSitemapIndex(req, true).then((sitemapIndex) => {
+    res.set('Content-Type', 'application/x-gzip');
+    res.set('Content-Disposition', 'attachment; filename="sitemap.xml.gz"');
+    res.send(sitemapIndex);
+  });
+};
+
 export default function sitemapMiddleware() {
   const middleware = express.Router();
 
-  middleware.all('**/sitemap.xml.gz', sitemap);
+  // For backwards compatibility, and allow a graceful transition for
+  // sites that are already set up on the Google Search Console, we continue delivering
+  // the new batched sitemap under the old sitemap.xml.gz name.
+  middleware.all('**/sitemap.xml.gz', sitemapIndexCompatibility);
   middleware.all('**/sitemap:batch.xml.gz', sitemap);
+  // For new setups, `sitemap-index.xml` should be added to the
+  // Google Search Console.
   middleware.all('**/sitemap-index.xml', sitemapIndex);
   middleware.id = 'sitemap.xml.gz';
   return middleware;
--- a/packages/volto/src/helpers/Sitemap/Sitemap.js
+++ b/packages/volto/src/helpers/Sitemap/Sitemap.js
@@ -61,7 +61,7 @@ export const generateSitemap = (_req, start = 0, size = undefined) =>
  * @param {Object} _req Request object
  * @return {string} Generated sitemap index
  */
-export const generateSitemapIndex = (_req) =>
+export const generateSitemapIndex = (_req, gzip = false) =>
   new Promise((resolve) => {
     const { settings } = config;
     const APISUFIX = settings.legacyTraverse ? '' : '/++api++';
@@ -88,7 +88,14 @@ export const generateSitemapIndex = (_req) =>
         const result = `<?xml version="1.0" encoding="UTF-8"?>
 <sitemapindex xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
 ${items.join('\n')}\n</sitemapindex>`;
-        resolve(result);
+
+        if (gzip) {
+          zlib.gzip(Buffer.from(result, 'utf8'), (_err, buffer) => {
+            resolve(buffer);
+          });
+        } else {
+          resolve(result);
+        }
       }
     });
   });


sneridagh (approved):
@plone/volto-team anybody else feedback?

ichim-david:
@sneridagh I asked at Edw and at least as a preliminary reply was that it doesn't affect us as we check also for  sitemap-index.xml. So at least for us it seems fine. Waiting to see what the other companies have to say about this but in the meanwhile I wanted to write to comment to know that we've discussed about this and we are ok with these changes

erral:
I would add a redirect from `sitemap.xml.gz` to `sitemap-index.xml.gz` to gracefully handle the change.

reebalazs:
> I would add a redirect from `sitemap.xml.gz` to `sitemap-index.xml.gz` to gracefully handle the change.

I think that a redirect, while technically possible, is problematic because `sitemap-index.xml` is not gzipped and the original one is. So we would serve a file ending with `.gz` but actually not gzipped. Whether or not google eats this or not, is not possible to find out from the documentation. We could figure out experimentally... but then what if they change things.

So if we don't redirect, then Google either finds the sitemap from robots.txt. or it has to be changed once in the console, but it's explicit and there is no magic behind that might break things in the future.

 (The reason for it being not gzipped is because of following the google documentation. It seemed simpler to do the same then to start experimenting what can work and what not.)

erral:
Sorry I have written it incorrectly, I meant a redirect between `sitemap.xml.gz` and `sitemap-index.xml`

sneridagh:
@erral Maybe we are not talking about the same, but if we create a redirect from `sitemap.xml.gz` to `sitemap-index.xml` it can bring some other problems, starting with the fact that it's not a `.gz` file. 

@reebalazs I can understand that people that have their consoles pointing to the old, can have problems and will require action, so it's not an easy one (and then, it would require to add it in a breaking change).

The main problem here is that we are providing pointers to both... and could lead to duplicated indexes. Not an easy one. We'll have to test if a issuing a 301 would work for the console.

erral:
Then, if we are removing an existing URL, I think it should be considered a breaking change.

reebalazs:
> @erral Maybe we are not talking about the same, but if we create a redirect from `sitemap.xml.gz` to `sitemap-index.xml` it can bring some other problems, starting with the fact that it's not a `.gz` file.
> 
> @reebalazs I can understand that people that have their consoles pointing to the old, can have problems and will require action, so it's not an easy one (and then, it would require to add it in a breaking change).
> 
> The main problem here is that we are providing pointers to both... and could lead to duplicated indexes. Not an easy one. We'll have to test if a issuing a 301 would work for the console and make sure we issue the correct mimetype too.

The problems are

1. the duplication
2. the old one is only working until a site has grown as big that Google will reject the sitemap.

erral:
If we have hit a maximum number of URLs that Google can consider in a single `sitemap.xml.gz` we can set a limit in the URLs generated in that file, and document that in the Upgrade guide.

This way, although still being a breaking change, the url will keep working, and those small sites that don't reach the limit can still be keep working.

We can set the limit to the limit you have set to build each of the files referenced in the `sitemap-index.xml`

tisto:
@erral setting a limit is very dangerous, IMO, since the admins of a site won't get notified if they hit the limit. I agree that this is a breaking change. However, I would fix this once and for all with a breaking change release and just support the index sitemap. Anything else will lead to confusion and make the situation worse than it is already (we lost our entire SEO rankings for plone.de thanks to this problem; I don't want to imagine what would happen if that happened in a client project).

erral:
Don't get me wrong, I have also faced picky SEO things, and I understand how hard this is and how important is.

I understand that if we have hit the limit we need to provide a way to fix it, and we have done so using the index thing.

I only say that just removing the sitemap.xml.gz can be a "hard" thing, so to say. That's why I proposed 2 possible solutions (redirect and limit the amount of the URLs).

If none of them is valid, or if the solution is worst than the problem, fair enough. I just wanted to raise the concern of removing a URL.

reekitconcept:
> Don't get me wrong, I have also faced picky SEO things, and I understand how hard and important is.
> 
> I understand that if we have hit the limit we need to provide a way to fix it, and we have done so using the index thing.
> 
> I only say that just removing the sitemap.xml.gz can be a "hard" thing, so to say. That's why I proposed 2 possible solutions (redirect and limit the amount of the URLs).
> 
> If none of them is valid, or if the solution is worst than the problem, fair enough. I just wanted to raise the concern of removing a URL.

Ok so if we want to do the redirect, we have to test it out. There is no other way.

- Does Google follow the redirect at all?
- Is it a problem for Google that we redirect a `.gz` file to an unzipped one, iow does Google process the file that ends with `.gz` but it's actually a plain xml?

If the answer to these is true then yes we can do the redirect. If no then no it does not work.

Does anyone have the capacity to test this out? I'm flooded right now so I can't make any promises.

reebalazs:
> Don't get me wrong, I have also faced picky SEO things, and I understand how hard and important is.
> 
> I understand that if we have hit the limit we need to provide a way to fix it, and we have done so using the index thing.
> 
> I only say that just removing the sitemap.xml.gz can be a "hard" thing, so to say. That's why I proposed 2 possible solutions (redirect and limit the amount of the URLs).
> 
> If none of them is valid, or if the solution is worst than the problem, fair enough. I just wanted to raise the concern of removing a URL.

Ok so if we want to do the redirect, we have to test it out. There is no other way.

- Does Google follow the redirect at all?
- Is it a problem for Google that we redirect a .gz file to an unzipped one, iow does Google process the file that ends with .gz but it's actually a plain xml?

If the answer to these is true then yes, we can do the redirect. If no, then no it does not work.

Does anyone have the capacity to test this out with the Google console, see if a file is actually picked up? I'm flooded right now so I can't make any promises.

sneridagh:
We have to document the issues `sitemap-index.html` / `sitemap.xml.gz` nuances.

stevepiercy:
> We have to document the issues `sitemap-index.html` / `sitemap.xml.gz` nuances.

Where should the docs go? I'd say at least a mention in the Upgrade Guide. Is any of this configurable or changeable? If so, then it's probably a how to guide, else it's probably explanation.

erral:
After talking with the guy that handles Search Console in our company, the solution may be easier than we thought:

The `sitemap-index.xml` file it's in itself a sitemap file, so instead of creating a new URL with the index, we can use the `sitemap.xml.gz` URL and serve the index file there.

This way, we don't need to change anything, we just change the contents of the `sitemap.xml.gz`.

My colleague says it doesn't matter to serve the file compress or uncompressed, he even suggests to compress every sitemap file, even the index one.

reebalazs:
> After talking with the guy that handles Search Console in our company, the solution may be easier than we thought:
> 
> The `sitemap-index.xml` file it's in itself a sitemap file, so instead of creating a new URL with the index, we can use the `sitemap.xml.gz` URL and serve the index file there.
> 
> This way, we don't need to change anything, we just change the contents of the `sitemap.xml.gz`.
> 
> My colleague says it doesn't matter to serve the file compress or uncompressed, he even suggests to compress every sitemap file, even the index one.

That should be pretty easy then. I also suspected that Google does not care if the file is compressed or uncompressed.

Still for future compatibility I suggest:

- We continue serving `sitemap-index.xml`, and we also put it to robot.txt just like in the current PR
- We add back serving `sitemap.xml.gz` but we actually serve the same content as `sitemap-index.xml` with this name.

That way every new site setup would see `sitemap-index.xml` which indicates that this is, in fact, a batched index. Also this is available for picking up from robots.txt. But all old deployments will continue to work even if the old `sitemap.xml.gz` is added to the console.

erral:
We have tested this approach (serve the index file gzipped in the sitemap.xml.gz file) in 2 sites of ours and Search Console reports that the files are OK and the reported number of URLs is correct.

reebalazs:
> We have tested this approach (serve the index file gzipped in the sitemap.xml.gz file) in 2 sites of ours and Search Console reports that the files are OK and the reported number of URLs is correct.

@erral I've added this, can you please test it out and approve? Thank you!

erral (changes requested):
I have tested it locally before deploying it, and I see that the `sitemap.xml.gz` contents are not gzipped, they directly contain the index file.

erral:
Mmm, strange.

The headers mark that the content is sent gzipped, which it happens, but the transferred file is not a gzipped file, but a text/plain:

```
erral@lindari:/tmp$ wget -S http://localhost:3000/sitemap.xml.gz
--2025-02-11 10:46:09--  http://localhost:3000/sitemap.xml.gz
Resolving localhost (localhost)... 127.0.0.1
Connecting to localhost (localhost)|127.0.0.1|:3000... connected.
HTTP request sent, awaiting response... 
  HTTP/1.1 200 OK
  Content-Type: application/x-gzip; charset=utf-8
  Content-Disposition: attachment; filename="sitemap.xml.gz"
  Content-Length: 199
  ETag: W/"c7-Jl9aJBoMZB2Ag5+54aah1NO77Sk"
  Date: Tue, 11 Feb 2025 09:46:09 GMT
  Connection: keep-alive
  Keep-Alive: timeout=5
Length: 199 [application/x-gzip]
Saving to: 'sitemap.xml.gz'

sitemap.xml.gz                              100%[========================================================================================>]     199  --.-KB/s    in 0s      

2025-02-11 10:46:09 (20.7 MB/s) - 'sitemap.xml.gz' saved [199/199]

```

And then:  

```
erral@lindari:/tmp$ more sitemap.xml.gz 
<?xml version="1.0" encoding="UTF-8"?>
<sitemapindex xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
  <sitemap>
    <loc>http://localhost:3000/sitemap1.xml.gz</loc>
  </sitemap>
</sitemapindex>
```

I would expect that not only the transfer is gzipped, but the file itself also, right?

This is the example with the current https://demo.plone.org/sitemap.xml.gz

```
erral@lindari:/tmp$ wget https://demo.plone.org/sitemap.xml.gz
--2025-02-11 10:47:24--  https://demo.plone.org/sitemap.xml.gz
Resolving demo.plone.org (demo.plone.org)... 104.25.83.118, 172.67.80.190, 104.25.84.118, ...
Connecting to demo.plone.org (demo.plone.org)|104.25.83.118|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: 927 [application/x-gzip]
Saving to: 'sitemap.xml.gz'

sitemap.xml.gz                              100%[========================================================================================>]     927  --.-KB/s    in 0s      

2025-02-11 10:47:24 (9.97 MB/s) - 'sitemap.xml.gz' saved [927/927]

```

And the contents:
```
erral@lindari:/tmp$ more sitemap.xml.gz 
�o�0���+P���m 	Oiz۩;-�v�<p�7�v��O��
aFQ���ы�H���`t��;�qg�1
�y#3�����<ԪhX�y�x�.I\������
��b��+��Y6B��	����:��W�׉��f�ׁ�Hp�sP�,ұ��x��KV�������5��$A<(g�E���� YN�L5Zp~DJ�x��A����b�(şH�Y�
R��O
erral@lindari:/tmp$ 
```

reebalazs:
@erral, sorry, my bad. I'll ping you when I've updated it.

reebalazs:
@erral please check it out now!

erral (approved):
Yes, now it works as expected

packages/volto/news/6561.bugfix
```diff
@@ -0,0 +1 @@
+Stop generating sitemap.xml.gz @reebalazs
```
davisagli, line 1:
```suggestion
Serve `sitemap.xml.gz` as an alias for `sitemap-index.xml`. @reebalazs
```

davisagli (approved):
LGTM

PR #6576, opened 2025-01-10
Improve the usability of the ObjectBrowser when inputting a manual value, checking it on blur, and adding a local validator

sneridagh:
Instead of requiring to press enter to validate the value, it does it on blur too.
Also, moved to use the URL validator from the main Validators.

### DIFF
--- a/packages/volto/news/6576.bugfix
+++ b/packages/volto/news/6576.bugfix
@@ -1,1 +1,1 @@
-Improve the usability of the ObjectBrowser when inputting a manual value, checking it on blur, and adding a local validator. @sneridagh
+Improve the usability of the `ObjectBrowser` when inputting a manual value, checking it on blur, and adding a local validator. @sneridagh
--- a/packages/volto/news/6576.bugfix
+++ b/packages/volto/news/6576.bugfix
@@ -1,0 +1,1 @@
+Improve the usability of the ObjectBrowser when inputting a manual value, checking it on blur, and adding a local validator. @sneridagh
--- a/packages/volto/src/components/manage/Widgets/ObjectBrowserWidget.jsx
+++ b/packages/volto/src/components/manage/Widgets/ObjectBrowserWidget.jsx
@@ -16,10 +16,10 @@ import { Image, Label, Popup, Button } from 'semantic-ui-react';
 import {
   flattenToAppURL,
   isInternalURL,
-  isUrl,
   normalizeUrl,
   removeProtocol,
 } from '@plone/volto/helpers/Url/Url';
+import { urlValidator } from '@plone/volto/helpers/FormValidation/validators';
 import { searchContent } from '@plone/volto/actions/search/search';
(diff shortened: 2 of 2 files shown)


packages/volto/src/components/manage/Widgets/ObjectBrowserWidget.jsx
```diff
@@ -341,9 +351,14 @@ export class ObjectBrowserWidgetComponent extends Component {
...
 
+    const props = {
+      ...this.props,
+      error: this.props.error.concat(this.state.errors),
+    };
+
     return (
       <FormFieldWrapper
-        {...this.props}
+        {...props}
```
davisagli, line 361: It's confusing to have a thing called `props` which is more than just the props. Can't you just pass `error` separately here?

sneridagh: @davisagli check now it's better.

Thanks to the errors I've also found out that the OBW errors is messed up because how we use it as a HOC in some places :/ The errors shape switches back and forth from `{}` to `[]`... We didn't realise because we never validated what's inside. Now we are dealing with a little validation inside, and since we do not handle outside validation, let's simplify it and do not take that into consideration. We can do it if we enable external validation in this widget at some point.

sneridagh:
@plone/volto-team please a look on this one?

packages/volto/news/6576.bugfix
```diff
@@ -0,0 +1 @@
+Improve the usability of the ObjectBrowser when inputting a manual value, checking it on blur, and adding a local validator. @sneridagh
```
stevepiercy, line 1:
```suggestion
Improve the usability of the `ObjectBrowser` when inputting a manual value, checking it on blur, and adding a local validator. @sneridagh
```

stevepiercy (approved):
News is OK. Needs technical review.

packages/volto/src/components/manage/Widgets/ObjectBrowserWidget.jsx
```diff
@@ -230,7 +231,16 @@ export class ObjectBrowserWidgetComponent extends Component {
...
     if (this.props.allowExternals) {
-      return isUrl(url);
+      const error = urlValidator({
+        value: url,
+        formatMessage: this.props.intl.formatMessage,
+      });
+      if (error && url !== '') {
+        this.setState({ errors: [error] });
+      } else {
+        this.setState({ errors: [] });
```
ichim-david, line 241: @sneridagh isn't this redundant due to the initial empty value at line 105?
https://github.com/plone/volto/pull/6576/files#diff-526a68594a32e6dd3a09f1243f7780736303aae2414bb98f8670cbdbc5f71703R105
Technically speaking react should know not to trigger a state chance if it's the same value but I 
still think this is unnecessary

sneridagh: We could reset the state at the top of the method too.

davisagli (approved):

PR #6557, opened 2024-12-25
added swedish translation

TechSubham:
- [ ✅] I signed and returned the [Plone Contributor Agreement](https://plone.org/foundation/contributors-agreement), and received and accepted an invitation to join a team in the Plone GitHub organization.
- [ ✅] I verified there aren't other open [pull requests](https://github.com/plone/volto/pulls) for the same change.
- [ ✅] I followed the guidelines in [Contributing to Volto](https://6.docs.plone.org/volto/contributing/index.html).
- [ ✅] I succesfully ran [code linting checks](https://6.docs.plone.org/volto/contributing/linting.html) on my changes locally.
- [ ✅] I succesfully ran [unit tests](https://6.docs.plone.org/volto/contributing/testing.html) on my changes locally.
- [ ✅] I succesfully ran [acceptance tests](https://6.docs.plone.org/volto/contributing/acceptance-tests.html) on my changes locally.
- [ ] If needed, I added new tests for my changes.
- [ ] If needed, I added [documentation](https://6.docs.plone.org/volto/contributing/documentation.html#narrative-documentation) for my changes, either in the Storybook or narrative documentation.
- [ ✅ ] I included a [change log entry](https://6.docs.plone.org/contributing/index.html#contributing-change-log-label) in my commits.

Closes #5957

Issue #5957, opened 2024-04-14
Add Swedish translation


AndreasJonsson:
**Is your feature request related to a problem? Please describe.**
Swedish translation missing.

**Describe the solution you'd like**
Please, translation file in pull request.

stevepiercy:
@AndreasJonsson I apologize for missing this pull request from so many months ago. Thank you for your work!

Our bot mr-roboto should have notified you to sign the [Plone Contributor Agreement](https://6.docs.plone.org/contributing/index.html#contributing-sign-and-return-the-plone-contributor-agreement-label), but did not. We require that of all contributors. Would you please sign that at your earliest convenience.

Reopening to resolve properly and give credit where it is due.

stevepiercy:
Whoops, wrong browser tab. Closing again.

The PR that we closed is at https://github.com/plone/volto/pull/5958 due to no response from the @AndreasJonsson after repeated attempts.

AndreasJonsson:
> Whoops, wrong browser tab. Closing again.
> 
> The PR that we closed is at #5958 due to no response from the @AndreasJonsson after repeated attempts.

Sorry about that! I will get back to you if I eventually sign the agreement.

PR #6578, opened 2025-01-10
Revert "added swedish translation"

sneridagh:
Reverts plone/volto#6557

Reverting this PR, under the suspicion that this PR was a copy of the https://github.com/plone/volto/issues/5957
reported by @AndreasJonsson on 10/01/25. We will open an investigation to clarify the details.


stevepiercy (approved):
Approved pending outcome of investigation.

PR #6586, opened 2025-01-13
Fix slate RichTextwidget to allow support slate extensions refs#6570

### LINKED ISSUE #6570: Allow passing slate extensions to RichTextWidget
**Describe the bug**
The text block configuration in the core can be extended using textBlockExtensions. However, we don't apply this to the RichtextWidget because, in the past, we faced issues with how it handled unordered lists (ul) and list items (li).

**To Reproduce**

1. Try adding a richText widget on a field and add a `ul` with some `li`'s. 
2. It seems you can't go out of lists, also on shift+Enter.

**Expected behavior**
The behaviour should be same as textlock where shift+Enter should go on a newline `\n`, Enter on empty li should come out of list. 

- Volto Version - 18.x also on 17.x

TextBlock: https://github.com/plone/volto/blob/e55b4a65170f0a2038d9d1db994748bca9f72f95/packages/volto-slate/src/blocks/Text/DefaultTextBlockEditor.jsx#L73

https://github.com/plone/volto/blob/e55b4a65170f0a2038d9d1db994748bca9f72f95/packages/volto-slate/src/blocks/Text/index.jsx#L41

RichTextWidget: https://github.com/plone/volto/blob/e55b4a65170f0a2038d9d1db994748bca9f72f95/packages/volto-slate/src/widgets/RichTextWidget.jsx#L63

nileshgulia1:
Closes #6570


nileshgulia1:
@davisagli Its something which we are also doing it for textBlock. slate-extensions just adds extra capabilities to slate wrapped components. To test this, you can create a RichText widget in your content type, try to create some lists(ul) with few li's, save the page and re-edit them, you see there is no way to get outside of lists, or actually there is no way to add anything else. 
I've created a sample here(on edit ): https://demo.plone.org/test-richtext

Perhaps I should create some E2E tests also 🤔

nileshgulia1:
There is a detached mode also but I think its for blocks not widgets.

stevepiercy:
@nileshgulia1 I don't follow. I can use the up and down arrows to move between blocks.


https://github.com/user-attachments/assets/5d70c9bc-0b92-4ac4-b78e-e80d9ac66ce8

davisagli (changes requested):
@nileshgulia1 @stevepiercy I think I understand the problem better now:
- This is specifically a problem with the rich text widget, not the slate block. To test it, you have to add a rich text field to a content type, and edit there, not in a block.
- The problem is that when you're typing in a list, you would expect to be able to hit Enter a couple times to end the list. But this just creates new list items, and there's no obvious way to add text that is not in the list.

Do I have it right? If so, I'd suggest a different changelog entry here: "Fix typing Enter to end a list in the RichTextWidget." Using a slate extension for this is just an implementation detail, not the point of the change.

Unfortunately, I tested with this PR locally and it doesn't actually seem to fix the problem.

davisagli:
@nileshgulia1 Your video looks like what I would expect! I'll try to step through in the debugger in my local site later this week and see if I can tell why it didn't work for me.

packages/volto-slate/src/widgets/HtmlSlateWidget.jsx
```diff
@@ -127,7 +132,10 @@ const HtmlSlateWidget = (props) => {
             block={block}
             selected={selected}
             properties={properties}
+            extensions={slateWidgetExtensions}
```
nileshgulia1, line 135: I fixed the issue with inconsistency,  htmlSlateWidget is used as `richtext` and we override this in our policy adddon. Thus, the same extension should be ported here as well.

I will write the cypress for this scenario.

nileshgulia1: -cc @davisagli

packages/volto-slate/news/6570.bugfix
```diff
@@ -0,0 +1 @@
+fix: add extensions in RichtextWidget @nileshgulia1
```
davisagli, line 1:
```suggestion
In `RichTextWidget` and `HtmlSlateWidget`, fix breaking a list by typing Enter. @nileshgulia1
```

davisagli (approved):
@nileshgulia1 Thanks, now it does what I expected!

nileshgulia1:
@sneridagh We can merge this please. Thank you.

packages/volto/news/6570.internal
davisagli, line 1: @nileshgulia1 Use `.internal` for changes that only involve tests. I updated it already.

nileshgulia1:
@davisagli Perhaps this is a candidate for volto 17.x backport?

nileshgulia1:
@davisagli Done, Pls check: https://github.com/plone/volto/pull/6743

PR #6597, opened 2025-01-16
Fix: a11y - edit page - add block/blocks-chooser

### LINKED ISSUE #5212: a11y - edit page - add block/blocks-chooser - cms ui
**Describe the bug**
Only the "most used" accordion is accessible (though with an invisible focus), while the other accordions are not. Besides that, after the 'last tab,' in this case when pressing tab on the 'system' item, the focus is completely lost. (see video)

**To Reproduce**

1- Go to https://demo.plone.org/edit
2- enter in edit mode by clicking on the edit button (top-left)
3- Press enter to create a new content
4- Click on the "add block" icon
5- Navigate using tab key

**Expected behavior**
After 'system,' the focus should move to the next accordion. Here, I see two possibilities:
1- The first accordion closes, the second one opens, and the focus goes to the first item of the second accordion.
2- The focus remains outside of the second accordion, on the accordion header, with the accordion closed. It opens only if the user wants. In this case, the next tab goes to the next accordion header.

**Screenshots**
https://github.com/plone/volto/assets/60133113/9ae96eb0-57ab-454b-8f92-a417709b1846

Manas-Kenge:
- [ ] I succesfully ran [acceptance tests](https://6.docs.plone.org/volto/contributing/acceptance-tests.html) on my changes locally.

-----

Closes #5212 

This pull request fixes accessibility in the "Add Block" modal. Focus now stays inside the modal, pressing Enter expands the accordion and moves focus inside, while ESC closes it and returns focus to the header.

### DIFF
--- a/packages/volto/news/5212.bugfix
+++ b/packages/volto/news/5212.bugfix
@@ -1,0 +1,2 @@
+Fixed accessibility issues in the "Add Block" modal of the Plone editor. Focus now stays inside the modal while navigating.
+@Manas-Kenge
--- a/packages/volto/news/5212.bugfix
+++ b/packages/volto/news/5212.bugfix
@@ -1,2 +1,1 @@
-Fixed accessibility issues in the "Add Block" modal of the Plone editor. Focus now stays inside the modal while navigating.
-@Manas-Kenge
+Fixed accessibility issues in the "Add Block" modal of the editor where focus now stays inside the modal while navigating. @Manas-Kenge
--- a/packages/volto/src/components/manage/BlockChooser/BlockChooser.jsx
+++ b/packages/volto/src/components/manage/BlockChooser/BlockChooser.jsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState, useEffect, useRef } from 'react';
 import useUser from '@plone/volto/hooks/user/useUser';
 import PropTypes from 'prop-types';
 import filter from 'lodash/filter';
@@ -37,10 +37,12 @@ const BlockChooser = ({
   properties = {},
   navRoot,
   contentType,
+  onClose,
 }) => {
   const intl = useIntl();
   const user = useUser();
   const hasAllowedBlocks = !isEmpty(allowedBlocks);
+  const accordionRefs = useRef([]);
 
   const filteredBlocksConfig = filter(blocksConfig, (item) => {
     // Check if the block is well formed (has at least id and title)
@@ -76,7 +78,7 @@ const BlockChooser = ({
 
   let blocksAvailable = {};
   const mostUsedBlocks = filter(filteredBlocksConfig, (item) => item.mostUsed);
-  if (mostUsedBlocks) {
+  if (mostUsedBlocks.length) {
     blocksAvailable.mostUsed = mostUsedBlocks;
   }
   const groupedBlocks = groupBy(filteredBlocksConfig, (item) => item.group);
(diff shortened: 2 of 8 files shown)


Manas-Kenge:
@stevepiercy Can you review this?

packages/volto/news/5212.bugfix
```diff
@@ -0,0 +1,2 @@
+Fixed accessibility issues in the "Add Block" modal of the Plone editor. Focus now stays inside the modal while navigating.
```
stevepiercy, line 1: Minor tweak.

```suggestion
Fixed accessibility issues in the "Add Block" modal of the editor where focus now stays inside the modal while navigating. @Manas-Kenge
```

packages/volto/news/5212.bugfix
```diff
@@ -0,0 +1,2 @@
...
+@Manas-Kenge
```
stevepiercy, line 2: Move to same line.

```suggestion
```

stevepiercy (changes requested):
I only triaged this PR. It still needs technical review. I'll add the teams to the request.

stevepiercy (commented):
The news looks good. This PR still needs technical review.

JeffersonBledsoe:
Hi @Manas-Kenge, thanks for the work on this. While the work in this PR is definitely an improvement on the existing functionality and is much appreciated, I think the multiple problems noted with the blocks chooser noted in the linked issue need further discussion and some more definitive criteria for what a 'fixed' version of the chooser should be. Let's put this PR on hold while we get a clearer picture of what all the problems are and what the best way of solving them should be. cc @plone/volto-accessibility .

Notes I made while doing a quick test of this PR using VoiceOver on mac via Safari and Google Chrome:
- "Add block button" markup should reflect what the button is going to control and it's state. We could `aria-expanded` and `aria-controls` to associate the button with the picker. We also need to add `aria-haspopup="dialog"`.
- When using the search, we should use a live region to convey the results are changing (e.g. list the number of results or say there are no results).
- We should include a close button. This isn't a simple enhancement for an input picker (like a calendar selection for a date input) and so the 'ESC to close' shortcut shouldn't be the only thing we provide to close this dialog. Even alot of accessible date pickers provide a dedicated close button. This does not currently exist.
- When closing the picker without selecting something, focus should return to the "Add block" button. This does not currently work.
- When the picker is closed through the selection of a block, the focus should move to the new block. This PR has this behaviour.
- Some users may expect focus to return to the start of the picker when pressing 'TAB' after being on the last focusable element. This is common behaviour for forms within dialogs. This needs further research so isn't a "MUST" requirement.

Manas-Kenge:
Thank you, @JeffersonBledsoe, for the detailed review and thoughtful feedback.

Regarding the close button, I initially implemented an 'X' button for closing the blocks chooser but noticed it negatively impacted the UI, as it increased the height of the accordion more than necessary. To maintain a clean design, I decided to remove it. However, I understand the importance of accessibility and am happy to add it if the accessibility team decides that including a close button is essential.

As for using `aria-expanded`, `aria-controls`, `aria-haspopup` I will look into it.

I'll await further input from the accessibility team to ensure these changes are properly implemented. Please let me know how you’d like to proceed after discussing these points further.

Wagner3UB:
@wesleybl @Humanaice Hey guys, we need to see if we can continue this work. It would be a shame not to finish this PR, especially since most of the work is already done.

PR #6599, opened 2025-01-17
Initial push of the Plone 7 base package

sneridagh:
This is the initial (and very early) implementation of Plone 7.
After the design and first implementations of all the required pieces (the `@plone/*` libraries) that will compose Plone 7, this package will concentrate all the development during the next years.

It is based on [React Router](https://reactrouter.com/dev/docs) 7, using the `@plone/*` libraries.

Disclaimer:

> [!WARNING]
> This package and all the efforts around it are not even in an alpha state and are experimental.
> The naming is by no means binding whatsoever both in the package name or the package folder and may change to a definitive one in the future once details have been clarified and decided. 
> The community offers no support whatsoever for it.
> Breaking changes may occur without notice.


.github/workflows/changelog.yml
```diff
@@ -49,7 +49,7 @@ jobs:
             theming:
               - 'packages/theming/**'
             plone7:
-              - 'packages/revolto/**'
+              - 'packages/voltotng/**'
```
stevepiercy, line 52: How do you pronounce that? 😝 

I still like "DEVOlto" better. Are we not men? We are DEVOlto!

sneridagh: @stevepiercy tip: you don't, because you'll never need to pronounce it. ;) We are bad ass trekkies, at most :)

stevepiercy: @sneridagh looks like Klingon to me.

erral: PloneNG2 :laughing:

sneridagh: 😂

pnicolli (approved):

PR #6600, opened 2025-01-17
Route registry

sneridagh:
Added `registerRoute` API for the config. 
It allows to add a ReactRouter route to the configuration, that we can process and load later.
Quite simple, it only saves the configuration given a specific object.

### DIFF
--- a/packages/coresandbox/src/index.ts
+++ b/packages/coresandbox/src/index.ts
@@ -236,11 +236,16 @@ const applyConfig = (config: ConfigType) => {
   });
 
   config.registerRoute({
+    type: 'route',
     path: '/hello',
     file: 'src/components/Views/NewsAndEvents/asd.tsx',
-    {index: true},
     children: [],
+    options: {
+      id: 'hello',
+      index: true,
+    },
   });
+
   return config;
 };
 
--- a/packages/coresandbox/src/index.ts
+++ b/packages/coresandbox/src/index.ts
@@ -239,7 +239,6 @@ const applyConfig = (config: ConfigType) => {
     type: 'route',
     path: '/hello',
     file: 'src/components/Views/NewsAndEvents/asd.tsx',
-    children: [],
     options: {
       id: 'hello',
       index: true,
--- a/packages/coresandbox/src/index.ts
+++ b/packages/coresandbox/src/index.ts
@@ -235,6 +235,12 @@ const applyConfig = (config: ConfigType) => {
     predicates: [ContentTypeCondition(['Document']), RouteCondition('/hello')],
   });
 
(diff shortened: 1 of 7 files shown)


davisagli (commented):
@sneridagh No documentation?

stevepiercy:
@sneridagh if docs are needed—and I think they are needed, especially for an API—then at least create a new issue referencing this PR. Otherwise it will go unnoticed and unused.

sneridagh:
@davisagli @stevepiercy incoming, I promise, I needed it to see if something I wanted to try would work... and it did :)

sneridagh:
@stevepiercy @davisagli Which brings the point on how to document something that will be only usable in the next Plone (Volto) major, or in this case, only relevant for React Router 7 projects.

stevepiercy:
@sneridagh if by "how to document" you mean "where to document", and assuming this would be a how-to guide to configure routing in Volto, then it would be a new page under [Development](https://6.docs.plone.org/volto/development/index.html). If a reference, then under [Configuration](https://6.docs.plone.org/volto/configuration/index.html). If explanation, then under [Conceptual guides](https://6.docs.plone.org/volto/conceptual-guides/index.html).

In all cases, we would need to add an admonition and use a title that includes "Plone 7". We could try a collapsible admonition, initially closed, so that the reader would have to click on the admonition title to reveal the content, effectively serving as a "I understand and accept the risks" thing.

https://myst-parser.readthedocs.io/en/latest/syntax/admonitions.html#collapsible-admonitions

````markdown
# Routes in Plone 7

```{admonition} Warning! This feature is experimental. Click to accept the risks and read details.
:class: warning

All the content here.
```
````

If that's too much, we can just use an admonition at the top of the page without a toggle.

We also have this ridiculously shameless document [Contributing to Volto > Routing](https://6.docs.plone.org/volto/contributing/routing.html), so whatever you write will not conflict with what does not exist. There may be other relevant docs, but I don't know which they would be from a [search for "routing"](https://6.docs.plone.org/search.html?q=routing).

sneridagh:
@stevepiercy let's talk about this, because I started to write it, and it's not that easy.

For example, let's assume we start to document all the new things in Plone 7. If we mix it up with the "Plone 6" things it will be a complete mess. Specially when we will have to "switch" to Plone 7 is default eg. "https://7.docs.plone.org"... 

So I think we really need a sound strategy with this. TBH, I'd start a new completely different "Canary docs" section, going full diataxis in there.

It's easier to document the new feature in `@plone/registry` alone for now. I'll start doing it.

sneridagh:
@stevepiercy @davisagli  https://github.com/plone/volto/pull/6604

PR #6617, opened 2025-01-24
Split RTD PR preview builds between Volto and plone/registry

stevepiercy:
This PR should be merged before:

- https://github.com/plone/volto/pull/6618
- https://github.com/plone/volto/pull/6619

PR #6618, opened 2025-01-24
Use build.commands to disable RTD build on 16.x.x

stevepiercy:


stevepiercy:
This takes care of the Volto RTD build failure, which is an improvement.

I also think that https://github.com/plone/volto/pull/6617 will prevent pull request preview builds against non-main branches.

davisagli (approved):

PR #6629, opened 2025-01-28
fix broken releations control panel

giuliaghisini:
in relations controlpanel, if there was some broken relation, controlpanel brokes.


stevepiercy:
What happens when there is a broken relationship? Did you observe a stack trace or error message?

davisagli (approved):
Looks okay to me.

giuliaghisini:
> What happens when there is a broken relationship? Did you observe a stack trace or error message?

one of two items was null and wasn't handled in this view

packages/volto/news/6629.bugfix
```diff
@@ -0,0 +1 @@
+Fixed Relations controlpanel when have broken relations. @giuliaghisini
```
stevepiercy, line 1: I grabbed your comment, and amended the log to make it clear how someone can identify the problem.

```suggestion
Correctly handle when one of the items is `null` in the Relations control panel. @giuliaghisini
```

stevepiercy (changes requested):
Let's merge after this change.

stevepiercy (approved):
Please merge when green. Thank you!

PR #6634, opened 2025-01-30
Fix blurry images

giuliaghisini:
If we have for example image size defined in volto images controlpanel as:
Screenshot 2025-01-30 alle 10 52 19

and so we have scales ... 300x, 400x, 600px

and we have an image with original sizes 587x459, the generated image code and srcSet was: 

``

where src attribute contains the original image url, but it seems browser ignore it and use the largest image provided by srcSet.

If i create an Image Block with default settings align: center and size: L
the image will results blurried, because it's width and height setted as attributes says that image will have that sizes in page, but srcSet only generates url until 400w (because restapi returned only that scales..), so it uses 400w scale inside a space of 597 and adapt it to that width with the result of blurried image.

The solution to avoid blurry images in this cases, is to add to srcSet the original size, so with this pr, the generated tag image with srcSet is:

``

and image will not blurried. 

Example before fix: 
Screenshot 2025-01-30 alle 11 05 22

Example after fix:
Screenshot 2025-01-30 alle 11 04 19

### DIFF
--- a/packages/volto/news/6634.bugfix
+++ b/packages/volto/news/6634.bugfix
@@ -1,0 +1,1 @@
+Fix blurry images (srcSet generation). @giuliaghisini
--- a/packages/volto/news/6634.bugfix
+++ b/packages/volto/news/6634.bugfix
@@ -1,1 +1,1 @@
-Fix blurry images (srcSet generation). @giuliaghisini
+Display the appropriately sized image to eliminate blurring from upsizing smaller images in the `srcSet` generation. @giuliaghisini
--- a/packages/volto/src/components/theme/Image/Image.jsx
+++ b/packages/volto/src/components/theme/Image/Image.jsx
@@ -54,7 +54,14 @@ export default function Image({
     attrs.className = cx(className, { responsive });
 
     if (!isSvg && image.scales && Object.keys(image.scales).length > 0) {
-      const sortedScales = Object.values(image.scales).sort((a, b) => {
+      const sortedScales = Object.values({
+        ...image.scales,
+        original: {
+          download: `${image.download}`,
+          width: image.width,
+          height: image.height,
+        },
+      }).sort((a, b) => {
         if (a.width > b.width) return 1;
         else if (a.width < b.width) return -1;
         else return 0;
--- a/packages/volto/src/components/manage/Blocks/LeadImage/__snapshots__/LeadImageSidebar.test.jsx.snap
+++ b/packages/volto/src/components/manage/Blocks/LeadImage/__snapshots__/LeadImageSidebar.test.jsx.snap
@@ -21,7 +21,7 @@ exports[`renders a Lead Image block Sidebar component 1`] = `
       height={400}
       sizes="188px"
       src="/image.png/@@images/image-1200.png"
-      srcSet="/image.png/@@images/image-400.png 400w"
+      srcSet="/image.png/@@images/image-400.png 400w, /image.png/@@images/image-1200.png 400w"
       width={400}
     />
   </div>
--- a/packages/volto/src/components/theme/Image/__snapshots__/Image.test.jsx.snap
+++ b/packages/volto/src/components/theme/Image/__snapshots__/Image.test.jsx.snap
@@ -7,7 +7,7 @@ exports[`renders an image component from a catalog brain 1`] = `
   fetchpriority="high"
   height={400}
   src="/image/@@images/image.png"
-  srcSet="/image/@@images/image-400.png 400w"
+  srcSet="/image/@@images/image-400.png 400w, /image/@@images/image.png 400w"
   width={400}
 />
 `;
@@ -19,7 +19,7 @@ exports[`renders an image component from a catalog brain using \`preview_image_l
   fetchpriority="high"
   height={400}
   src="/image.png/@@images/image.png"
-  srcSet="/image.png/@@images/image-400.png 400w"
+  srcSet="/image.png/@@images/image-400.png 400w, /image.png/@@images/image.png 400w"
   width={400}
 />
 `;
@@ -40,7 +40,7 @@ exports[`renders an image component with fetchpriority high 1`] = `
   fetchpriority="high"
   height={400}
   src="/image/@@images/image.png"
-  srcSet="/image/@@images/image-400.png 400w"
+  srcSet="/image/@@images/image-400.png 400w, /image/@@images/image.png 400w"
   width={400}
 />
 `;
@@ -53,7 +53,7 @@ exports[`renders an image component with lazy loading 1`] = `
   height={400}
   loading="lazy"
   src="/image/@@images/image.png"
-  srcSet="/image/@@images/image-400.png 400w"
+  srcSet="/image/@@images/image-400.png 400w, /image/@@images/image.png 400w"
   width={400}
 />
 `;
@@ -65,7 +65,7 @@ exports[`renders an image component with responsive class 1`] = `
   fetchpriority="high"
   height={400}
   src="/image/@@images/image-1200.png"
-  srcSet="/image/@@images/image-400.png 400w"
+  srcSet="/image/@@images/image-400.png 400w, /image/@@images/image-1200.png 400w"
   width={400}
 />
 `;
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -18546,7 +18546,7 @@ snapshots:
   '@babel/plugin-syntax-json-strings@7.8.3(@babel/core@7.25.8)':
     dependencies:
       '@babel/core': 7.25.8
-      '@babel/helper-plugin-utils': 7.25.7
+      '@babel/helper-plugin-utils': 7.24.0
 
   '@babel/plugin-syntax-jsx@7.24.1(@babel/core@7.24.4)':
     dependencies:


packages/volto/news/6634.bugfix
```diff
@@ -0,0 +1 @@
+Fix blurry images (srcSet generation). @giuliaghisini
```
stevepiercy, line 1:
```suggestion
Display the appropriately sized image to eliminate blurring from upsizing smaller images in the `srcSet` generation. @giuliaghisini
```

stevepiercy (changes requested):
Tweak news to describe the issue this PR fixes.

stevepiercy (approved):
News looks good. Still needs tech review. Thank you!

davisagli (approved):
The solution makes sense to me. We also did a similar fix in at least one project.

packages/volto/src/components/theme/Image/Image.jsx
```diff
@@ -54,7 +54,14 @@ export default function Image({
     attrs.className = cx(className, { responsive });
 
     if (!isSvg && image.scales && Object.keys(image.scales).length > 0) {
-      const sortedScales = Object.values(image.scales).sort((a, b) => {
+      const sortedScales = Object.values({
+        ...image.scales,
+        original: {
```
MrTango, line 59: I'm sorry, but this is really wrong.
Imagine a user uploads an image directly coming from a digital camera.
This image will be bigger than 10MB. With this setting in place, the browser will ofter take this image.

The original image should never be loaded in templates.
If we need bigger scales, then they have to be created and used.
The original image is uncompressed and unscaled. 
Which is useful for some rare cases.
But for templates we should define scales which are big enough, for example in Volto Light Theme with the Image block, we need around 2500px in width, to fill the whole screen, but never 6000px or more as we have in camera images.

sneridagh: @MrTango how the situation that this PR is trying to solve is solved in Classic? Did you hit that problem?

MrTango: In my prototype i was working on during the sprint, i implemented it, so that all scales where defined, but not the original image. By default that means, the largest scale is `huge 1600:65536`, which might be a bit to small if you use the Image block in full-width. Screens can have resolutions around 2000 up to 2500px in width.
The solution here, would be to add another scale around 2500px, which should be the largest version we have.
The benefit is, we have then a correctly scaled and compressed version and not a >10MB uncompressed image.

In classicui we go one step further with the picture variations defined in the image control panel (field is hidden in Volto). There one can define different srcset's by name, which than can be used in different places.
This way, you can have some srcset definitions which only provide a smaller amount of scales in the srcset-list and also implement art-direction, with the help of plone.app.imagescaling and some extra picture-variants.
https://6.docs.plone.org/classic-ui/images.html#default-configuration

This configuration is defined in json-schema and could also be used by Volto.
I think @pnicolli was looking into that for a project, but that didn't happen in the end. 

But the most important thing is, that the original is taken out there.
Users will upload such pictures and there are rightfully think that it is ok, it always was in Plone.

MrTango:
One thing i don't understand is, why there where only such small scales defined, I'm not seeing the bigger default scales.
Is this custom usage of a n Image Block?

this is the larges picture variant, we have defined for classic ui:

```json
 "large": {
  "sourceset": [
   {
    "additionalScales": [
     "preview",
     "teaser",
     "large",
     "great",
     "huge"
    ],
    "scale": "larger"
   }
  ],
  "title": "Large"
 },
```
which leads to a srcset containing all scales in:
"preview",
"teaser",
"large",
"great",
"huge"
"larger"

and we use "larger" as the fallback (default), which on Desktop depending on picture variant, will be a bigger or smaller image. But that's more the default size and matters more for smaller images and is used in TinyMCE to allow the user to choose between small, medium and large images. Similar to what we have in Volto with the `field-size` or `block-width` settings.

Also we control a bit how many scales we provide for the Browser, if the image is a smaller one, for example in listings, we don't need to provide scales up to huge, because it never makes sense to load them ans depending on the screen DPI, this can prevent the Browser from loading to large images.

packages/volto/src/components/theme/Image/Image.jsx
giuliaghisini: > @MrTango how the situation that this PR is trying to solve is solved in Classic? Did you hit that problem?
> 
> @giuliaghisini @davisagli correct me if I'm wrong, but here entangled we have the problem that PIL does not create scales for sizes above the original, right? Thus, we have to rely in using the original....

yes

giuliaghisini:
> One thing i don't understand is, why there where only such small scales defined, I'm not seeing the bigger default scales. Is this custom usage of a n Image Block?
> 
> this is the larges picture variant, we have defined for classic ui:
> 
> ```json
>  "large": {
>   "sourceset": [
>    {
>     "additionalScales": [
>      "preview",
>      "teaser",
>      "large",
>      "great",
>      "huge"
>     ],
>     "scale": "larger"
>    }
>   ],
>   "title": "Large"
>  },
> ```
> 
> which leads to a srcset containing all scales in: "preview", "teaser", "large", "great", "huge" "larger"
> 
> and we use "larger" as the fallback (default), which on Desktop depending on picture variant, will be a bigger or smaller image. But that's more the default size and matters more for smaller images and is used in TinyMCE to allow the user to choose between small, medium and large images. Similar to what we have in Volto with the `field-size` or `block-width` settings.
> 
> Also we control a bit how many scales we provide for the Browser, if the image is a smaller one, for example in listings, we don't need to provide scales up to huge, because it never makes sense to load them ans depending on the screen DPI, this can prevent the Browser from loading to large images.

the problem is not only in ImageBlock, but in all places where Image component is used

packages/volto/src/components/theme/Image/Image.jsx
davisagli: Our goal is to serve the smallest available image which is large enough to display with good resolution in the available space and device pixel ratio.

If the original image is the best available option, it should be used. So I agree with including it here.

I also agree that it is good to make sure a very large original image is not used, which is best done by making sure there is an image scale defined which is large enough to be selected for full-width images on a high pixel density device.

davisagli:
> One thing i don't understand is, why there where only such small scales defined, I'm not seeing the bigger default scales.

The image block scales come from the image_scales metadata in the catalog, which only includes scale sizes smaller than the original image.

MrTango:
> > One thing i don't understand is, why there where only such small scales defined, I'm not seeing the bigger default scales.
> 
> The image block scales come from the image_scales metadata in the catalog, which only includes scale sizes smaller than the original image.

Ah now i understand the problem, you have.
What about only using the original image if it is smaller then the largest defined scale?

pnicolli:
> Ah now i understand the problem, you have. What about only using the original image if it is smaller then the largest defined scale?

I don't think we have access to the list of scales set in the imaging controlpanel on the frontend. In other words, we can only rely on what we get from the catalog. If we have `preview`, `great` and `huge` scales and the image is bigger than `preview` and `great` but smaller than `huge`, we don't get any `huge` scale from the catalog, therefore we don't know that it exists.

pnicolli:
On the other hand, I fully agree that we should find a way to avoid using the original image, as an optimization. Maybe we should really consider adding a 2000 and a 4000 image scale by default? Or maybe just documenting it for optimization purposes? I wouldn't mind adding those, but I'm not sure if the fact that it creates more disk usage, bigger catalogs and bigger responses can be a problem. We are proving a way to have full size images in core, with image blocks, so this is not just a problem for specific people who have big images.

davisagli:
@pnicolli @MrTango I don't agree that the image view is the only place where we would ever want the original image. Think of a full-width banner or background image...in a large window on a high pixel density display, it would need a lot of pixels to look good. But it depends on whether you're optimizing for good resolution or for using less bandwidth, and I'm not sure there's a single right answer for all use cases.

Defining a larger scale size by default would solve most of this. It would be big enough to be picked by the browser instead of the original image in most cases. In Plone 6 the scales are not actually generated until they are loaded, so while it would contribute to disk use, it would only do so in cases where it's actually used.

MrTango:
@pnicolli 
As far as i understand, we don't really have more images in the catalog, thats only metadata and i would rather give up some of the tiny once, nobody uses. And scales which are never loaded are not created, or is Volto still doing the creation upfront?
One bigger scale will not bring harm, but not having this solved will harm us.
It is not about optimization, rather then fulfilling expectations on a modern CMS. And most of all, preventing users from building very slow websites. If we don't solve this in the default settings, users will build those and the result will be Plone is slow ;).

@davisagli 
The original image can be way bigger than 3000px, the standard is around 6000px width or higher.
Nobody needs more than 3000px width, even for full-width.
The other side is, the original image is a high quality version of the image, where a scale is usually not only scaled down in dimensions, but also compressed in quality. 

So yes, i stand with my statement that we should not use the original, like we do it right now.
As i said, the fix can be adjusted to only use the original image, if the original image is smaller than the scales.
I think this should even better be solved in the backend (core).
So we could have plone.scale/namedfile handle this for us.
There we have all the scales, from the controlpanel (plone registry).
This way classicUI would benefit too.

To summarize:

- in the case of an original image being smaller than the defined scales
- plone.scale could create the next higher scale using the original image dimensions

so if we upload an image with 580px in width, plone.scale would give us all the scales up to 600px, which should be the next higher scale.

In Volto and classic, we would then have all available scales up to 600px and don't need the original image.
If the original image is larger than the larges scale, we would just have all our defined scales. But not the potentially >10MB images coming from a digital camera.

@mauritsvanrees what do you think, should we handle this in plone.scale?

packages/volto/src/components/theme/Image/Image.jsx
MrTango: @erral yes normally thats the case, but in this Volto side, i have different results:
https://tagung.plone.de/2025
You can see that the loaded images in the 4 teasers in the grid block, are all to big and up to 2000px, the one with Victor.
It seems that the width on the img-tag (which is for fallback and aspect ratio) is used to decide on the sizes.
But the actual images is set via CSS to a much smaller size. But somehow the loading uses the bigger sizes.
What i also don't understand is, why one image is around 800px and other 2000px.
But thats probably also the original size here.
I'll do more testing, on that topic in general soon.
But with my last tests back then when i implemented this for classic UI, the size defined by CSS should work.
And the width/height defined on the img tag is purely used for aspect ratio.

erral: > @erral yes normally thats the case, but in this Volto side, i have different results: https://tagung.plone.de/2025 You can see that the loaded images in the 4 teasers in the grid block, are all to big and up to 2000px, the one with Victor. It seems that the width on the img-tag (which is for fallback and aspect ratio) is used to decide on the sizes. But the actual images is set via CSS to a much smaller size. But somehow the loading uses the bigger sizes. What i also don't understand is, why one image is around 800px and other 2000px. But thats probably also the original size here. I'll do more testing, on that topic in general soon. But with my last tests back then when i implemented this for classic UI, the size defined by CSS should work. And the width/height defined on the img tag is purely used for aspect ratio.

But the point there is that the `<img` tag doesn't have the corresponding `sizes` attribute and the browser doesn't know that it should load the image from the `srcset` so it takes it from the `src` tag, right?

The fix should go in the Teaser block usage of the Image component, where a correct `sizes` should be passed like it is done in the "Tagungsort & Lageplan" columns, where you have `sizes="100vw"`. You can even put a smaller value there to have a smaller image loaded.

erral:
> 
> To summarize:
> 
>     * in the case of an original image being smaller than the defined scales
> 
>     * plone.scale could create the next higher scale using the original image dimensions
> 
> 
> so if we upload an image with 580px in width, plone.scale would give us all the scales up to 600px, which should be the next higher scale.
> 
> In Volto and classic, we would then have all available scales up to 600px and don't need the original image. If the original image is larger than the larges scale, we would just have all our defined scales. But not the potentially >10MB images coming from a digital camera.
> 
> @mauritsvanrees what do you think, should we handle this in plone.scale?

But that wouldn't solve your problem, because plone.scale could create a scale called like the next higher scale, but in any case its width will be the same (580px), and it will be rendered like that is in the `srcset`. No change there. The point should be adjusting the `sizes` attribute of the `img` tag.

MrTango:
@erral 
> But that wouldn't solve your problem, because plone.scale would create a scale called like the next higher scale, but in any case its width will be the same (580px), and it will be rendered like that is in the srcset. 

No the idea would be to generate a 600 scale, but use (upscale) the original image.
This way we would still only have defined scales up to the larges one, which in this case a bit larger than the original image. This scale would then be used and we would not have any unsharp images. That would solve the problem of this PR, without introducing the new one. 
And we could benefit in classic UI from it too.

packages/volto/src/components/theme/Image/Image.jsx
MrTango: No the img tag is just a fallback, for when the browser doesn't know how to deal with the source-tags ;).
Also it uses the width/height to calculate the aspect ratio.
In the source tags, you have the Srcset Width Descriptors, thats the same as inside an img-tag with that attribute.
But the missing sizes-attribute is the reason, why the browser is loading the big one. The viewport is huge and thats the only thing the browser has now.
I need to improve this in classic ui too.
So with the correct sizes set, large enough scales available, it not load the original image.
But in the long run, i think handling it in the backend and not including the original is a good thing to do.
But the most important thing is, that we get the sizes attribute in all our images.

BTW a source-tag has the advantage that you can also provide different media-type, like AVIF images which are much smaller than jpegs or png. If the browser support the first source-tag image, it loads it if not tries the next one. By default we don't use this in Plone but that will change soon i guess.
AVIF is supported by all modern browsers for a while now and does not have the downsides of webp with blurry text.
https://caniuse.com/?search=avif

PR #6476, opened 2024-11-07
Dutch Translations update. 

fredvd:
Contributed to by the Flemish Environment…

- [x] I signed and returned the [Plone Contributor Agreement](https://plone.org/foundation/contributors-agreement), and received and accepted an invitation to join a team in the Plone GitHub organization.
- [x] I verified there aren't other open [pull requests](https://github.com/plone/volto/pulls) for the same change.
- [x] I followed the guidelines in [Contributing to Volto](https://6.docs.plone.org/volto/contributing/index.html).

These are translations only. I tried to run the linting etc. by following the instructions at https://6.docs.plone.org/volto/contributing/developing-core.html and then run 'pnpm lint', in the packages/volto directory but I get a message that more than 10 errors are present.  Also my pnpm-lock.yaml file gets local changes when I have pulled out the latest main and am using corepack to pick the correct pnpm version:

```
diff --git a/pnpm-lock.yaml b/pnpm-lock.yaml
index d17950b41..f4ff0050c 100644
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -25399,7 +25399,7 @@ snapshots:

   babel-plugin-polyfill-corejs2@0.4.10(@babel/core@7.25.8):
     dependencies:
-      '@babel/compat-data': 7.25.8
+      '@babel/compat-data': 7.24.4
```

- [ ] I succesfully ran [code linting checks](https://6.docs.plone.org/volto/contributing/linting.html) on my changes locally.
- [ ] I succesfully ran [unit tests](https://6.docs.plone.org/volto/contributing/testing.html) on my changes locally.
- [ ] I succesfully ran [acceptance tests](https://6.docs.plone.org/volto/contributing/acceptance-tests.html) on my changes locally.
- [ ] If needed, I added new tests for my changes.
- [ ] If needed, I added [documentation](https://6.docs.plone.org/volto/contributing/documentation.html#narrative-documentation) for my changes, either in the Storybook or narrative documentation.
- [ ] I included a [change log entry](https://6.docs.plone.org/contributing/index.html#contributing-change-log-label) in my commits.

-----

PR #6688, opened 2025-02-06
Update Dutch translations.

mauritsvanrees:
This turns two Dutch terms introduced in PR #6476 back to their English version: teaser and grid. With @fredvd and the client we agreed that in this case the English terms are better, less surprising. So this reverts a small part of that PR.

@fredvd I left one case of 'voorproefje' untouched, as it seems a fine translation of 'preview':

```
msgid "Results preview"
msgstr "Resultaten voorproefje"
```

### DIFF
--- a/packages/volto/locales/nl/LC_MESSAGES/volto.po
+++ b/packages/volto/locales/nl/LC_MESSAGES/volto.po
@@ -3020,7 +3020,7 @@ msgstr "Resultaten limiet"
 #. Default: "Results preview"
 #: components/manage/Blocks/Listing/Edit
 msgid "Results preview"
-msgstr "Resultaten voorproefje"
+msgstr "Resultaten voorvertoning"
 
 #. Default: "Results template"
 #: components/manage/Blocks/Search/SearchBlockEdit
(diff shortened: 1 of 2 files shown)


fredvd:
> This turns two Dutch terms introduced in PR #6476 back to their English version: teaser and grid. With @fredvd and the client we agreed that in this case the English terms are better, less surprising. So this reverts a small part of that PR.
> 
> @fredvd I left one case of 'voorproefje' untouched, as it seems a fine translation of 'preview':
> 
> ```
> msgid "Results preview"
> msgstr "Resultaten voorproefje"
> ```

Lol, yes there it is fine again. But it still is too playful IMHO, ‘voorvertoning’ would be more neutral but icky as well.  And maybe we’re dealing with Flemish/Dutch as well.  Ah linguistic yak shaving  :-P

fredvd (approved):
Comfirmed

mauritsvanrees:
@fredvd Ah yes, 'voorvertoning' is better there.  Also from a Flemish viewpoint I think.  It is used a few more times.  I have pushed a commit.

PR #6751, opened 2025-02-13
fix: block chooser position in experimental.addBlockButton

giuliaghisini:
problem: with 
config.experimental.addBlockButton.enabled=true prop
the placement of block chooser was wrong sometimes and search bar or some portion of block ckooser was not visible and usable (expecially when you have a lot of blocks in a block chooser section). See the problem here:

https://github.com/user-attachments/assets/fb37ec0e-9027-4103-a210-ea6522d73fa9

With this fix, block chooser will placed bottom to the block, without mentioned problems:

https://github.com/user-attachments/assets/3337bdee-5ad1-44ab-81e2-57bc730c2cf2

### DIFF
--- a/packages/volto/news/6751.bugfix
+++ b/packages/volto/news/6751.bugfix
@@ -1,0 +1,1 @@
+Move the block chooser beneath the block, instead of covering the block and preventing making a choice, when `config.experimental.addBlockButton` is enabled. @giuliaghisini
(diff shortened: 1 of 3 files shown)


ichim-david (approved):
@giuliaghisini you've misspelled bugfix this is why the changelog test is crashing. otherwise I agree with this change

davisagli (approved):
Thanks!

Issue #4290, opened 2023-01-21
Allow to publish a Volto site under a relative-path domain (e.g. "www.example.com/foo")


nileshgulia1:
## Abstract

Allows a Volto project to be published under relative-path. For eg: `www.mysite.com/foo,`

## Motivation

We came across projects that manifest the need to have multiple sites published under one domain. For eg: https://water.europa.eu/ is the static webpage that links to two "subsites". The page is static with two sites hosted inside their own folders. For eg: /Plone/marine and /freshwater. 

There is a need to replace the build path every time, eg for freshwater, we are using PUBLIC_PATH https://github.com/eea/freshwater-frontend/blob/master/entrypoint-prod.sh#L16

We might want a more robust solution that can handle the prefixing of build files at runtime under one env var. So something like `RAZZLE_PREFIX_PATH="/foo"` will ensure that plone site root is at `/foo`.

## Proposal & Implementation

This has been nicely already put on top of the table by red turtle folks @giuliaghisini @cekk @pnicolli in the form of https://github.com/plone/volto/pull/3592/. 

Based on basename react-router prop which adds an initial entry to browser "history" stack and allows all links to be prefixed with basename by react-router. The static assets like images need to be explicitly prefixed. @nileshgulia1  amended an express-middleware which re-directs to app base path on initial load.

When we use basename in react-router, we basically have two situations where we need to manually prefix:

1. When we use the `a` tag.
2. When we use the `img` tag

About the `a` tag, it is best to avoid it and use the `Link` component of `react-router-dom` or the `UniversalLink` component of Volto. This way, the prefix will be added automatically. There is a situation where the `Link` component will not work, which is for downloading files. In this situation, the best thing to do is to use the `UniversalLink` component and manually prefix the `a` tag in this component, in the case of downloads. This way, the prefix will be centered.

About the `img` tag, the proposal is to replace all uses of it with the `Image` component of Volto and prefix in the component.

There is another approach in https://github.com/plone/volto/pull/3190 PR, which uses a store enhancer middleware to prefix all the router paths(amending history accordingly) and modified flattenToAppUrl method.

## Deliverables
- [ ] `./foo` as plone site root
- [ ] serve build assets from RAZZLE_PREFIX_PATH
- [ ] prefix all Images with RAZZLE_PREFIX_PATH
- [ ] Url links to be prefixed as well.
- [ ] Need to take care of flattenToAppUrl. Shouldn't add multiple prefixes.
[...]

PR #6754, opened 2025-02-14
Use Image from @plone/volto/components/theme/Image/Image instead of semantic-ui-react

wesleybl:
This is part of #3592, to address #4290.

That PR is to make Volto run on a subpath. For example:

http://mydomain.com/site

The idea is to use `react-router`'s basename property to prefix routes. However, it doesn't add a prefix to the src of the img tag. So, to solve this, the idea is to insert the prefix in the `Image` component of `@plone/volto/components/theme/Image/Image`. Then all Volto images would be rendered with the Volto component.

The Image component of semantic-ui-react generates an img tag. That's why we want to stop using it.

Some points of this PR:

- The only thing that the `size` property of the `Image` of semantic-ui-react does is add a class to the img tag. So we use the `className` property of Volto's `Image` to simulate this behavior.

- The semantic-ui-react `Image` adds the "ui" and "image" classes. So we use the `className` property of the Volto Image for this.

- If you agree, we can use an eslint check to prevent the semantic-ui-react component from being used in the future.


packages/volto/src/components/theme/Logo/Logo.jsx
```diff
@@ -4,13 +4,13 @@
...
 import { defineMessages, useIntl } from 'react-intl';
 import { useEffect } from 'react';
-import { Image } from 'semantic-ui-react';
 import LogoImage from '@plone/volto/components/theme/Logo/Logo.svg';
 import { useSelector, useDispatch } from 'react-redux';
 import { Link, useLocation } from 'react-router-dom';
 import { getNavroot } from '@plone/volto/actions/navroot/navroot';
 import { flattenToAppURL, getBaseUrl } from '@plone/volto/helpers/Url/Url';
 import { hasApiExpander } from '@plone/volto/helpers/Utils/Utils';
+import Image from '@plone/volto/components/theme/Image/Image';
```
nileshgulia1, line 13: @wesleybl Aren't we tackling this in the main PR?
Also  I found that after we upload the logo through controlpanel this fails to work as `plone_site_logo` comes as site root URL and Plone doesn't know about the prefix. 
https://github.com/plone/Products.CMFPlone/blob/6e2b2f55fc251a0e68b5ba65c09de6be867a9d39/Products/CMFPlone/utils.py#L508

wesleybl: > Aren't we tackling this in the main PR?

@nileshgulia1 Yes, but I'm following @sneridagh's idea of ​​redoing PR #3592 to explain things better. So I decided to break it down into smaller, easier-to-understand PRs.

This PR practically doesn't change anything about the Volto, but it already prepares the framework for the prefix.

> Also I found that after we upload the logo through controlpanel this fails to work as plone_site_logo comes as site root URL and Plone doesn't know about the prefix.

The idea here is not yet to make the prefix work.

nileshgulia1: @wesleybl I will put it in this agenda in upcoming volto team meeting. Also what about the links?

wesleybl:
> Also what about the links?

@nileshgulia1 Others PRs will come if this one is approved. But for Links, the idea is to use the `basename` property of `react-router`. Then we will use the `Link` component of `react-router`. This will cause `react-router` to add the prefix itself. The only situation where we cannot use the `Link` component is for downloading files. Then the idea is to centralize the `UniversalLink` component and use a function to prefix it:

https://github.com/plone/volto/pull/3592/files#diff-04954b3fe24f3d7f20d9a599b9b4c987d3459053d2e2f32759114e55df6eb228R112

nileshgulia1 (approved):

nileshgulia1:
@pnicolli @sneridagh This will offload some changes in the main PR. We have to do it anyway in future. Would be great if you can review this please. Thank you!

sneridagh (approved):

sneridagh:
@plone/volto-team  a last quick review on this?

ichim-david (approved):

sneridagh:
@wesleybl I think we can have it in 18, if @plone/volto-team agrees with it. Given the tests snapshots, I think that the changes in there are not breaking, and if someone has a customization, it will be already using the shadow.

@wesleybl can you please take care?

PR #6758, opened 2025-02-16
Fix redirect for MDN's Using CSS custom properties (variables)

stevepiercy:
@sneridagh is it OK with you if I no longer include change log notes for linkcheck fixes? It would save me a good amount of time going forward, and I don't think anyone wants to read about them in the change log.

----
📚 Documentation preview 📚: https://volto--6758.org.readthedocs.build/


davisagli (approved):

davisagli:
@stevepiercy I agree, this sort of thing would just be noise in the release notes.

PR #6763, opened 2025-02-18
REmove the corepack fix from prettier

sneridagh:


sneridagh:
@plone/volto-team confirmed: The corepack armageddon is over. Node 20.19.0 fixes it in 20.x and Github rolled it out to the current images. Node 22 already fixed it a couple of weeks ago.

We have to decide if we leave the workaround or not... at least from my point of view I would remove it. However, I'd leave it in the amended documentation for now, in case someone has an outdated Node version installed. Maybe also update the recommended versions to the "good known" Node.js?

stevepiercy:
I'd update it, and not drag it out. It's an easy change, reverting that made in:

https://github.com/plone/documentation/pull/1879/files#diff-c3cafea9f90319caf6817a7072e0ac47aef9428887ebea1b656e3c234b0ebb49R101

That would update https://6.docs.plone.org/install/create-project-cookieplone.html#node-js and close https://github.com/plone/documentation/issues/1880.

FTR, we summarized that work after much discussion in this one comment https://github.com/plone/documentation/issues/1877#issuecomment-2692657492.

I think we need to update [Contributing to Volto > Version policy > Node.js](https://6.docs.plone.org/volto/contributing/version-policy.html#node-js) with a `warning` admonition. Would you specify excluded bad versions or included good versions?

I would also add the same admonition to the [install via Cookieplone](https://6.docs.plone.org/install/create-project-cookieplone.html#node-js).

stevepiercy:
@sneridagh it looks like `npm install -g corepack` is here to stay if you want to use pnpm. Is that correct? See this article:

https://socket.dev/blog/node-js-tsc-votes-to-stop-distributing-corepack

davisagli:
Based on the article Steve linked, this is no longer relevant: in the long term we either need to keep explicitly installing corepack, or stop using it.

PR #6766, opened 2025-02-18
Fix QueryWidget focus issue by checking this.node before calling focus

### LINKED ISSUE #2281: Can't use "query" as first widget in a form
The querystring widget (hooked for the field with id "query") will break if it's the first widget in a form, because first widgets get focused. 

https://github.com/plone/volto/blob/master/src/components/manage/Widgets/QuerystringWidget.jsx#L135

There's no `this.node` defined in that class

Haiderkhan64:
This PR fixes an issue where `QueryWidget` tries to call `focus()` on `this.node` 
without ensuring `this.node` is defined. This caused a crash when the widget 
was the first in a form and automatically focused.

- Added a check for `this.node` before calling `focus()`.
- Fixes issue #2281.

Closes #2281

### DIFF
--- a/packages/volto/src/components/manage/Widgets/QueryWidget.jsx
+++ b/packages/volto/src/components/manage/Widgets/QueryWidget.jsx
@@ -105,7 +105,7 @@ export class QuerystringWidgetComponent extends Component {
    * @returns {undefined}
    */
   componentDidMount() {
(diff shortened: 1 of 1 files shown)


stevepiercy:
Closing for not following Contributor guidelines.

PR #6779, opened 2025-02-24
[components] Rationalize and improve @plone/components icons handling. Add own Vite SVGR plugin proxy.

sneridagh:


packages/components/vite-plugin-sgvr.js
pnicolli, line 1: This name should be `vite-plugin-svgr`. Also, it could actually be completely different though, since it can be confusing to have a plugin with the same name as a famous vite plugin. It took me some time while reading the PR to realize the plugin used was a different one from the original one. What if we just export the config instead of reexporting the whole plugin? I feel like it's something you already considered and I am curious to understand what's behind this choice.

packages/components/tsup.config.ts
```diff
@@ -1,7 +1,7 @@
 import { defineConfig } from 'tsup';
 
 export default defineConfig({
-  entryPoints: ['src/index.ts'],
+  entryPoints: ['src/index.ts', 'src/components/Icons/index.ts'],
```
pnicolli, line 4: Proposal: what if we started using lowercase naming for folders? I think it adheres more to the community standard. Folders with a capital first letter, if used at all, are usually for folder containing a single React component and its tests or closely related files.

pnicolli (commented):
In addition to the file specific comments, we should test that the rendered `` elements have all the attributes needed for good web vitals (width and height to say the least).

packages/components/vite-plugin-sgvr.js
sneridagh: In fact, we are exporting the instantiated config that works with the Icon in @plone/components.

It's for convenience, so people does not have to have a dep on the original plugin, then configure it. Also, there's the TS config for dealing with the imports with `?react`... any suggestion instead of this approach?

pnicolli: Not really, we could have just exported a default config and leave it to the project to import it and use it. A pro of this approach is that users are more aware of what's happening and have more control, a con is for sure having less batteries included. I don't have a strong preference here, it's fine either way to me.

pnicolli: I would definitely correct sgvr to svgr everywhere in our code though, even if we keep this approach.

sneridagh: The con is that the app has to have the deps on `vite-plugin-sgvr` and `@sgvo/plugin-svgo`...

sneridagh: But I agree that `@plone/components` then is not the best place for them... I even evaluated having the icons in their own package.

packages/components/vite-plugin-sgvr.js
```diff
@@ -0,0 +1,47 @@
...
+            {
+              name: 'removeAttrs',
+              params: {
+                attrs: 'fill',
+              },
+            },
+          ],
+        },
+      },
+    }),
```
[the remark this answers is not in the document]
sneridagh: Done! Take a look now please!

packages/components/tsup.config.ts
sneridagh: Done!

packages/components/src/components/Icon/Icon.tsx
```diff
@@ -37,38 +27,33 @@ export interface IconProps extends DOMProps, AriaLabelingProps, StyleProps {
...
-
   if (!ariaHidden) {
     ariaHidden = undefined;
   }
 
   const iconSize = size ? size : 'M';
+  const color = props.color?.startsWith('--')
+    ? `var(${props.color})`
+    : props.color;
+  console.log(color);
```
pnicolli, line 53: Cleanup

packages/components/src/components/Icon/Icon.tsx
```diff
@@ -77,7 +62,8 @@ export function Icon(props: IconProps) {
       'q icon',
       `icon--size${iconSize}`,
       children.props.className,
-      styleProps.className,
+      props.className,
```
pnicolli, line 65: I think we should probably also spread `otherProps.style` here

packages/components/tsup.config.ts
sneridagh: Done

packages/components/src/components/Icon/Icon.tsx
sneridagh: Fixed

packages/components/src/components/Icon/Icon.tsx
sneridagh: Done too!

sneridagh:
@pnicolli I think this is ok now.

pnicolli:
@sneridagh I like the implementation, I still think we should fix the naming error: sgvr should be svgr everywhere in this PR imho

sneridagh:
@pnicolli wut? is there a typo there? can you believe I haven't realised of it all this time?

sneridagh:
@pnicolli sorryyyy, it's fixed!

pnicolli (approved):

PR #6786, opened 2025-02-26
Token Widget Label

Wagner3UB:
Added aria-label to CreatableSelect inside the TokenWidget to better accessibility.

PR for Volto 18: https://github.com/plone/volto/pull/6804


news/6786.bugfix
```diff
@@ -0,0 +1 @@
+Added an aria-label to CreatableSelect inside TokenWidget to improve the component’s accessibility. @Wagner3UB
```
stevepiercy, line 1:
```suggestion
Added an `aria-label` to `CreatableSelect` inside `TokenWidget` to improve the component’s accessibility. @Wagner3UB
```

stevepiercy (changes requested):
Sprinkle a little MyST in the change log.

davisagli:
> PS. The changes made in this PR will also be applied to the ‘main’ branch as soon as the PR for the current version is approved.

I'd really prefer to review and merge changes first for the master branch, then backport them to older branches. I don't want a situation where we merge something for an old branch and then forget to port it forward.

Wagner3UB:
@davisagli This should be correct now. I’ll reply here with what I said on the other related PR: the problem was that this file was already fixed in Volto 18, so the modification I made there was not necessary. I will close the v18 PR and leave only the v17 one open. I will replicate the modification made here to v17, to be compliant.

davisagli (approved):

PR #6790, opened 2025-02-27
Fix Display Warning When TextWidget Exceeds maxLength Limit

### LINKED ISSUE #6783: Display Warning When TextWidget Exceeds maxLength Limit
*Describe the bug*
Currently we are only disabling user to not type more than the given maxLength property. But we are not showing any visual warning. It will be nice to have a visual warning as Textarea widget.

*To Reproduce*
1. Set the maxLength property for TextAreaWidget.
2. You will not able to see any user feedback.

*Expected behavior*
1. It should show the user feedback like Textareawidget.

iRohitSingh:
Fix #6783 

screencast: 

https://github.com/user-attachments/assets/e3d0f8bf-01ef-4192-9f50-afc15a3fe961

### DIFF
--- a/packages/volto/news/6783.bugfix
+++ b/packages/volto/news/6783.bugfix
@@ -1,0 +1,1 @@
+Fix Display Warning When TextWidget Exceeds maxLength Limit. @iRohitSingh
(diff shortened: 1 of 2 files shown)


iFlameing (approved):
@sneridagh Functionality-wise, it looks good. It follows a similar approach to the TextAreaWidget.

iFlameing:
@sneridagh I have looked into the validation as well as this issue. Validation kicks in when you submit the form. Doesn't we should provide the visual feedback during typing for constraint like maxLength and minLength? So that user is already aware whether they passed the maxLength constraints or not? If it is the case this pr is also fine.

Issue #5662, opened 2024-01-19
Scroll down option for selecting months and years


07tAnYa:
**Is your feature request related to a problem? Please describe.**
When I go to select a date for my event, I have to keep on right/left swiping to select the appropriate date.

**Describe the solution you'd like**
When I hover on the year, it should show the years in a scroll-down-like structure, so that users can select easily.

**Additional context**
Here is the screen recording of the above-discussed issue: 

https://github.com/plone/volto/assets/140234197/5f157d0f-7ad3-44ad-ad33-ffc8f9be2053

stevepiercy:
@07tAnYa we use the datetime picker as shown in https://6.docs.plone.org/storybook/?path=/story/edit-widgets-datetime--datetime.

However, there may be better datetime pickers available. I brought this up for discussion in a PLIP at https://github.com/plone/volto/issues/5629. I agree that clicking through it, with limited keyboard shortcuts, is bad UX and accessibility. If you have an idea for a better off-the-shelf widget, please add your comment to that issue. Thank you!

07tAnYa:
@stevepiercy Sir, so should I close this issue over here?

stevepiercy:
No.

> If you have an idea for a better off-the-shelf widget, please add your comment to that issue.

vr-varad:
@07tAnYa is this issue still open

stevepiercy:
@vr-varad please read and follow [Things not to do](https://6.docs.plone.org/contributing/first-time.html#things-not-to-do) item 4.

vikash-prem123:
can i change the date widget like this so it will be easy to select.

![Screenshot from 2024-02-05 10-15-36](https://github.com/plone/volto/assets/106796672/dfc3f63a-1f43-41a2-a5bb-5074bbf2e489)

davisagli:
@vikash-prem123 No, that will hurt usability in other ways. We'd like to keep the existing calendar, but add a way to jump to a different month/year.

vikash-prem123:
Okay i have an idea. So when i click on the year ( on top )  and then  left arrow or right arrow  will change the year and click back it will be normal (or change the months).
By default it will change months.

![Screenshot from 2024-02-05 10-23-23](https://github.com/plone/volto/assets/106796672/49b51036-a9dc-4db5-848c-0fc32e14dcac)

stevepiercy:
[...]

PR #6793, opened 2025-02-28
Datetime widget a11y

Wagner3UB:
TODO: Description

### DIFF
--- a/locales/ca/LC_MESSAGES/volto.po
+++ b/locales/ca/LC_MESSAGES/volto.po
@@ -4707,6 +4707,12 @@ msgstr ""
 msgid "private"
 msgstr "privat"
 
+#. Default: "Navigate forward to interact with the calendar and select a date. Press the question mark key to get the keyboard shortcuts for changing dates."
+#: components/manage/Widgets/DatetimeWidget
+msgid "publicationDateMessage"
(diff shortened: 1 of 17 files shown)


stevepiercy:
Is there any reason not to go with React Aria's date, time, and datetime pickers instead?

https://github.com/plone/volto/issues/5662#issuecomment-2200104509

pnicolli:
> Is there any reason not to go with React Aria's date, time, and datetime pickers instead?
> 
> [#5662 (comment)](https://github.com/plone/volto/issues/5662#issuecomment-2200104509)

@stevepiercy It's a change to be considered more carefully. Not only it is a big change for us to implement, it is also changing the user interface for editors and needs to be communicated and tested properly. I would be happy if it happened, of course :) But I would accept intermediate steps like this one honestly untile a bigger time budget shows up for this.

Wagner3UB:
> Is there any reason not to go with React Aria's date, time, and datetime pickers instead?
> 
> [#5662 (comment)](https://github.com/plone/volto/issues/5662#issuecomment-2200104509)

I know it would be nice, but we don’t have the time (budget) to make the full modification at the moment. We need to keep things short for now.

davisagli:
@Wagner3UB I would like to review and merge this for the main branch first (Volto 18), then backport to older branches.

PR #6734, opened 2025-02-08
[Components] Tailwind Support

sneridagh:
PoC for adding Tailwind support by default in `@plone/components` to investigate possibilities and collisions.

PR #6795, opened 2025-03-01
[components] Tailwind support

sneridagh:
PoC for adding Tailwind support by default in @plone/components to investigate possibilities and collisions.

UPDATE (08/03/2025): A few words on this one. 
I was openly against Tailwind and its approach mainly because I saw it as yet another one CSS-in-JS solution, that was more focused in take away CSS from the developers and put it on the JS side than making the C in CSS work for them. 

However, they moved lately into the right direction, in v4. They embraced the latest improvements in CSS spec for their own benefit. They fully embraced CSS layers, and simplified the overall hackyness feeling around the solution.

This is one of the things that you have to feel and test it out yourself, until it "clicks". It happened for me with TypeScript, and I have to admit that it happened to me now with Tailwind. 

We cannot deny what the ecosystem and the industry is driving. Same that we have to embrace how React 19 deals with data acquiring, and all the new features, we have to stop being blind to what happens in the React and JS brave new world. When such an amount of people praises something... it has a reason behind. Could be that you still haven't seen it yet, or understood why, but normally, it has its reasons to be. Happened to me with TypeScript, happened now with Tailwind.

From my personal point of view, I've come to think in Tailwind as another pre-processor. You have the decission of taking it or not. They embraced [lightningCSS](https://lightningcss.dev/) which provides still some goodies that you might miss from a traditional preprocessor. As soon you use a utility class, the corresponding CSS is generated. No more, no less. You no longer requires it, it gets removed. They also fully embraced CSS custom properties, so you can add them and it seamlessly adds the corresponding utility class to it ready to be. 

In summary, I think this is worth to discuss, explore, and test. We have to make a list all the concerns and try to find the solutions to these concerns.

Supersedes: #6734


sneridagh:
@pnicolli I polished this one, my plan: Add the basics and foundation for them, then merge. We will do the work "as we develop" quanta.

pnicolli (approved):

sneridagh:
I had a talk with @davisagli yesterday, and he expressed some valid concerns, that we will have to address and give answer before taking the decision of going with this approach. This is also valid for: https://github.com/plone/volto/pull/6797 and having it by default in the CMSUI components.

- What is the override story of a Tailwind based component?
- What is the implications if you want to mix one design system using a non-Tailwind approach (like IO-Comune bootstrap)
- and related: How this will work with the public UI view, in case that the public UI is not using a Tailwind approach?

@davisagli it would be great to elaborate them more, in order to find a valid good answer, and a PoC for each one. Could be that I'm also forgetting things that we talked yesterday. Let's write them down, so we can work and discuss on this during the Bucharest sprint.

sneridagh:
@davisagli I added my thoughts to the description, so we have it as reference too.

@pnicolli if you think there's something to add, please do.

I would refrain to merge this until we have an educated decision happening during or from the outcome of the sprint.

davisagli:
I'm not necessarily opposed to using Tailwind for the components, especially now that I learned a bit more about it. Here's how I would answer those questions based on my current understanding:

> What is the override story of a Tailwind based component?

It's no different from any other component: you can shadow it.

But if `@plone/components` will use Tailwind, I'm just saying we should be very clear about that in the docs, that it has an opinionated approach to styles and can't easily have other styles applied (aside from what can be done with Tailwind theme variables) unless you override the components.

> How this will work with the public UI view, in case that the public UI is not using a Tailwind approach?

It won't. If we use Tailwind for the components, then either the public UI has to support Tailwind or it needs to use different components. This includes block views.

pnicolli:
I think it's correct to say that we need to provide answers to these questions beforehand, in order to have a more defined goal.

My opinion on the components is that we should export an unstyled version of the components, plain CSS files to be imported if needed, and make a tailwind version available on the side in some way, for example with a shadcn registry or with a different set of exports.
This being said, here are the answers:

> What is the override story of a Tailwind based component?

Same as today. We should probably also think about "when are devs overriding our components?"
This question will need a better answer when we define theming, I would not expect too much overriding in the CMSUI, but as long as shadowing remains available, we should be fine.

> What is the implications if you want to mix one design system using a non-Tailwind approach (like IO-Comune bootstrap)
> and related: How this will work with the public UI view, in case that the public UI is not using a Tailwind approach?

We should use css custom properties for defining `@plone/components` styles as much as possible, in order to allow easy customization when `body.public-ui`. If we do this, it's a matter of theming again: we should have clearly separated default themes for CMSUI and public ui in order to just be able to build the second one from scratch if needed.

davisagli:
Say I'm creating an add-on with a block. The block view has some form components (for example) so I want to make sure they're rendered in "the standard way" i.e. they fit the (public) site design. The add-on wants to support any site's public UI so doesn't want to be prescriptive about using Tailwind or not. What components should I use in the block view?

pnicolli:
Well yeah this is probably the biggest challenge and we know this exact example will absolutely happen. Challenge accepted :)

The way I see this whole thing is with Tailwind as an added bonus, meaning the easy and straightforward import from `@plone/components` would be the unstyled ones, with easy addition of a styleesheet. Devs are used to this, many react libraries exporting widgets work this way or used to work this way. Tailwind should be something extra, something you have to explicitely import. This being said, let's answer the question.

> Say I'm creating an add-on with a block. The block view has some form components (for example) so I want to make sure they're rendered in "the standard way" i.e. they fit the (public) site design. The add-on wants to support any site's public UI so doesn't want to be prescriptive about using Tailwind or not. What components should I use in the block view?

This is perfect, it's how I would want block add-ons to work. The easy and default solution for the add-on dev would be to import components from `@plone/components` and style them with basic styling (with or without the styles exported from `@plone/components`), then export basic widgets from the add-on, together with the rest of the block/add-on exports. Maybe it should also just use the config registry, meaning it sets its own widgets there therefore allowing customization from there. This story has yet to be tested with rr7 but if it splits code correctly it could be a huge improvement.
If the add-on dev wants to use tailwind, I would assume they would need to export the styles separately in case the public theme does not use tailwind. This is an area that needs to be explored with more PoCs imho, but it should be doable.

All in all, I would be happy to be able to export tailwind-based components as an extra option from the library in order to use those for the cmsui, for now, and then experiment more. As I always said, it's a requirement that we make everything work seamlessly without tailwind for now and build on top of it.

PR #6797, opened 2025-03-02
Quanta CMSUI app

sneridagh:
This PR provides the CMSUI detached app. We decided to name it `quanta` for now, since it made very much sense.
It has Tailwind support by default.

This PR provides also a `@plone/cmsui` add-on which will contain the CMSUI as an add-on.

This PR includes also an improvement in the Plone Vite plugin so it creates a `@source` entry for each add-on that finds that has Tailwind as a dependency.

Supersedes: #6756


apps/quanta/README.md
```diff
@@ -0,0 +1,33 @@
...
+Even in experimental phase, this package will be soft released periodically, under a tag.
+This will provide a way to try it out in real development and deploy scenarios.
+
+## Development
+
+To start, from the root of the monorepo, issue the following commands.
+
+```shell
+pnpm install
+pnpm --filter quanta run dev
```
pnicolli, line 26:
```suggestion
pnpm --filter quanta dev
```

apps/quanta/app/loaders/content.ts
```diff
@@ -0,0 +1,64 @@
...
+function flattenToAppURL(data: Content, request: Request) {
+  const currentUrl = new URL(request.url);
+  const baseUrl = `${currentUrl.protocol}//${currentUrl.host}`;
+
+  // Convert data to string to perform replacements
+  let stringData = JSON.stringify(data);
+
+  // Replace all occurrences of backend URLs
+  stringData = stringData.replace(
+    new RegExp(config.settings.apiPath, 'g'),
```
pnicolli, line 23: Do we really need to use a regexp here? I understand this might be just copied over from the current Volto implementation, but we don't really expect to have regular expressions in `apiPath`, actually the `RegExp` constructor might break for some special characters if they are used in `apiPath`.

apps/quanta/app/utils.ts
```diff
@@ -0,0 +1,17 @@
+import config from '@plone/registry';
+
+/**
+ * Flatten to app server URL - Given a URL if it starts with the API server URL
+ * this method flattens it (removes) the server part
+ * TODO: Update it when implementing non-root based app location (on a
+ * directory other than /, eg. /myapp)
+ * @method flattenToAppURL
+ */
+export function flattenToAppURL(url: string | undefined): string | undefined {
```
pnicolli, line 10: I would argue that with our current policy of handling all the backend requests through our server, we don't need this anymore. Or, better, we should make it so that we don't need this.

packages/cmsui/index.ts
```diff
@@ -0,0 +1,42 @@
+import type { ConfigType } from '@plone/registry';
+
+export default function install(config: ConfigType) {
+  // Translation factory
+  config.registerUtility({
+    name: 'translation',
+    type: 'factory',
+    method: (id: string) => id,
+  });
```
pnicolli, line 9: I'm not sure this belongs in this package, but let's say it does, for now

pnicolli (approved):

PR #6798, opened 2025-03-03
File widget a11y

Wagner3UB:
The dropdown file widget isn’t properly reading the name of the field and is only reading the message in the content. Ideally, it should first read the name of the field, like ‘Header Image’, followed by the message, such as ‘Drop file here to replace the existing file’, or any of the other messages already set up.

PS. The changes made in this PR will also be applied to the ‘main’ branch as soon as the PR for the current version is approved.

### DIFF
--- a/news/6798.bugfix
+++ b/news/6798.bugfix
@@ -1,0 +1,1 @@
+Improved the accessibility of the ‘FileWidget’ component by adding ‘aria-labelledby’ and ‘aria-describedby’ to the dropzone area. @Wagner3UB
--- a/src/components/manage/Widgets/FileWidget.jsx
+++ b/src/components/manage/Widgets/FileWidget.jsx
@@ -141,7 +141,7 @@ const FileWidget = (props) => {
                 src={imgsrc}
               />
             ) : (
-              <div id={'abc'} className="dropzone-placeholder">
+              <div className="dropzone-placeholder">
                 <p id="dropzone-description" className="dropzone-text">
                   {isDragActive
                     ? intl.formatMessage(messages.releaseDrag)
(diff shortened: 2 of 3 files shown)


Wagner3UB:
@JeffersonBledsoe @ichim-david Guys, I need help here. This dropzone needs two ‘labels’: one for the field name and one to provide a description for the user. The screen reader should read something like ‘Header image, drop your file here,’ although I think the text is wrong. It should be something like ‘Click or press Enter to choose a file, or drop the file here.’

The issue is that the ‘aria-label’ I set works fine, but the ‘aria-describedby’ doesn’t. I’ve tried many different approaches: inside the ‘getRootProps’, outside, with a , with ‘aria-label’, and I can’t resolve it. Any suggestions?

stevepiercy:
This looks like a good example to emulate. https://ark-ui.com/react/docs/components/file-upload

stevepiercy:
React Aria has another example. https://react-spectrum.adobe.com/react-aria/DropZone.html#filetrigger

wesleybl:
@Wagner3UB I think it would be better to close this PR and open another one for the `main` branch. It would also be good to open an issue.

Wagner3UB:
> @Wagner3UB I think it would be better to close this PR and open another one for the `main` branch. It would also be good to open an issue.

@wesleybl This is a stale PR. It was left behind by me unintentionally.

We already have had merged on main and 18 the same fix:
PR Volto main: https://github.com/plone/volto/pull/7956 - 19.0.0-alpha.27
PR Volto 18: https://github.com/plone/volto/pull/7979 - 18.33.0

I'll close this one, thank you!

wesleybl:
@Wagner3UB I think it's not showing the field name or the keyboard interaction text, as you suggested:

> The screen reader should read something like ‘Header image, drop your file here,’ although I think the text is wrong. It should be something like ‘Click or press Enter to choose a file, or drop the file here.’

Issue #6602, opened 2025-01-18
Volto is missing content alternate hreflangs


erral:
**Is your feature request related to a problem? Please describe.**
When a Volto site is multilingual, we provide language change URLs, but we can also provide them in the form of `<link rel="alternate" .../>` links in the header to let search engines know of alternate versions of the same content in different languages.

**Describe the solution you'd like**
The whole point is to add those links in the `<head>` when the site is multilingual and translations exist.

**Describe alternatives you've considered**
None

**Additional context**
We already [have this in Classic](https://github.com/plone/plone.app.multilingual/blob/master/src/plone/app/multilingual/browser/templates/alternate-languages.pt) and it is an easy task that involves no visual HTML change.

sneridagh:
@erral Question: @tisto pointed me to:

https://www.seoclarity.net/blog/self-referencing-hreflang

Right now, we are not rendering the current one, and it seems that it's something that people do. Do you have any insight in this?

erral:
I will check that, because in our classic sites we also have the same
implementation, so skiping the "current" self link

I will check it with our SEO person.

erral:
It looks like nowadays the recommendation is to present also a "self-link".

I have created a PR to add it.

sneridagh:
@erral it seems that for SSR it does return backend URLs in production 😞

<img width="1196" alt="Image" src="https://github.com/user-attachments/assets/0e900107-3b64-418f-b564-ee16b722ff69" />

We will have to revisit this.

PR #6799, opened 2025-03-03
Fix deployment issue, flattenToAppURL needed

sneridagh:
Closes: https://github.com/plone/volto/issues/6602

PD: In Seven we won't have this any more (hopefully). We could backport the general solution in there to Volto as well.


erral:
This will create URLs without the domain part, which I don't know if it is required

sneridagh:
@erral I think now it covers all the bases. Please take a look!

erral (approved):

erral:
yes, I was also doing a test with `toPublicURL` :+1:

PR #6800, opened 2025-03-03
Display field form errors fix #6787

### LINKED ISSUE #6787: display field form errors in ObjectBrowserWidget.
**Describe the bug**
When using the ObjectBrowserWidget, external errors (e.g. required field) do not display visual alerts in the form field. This prevents the manager from clearly identifying the issues.

**Additional context**
The problem is related to the validateManualLink and onSubmitManualLink methods, where internal errors are not being merged with external errors.
https://github.com/plone/volto/blob/91b2258d64d72f4a3aef7b2298428f06f968281b/packages/volto/src/components/manage/Widgets/ObjectBrowserWidget.jsx#L238-L242

https://github.com/plone/volto/blob/91b2258d64d72f4a3aef7b2298428f06f968281b/packages/volto/src/components/manage/Widgets/ObjectBrowserWidget.jsx#L358

A possible solution is to combine this.state.errors with this.props.error in the render method before passing them to the FormFieldWrapper.

VarshanMaj1:
#6787


davisagli:
This PR is not helpful. You've committed a lot of files that should not be added to the repository.

VarshanMaj1:
my bad, i was facing some issues with few dependencies which screwed up the pr

PR #6772, opened 2025-02-21
Slate FormFieldWrapper to proper identify the field to screen readers

Wagner3UB:

PR #6803, opened 2025-03-04
HtmlSlateWidget to proper identify the field to screen readers

Wagner3UB:
Correction applied to ensure proper screen reader identification of text fields created through the HtmlSlateWidget.

PR volto 17 https://github.com/plone/volto/pull/6772

### DIFF
--- a/packages/volto-slate/news/6803.bugfix
+++ b/packages/volto-slate/news/6803.bugfix
@@ -1,1 +1,1 @@
-To ensure accessibility, an ’aria-labelledby’ attribute was added to the editable part of the ’Slate’ component. @Wagner3UB
+To ensure accessibility, an `aria-labelledby` attribute was added to the editable part of the `Slate` component. @Wagner3UB
--- a/packages/volto-slate/news/6803.bugfix
+++ b/packages/volto-slate/news/6803.bugfix
@@ -1,0 +1,1 @@
+To ensure accessibility, an ’aria-labelledby’ attribute was added to the editable part of the ’Slate’ component. @Wagner3UB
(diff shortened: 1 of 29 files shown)


Wagner3UB:
I need to check the ID for htmlFor. The way I did it created problems for other labels

packages/volto-slate/news/6803.bugfix
```diff
@@ -0,0 +1 @@
+To ensure accessibility, an ’aria-labelledby’ attribute was added to the editable part of the ’Slate’ component. @Wagner3UB
```
davisagli, line 1:
```suggestion
To ensure accessibility, an `aria-labelledby` attribute was added to the editable part of the `Slate` component. @Wagner3UB
```

davisagli (dismissed):

packages/volto-slate/src/editor/SlateEditor.jsx
```diff
@@ -340,6 +341,7 @@ class SlateEditor extends Component {
                 onKeyDown && onKeyDown({ editor, event });
               }}
               {...editableProps}
+              aria-labelledby={`field-${id}`}
```
davisagli, line 344: @Wagner3UB This isn't the same as the change in #6772, which also uses `fieldSet`...what's the reason for the difference?

Wagner3UB: @davisagli No, you are correct, it is the same. In #6772 there was an acceptance test failure, which I have already solved here. I’ll try to take a look today at all my new PRs.

davisagli (approved):

PR #6804, opened 2025-03-04
Token Widget Label

Wagner3UB:
Added aria-label to CreatableSelect inside the TokenWidget to better accessibility.

PR for Volto 17: https://github.com/plone/volto/pull/6786

### DIFF
--- a/packages/volto/src/components/manage/Widgets/__snapshots__/TokenWidget.test.jsx.snap
+++ b/packages/volto/src/components/manage/Widgets/__snapshots__/TokenWidget.test.jsx.snap
@@ -64,6 +64,7 @@ exports[`renders a token widget component 1`] = `
                 >
                   <input
                     aria-autocomplete="list"
+                    aria-label="My field"
                     aria-labelledby="fieldset-default-field-label-my-field"
                     autoCapitalize="none"
                     autoComplete="off"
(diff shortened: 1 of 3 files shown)


packages/volto/src/components/manage/Widgets/__snapshots__/TokenWidget.test.jsx.snap
```diff
@@ -64,6 +64,7 @@ exports[`renders a token widget component 1`] = `
                 >
                   <input
                     aria-autocomplete="list"
+                    aria-label="My field"
                     aria-labelledby="fieldset-default-field-label-my-field"
```
davisagli, line 68: Is it okay to have both aria-label and aria-labelledby?

Wagner3UB: No, you are right. The problem was that this file is already fixed, so the modification I made now was not necessary. I will close this one and leave only v17 open. I will replicate the modification made to this file for the v17 one, to be compliant.

davisagli: @Wagner3UB Ah okay!

PR #6807, opened 2025-03-04
Get `defaultLanguage` from the backend `@site` endpoint

sneridagh:


packages/volto/src/server.jsx
```diff
@@ -224,9 +224,33 @@ server.get('/*', (req, res) => {
 
   loadOnServer({ store, location, routes, api })
     .then(() => {
+      const defaultLanguage =
+        store.getState().site?.data['plone.default_language'] || 'en';
+
+      const supportedLanguages =
+        store.getState().site?.data['plone.available_languages'] || 'en';
```
pnicolli, line 231: Should we maybe default to what's in `config.settings`? I feel like this creates a breaking change otherwise for sites that maybe don't have those keys exported from `@site`? Honestly though I don't know if this is a scenario that is plausible.

packages/volto/src/server.jsx
```diff
@@ -224,9 +224,33 @@ server.get('/*', (req, res) => {
...
+
+      // Load the lang files when they matter
+      // We have to investigate if the L46 block is really necessary then
```
pnicolli, line 234: This is a good question. At a first glance I'd like to remove L46 but it looks like more refactoring would be needed. On the other hand, without removing that part, the import part from this new code looks useless.

davisagli:
@pnicolli Yeah, I would agree with being cautious and aiming to include this in Volto 19.

sneridagh:
@pnicolli @davisagli I do agree, since could be that people really didn't pay attention to the config in the Plone side, then all of a suddent their default languages shift in their sites. Let's leave it for 19 then, make it a (potential) breaking.

sneridagh:
Closing since it was addressed in #7125

PR #6361, opened 2024-10-01
Update a11y guidelines

stevepiercy:
@plone/volto-accessibility please review.

----
📚 Documentation preview 📚: https://volto--6361.org.readthedocs.build/

## Docs review/ updates to make

- We should use https://volto.demo.plone.org for testing
- Review tool recommendations
- Discuss contrast
- Discuss Screen-reader testing
- Should we use "tap", "press", or "click" in language strings for that gesture across desktop and mobile devices?
- How should we denote keystrokes in language strings across operating systems? For example, in MyST:
  ```
  {kbd}`Return`
  ```

PR #6808, opened 2025-03-04
Datetime widget a11y

Wagner3UB:
Improved accessibility for the ‘Datetime-Widget’ component by adding an ‘aria-label’ and a translation for the ‘screenReaderInputMessage’.

PR Volto 17: https://github.com/plone/volto/pull/6793/

### DIFF
--- a/packages/volto/locales/volto.pot
+++ b/packages/volto/locales/volto.pot
@@ -1,7 +1,7 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: Plone\n"
-"POT-Creation-Date: 2024-12-06T13:45:21.266Z\n"
+"POT-Creation-Date: 2025-03-04T11:40:01.749Z\n"
 "Last-Translator: Plone i18n <plone-i18n@lists.sourceforge.net>\n"
 "Language-Team: Plone i18n <plone-i18n@lists.sourceforge.net>\n"
 "Content-Type: text/plain; charset=utf-8\n"
@@ -4833,6 +4833,11 @@ msgstr ""
 #. Default: "Private"
 #: components/manage/Contents/ContentsItem
 msgid "private"
+msgstr ""
+
+#. Default: "Navigate forward to interact with the calendar and select a date. Press the question mark key to get the keyboard shortcuts for changing dates."
+#: components/manage/Widgets/DatetimeWidget
+msgid "publicationDateMessage"
 msgstr ""
 
 #. Default: "Published"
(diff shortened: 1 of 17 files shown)


packages/volto/locales/volto.pot
```diff
@@ -4835,6 +4835,11 @@ msgstr ""
 msgid "private"
 msgstr ""
 
+#. Default: "Navigate forward to interact with the calendar and select a date. Press the question mark key to get the keyboard shortcuts for changing dates."
```
stevepiercy, line 4838: I don't understand. How do I "navigate forward"? Is that a mouse click, tap, or keyboard shortcut?

packages/volto/news/6808.bugfix
```diff
@@ -0,0 +1 @@
+Improved accessibility for the ‘Datetime-Widget’ component by adding an ‘aria-label’ and a translation for the ‘screenReaderInputMessage’. @Wagner3UB
```
stevepiercy, line 1: Replace smart quotes with backticks for inline literal formatting.

```suggestion
Improved accessibility for the `Datetime-Widget` component by adding an `aria-label` and a translation for the `screenReaderInputMessage`. @Wagner3UB
```

stevepiercy (changes requested):
A syntax change on the changelog, and a question about the message.

packages/volto/locales/volto.pot
stevepiercy: Is this the library we're using for a date picker? Sorry, I don't know.

See this line and its key.

https://github.com/react-dates/react-dates/blob/b7bad38dcff024d374ee98f972b55b3de9e61289/src/defaultPhrases.js#L26C7-L26C44

```js
const keyboardForwardNavigationInstructions = 'Navigate forward to interact with the calendar and select a date. Press the question mark key to get the keyboard shortcuts for changing dates.';
```

We also have existing messages for what appears to be the same thing.

It has a backward counterpart in the next line.

`Move backward to switch to the previous month.`

Screenshot 2025-03-04 at 1 22 03 PM

I assume that navigation messages should be more descriptive and only for the icon. If so, then I'd use something like the following messages.

- back: `Tap or left arrow key to navigate backward in the calendar. On a selected date, tap or enter or return key to pick it.`
- forward: `Tap or right arrow key to navigate forward in the calendar. On a selected date, tap or enter or return key to pick it.`
- info: `Tap or question mark key to reveal keyboard shortcuts. Tap the close or "X" icon or the esc key to hide keyboard shortcuts.`

I'm not sure whether we've adopted "tap", "press", or "click" for that gesture to satisfy both desktop and mobile UIs. Another thing for the [accessibility guide](https://github.com/plone/volto/pull/6361)! I'm also not sure whether we do anything to the names of keys to make them more accessible, such as in MyST I would do the following.

```
{kbd}`Return`
```

davisagli:
Also needs an update to the unit test snapshots

PR #6811, opened 2025-03-05
Add validator for `default_language` in language control panel

sneridagh:
Fixes the use case that you can save a new default language without adding it to the list of available languages.
Now it complains, on blur and on save.
Probably, it would need a server side validator too.

Depends on https://github.com/plone/plone.i18n/pull/73 but not requires it.

Also fixes a section of the docs that was incorrect.

----
📚 Documentation preview 📚: https://volto--6811.org.readthedocs.build/

### DIFF
--- a/packages/volto/locales/volto.pot
+++ b/packages/volto/locales/volto.pot
@@ -1,7 +1,7 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: Plone\n"
-"POT-Creation-Date: 2024-12-06T13:45:21.266Z\n"
+"POT-Creation-Date: 2025-03-05T10:53:16.765Z\n"
 "Last-Translator: Plone i18n <plone-i18n@lists.sourceforge.net>\n"
 "Language-Team: Plone i18n <plone-i18n@lists.sourceforge.net>\n"
 "Content-Type: text/plain; charset=utf-8\n"
@@ -3763,6 +3763,11 @@ msgstr ""
 #. Default: "The registration process has been successful. Please check your e-mail inbox for information on how activate your account."
 #: components/theme/Register/Register
 msgid "The registration process has been successful. Please check your e-mail inbox for information on how activate your account."
+msgstr ""
+
+#. Default: "The selected default language must be in the list of the field "Available languages""
+#: helpers/MessageLabels/MessageLabels
+msgid "The selected default language must be in the list of the field "Available languages""
 msgstr ""
 
 #. Default: "The site configuration is outdated and needs to be upgraded."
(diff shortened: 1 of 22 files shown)


packages/volto/locales/volto.pot
```diff
@@ -3765,6 +3765,11 @@ msgstr ""
 msgid "The registration process has been successful. Please check your e-mail inbox for information on how activate your account."
 msgstr ""
 
+#. Default: "The selected default language must be in the list of the field "Available languages""
+#: helpers/MessageLabels/MessageLabels
+msgid "The selected default language must be in the list of the field "Available languages""
```
stevepiercy, line 3770: Escape double quotes.

```suggestion
msgid "The selected default language must be in the list of the field \"Available languages\""
```

packages/volto/locales/volto.pot
```diff
@@ -3765,6 +3765,11 @@ msgstr ""
 msgid "The registration process has been successful. Please check your e-mail inbox for information on how activate your account."
 msgstr ""
 
+#. Default: "The selected default language must be in the list of the field "Available languages""
```
stevepiercy, line 3768: Escape double quotes.

```suggestion
#. Default: "The selected default language must be in the list of the field \"Available languages\""
```

stevepiercy (changes requested):

stevepiercy (approved):

packages/volto/locales/volto.pot
stevepiercy: Ah, tooling that gets in the way of what should be. This is fine.

packages/volto/src/config/validation.ts
```diff
@@ -150,6 +151,13 @@ const registerValidators = (config: ConfigType) => {
     dependencies: { behaviorName: 'plone.eventbasic', fieldName: 'end' },
     method: endEventDateRangeValidator,
   });
+
+  config.registerUtility({
+    name: 'default_language',
+    type: 'validator',
+    dependencies: { format: 'default_language' },
```
davisagli, line 158: Can you use `fieldName: 'default_language'` so https://github.com/plone/plone.i18n/pull/73 isn't needed?

sneridagh: We don't have this feature (yet). Do you really think it's the way to go? For me, the problem is the lack of expressivity in the definition of the fields in the backend source, which is not Volto to blame. And it's not only here. It's all over the place in the backend fields definitions.

pnicolli (approved):

davisagli (approved):

PR #6813, opened 2025-03-07
Leaking React19 in `volto-testing` package

sneridagh:
I've discovered that `@plone/volto-testing` is leaking React 19 in its resolutions.


pnicolli:
Thanks! Got bit by this just one hour ago. ❤️

PR #6817, opened 2025-03-07
Fix Form validation ignores backend validator props passed to fields

### LINKED ISSUE #6816: Form Validation ignores backend passed validators for Fields (String, Integer, Array)
**Describe the bug**

Form validation does not recognize validators passed from the backend as widgetProps. This is affecting various field types, including string, integer, and array fields. As a result, constraints like maxLength, minLength, maxItems, and other validation rules are not working.

**To Reproduce**

1. Go to your backend schema implementation and add one of the directives like this one.

```

    directives.widget(
        "seo_description",
        frontendOptions={
            "widgetProps":{
            "maxLength":55,
            }
        },
    )

```

2. Remove the maxLength property from Textwidget(since it doesn't let you type greater than maxLength, you can clearly see in TextAreaWidget more clearly).
3. Go to sidebar and fill the field with words greater than the maxLength.
4. Save the page
5. Page save without any error but it should give error since maxLength is there.

**Expected behavior**
It should throw an error like Title field when it is left empty.

iRohitSingh:
Fix #6816 

screenshot


packages/volto/news/6816.bugfix
```diff
@@ -0,0 +1 @@
+Fix Form validation ignores backend validator props passed to fields @iRohitSingh.
```
stevepiercy, line 1:
```suggestion
For form validation, use the backend validator props passed to fields. @iRohitSingh.
```

sneridagh:
@iFlameing please, add some unit tests for this one.

sneridagh (changes requested):
LGTM, but @iRohitSingh please, add some unit tests for this one.

iRohitSingh:
@sneridagh  Added the unit test.

sneridagh (approved):

davisagli:
@iRohitSingh @sneridagh It shouldn't be necessary to pass maxLength in widgetProps.

1. max_length can be specified on a TextLine field in the backend schema, e.g.
```
my_field = schema.TextLine(
    title="My field",
    max_length = 15,
)
```
2. It's supported as a standard property in JSON Schema (https://cswr.github.io/JsonSchema/spec/basic_types/#string-schemas) and plone.restapi includes it when it serializes the field in the `@types` endpoint:
```
properties: {
    my_field: {
        title: 'My field',
        maxLength: 15
    }
}
```
4. Volto already validates maxLength when it's on the field instead of the widget.

This way has the benefit that the backend also validates the length, in case someone sends a bad value to the API who is not using Volto.

Having said all that, I don't think we necessarily need to revert this change. It's just not clear to me that it is the best way to do it.

PR #6801, opened 2025-03-03
Fix QueryWidget focus issue by checking this.node before calling focus

Haiderkhan64:
This PR fixes an issue where the QueryWidget component attempts to call the focus() method on this.node without ensuring that this.node is defined. This caused a crash when the widget was the first in a form and automatically focused.

Changes
Added a check to ensure this.node is defined before calling focus() in the componentDidMount lifecycle method.

Impact
Prevents the application from crashing when this.node is undefined.

Improves the stability of forms using the QueryWidget.

PR #6819, opened 2025-03-07
Fix QueryWidget focus crash by checking this.node [closes #2281]

### LINKED ISSUE #2281: Can't use "query" as first widget in a form

Haiderkhan64:
**Description**
Fixes a crash in QueryWidget (issue #2281) where `focus()` was called on `this.node` without checking its existence, breaking forms when it’s the first widget. Added a null check.

**Changes**
- Added `this.node` check in `componentDidMount` before `focus()`.
- Included news entry in `news/2281.bugfix`.

**Impact**
- Prevents crashes when QueryWidget is auto-focused.
- Closes #2281.

Replaces #6766 and #6801. @stevepiercy


davisagli:
I can't reproduce this issue this tries to fix.

PR #6824, opened 2025-03-10
Validation for internal blocks.

### LINKED ISSUE #6776: Validation internal blocks onSubmit form.
### Initial issue:
Block validations are not working properly inside inline blocks, such as Grid, Column, and others.

### Root cause:
Validation in the `onSubmit` method inside the Form is only applied to blocks at the root of the structure, meaning that inner blocks inside inline components are not validated.

https://github.com/plone/volto/blob/91b2258d64d72f4a3aef7b2298428f06f968281b/packages/volto/src/components/manage/Form/Form.jsx#L517-L566

### Passing errors from props in Container Edit:
Another issue is that error properties inside ContainerEdit are not being passed to BlockForm.
https://github.com/plone/volto/blob/91b2258d64d72f4a3aef7b2298428f06f968281b/packages/volto/src/components/manage/Blocks/Container/Edit.jsx#L123

## Solution suggestion

A potential solution would be to create a recursive helper function, such as `blockValidator`, that checks for the presence of the `blocks` and `blocks_layout` fields in a block. This function should be able to validate inner blocks up to a depth of level 1, ensuring that inner errors are detected and handled better.

#### Error differentiation:
It is important to differentiate internal errors from regular block errors. One possible approach could be to give internal errors a specific key, such as `@internal_blocks`. This would allow for proper handling, especially when setting up the error toast and moving focus to the correct element.

https://github.com/plone/volto/blob/91b2258d64d72f4a3aef7b2298428f06f968281b/packages/volto/src/components/manage/Form/Form.jsx#L601-L624

#### Visibility of errors in internal block forms:
Another problem arises when internal block forms fail to detect the associated errors due to the dependency on element IDs in the root `blocksErrors` object. A fix could be to add internal block errors to the root structure, (this would be duplicated both at the root and within the parent block with the `@internal_blocks` key) ensuring that they are identified and displayed correctly. (This workaround may not be ideal, but it avoids major changes and significant impacts to the rest of the system.)

alexandreIFB:
Hello, This may be a more drastic change, and there may be better ways to work around the problem. However, I leave the suggestion here for analysis and consideration.

Although the solutions described below have solved the problems encountered, I recognize that there may be more appropriate places within the project architecture to implement these functions. Due to lack of experience with the current architecture, I chose to implement them in the places I considered most appropriate. With that in mind, I open it for comments and analysis and suggestions for improvements and adjustments, if necessary, to better adapt to the project flow.

## Initial Problem:

Block validations are not working correctly within aligned blocks, such as Grid, Column, among others.

## Root of the Problem:

Validation, in the onSubmit method within Form, is being applied only to the blocks that are at the root of the structure. This results in the failure to validate internal blocks within aligned components.

### Internal Block Validation:

To work around the aforementioned problem, an auxiliary function called blockValidator was created, which checks the presence of the blocks and blocks_layout fields in a recursive block.

https://github.com/alexandreIFB/volto/blob/8d45c5bd25a0a85e534af3a7fc27d54820283d83/packages/volto/src/components/manage/Form/Form.jsx#L539-L546

### Error Differentiation:

The second challenge was to identify and differentiate internal errors from common block errors. To do this, an approach was implemented where internal errors are assigned to the @internal_blocks key. This distinction allows error messages to be analyzed and displayed correctly, associating them with their respective internal elements.

### Error Forwarding in Aligned Blocks:

The third problem identified was that errors in aligned blocks (such as Grid, Column) were not being forwarded correctly. To solve this, errors forwarding for aligned blocks were added to the ContainerEdit component, ensuring that it is possible to access 'blocksErrors'.

https://github.com/alexandreIFB/volto/blob/8d45c5bd25a0a85e534af3a7fc27d54820283d83/packages/volto/src/components/manage/Blocks/Container/Edit.jsx#L123-L125

### Visibility of Errors in Internal Block Forms:

The last problem arose when internal block forms were unable to see the errors associated with them, since the validation of the errors depended on the ID of the elements in the root of the 'blockErrors;' object. To solve this, internal block errors were also added to the root of the structure, allowing them to be recognized and displayed correctly. (I didn't find the alternative elegant but I couldn't think of another way without too much impact)

https://github.com/alexandreIFB/volto/blob/8d45c5bd25a0a85e534af3a7fc27d54820283d83/packages/volto/src/components/manage/Form/Form.jsx#L548-L552

Closes #6776

### DIFF
--- a/packages/volto/src/components/manage/Blocks/Container/Edit.jsx
+++ b/packages/volto/src/components/manage/Blocks/Container/Edit.jsx
@@ -121,6 +121,8 @@ const ContainerBlockEdit = (props) => {
       )}
 
       <BlocksForm
+        errors={props.erros}
+        blocksErrors={props.blocksErrors}
         metadata={metadata}
         properties={properties}
         direction={direction}
--- a/packages/volto/src/components/manage/Form/Form.jsx
+++ b/packages/volto/src/components/manage/Form/Form.jsx
@@ -56,6 +56,9 @@ import { setFormData, setUIState } from '@plone/volto/actions/form/form';
 import { compose } from 'redux';
 import config from '@plone/volto/registry';
 import SlotRenderer from '@plone/volto/components/theme/SlotRenderer/SlotRenderer';
+import BlockValidation, {
+  extractFirstMessageToast,
+} from '@plone/volto/helpers/BlockValidation/BlockValidation';
 
 /**
  * Form container class.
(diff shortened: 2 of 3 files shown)


wesleybl:
@alexandreIFB please add an entry in the news folder: https://6.docs.plone.org/contributing/index.html#contributing-change-log-label

wesleybl:
@alexandreIFB Can you please suggest a change to some core block to simulate the problem? Or does some core block that can be inserted into a Grid already require validation?

alexandreIFB:
> @alexandreIFB Can you please suggest a change to some core block to simulate the problem? Or does some core block that can be inserted into a Grid already require validation?

It is possible to simulate with any block that has validation within the Grid. I recommend using the Teaser block that comes natively, as it requires an href.

Another test I performed was the validation within the Column block from the @eeacms/volto-columns-block community.

wesleybl (changes requested):

wesleybl:
> I recommend using the Teaser block that comes natively, as it requires an href.

I can confirm this error on the demo site with the Teaser block.

@alexandreIFB  could you please include this information in the issue as well?

PR #6826, opened 2025-03-11
feat: refactor UniversalLink component using typescript

tomschall:
- Refactor the UniversalLink component using typescript.
- Modify lint-staged.config.js to exclude d.ts files
- Modify tsconfig, include types there, otherwise d.ts files will noch be recognized in .tsx files, e.g. inside components
- Use newest version of classnames


packages/volto/src/components/manage/UniversalLink/UniversalLink.tsx
```diff
@@ -0,0 +1,183 @@
...
+import {
+  flattenToAppURL,
+  isInternalURL,
+  URLUtils,
+} from '@plone/volto/helpers/Url/Url';
+import config from '@plone/volto/registry';
+import cx from 'classnames';
+
+interface UniversalLinkProps {
+  href: string;
```
sneridagh, line 13: href is optional too, it's either href or item.

packages/volto/src/components/manage/UniversalLink/UniversalLink.tsx
```diff
@@ -0,0 +1,183 @@
...
+const UniversalLink = React.memo<UniversalLinkProps>(
+  ({
+    href,
+    item,
+    openLinkInNewTab,
+    download,
+    children,
+    className,
+    title,
+    smooth,
```
sneridagh, line 64: `smooth` comes from the config always below. If you want it to be overridable, it's ok, but then enable it below.

packages/volto/src/components/manage/UniversalLink/UniversalLink.tsx
```diff
@@ -0,0 +1,183 @@
...
+  };
+  intl: {
+    locale: string;
+  };
+  userSession: {
+    token: string;
+  };
+}
+
+const UniversalLink = React.memo<UniversalLinkProps>(
```
sneridagh, line 55: What is the reasoning behind the `React.memo`?

sneridagh (changes requested):
Overall LGTM, just small adjustments and questions.

@plone/volto-team Please can I have another review?

pnicolli:
Love the work on converting this very important component to TS, thanks!

@sneridagh I will definitely review this over the weekend 👍

packages/volto/src/components/manage/UniversalLink/UniversalLink.tsx
tomschall: Ok i use now Union Types here to differentiate between href or item.

packages/volto/src/components/manage/UniversalLink/UniversalLink.tsx
tomschall: It's a performance optimization.

The component gets memoized and if we have for example 50 links in a list and only one changes, with React.memo only the changed link get rendered new.

We can use it here because UniversalLink changes only over props, we have no local states or side effects.

tomschall: I made some tests and negative tests to show this optimized behavior and also show when it's not working (unstable props). We always have to use stable references if we want the optimized behavior, that means not passing inline objects to props or inline jsx to children. 

I made some detailed examples in the tests, maybe we should also document this.

sneridagh (approved):
LGTM!

packages/volto/src/components/manage/UniversalLink/UniversalLink.tsx
```diff
@@ -0,0 +1,205 @@
...
+      navroot: {
+        id: string;
+      };
+    };
+  };
+  intl: {
+    locale: string;
+  };
+  userSession: {
+    token: string;
```
pnicolli, line 60:
```suggestion
    token: string | null;
```

The default value for `token` is `null`. See https://github.com/plone/volto/blob/main/packages/volto/src/reducers/userSession/userSession.js#L14

packages/volto/src/components/manage/UniversalLink/UniversalLink.tsx
```diff
@@ -0,0 +1,205 @@
...
+    item,
+    ...rest
+  } = props;
+  __test.renderCounter();
+
+  const token = useSelector<AppState, string>(
+    (state) => state.userSession?.token,
+  );
+
+  function getUrl(props: UniversalLinkProps, token: string): string {
```
pnicolli, line 89: I'm fine with turning the previous code into a function, if we go this way I would extract the function from the component completely so it's also easily testable. This is not mandatory but would be nice imho.

pnpm-lock.yaml
```diff
@@ -7584,6 +7584,9 @@ packages:
   classnames@2.2.6:
     resolution: {integrity: sha512-JR/iSQOSt+LQIWwrwEzJ9uk0xfN3mTVYMwt1Ir5mUcSN6pU+V4zQFFaJsclJbPuAUQH+yfWef6tm7l1quW3C8Q==}
```
pnicolli, line 7585: I would strongly suggest installing classnames 2.5.1 in @plone/volto-slate as well in order to dedupe the dependency

pnicolli (changes requested):

sneridagh:
@tomschall Hello! Could you please take a look at the comments and suggestions? Thanks!

pnpm-lock.yaml
sneridagh: Done.

packages/volto/src/components/manage/UniversalLink/UniversalLink.tsx
sneridagh: Done. (not tests, but I separated it).

packages/volto/src/components/manage/UniversalLink/UniversalLink.tsx
sneridagh: Done.

sneridagh:
@tomschall I go forward and fixed the suggestions, I hope you won't mind.

PR #6830, opened 2025-03-12
Allow setting the default language via a RUNTIME environment variable

sneridagh:
This PR allows you to setup the default language (`config.settings.defaultLanguage`) from a RUNTIME variable `SITE_DEFAULT_LANGUAGE`.

This is a non-breaking change workaround for 18, since we decided that #6807 will be targeted for 19. 

The idea is to incorporate this also in #6807 since it might be valuable to have both ways of doing it.

### DIFF
--- a/docs/source/configuration/environmentvariables.md
+++ b/docs/source/configuration/environmentvariables.md
@@ -186,7 +186,7 @@ You can also generate builds on your continuous integration, then deploy them an
     ```
 
 `SITE_DEFAULT_LANGUAGE`
-    This is a runtime environment variable that sets the `config.settings.defaultLanguage`, allowing you to specify the default language of a site.
+    This is a runtime environment variable that sets the `config.settings.defaultLanguage`, allowing you to specify the default language of a site. It needs to match the default language that is configured in the backend's Language control panel.
 
     ```shell
     SITE_DEFAULT_LANGUAGE=ca pnpm start
--- a/docs/source/configuration/environmentvariables.md
+++ b/docs/source/configuration/environmentvariables.md
@@ -184,6 +184,13 @@ You can also generate builds on your continuous integration, then deploy them an
     ```shell
     VOLTOCONFIG=../../volto.config.js pnpm start
     ```
+
+`SITE_DEFAULT_LANGUAGE`
+    This is a runtime environment variable that setups the `config.settings.defaultLanguage` allowing you to specify the default language of a site.
+
+    ```shell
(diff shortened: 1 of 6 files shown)


pnicolli (approved):

pnicolli:
I'm not sure it's valuable to have this after we merge #6807 because I kind of want to have people doing less config on the js side, when there are ways to do that on the backend.

stevepiercy (changes requested):
Needs to be added to the docs. https://github.com/plone/volto/blob/main/docs/source/configuration/settings-reference.md

sneridagh:
@pnicolli yes, could be... we will have to think about it. Still, for batch deployments could be useful. And adding something for immediately removing it feels wrong :/ We'll see.

sneridagh:
@stevepiercy sorry! done!

stevepiercy (changes requested):
Little tweak. Otherwise LGTM. Thank you!

packages/volto/news/6830.feature
```diff
@@ -0,0 +1 @@
+Allow setting the default language via a RUNTIME environment variable. @sneridagh
```
davisagli, line 1:
```suggestion
Allow setting the default language via a runtime environment variable. @sneridagh
```

docs/source/configuration/environmentvariables.md
```diff
@@ -184,6 +184,13 @@ You can also generate builds on your continuous integration, then deploy them an
     ```shell
     VOLTOCONFIG=../../volto.config.js pnpm start
     ```
+
+`SITE_DEFAULT_LANGUAGE`
+    This is a runtime environment variable that sets the `config.settings.defaultLanguage`, allowing you to specify the default language of a site.
```
davisagli, line 189:
```suggestion
    This is a runtime environment variable that sets the `config.settings.defaultLanguage`, allowing you to specify the default language of a site. It needs to match the default language that is configured in the backend's Language control panel.
```

davisagli (changes requested):
Overall this looks like a helpful short-term improvement until we get #6807 in Volto 19.

We have an inconsistency in naming. This uses SITE_DEFAULT_LANGUAGE, but the plone-backend image uses SITE_LANGUAGE. We made that change in plone-backend recently and it hasn't been released yet, so we can still change it there. Let's go with SITE_DEFAULT_LANGUAGE, because it's a bit more specific. I'll make PRs for plone-backend.

A few small tweaks...

davisagli (approved):

PR #6834, opened 2025-03-13
fix Form error handling to show field title in Toast

### LINKED ISSUE #6833: Display descriptive error message in the toast for block errors.
**Describe the bug**
The Toast in the Form blockErrors displays the properties name of the field that threw the error, which is not a clear message to the editor. This results in less descriptive error messages.

**To Reproduce**
1. Go to a form with block schema validation [e.g Teaser block].
2. Submit the form without filling in a required field.
3. Note the error message that is displayed.

**Expected behavior**
The error message should display the error field title as defined in the schema, providing a more descriptive reference to the field.

alexandreIFB:
The implementation has been modified to improve error handling in the `Form` component. Now, when encountering an error in a block, the code does the following:

1. Look up the schema of the first block with an error using the block ID.

2. Get the title of the first field with an error from the block schema properties (`blockSchema.properties`).

Aditional:
- Corrected the translation of the error message in Portuguese for better clarity

This ensures that the error message displayed in the `Toast` component is more descriptive and clear to the user, showing the title of the field that caused the error instead of the field ID.

Closes #6833


davisagli (commented):
It's a good idea. I think it might be better to add the title in validateFieldsPerFieldset which already has access to the schema. This would avoid building the schema an additional time.

Specifically we could add it [here](https://github.com/plone/volto/blob/main/packages/volto/src/helpers/FormValidation/FormValidation.jsx#L234) and set `errors[fieldId].title = field.title`

(This might look strange since errors[fieldId] is an array. But Javascript arrays are also objects, and can still have properties set on them in addition to the array items.)

sneridagh (approved):
LGTM

@plone/volto-team please can we have another review?

wesleybl:
@alexandreIFB can you please backport this fix to the 18.x.x branch:

https://github.com/plone/volto/tree/18.x.x

wesleybl:
Thanks @alexandreIFB !

PR #6835, opened 2025-03-14
Fix redirects and 404s

stevepiercy:
----
📚 Documentation preview 📚: https://volto--6835.org.readthedocs.build/


stevepiercy:
This is just routine maintenance on broken links and redirects. I'd rather not clutter up the change log with such tasks, and only care that the docs build and pass linkcheckbroken.

PR #6841, opened 2025-03-15
PoC - mixing Tailwind with react-bootstrap

sneridagh:
😬

image

The code:

```tsx
import Button from 'react-bootstrap/Button';
import { Button as ButtonTW } from '@plone/cmsui/components/Button';

export default function Test() {
  return (
    <div className="flex flex-col text-center items-center justify-center min-h-screen">
      <div className="bootstrap-wrapper">
        <h1>Bootstrap button</h1>
        <Button variant="danger">Primary</Button>
      </div>

      <div className="plone-components-tw-wrapper mt-10">
        <h1>
          @plone/components TW button (Notice this is an TW H1, not styled!)
        </h1>
        <ButtonTW variant="primary" className="mt-4">
          Primary
        </ButtonTW>
      </div>
    </div>
  );
}
```

The magic:

```css
@layer theme, base, components, plone-components, utilities, custom; /* Add custom layer before tailwind defines them */
@layer custom {
  .bootstrap-wrapper {
    @import 'bootstrap/dist/css/bootstrap.min.css';
  }
}

@import 'tailwindcss';
@import '../addons.styles.css';
```

image


pnicolli:
Soooo yes having been there for our custom theme that requires bootstrap, I can tell you it _looks_ that easy, but we definitely need to check a few cases where just nesting didn't work for us back in the day.

For reference, this is our custom bootstrap import to nest it inside `.public-ui` to try and avoid clashing with Sematic-UI: https://github.com/RedTurtle/design-comuni-plone-theme/blob/main/src/theme/bootstrap-override/_bootstrap-italia-site.scss

As you can see, not all bootstrap was nested in `.public-ui`. Some of it is because of scss, meaning we needed the variables files, the mixins and all that stuff to be global. For the rest, the baseline is that nesting won't work for all those styles that apply to `html` and `body`, for example the "reboot" styles that css frameworks usually have. Tailwind has its own reboot styles.
I think this approach is perfectly fine for building public ui themes, because in Seven we should really avoid leaking cmsui styles to the publicui theme (and I'm worries about the TW reset styles mostly). The big thing that needs to be tested is the rendering of blocks in the cmsui, and the rendering of the toolbar in the public ui.

sneridagh:
Indeed, the resets are the painpoint here.
I am expecting that the layers are the game changer, though. Granted, the top level styles the resets apply will have to be moved to the filtering classname and crossfingers. But of course, you are the one that has been dealing with this the last years the most, and this is indeed the worse case scenario. Let's continue exploring!

pnicolli:
@sneridagh I would say this can be closed now, or do you have any other plans for it?

PR #6843, opened 2025-03-18
Fixed redirection logic which now checks for edit permission

pnicolli:
This is both a bugfix and a breaking change imho. It fixes the behavior of the link view, which is checking for the user to be logged in, while in classic plone actually checks for editing permissions.
Since this could be quite a big breaking change for some and not a severe bug, I would argue this should be targeted at Volto 19. Let me know what you think @plone/volto-team


sneridagh:
@pnicolli yeah, a controversial one. I was completely unaware of that difference. One could argue that this is a bugfix, but to play it safe, yes, maybe slate it for 19.

/cc @tisto

sneridagh (approved):

PR #6370, opened 2024-10-03
Volto form block fixes

robgietema:
- [x] I signed and returned the [Plone Contributor Agreement](https://plone.org/foundation/contributors-agreement), and received and accepted an invitation to join a team in the Plone GitHub organization.
- [x] I verified there aren't other open [pull requests](https://github.com/plone/volto/pulls) for the same change.
- [x] I followed the guidelines in [Contributing to Volto](https://6.docs.plone.org/volto/contributing/index.html).
- [x] I succesfully ran [code linting checks](https://6.docs.plone.org/volto/contributing/linting.html) on my changes locally.
- [x] I succesfully ran [unit tests](https://6.docs.plone.org/volto/contributing/testing.html) on my changes locally.
- [x] I succesfully ran [acceptance tests](https://6.docs.plone.org/volto/contributing/acceptance-tests.html) on my changes locally.
- [x] If needed, I added new tests for my changes.
- [x] If needed, I added [documentation](https://6.docs.plone.org/volto/contributing/documentation.html#narrative-documentation) for my changes, either in the Storybook or narrative documentation.
- [x] I included a [change log entry](https://6.docs.plone.org/contributing/index.html#contributing-change-log-label) in my commits.

-----

If your pull request closes an open issue, include the exact text below, immediately followed by the issue number. When your pull request gets merged, then that issue will close automatically.

Closes #

PR #6858, opened 2025-03-24
Fix return type of emailValidator.

### LINKED ISSUE #6857: test on return type of emailValidator fails
it's `string | null`, not `string`

ksuess:
fix #6857

### DIFF
--- a/packages/volto/news/6857.bugfix
+++ b/packages/volto/news/6857.bugfix
@@ -1,0 +1,1 @@
+Fix return type of emailValidator. @ksuess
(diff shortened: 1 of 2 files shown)


packages/volto/news/6857.bugfix
```diff
@@ -0,0 +1 @@
+Fix return type of emailValidator. @ksuess
```
stevepiercy, line 1:
```suggestion
Fix return type of `emailValidator`. @ksuess
```

stevepiercy (changes requested):
MySTify me!

davisagli:
@ksuess Coincidentally this was also fixed in https://github.com/plone/volto/pull/6370 which was merged today, so (I think) this is no longer needed.

PR #6886, opened 2025-03-27
Integrate DatePicker and TimeField from @plone/components into Dateti…

tedw87:
Integrates the new DatePicker and TimeField components from @plone/components into the existing Volto DatetimeWidget.
Converts values between DateValue and ISO strings to maintain compatibility with Volto forms
Tests rendering behavior, value handling, and change events.
Notes:
This is a work in progress.
Final design/styling and full localization are not included yet.

See plone/aurora#94


stevepiercy:
@tedw87 would you please add a comment with what you would like us to do next? I put this PR in draft state, and cross-referenced the original issue.

Issue #6354, opened 2024-09-30
Disable VSCode's format-on-save feature for markdown


stevepiercy:
This should help avoid contributors accidentally reformatting (the default formatter makes changes we don't want, like converting to the wrong quotes).

Is the right location at the project root? I don't use VSCode.

See https://github.com/plone/documentation/pull/1713/

## .vscode/settings.json

```json
{
  "[markdown]": {
    "editor.formatOnSave": false
  }
}
```

aadityaforwork:
I am working on this issue in the pull request #6357

davisagli:
Fixed in #6357

PR #6893, opened 2025-03-27
Don't reformat Markdown or MyST in VSCode

stevepiercy:
See https://github.com/plone/volto/pull/6889/files#diff-0fc41c579fc3fe8ffca36f60af7c3bc6be3a0a0d15b0f7c5d18e0698e785f514.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,6 +21,7 @@ build
 !.yarn/versions
 
 # Other
+.vscode
 /cache
 .yarn/
 .DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -21,7 +21,6 @@ build
 !.yarn/versions
 
 # Other
-.vscode
 /cache
 .yarn/
 .DS_Store
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,5 @@
+{
+  "[markdown]": {
+    "editor.formatOnSave": false
+  }
+}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,5 +1,0 @@
-{
-  "[markdown]": {
-    "editor.formatOnSave": false
-  }
-}
--- a/docs/source/contributing/developing-core.md
+++ b/docs/source/contributing/developing-core.md
@@ -1,10 +1,10 @@
 ---
 myst:
   html_meta:
-    'description': 'Learn how to set up an environment to develop Volto core and the basics of the Volto monorepo.'
-    'property=og:description': 'Learn how to set up an environment to develop Volto core and the basics of the Volto monorepo.'
-    'property=og:title': 'Develop Volto core'
-    'keywords': 'pnpm, monorepo, develop, core, Volto, Plone, frontend, typescript'
+    "description": "Learn how to set up an environment to develop Volto core and the basics of the Volto monorepo."
+    "property=og:description": "Learn how to set up an environment to develop Volto core and the basics of the Volto monorepo."
+    "property=og:title": "Develop Volto core"
+    "keywords": "pnpm, monorepo, develop, core, Volto, Plone, frontend, typescript"
 ---
 
 # Develop Volto core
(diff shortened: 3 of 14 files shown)


pnicolli:
I'm not sure if we should commit the `.vscode` folder. I see the value of it, but unfortunately several devs have personal settings there because they cannot be in a separate file afaik. This means that they have custom plugin configuration for the project, or even custom editor colors, I would not want those commited in the repo.

wesleybl (changes requested):
I don't think it's a good idea to commit IDE configuration. Each user has their own settings. This should be just documentation. If possible, in lint.

stevepiercy:
AFAIK, there is no enforceable automated linter for MyST files in CI or editors. There is for Markdown, but MyST is a superset of Commonmark which is an extension of Markdown. All these files use `.md` as their extension. This is why we don't want every developer to have their own personal preference.

If your editor reformats a `.md` file after you touch it, then your editor should be configured not to do that, and you should revert the reformatting. I don't know how to enforce that without adding this file.

How does VSCode treat a hierarchy of `.vscode/settings.json` files? Most editors allow setting a global or user default that you can override per project.

Finally, we have to stop rushing through merging PRs where these changes occur, even during sprints. It can break rendering of documentation.

sneridagh:
@stevepiercy I also think that it's bad, unless strictly necessary. I myself have custom things in there for Volto (and other projects). People can add a custom theme color (as I do) and mistakenly push it without notice.

Is there any md linter that we can use, so can we detect them on CI? We can document recommended vscode settings, that people can include in their main global settings, instead of per workspace. We will need some of these for Tailwind integration, for example.

stevepiercy:
> How does VSCode treat a hierarchy of `.vscode/settings.json` files? Most editors allow setting a global or user default that you can override per project.

That is, in fact, how VSCode treats settings.

https://code.visualstudio.com/docs/configure/settings

This PR is an improvement that will avoid unintentional reformatting of MyST `.md` files throughout the monorepo, which may cause documentation build warnings, errors, or visual issues when rendered to HTML. It will not prevent a user from manually invoking a reformat command on `.md` files. If a developer mistakenly commits their project scope `.vscode/settings.json` file in a PR, we can also review whether the settings are useful to share or reject them.

stevepiercy:
> Is there any md linter that we can use, so can we detect them on CI?

https://github.com/plone/volto/pull/6893#issuecomment-2762852682

> AFAIK, there is no enforceable automated linter for MyST files in CI or editors. There is for Markdown, but MyST is a superset of Commonmark which is an extension of Markdown. All these files use `.md` as their extension.

> I myself have custom things in there for Volto (and other projects). People can add a custom theme color (as I do) and mistakenly push it without notice.

Are you doing that for user/global or per workspace? IMO, personalized preferences belong in the former and workspace specific settings in the latter.

This PR should not negatively impact your personal settings. Please check out this branch, or just copy the file, and try for yourself.

stevepiercy:
Victor forgot, and then remembered, that we have `make install`, which calls `node packages/scripts/vscodesettings.js`, which actually generates `.vscode/settings.json`. I also forgot that @peralta added this in #6354. (We're both getting forgetful 👴 .) And because it's in `.gitignore`, and I don't develop Volto core or add-ons and invoke `make install`, I thought the settings file was not present. Of course, if you use PyCharm (me) or any other non-VSCode editor, then the generated file has no effect. For these contributors, we need to still tell them to turn off `.md` reformatting.

So I changed course in this PR, which should make everyone a wee bit more aware and happy.

pnicolli:
I understand where you're coming from @stevepiercy, I feel the pain when having issues with formatting and similar things when I review PRs myself. I also did not know there was the `vscodesettings` script, which I find amazing.
FWIW I have also seen other tooling do the same, generating settings for you or documenting how to properly setup the editor, for example the tools that create a Next.js project, but I understand that they solve a different problem. I like the added docs in this PR, thanks for updating this to the current shape.

pnicolli (approved):

PR #6897, opened 2025-03-28
prefix-path with minimal changes

nileshgulia1:
Related PLIP #https://github.com/plone/volto/issues/4290

The idea is to use React-router basename which takes care of 90% of our tasks. The only caveat is how it handles the root path when basename is set. By default <Link to='/'/> will point to /prefix/ (with a trailing slash). RRv6 has a fix for this but its not backported to v5.x.

### DIFF
--- a/packages/volto/razzle.config.js
+++ b/packages/volto/razzle.config.js
@@ -417,6 +417,20 @@ const defaultModify = ({
         ]
       : [];
 
+  //prefix-path
+  const prefixPath = process.env.RAZZLE_PREFIX_PATH || '';
+  if (prefixPath) {
+    if (target === 'web' && dev) {
+      if (config.devServer.devMiddleware)
+        config.devServer.devMiddleware.publicPath = prefixPath;
+      else config.devServer.publicPath += `${prefixPath.slice(1)}/`;
+    }
+    const pp = config.output.publicPath;
+
+    //check if publicPath already has the prefixPath
+    if (pp.indexOf(prefixPath) === -1)
+      config.output.publicPath = `${pp}${prefixPath.slice(1)}/`;
+  }
   return config;
 };
 
--- a/packages/volto/src/components/theme/Image/Image.jsx
+++ b/packages/volto/src/components/theme/Image/Image.jsx
@@ -1,6 +1,10 @@
 import PropTypes from 'prop-types';
 import cx from 'classnames';
-import { flattenToAppURL, flattenScales } from '@plone/volto/helpers/Url/Url';
+import {
+  flattenToAppURL,
+  flattenScales,
+  addPrefixPath,
+} from '@plone/volto/helpers/Url/Url';
 
 /**
  * Image component
@@ -30,7 +34,7 @@ export default function Image({
   attrs.className = cx(className, { responsive }) || undefined;
 
   if (!item && src) {
-    attrs.src = src;
+    attrs.src = addPrefixPath(src);
   } else {
     const isFromRealObject = !item.image_scales;
     const imageFieldWithDefault = imageField || item.image_field || 'image';
@@ -46,9 +50,11 @@ export default function Image({
 
     const isSvg = image['content-type'] === 'image/svg+xml';
     // In case `base_path` is present (`preview_image_link`) use it as base path
-    const basePath = image.base_path || item['@id'];
+    const basePath = addPrefixPath(
+      flattenToAppURL(image.base_path || item['@id']),
+    );
+    attrs.src = `${basePath}/${image.download}`;
 
-    attrs.src = `${flattenToAppURL(basePath)}/${image.download}`;
     attrs.width = image.width;
     attrs.height = image.height;
 
@@ -67,10 +73,7 @@ export default function Image({
       });
 
       attrs.srcSet = sortedScales
-        .map(
-          (scale) =>
-            `${flattenToAppURL(basePath)}/${scale.download} ${scale.width}w`,
-        )
+        .map((scale) => `${basePath}/${scale.download} ${scale.width}w`)
         .join(', ');
     }
   }
--- a/packages/volto/src/helpers/Api/Api.js
+++ b/packages/volto/src/helpers/Api/Api.js
@@ -7,7 +7,10 @@ import superagent from 'superagent';
 import Cookies from 'universal-cookie';
 import config from '@plone/volto/registry';
 import { addHeadersFactory } from '@plone/volto/helpers/Proxy/Proxy';
-import { stripQuerystring } from '@plone/volto/helpers/Url/Url';
+import {
+  stripQuerystring,
+  stripPrefixPath,
+} from '@plone/volto/helpers/Url/Url';
 
 const methods = ['get', 'post', 'put', 'patch', 'del'];
 
@@ -23,13 +26,15 @@ export function formatUrl(path) {
 
   if (path.startsWith('http://') || path.startsWith('https://')) return path;
 
-  const adjustedPath = path[0] !== '/' ? `/${path}` : path;
   let apiPath = '';
   if (settings.internalApiPath && __SERVER__) {
     apiPath = settings.internalApiPath;
   } else if (settings.apiPath) {
     apiPath = settings.apiPath;
   }
+
+  let adjustedPath = path[0] !== '/' ? `/${path}` : path;
+  adjustedPath = stripPrefixPath(adjustedPath);
 
   return `${apiPath}${APISUFIX}${adjustedPath}`;
 }
--- a/packages/volto/src/server.jsx
+++ b/packages/volto/src/server.jsx
@@ -18,7 +18,7 @@ import { CookiesProvider } from 'react-cookie';
 import cookiesMiddleware from 'universal-cookie-express';
 import debug from 'debug';
 
-import routes from '@root/routes';
+import routes from '@plone/volto/routes';
 import config from '@plone/volto/registry';
 
 import { flattenToAppURL } from '@plone/volto/helpers/Url/Url';
@@ -39,7 +39,10 @@ import ErrorPage from '@plone/volto/error';
 import languages from '@plone/volto/constants/Languages.cjs';
 
 import configureStore from '@plone/volto/store';
-import { ReduxAsyncConnect, loadOnServer } from './helpers/AsyncConnect';
+import {
+  ReduxAsyncConnect,
+  loadOnServer,
+} from '@plone/volto/helpers/AsyncConnect';
 
 let locales = {};
 
@@ -68,10 +71,26 @@ const server = express()
   })
   .use(cookiesMiddleware());
 
+if (process.env.RAZZLE_PREFIX_PATH) {
+  server.use(
+    process.env.RAZZLE_PREFIX_PATH,
+    express.static(
+      process.env.BUILD_DIR
+        ? path.join(process.env.BUILD_DIR, 'public')
+        : process.env.RAZZLE_PUBLIC_DIR,
+      {
+        redirect: false, // Avoid /my-prefix from being redirected to /my-prefix/
+      },
+    ),
+  );
+}
+
 const middleware = (config.settings.expressMiddleware || []).filter((m) => m);
 
 server.all('*', setupServer);
-if (middleware.length) server.use('/', middleware);
+if (middleware.length) {
+  server.use('/', middleware);
+}
 
 server.use(function (err, req, res, next) {
   if (err) {
@@ -254,7 +273,11 @@ server.get('/*', (req, res) => {
         <ChunkExtractorManager extractor={extractor}>
           <CookiesProvider cookies={req.universalCookies}>
             <Provider store={store} onError={reactIntlErrorHandler}>
-              <StaticRouter context={context} location={req.url}>
+              <StaticRouter
+                context={context}
+                location={req.url}
+                basename={process.env.RAZZLE_PREFIX_PATH}
+              >
                 <ReduxAsyncConnect routes={routes} helpers={api} />
               </StaticRouter>
             </Provider>
--- a/packages/volto/src/start-client.jsx
+++ b/packages/volto/src/start-client.jsx
@@ -18,7 +18,9 @@ import Api from '@plone/volto/helpers/Api/Api';
 import { persistAuthToken } from '@plone/volto/helpers/AuthToken/AuthToken';
 import ScrollToTop from '@plone/volto/helpers/ScrollToTop/ScrollToTop';
 
-export const history = createBrowserHistory();
+export const history = createBrowserHistory({
+  basename: config.settings.prefixPath ? config.settings.prefixPath : '/',
+});
 
 function reactIntlErrorHandler(error) {
   debug('i18n')(error);
--- a/packages/volto/src/config/index.js
+++ b/packages/volto/src/config/index.js
@@ -97,6 +97,7 @@ let config = {
     // apiPath: process.env.RAZZLE_API_PATH || 'http://localhost:8081/db/web', // for guillotina
     actions_raising_api_errors: ['GET_CONTENT', 'UPDATE_CONTENT'],
     internalApiPath: process.env.RAZZLE_INTERNAL_API_PATH || undefined,
+    prefixPath: process.env.RAZZLE_PREFIX_PATH || '',
     websockets: process.env.RAZZLE_WEBSOCKETS || false,
     // TODO: legacyTraverse to be removed when the use of the legacy traverse is deprecated.
     legacyTraverse: process.env.RAZZLE_LEGACY_TRAVERSE || false,
--- a/packages/volto/src/helpers/Api/APIResourceWithAuth.js
+++ b/packages/volto/src/helpers/Api/APIResourceWithAuth.js
@@ -6,6 +6,7 @@
 import superagent from 'superagent';
 import config from '@plone/volto/registry';
 import { addHeadersFactory } from '@plone/volto/helpers/Proxy/Proxy';
+import { stripPrefixPath } from '@plone/volto/helpers/Url/Url';
 
 /**
  * Get a resource image/file with authenticated (if token exist) API headers
@@ -17,8 +18,8 @@ export const getAPIResourceWithAuth = (req) =>
   new Promise((resolve, reject) => {
     const { settings } = config;
     const APISUFIX = settings.legacyTraverse ? '' : '/++api++';
-
     let apiPath = '';
+
     if (settings.internalApiPath && __SERVER__) {
       apiPath = settings.internalApiPath;
     } else if (__DEVELOPMENT__ && settings.devProxyToApiPath) {
@@ -26,8 +27,13 @@ export const getAPIResourceWithAuth = (req) =>
     } else {
       apiPath = settings.apiPath;
     }
+
+    let path = req.path;
+    //strip prefix if any
+    path = stripPrefixPath(path);
+
     const request = superagent
-      .get(`${apiPath}${__DEVELOPMENT__ ? '' : APISUFIX}${req.path}`)
+      .get(`${apiPath}${__DEVELOPMENT__ ? '' : APISUFIX}${path}`)
       .maxResponseSize(settings.maxResponseSize)
       .responseType('blob');
     const authToken = req.universalCookies.get('auth_token');
--- a/packages/volto/src/helpers/Html/Html.jsx
+++ b/packages/volto/src/helpers/Html/Html.jsx
@@ -129,14 +129,37 @@ class Html extends Component {
             }}
           />
 
-          <link rel="icon" href="/favicon.ico" sizes="any" />
-          <link rel="icon" href="/icon.svg" type="image/svg+xml" />
+          <link
+            rel="icon"
+            href={
+              (config.settings.prefixPath ? config.settings.prefixPath : '') +
+              '/favicon.ico'
+            }
+            sizes="any"
+          />
+          <link
+            rel="icon"
+            href={
+              (config.settings.prefixPath ? config.settings.prefixPath : '') +
+              '/icon.svg'
+            }
+            type="image/svg+xml"
+          />
           <link
             rel="apple-touch-icon"
             sizes="180x180"
-            href="/apple-touch-icon.png"
+            href={
+              (config.settings.prefixPath ? config.settings.prefixPath : '') +
+              '/apple-touch-icon.png'
+            }
           />
-          <link rel="manifest" href="/site.webmanifest" />
+          <link
+            rel="manifest"
+            href={
+              (config.settings.prefixPath ? config.settings.prefixPath : '') +
+              '/site.webmanifest'
+            }
+          />
           <meta name="generator" content="Plone 6 - https://plone.org" />
           <meta name="viewport" content="width=device-width, initial-scale=1" />
           <meta name="mobile-web-app-capable" content="yes" />
--- a/packages/volto/src/helpers/Url/Url.js
+++ b/packages/volto/src/helpers/Url/Url.js
@@ -259,6 +259,36 @@ export function isUrl(url) {
 }
 
 /**
+ * Add prefix path if set in settings
+ * @method addPrefixPath
+ * @param {string} src pathname
+ * @returns {string} prefixed pathname
+ */
+export function addPrefixPath(src) {
+  let url = src;
+  const { prefixPath } = config.settings;
+  if (isInternalURL(src) && prefixPath && !src.startsWith(prefixPath)) {
+    url = prefixPath + src; //add prefixPath to src if it's an internal url and not a static resource.
+  }
+  return url;
+}
+
+/**
+ * strip prefix path particulary from api calls
+ * @method stripPrefixPath
+ * @param {string} src pathname
+ * @returns {string} pathname
+ */
+export function stripPrefixPath(src) {
+  let url = src;
+  const { prefixPath } = config.settings;
+  if (prefixPath && src.match(new RegExp(`^${prefixPath}(/|$)`))) {
+    url = src.slice(prefixPath.length);
+  }
+  return url;
+}
+
+/**
  * Get field url
  * @method getFieldURL
  * @param {object} data


packages/volto/razzle.config.js
```diff
@@ -417,6 +417,20 @@ const defaultModify = ({
         ]
       : [];
 
+  //prefix-path
+  const prefixPath = process.env.RAZZLE_PREFIX_PATH || '';
+  if (prefixPath) {
+    if (target === 'web' && dev) {
```
nileshgulia1, line 423: TODO: prefix assets on prod only. rename vars

packages/volto/src/helpers/Api/Api.js
```diff
@@ -23,14 +26,16 @@ export function formatUrl(path) {
...
-  const adjustedPath = path[0] !== '/' ? `/${path}` : path;
   let apiPath = '';
   if (settings.internalApiPath && __SERVER__) {
     apiPath = settings.internalApiPath;
   } else if (settings.apiPath) {
     apiPath = settings.apiPath;
   }
 
+  let adjustedPath = path[0] !== '/' ? `/${path}` : path;
+  adjustedPath = stripPrefixPath(adjustedPath);
```
sneridagh, line 37: I have doubts about the `stripPrefixPath`... if the deployment is configured also with the path... then the URLs of the backend will match the frontend config, then there will be no need of stripping anything, right? Or I am missing something? True, that then the deployment will be the same as if there will be no prefix, but I don't know if not doing that has any other hidden drawback...

/cc @ericof @davisagli

sneridagh (commented):
Looks good so far! Great polishment, thanks @claudiaifrim and @nileshgulia1 !!

We need to port the Cypress tests, right? I guess that then we will assess more things in there.

I don't get why the root returning an `/` is a problem. Can we catch it in the universal link somehow?

wesleybl:
> We need to port the Cypress tests, right? I guess that then we will assess more things in there.

@sneridagh I think the tests won't work, before everything in #3592 is here, or in other PRs. See: https://github.com/plone/volto/pull/3592#issuecomment-2754312281

sneridagh:
@wesleybl well, the idea is to have the minimum in here that works, we won't know what does work and what fails, and what's left until we do it.

packages/volto/razzle.config.js
wesleybl: @nileshgulia1 @sneridagh I still don't understand the advantage of the prefix working only in production mode. When we are developing the prefix functionality, it is necessary for it to work in development mode. For example, when we are fixing a Cypress test.

sneridagh:
@wesleybl yeah, I would do that. I don't know if @nileshgulia1 is on it already, please ask him just to make sure you don't step on each other.

packages/volto/razzle.config.js
```diff
@@ -414,6 +414,20 @@ const defaultModify = ({
...
+      if (config.devServer.devMiddleware)
+        config.devServer.devMiddleware.publicPath = prefixPath;
+      else config.devServer.publicPath += `${prefixPath.slice(1)}/`;
+    }
+    const pp = config.output.publicPath;
```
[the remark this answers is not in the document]
nileshgulia1: [feab1ae](https://github.com/plone/volto/pull/6897/commits/feab1aefde999b5d764f02adda140939f64f98a2)

packages/volto/src/helpers/Api/Api.js
davisagli: This seems right to me. We strip the frontend prefixPath to get adjustedPath, which is the path relative to the portal root. Then we add apiPath which is the path to the portal root as a backend (API) URL.

davisagli:
I've been testing this some today. If Volto is on a subpath http://somehost/foo, then I would expect that there is often a different website served at http://somehost. So, in seamless mode we should try to also serve the API under the prefix, at http://somehost/foo/++api++, instead of at http://somehost/++api++

I am experimenting with doing this in a new branch: https://github.com/plone/volto/compare/prefix-path-revised...davisagli-prefix-path?expand=1

packages/volto/razzle.config.js
davisagli: I agree, we should try to have the paths be the same between dev and prod, and that seems to already be the case.

wesleybl:
> I've been testing this some today. If Volto is on a subpath http://somehost/foo, then I would expect that there is often a different website served at http://somehost/. So, in seamless mode we should try to also serve the API under the prefix, at http://somehost/foo/++api++, instead of at http://somehost/++api++

I do this with Nginx. Do you mean that Volto itself should do this?

wesleybl:
> I am experimenting with doing this in a new branch: https://github.com/plone/volto/compare/prefix-path-revised...davisagli-prefix-path?expand=1

@davisagli  I tested your branch and it worked! Requests are made to `/foo/++api++`

packages/volto/src/server.jsx
```diff
@@ -69,10 +72,26 @@ const server = express()
   })
   .use(cookiesMiddleware());
 
+if (process.env.RAZZLE_PREFIX_PATH) {
+  server.use(
+    process.env.RAZZLE_PREFIX_PATH,
+    express.static(
+      process.env.BUILD_DIR
+        ? path.join(process.env.BUILD_DIR, 'public')
+        : process.env.RAZZLE_PUBLIC_DIR,
```
wesleybl: Not only the static file middleware needs to be prefixed but other middlewares as well. And the `devProxy` and `ok` middlewares cannot be prefixed.

And as it is, it is not considering overwriting Volto files in the `public` folder of the addon root.

packages/volto/src/server.jsx
```diff
@@ -69,10 +72,26 @@ const server = express()
...
+      {
+        redirect: false, // Avoid /my-prefix from being redirected to /my-prefix/
```
wesleybl, line 87: This causes a 404 in the portal root SSR.

wesleybl:
> I am experimenting with doing this in a new branch: https://github.com/plone/volto/compare/prefix-path-revised...davisagli-prefix-path?expand=1

@davisagli I tested your branch and it worked. Do you plan to make a PR? It would be very welcome!

davisagli:
@wesleybl Thanks for the reminder. I'll try to find time for it soon.

mamico:
Really impressive! In my initial tests everything just worked fine. Just one note for the documentation: in production the build needs to be done with the environment variables set. For example:
```
RAZZLE_PREFIX_PATH=/testme pnpm build
NODE_ENV=production RAZZLE_PREFIX_PATH=/testme node build/server.js
```

Otherwise, the built assets will be missing the prefix, and the site will try to load JavaScript files from `/static/assets/js/client..js` instead of `/testme/static/....`

mamico:
There’s an edge case with prefix paths like /foo/bar that requires a fix in devproxy if you want to support it. This solution works for me.

```
diff --git a/packages/volto/src/express-middleware/devproxy.js b/packages/volto/src/express-middleware/devproxy.js
index e219edfd8..123a97fb8 100644
--- a/packages/volto/src/express-middleware/devproxy.js
+++ b/packages/volto/src/express-middleware/devproxy.js
@@ -85,7 +85,7 @@ export default function devProxyMiddleware() {
         config.settings.proxyRewriteTarget ||
         `/VirtualHostBase/${apiPathURL.protocol.slice(0, -1)}/${
           apiPathURL.hostname
-        }:${apiPathURL.port}${instancePath}/++api++/VirtualHostRoot${config.settings.prefixPath ? '/_vh_' + config.settings.prefixPath.slice(1) : ''}`;
+        }:${apiPathURL.port}${instancePath}/++api++/VirtualHostRoot${config.settings.prefixPath ? config.settings.prefixPath.split('/').filter(Boolean).map(part => '/_vh_' + part).join('') : ''}`;
 
       return `${target}${path.replace(`${config.settings.prefixPath}/++api++`, '')}`;
     },
```

davisagli:
@mamico 

> Just one note for the documentation: in production the build needs to be done with the environment variables set.

I tried to find a way to avoid this, but it's not trivial. Setting `__webpack_public_path__` in the entry point makes it mostly work, but then I'm still having trouble with the CSS files which are extracted at build time using mini-css-extract-plugin. I suggest we document this as is, and maybe see if we can improve it later.

> There’s an edge case with prefix paths like /foo/bar that requires a fix in devproxy if you want to support it. This solution works for me.

Thanks, I added this.

mamico:
> @mamico 
> 
> 
> 
> > Just one note for the documentation: in production the build needs to be done with the environment variables set.
> 
> I tried to find a way to avoid this, but it's not trivial...

Ok. I think could be not a problem. Need only to be documented.

> > There’s an edge case with prefix paths like /foo/bar ...
> 
> Thanks, I added this.

Thanks

docs/source/deploying/prefix-path.md
```diff
@@ -0,0 +1,28 @@
...
+  html_meta:
+    "description": "Serve Volto on a subpath"
+    "property=og:description": "Serve Volto on a subpath"
+    "property=og:title": "Serve Volto on a subpath"
+    "keywords": "Volto, Plone, path, prefix, virtual hosting"
+---
+
+# Serve Volto on a subpath
+
+Volto can be served on a subpath, e.g. `https://mysite.com/subpath`.
```
sneridagh: How are we calling the feature? "prefix-path" or "subpath". I see a discrepancy in several places.

packages/volto/Makefile
```diff
@@ -107,7 +107,7 @@ release-notes-copy-to-docs: ## Copy release notes into documentation
 
 .PHONY: backend-docker-start
 backend-docker-start: ## Starts a Docker-based backend for development
-	docker run -it --rm --name=backend -p 8080:8080 -v volto-backend-data:/data -e SITE=Plone -e ADDONS='$(KGS)' -e SITE_DEFAULT_LANGUAGE='$(SITE_DEFAULT_LANGUAGE)' $(DOCKER_IMAGE)
+	docker run -it --rm --name=backend -p 8080:8080 -v volto-backend-data:/data -e SITE=Plone -e ADDONS='$(KGS)' -e SITE_DEFAULT_LANGUAGE='$(SITE_DEFAULT_LANGUAGE)' -e CORS_ENABLE=1 $(DOCKER_IMAGE)
```
sneridagh: Why the `-e CORS_ENABLE=1`? this is the normal dev backend... It could hide other issues.

packages/volto/Makefile
```diff
@@ -49,7 +49,7 @@ start: ## Starts Volto, allowing reloading of the add-on during development
 
 .PHONY: build
 build: ## Build a production bundle for distribution
-	pnpm i && RAZZLE_API_PATH=http://127.0.0.1:55001/plone pnpm build
+	pnpm i && pnpm build
```
sneridagh, line 52: Uh, is this a typo that was in there since forever?

packages/volto/Makefile
```diff
@@ -174,11 +174,11 @@ deployment-ci-acceptance-test-run-all: ## With a single command, run the backend
 
 .PHONY: cookieplone-acceptance-frontend-prod-start
 cookieplone-acceptance-frontend-prod-start: build-deps ## Start acceptance frontend in production mode for cookieplone tests
-	(cd ../../volto-add-on && RAZZLE_API_PATH=http://127.0.0.1:55001/plone pnpm build && pnpm start:prod)
+	(cd ../../volto-add-on && RAZZLE_INTERNAL_API_PATH=http://127.0.0.1:55001/plone pnpm build && pnpm start:prod)
 
 .PHONY: project-acceptance-frontend-prod-start
 project-acceptance-frontend-prod-start: build-deps ## Start acceptance frontend in production mode for project tests
-	(cd ../../my-volto-app && RAZZLE_API_PATH=http://127.0.0.1:55001/plone yarn build && yarn start:prod)
+	(cd ../../my-volto-app && RAZZLE_INTERNAL_API_PATH=http://127.0.0.1:55001/plone yarn build && yarn start:prod)
```
sneridagh, line 181: Is this being used somewhere nowadays? If not I would remove it. (eg. it has `yarn` in it still)

packages/volto/src/express-middleware/devproxy.js
```diff
@@ -78,17 +81,24 @@ export default function devProxyMiddleware() {
...
         config.settings.proxyRewriteTarget ||
         `/VirtualHostBase/${apiPathURL.protocol.slice(0, -1)}/${
           apiPathURL.hostname
-        }:${apiPathURL.port}${instancePath}/++api++/VirtualHostRoot`;
+        }:${apiPathURL.port}${instancePath}/++api++/VirtualHostRoot${vhSubpath}`;
 
-      return `${target}${path.replace('/++api++', '')}`;
+      return `${target}${path.replace(`${config.settings.prefixPath}/++api++`, '')}`;
     },
+    changeOrigin: true,
```
sneridagh, line 99: Not that I am against it, but why we do this always now?

packages/volto/src/server.jsx
```diff
@@ -18,7 +18,7 @@ import { CookiesProvider } from 'react-cookie';
 import cookiesMiddleware from 'universal-cookie-express';
 import debug from 'debug';
 
-import routes from '@root/routes';
+import routes from '@plone/volto/routes';
```
sneridagh, line 21: Let's remind us that we need to deprecate @root usage in 19.

packages/volto/src/start-client.jsx
```diff
@@ -18,15 +18,20 @@ import Api from '@plone/volto/helpers/Api/Api';
 import { persistAuthToken } from '@plone/volto/helpers/AuthToken/AuthToken';
 import ScrollToTop from '@plone/volto/helpers/ScrollToTop/ScrollToTop';
 
-export const history = createBrowserHistory();
```
sneridagh, line 21: Why is it exported? I can't remember, but it had a reason...

packages/volto/src/components/theme/Image/Image.jsx
```diff
@@ -30,7 +34,7 @@ export default function Image({
   attrs.className = cx(className, { responsive }) || undefined;
 
   if (!item && src) {
-    attrs.src = src;
+    attrs.src = addPrefixPath(src);
```
sneridagh: When will be `addPrefixPath()` needed? Do we need to document that?

sneridagh (commented):
I added some comments.

A final one:
Don't we have to amend also the middlewares? eg. robots.txt?

For the rest LGTM! Great work guys!!!

sneridagh:
A last one, do we need a special documentation note on production deployments?

wesleybl:
> A final one:
Don't we have to amend also the middlewares? eg. robots.txt?

Yes, it is necessary to prefix middleware. See:

https://github.com/plone/volto/pull/3592/commits/5ae6259dc833f5eab06beb73bee2f36a524cb005

At the time, I had problems prefixing `devProxyMiddleware`. Another thing that shouldn't be prefixed is `/ok`. If Volto is in a container, and we're monitoring Volto, the monitoring doesn't know what the prefix is.

davisagli:
@sneridagh @wesleybl I think the middlewares are all handled correctly. Some of them like the robots.txt one match on a path starting with `**/`, so it doesn't matter what the prefix is.

docs/source/deploying/prefix-path.md
davisagli: From chatgpt:

> What would you call a feature where a website is served at a location other than the root path?

Good question — there are a few terms that could describe that situation depending on context. Here are the most common ones:

🧭 General web terminology

Subpath deployment – a general and modern way to say the site is hosted under a path, e.g. example.com/myapp/ instead of example.com/.

“The app supports subpath deployment.”

Subdirectory deployment – similar meaning; emphasizes the directory structure of the URL.

“The docs site runs in a subdirectory.”

Non-root path or non-root mount – explicitly says it’s not at /.

“The site is mounted on a non-root path.”

⚙️ In framework or server contexts

Mount point – used when describing how an app is attached to a web server or WSGI pipeline.

“The Plone instance is mounted at /intranet.”
(Common in Zope, Flask, Express, etc.)

Base path – sometimes refers to the configured URL prefix the app assumes.

“The frontend’s base path is /myapp.”

Context path – Java EE / servlet term, but sometimes borrowed elsewhere.

“The application’s context path is /portal.”

If you’re writing documentation for developers, “subpath deployment” or “mounted under a subpath” are the clearest and most framework-neutral phrases.

packages/volto/Makefile
davisagli: I don't remember but it's probably left from testing with RAZZLE_API_PATH rather than RAZZLE_INTERNAL_API_PATH. Removed in 2288c54e9ccdf703d2bbeef71f0fa80c17972cf1

packages/volto/Makefile
davisagli: No idea, but it really surprised me when I was testing, and it shouldn't be needed these days.

packages/volto/Makefile
davisagli: This is unrelated to prefix paths, but yeah, I think we can remove it.

packages/volto/src/express-middleware/devproxy.js
davisagli: I don't remember why it got included here, but it's needed in cases where the API is at a different origin and enforces a CORS policy.

packages/volto/src/server.jsx
davisagli: Okay but let's not add that to this PR.

packages/volto/src/start-client.jsx
davisagli: @sneridagh We can no longer initialize it at import time, because it depends on config.settings.prefixPath

So if something else needs to get it then we have to provide it a different way than a direct import.

wesleybl:
> I think the middlewares are all handled correctly. Some of them like the robots.txt one match on a path starting with **/, so it doesn't matter what the prefix is.

@davisagli I remember that when we added the portal logo through the control panel, we needed to prefix some middleware for it to work. We need to test this. In any case, I prefer to prefix the middleware, in case some new middleware is created, or an addon creates one.

packages/volto/src/components/theme/Image/Image.jsx
davisagli: It's needed if you have a URL from the backend that will be served via a frontend route (which is done here, for internal image URLs, using the images express middleware ... it matches the route `**/@@images/*`, then calls getAPIResourceWithAuth which strips the prefix path before proxying it to the backend)

davisagli:
> A last one, do we need a special documentation note on production deployments?

@sneridagh I thought I covered this here: 

https://github.com/plone/volto/pull/6897/files#diff-feef314560bd269daadb31268babb163ecab44efa5b18b9d575276ac254417f6R22

davisagli:
> @davisagli I remember that when we added the portal logo through the control panel, we needed to prefix some middleware for it to work. We need to test this.

@wesleybl Aha, fixed in https://github.com/plone/volto/pull/6897/commits/a94fe3dfe6e4e8abb8abcbaec593dee303f0003f

> In any case, I prefer to prefix the middleware, in case some new middleware is created, or an addon creates one.

I see your point. On the other hand, if we don't prefix all the middleware when it is used, then it's easier to handle the exceptions where the middleware shouldn't have a prefix. I would personally like to merge this as is, and then consider this in a separate PR.

wesleybl:
> I see your point. On the other hand, if we don't prefix all the middleware when it is used, then it's easier to handle the exceptions where the middleware shouldn't have a prefix. I would personally like to merge this as is, and then consider this in a separate PR.

@davisagli I still don't see any advantage in not prefixing. I think most cases should be prefixed. Actually, I can only think of `/ok` and `devProxyMiddleware`. I think we'll make fewer mistakes if we prefix. If I didn't know about this PR and wanted to create middleware that only made sense in the context of the portal root, I would create something like `/my-middleware`, not `**/my-middleware`. This doesn't include middleware that might already be created in addons.

But this isn't a block. We can continue like this if you want and discuss it in another issue.

davisagli:
@wesleybl I was thinking about this after my comment and also realizing that it would make it easier to add custom middleware if you don't have to think about the prefix there. So I think we should give it a try the way you showed in https://github.com/plone/volto/commit/5ae6259dc833f5eab06beb73bee2f36a524cb005 -- I just don't want to delay merging this much longer.

docs/source/deploying/prefix-path.md
stevepiercy: "Subpath" works for me, but let's use words and not Latin abbreviations.

```suggestion
A Volto site can be served on a subpath, for example, `https://mysite.com/subpath`.
```

docs/source/deploying/subpath.md
```diff
@@ -0,0 +1,28 @@
...
+    "property=og:description": "Serve Volto on a subpath"
+    "property=og:title": "Serve Volto on a subpath"
+    "keywords": "Volto, Plone, path, prefix, virtual hosting"
+---
+
+# Serve Volto on a subpath
+
+Volto can be served on a subpath, e.g. `https://mysite.com/subpath`.
+
+When this is enabled, both content and static assets are served on the subpath, which means a different website can be served at the root path.
```
stevepiercy: I think this is more explicit.

Can we serve multiple sites on multiple subpaths, or is it limited to just one?

```suggestion
When this is enabled, both content and static assets are served on the subpath.
```

docs/source/deploying/prefix-path.md
```diff
@@ -0,0 +1,28 @@
...
+
+When this is enabled, both content and static assets are served on the subpath, which means a different website can be served at the root path.
+
+To use a subpath, run Volto with the `RAZZLE_PREFIX_PATH` environment variable set:
+
+```shell
+RAZZLE_PREFIX_PATH=/subpath make start
+```
+
+If you are building bundles to run in production mode, the environment variable must be set at both build and run time:
```
stevepiercy:
```suggestion
If you build bundles to run in production mode, you must set the environment variable at both build and run times as shown.
```

docs/source/deploying/prefix-path.md
```diff
@@ -0,0 +1,28 @@
...
+    "keywords": "Volto, Plone, path, prefix, virtual hosting"
+---
+
+# Serve Volto on a subpath
+
+Volto can be served on a subpath, e.g. `https://mysite.com/subpath`.
+
+When this is enabled, both content and static assets are served on the subpath, which means a different website can be served at the root path.
+
+To use a subpath, run Volto with the `RAZZLE_PREFIX_PATH` environment variable set:
```
stevepiercy:
```suggestion
To use a subpath, run Volto with the `RAZZLE_PREFIX_PATH` environment variable set to the subpath.
```

docs/source/deploying/prefix-path.md
```diff
@@ -0,0 +1,28 @@
...
+RAZZLE_PREFIX_PATH=/subpath make start
+```
+
+If you are building bundles to run in production mode, the environment variable must be set at both build and run time:
+
+```shell
+RAZZLE_PREFIX_PATH=/subpath make build && pnpm prod:start
+```
+
+Now you can access http://localhost:3000/subpath in your browser.
```
stevepiercy: Make the URL not be clickable as an inline literal.

```suggestion
In the above example, you would access the site at `http://localhost:3000/subpath` in your browser.
```

packages/volto/news/+subpath.feature
```diff
@@ -0,0 +1 @@
+Add option to serve Volto on a subpath using the `RAZZLE_PREFIX_PATH` environment variable. @nileshgulia1, @wesleybl, @davisagli
```
stevepiercy:
```suggestion
Add option to serve a Volto site on a subpath using the `RAZZLE_PREFIX_PATH` environment variable. @nileshgulia1, @wesleybl, @davisagli
```

packages/volto/news/6976.feature
```diff
@@ -0,0 +1 @@
+Add cypress for prefix-path @wesleybl
```
stevepiercy:
```suggestion
Add Cypress for a subpath. @wesleybl
```

packages/volto/news/7326.feature
```diff
@@ -0,0 +1 @@
+Serve API requests from prefixpath @davisagli
```
stevepiercy:
```suggestion
Serve API requests from a subpath. @davisagli
```

packages/volto/razzle.config.js
```diff
@@ -414,6 +414,22 @@ const defaultModify = ({
         ]
       : [];
 
+  // If Volto is served under a prefix path,
```
stevepiercy:
```suggestion
  // If Volto is served under a subpath,
```

packages/volto/razzle.config.js
```diff
@@ -414,6 +414,22 @@ const defaultModify = ({
...
+  // we have to adjust where Webpack assets are served too.
+  const prefixPath = process.env.RAZZLE_PREFIX_PATH || '';
```
stevepiercy: I would suggest refactoring this `const` to use the documented name of `subpath`. Otherwise when people search for `subpath` in the source code, they won't find anything. Razzle, of course, would continue to use its name of `RAZZLE_PREFIX_PATH`.

```suggestion
  const prefixPath = process.env.RAZZLE_PREFIX_PATH || '';
```

stevepiercy (changes requested):
I would suggest refactoring the `const` of `prefexPath` and other Volto implementations of a subpath to use the documented name of `subpath`. Otherwise when people search for `subpath` in the source code, they won't find anything. Razzle, of course, would continue to use its name of `RAZZLE_PREFIX_PATH`.

docs/source/deploying/subpath.md
davisagli: @stevepiercy I got caught on the wording of your second sentence (if they're different, then it's odd to use the word "both"). We're also making it sound like maybe both of these can be served through one Volto instance, but no, I'm saying it leaves the root path free in case you want to route it to different software via a frontend webserver.

How about:
```suggestion
When this is enabled, both content and static assets are served on the subpath.
This leaves the root path free in case you want to route it to a different website.
```

davisagli: > Can we serve multiple sites on multiple subpaths, or is it limited to just one?

Just one.

packages/volto/razzle.config.js
```diff
@@ -414,6 +414,22 @@ const defaultModify = ({
         ]
       : [];
 
+  // If Volto is served under a subpath,
+  // we have to adjust where Webpack assets are served too.
+  const subpathPrefix = process.env.RAZZLE_PREFIX_PATH || '';
```
wesleybl: Shouldn't the environment variable name be updated as well?

davisagli: Steve is under the impression that it's an existing variable from Razzle that can't be changed, and I didn't point out that it's our own variable that we have the option of changing because, frankly, I'm lazy and don't want to do the work.

sneridagh:
> > A last one, do we need a special documentation note on production deployments?
> 
> @sneridagh I thought I covered this here:
> 
> https://github.com/plone/volto/pull/6897/files#diff-feef314560bd269daadb31268babb163ecab44efa5b18b9d575276ac254417f6R22

@davisagli I was referring to actual deployment example, as in Docker/Traefik config.

davisagli:
> I was referring to actual deployment example, as in Docker/Traefik config.

I'm planning to add a demo site once this is merged and released. I'd like to use that to confirm we have all the pieces right*; then we can add it to the docs and/or point to the demo as an example.

*(It's mostly just setting the environment variable, but we probably have to adjust the traefik rules too)

packages/volto/razzle.config.js
stevepiercy: That's exactly what I thought, but let's be thorough, not lazy, for future us.

docs/source/deploying/subpath.md
stevepiercy: Ah, that almost clears it up for me, as I originally read it as two different Volto sites could be served. Your suggested wording still makes it sound like two different Volto sites to me, though. I don't see the advantage of serving a site on a subpath. I'm missing something important, but I don't know what it is.

stevepiercy:
> I'm planning to add a demo site once this is merged and released. I'd like to use that to confirm we have all the pieces right*; then we can add it to the docs and/or point to the demo as an example.
> 
> *(It's mostly just setting the environment variable, but we probably have to adjust the traefik rules too)

Makes sense to me. Would you please create a new issue, and self-assign?

nileshgulia1:
@davisagli Thank you for taking it forward. I took care of renaming things as per @stevepiercy's suggestions. Let's go go go...🎉

davisagli:
@nileshgulia1 Thank you, I really appreciate it. I'll check it later today.

docs/source/deploying/subpath.md
davisagli: @stevepiercy Here are some scenarios:

1. I want to serve a Drupal website on mycompany.com and a Volto intranet on mycompany.com/intranet (this is the scenario I was trying to describe here)
2. I want to serve a Volto website on mycompany.com, and a different Volto website on mycompany.com/special-event (maybe with a different theme and different Volto version)

docs/source/deploying/subpath.md
```diff
@@ -0,0 +1,28 @@
...
+  html_meta:
+    "description": "Serve Volto on a subpath"
+    "property=og:description": "Serve Volto on a subpath"
+    "property=og:title": "Serve Volto on a subpath"
+    "keywords": "Volto, Plone, path, prefix, virtual hosting"
+---
+
+# Serve Volto on a subpath
+
+A Volto site can be served on a subpath, for example, `https://mysite.com/subpath`.
```
stevepiercy, line 12: I think this gives the context I was missing. @davisagli please let me know.

```suggestion
A Volto site can be served on a subpath, for example, `https://mysite.com/subpath`.
This is useful when you want to leave the root path available for another website that's not the same version or theme of the Volto site served on the subpath.
By using a subpath for your Volto site, you could serve another CMS, a Plone Classic UI site, or even a static site from the root of the hostname.
Routing is typically handled upstream of Volto in a router server, such as Traefik.
```

docs/source/deploying/subpath.md
```diff
@@ -0,0 +1,28 @@
...
+
+When this is enabled, both content and static assets are served on the subpath, which means a different website can be served at the root path.
+
```
stevepiercy:
```suggestion
```

docs/source/deploying/subpath.md
stevepiercy: OK, with the mention of Drupal, as well as Traefik configuration elsewhere in this PR, I think it's important to mention in this section both (1) why this feature would be used, and (2) that such routing is handled upstream of Volto. I've added two new suggestions, and edited this one. Please take another look and refine. Thank you!

stevepiercy (changes requested):

stevepiercy (approved):
Thank you!

stevepiercy:
@datakurre @rioksane did you recently add security to your network to block certain requests to https://tech.blog.jyu.fi/2020/06/plone-volto-hasura-gatsbyjs-mashup/ ? The page serves fine under a GET request, but Lychee, our README linkchecker, reports a network error https://github.com/plone/volto/actions/runs/18709350106/job/53353804838?pr=6897#step:3:98.

stevepiercy:
@datakurre @rioksane I ran through the Lychee troubleshooting guide at https://lychee.cli.rs/troubleshooting/network-errors/ and it appears that your certificate is not valid or recognized per https://lychee.cli.rs/troubleshooting/network-errors/#use-the---insecure-flag.

```shell
$  lychee --insecure packages/volto/README.md 
  126/126 ━━━━━━━━━━━━━━━━━━━━ Finished extracting links                                                                                                                                                                                                                                                                                       Issues found in 1 input. Find details below.

[packages/volto/README.md]:
     [403] https://www.npmjs.com/package/@plone/client | Rejected status code (this depends on your "accept" configuration): Forbidden
     [403] https://www.npmjs.com/package/@plone/components | Rejected status code (this depends on your "accept" configuration): Forbidden
     [403] https://www.npmjs.com/package/@plone/helpers | Rejected status code (this depends on your "accept" configuration): Forbidden
     [403] https://www.npmjs.com/package/@plone/providers | Rejected status code (this depends on your "accept" configuration): Forbidden
     [403] https://www.npmjs.com/package/@plone/registry | Rejected status code (this depends on your "accept" configuration): Forbidden
     [403] https://www.npmjs.com/package/@plone/scripts | Rejected status code (this depends on your "accept" configuration): Forbidden
     [403] https://www.npmjs.com/package/@plone/types | Rejected status code (this depends on your "accept" configuration): Forbidden
     [403] https://www.npmjs.com/package/@plone/volto | Rejected status code (this depends on your "accept" configuration): Forbidden
     [403] https://www.npmjs.com/package/@plone/volto-slate | Rejected status code (this depends on your "accept" configuration): Forbidden
     [403] https://www.npmjs.com/search?q=keywords%3Avolto-addon%2Cvolto | Rejected status code (this depends on your "accept" configuration): Forbidden

🔍 126 Total (in 5s) ✅ 116 OK 🚫 10 Errors
```

To resolve the issue, see https://lychee.cli.rs/troubleshooting/network-errors/#update-the-ca-certificates.

datakurre:
> @datakurre @rioksane I ran through the Lychee troubleshooting guide at https://lychee.cli.rs/troubleshooting/network-errors/ and it appears that your certificate is not valid or recognized per https://lychee.cli.rs/troubleshooting/network-errors/#use-the---insecure-flag.
> 
> ```shell
> $  lychee --insecure packages/volto/README.md 
>   126/126 ━━━━━━━━━━━━━━━━━━━━ Finished extracting links                                                                                                                                                                                                                                                                                       Issues found in 1 input. Find details below.
> 
> [packages/volto/README.md]:
>      [403] https://www.npmjs.com/package/@plone/client | Rejected status code (this depends on your "accept" configuration): Forbidden
>      [403] https://www.npmjs.com/package/@plone/components | Rejected status code (this depends on your "accept" configuration): Forbidden
>      [403] https://www.npmjs.com/package/@plone/helpers | Rejected status code (this depends on your "accept" configuration): Forbidden
>      [403] https://www.npmjs.com/package/@plone/providers | Rejected status code (this depends on your "accept" configuration): Forbidden
>      [403] https://www.npmjs.com/package/@plone/registry | Rejected status code (this depends on your "accept" configuration): Forbidden
>      [403] https://www.npmjs.com/package/@plone/scripts | Rejected status code (this depends on your "accept" configuration): Forbidden
>      [403] https://www.npmjs.com/package/@plone/types | Rejected status code (this depends on your "accept" configuration): Forbidden
>      [403] https://www.npmjs.com/package/@plone/volto | Rejected status code (this depends on your "accept" configuration): Forbidden
>      [403] https://www.npmjs.com/package/@plone/volto-slate | Rejected status code (this depends on your "accept" configuration): Forbidden
>      [403] https://www.npmjs.com/search?q=keywords%3Avolto-addon%2Cvolto | Rejected status code (this depends on your "accept" configuration): Forbidden
> 
> 🔍 126 Total (in 5s) ✅ 116 OK 🚫 10 Errors
> ```
> 
> To resolve the issue, see https://lychee.cli.rs/troubleshooting/network-errors/#update-the-ca-certificates.

Thanks. It appeared that we actually had configuration error for that specific site after we changed our certificate provider and everything changed.

sneridagh:
@nileshgulia1 @wesleybl @davisagli Thanks for pushing for this during the last months!! Finally, we got this in core. Congrats!

Merged and released! 

https://github.com/plone/volto/releases/tag/19.0.0-alpha.9

wesleybl:
@sneridagh @davisagli Can we have this in Volto 18?

davisagli:
@wesleybl No, it's one of the features we are marketing as new in Plone 6.2 / Volto 19. And it's quite a big change, I'm not confident that everything is backwards compatible.

PR #6928, opened 2025-04-01
Storybook for `Container` component

sneridagh:
image

image

image

image


stevepiercy (changes requested):
Will there also be a reference in the main Volto docs, not just its Storybook? I'm not sure where it would go yet. Maybe https://6.docs.plone.org/volto/development/customizing-components.html ?

sneridagh:
@stevepiercy sure, let's talk tomorrow about it. We should start building this branch, instead of main one.

PR #6964, opened 2025-04-09
Fix translations default of ContentsDeleteModal: 'linkintegrity: delete' -> 'delete' if no link to break. 

ksuess:
grafik


davisagli (approved):
Danke!

PR #6976, opened 2025-04-11
Add Cypress prefix tests on CI

wesleybl:
This PR has #6897 as target.

> [!CAUTION]

-----

- [ ] I signed and returned the [Plone Contributor Agreement](https://plone.org/foundation/contributors-agreement), and received and accepted an invitation to join a team in the Plone GitHub organization.
- [ ] I verified there aren't other open [pull requests](https://github.com/plone/volto/pulls) for the same change.
- [ ] I followed the guidelines in [Contributing to Volto](https://6.docs.plone.org/volto/contributing/index.html).
- [ ] I succesfully ran [code linting checks](https://6.docs.plone.org/volto/contributing/linting.html) on my changes locally.
- [ ] I succesfully ran [unit tests](https://6.docs.plone.org/volto/contributing/testing.html) on my changes locally.
- [ ] I succesfully ran [acceptance tests](https://6.docs.plone.org/volto/contributing/acceptance-tests.html) on my changes locally.
- [ ] I included a [change log entry](https://6.docs.plone.org/contributing/index.html#contributing-change-log-label) in my commits.

-----

Closes #


wesleybl:
@nileshgulia1 @sneridagh Can we merge this and go fixing the errors?

nileshgulia1:
> @nileshgulia1 @sneridagh Can we merge this and go fixing the errors?

@wesleybl Since this is based on new prefix branch, I think we should make it work with cypress(and CI) before pushing into feature.

wesleybl:
@nileshgulia1 I think CI tests will only pass when everything is implemented.

nileshgulia1:
@davisagli @wesleybl Since we set the baseUrl in cypress, the tests like these will eventually fail: https://github.com/plone/volto/blob/05e4cbd95207331c957f94261f5ae5620f0931dc/packages/volto/cypress/tests/core/content/content.js#L225 because the assertion stands wrong. 

We need to update the tests to use prefixPath especially for the links that hardcoded or maybe test in a way that takes care of both the scenarios.

nileshgulia1:
@wesleybl I took care of most of the tests. The remaining failing tests I see are not very much related to prefix feature, I think they are flaky which should be fixed in some or the other way. Could you take a look please? Thank you.

wesleybl:
> @wesleybl I took care of most of the tests. The remaining failing tests I see are not very much related to prefix feature, I think they are flaky which should be fixed in some or the other way. Could you take a look please? Thank you.

@nileshgulia1 routes must be prefixed for SSR to work with prefixes. See:

https://github.com/plone/volto/commit/149c8998babe318387722f9b7d26240c1c8615e1

nileshgulia1:
> > @wesleybl I took care of most of the tests. The remaining failing tests I see are not very much related to prefix feature, I think they are flaky which should be fixed in some or the other way. Could you take a look please? Thank you.
> 
> 
> 
> @nileshgulia1 routes must be prefixed for SSR to work with prefixes. See:
> 
> 
> 
> https://github.com/plone/volto/commit/149c8998babe318387722f9b7d26240c1c8615e1

We are using this in our configuration too but then not using the RR basename. iirc, the entire point of using the basename is avoiding the manual prefixes and the links.
You really think manual prefixing routes could solve the underlying problem?

wesleybl:
> We are using this in our configuration too but then not using the RR basename. iirc, the entire point of using the basename is avoiding the manual prefixes and the links.
You really think manual prefixing routes could solve the underlying problem?

@nileshgulia1 Not necessarily. The problem here is that if the routes aren't prefixed, the components' `asyncConnects` won't execute in SSR. This happens because the URL won't match the route. For example, when accessing:

```
http://localhost:3000/foo/search
```

In SSR, the location is `/foo/search` but the route is `/search`. So the match doesn't occur and the `asyncConnect` doesn't execute.

When the request is made by the client, the `basename` removes the `/foo` prefix, and the requests work.

The [loadOnServer](https://github.com/plone/volto/blob/4ef468c57298038dd0df623b6abef4efdbaf7b82/packages/volto/src/server.jsx#L224) handles the `asyncConnect` executions. To determine whether the `asyncConnect` should execute, it matches the `location` it receives with the routes. But at this point, the `location` still has the prefix. So, if we remove the `location` prefix before passing it to `loadOnServer`, the route will match and `asyncConnect` will be executed. I was hesitant to change `location` here, so I prefixed the routes. But thinking about it, I don't think removing the `location` prefix would have any side effects.

PR #6988, opened 2025-04-14
Add username field to user creation in registration form

### LINKED ISSUE #6987: Registration form fails with 404 in Volto
**Describe the bug**
The registration form in Volto is failing because it doesn't include the required `username` field when submitting to the backend API, resulting in a 404 error. The form currently only sends `fullname` and `email` fields, but the backend API requires an explicit `username` field.
When submitting the registration form, the following error is returned:
```json
{
  "error": {
    "errors": [
      {
        "field": "username", 
        "message": "Property 'User Name' is required."
      }
    ], 
    "message": "Error in fields. Property 'User Name' is required.", 
    "type": "WrongParameterError"
  }
}
```

**To Reproduce**

1. Navigate to the registration page (/register)
2.  in the fullname and email fields
3.  the form
4.  the 404 error in the network tab

**Expected behavior**
Since the email field description states "This will be your login name", the form should automatically use the email as the username when submitting to the API:

**Screenshots**

Image

- Volto 18.8.1
-  Plone 6.1.1
- restapi 9.13.3
-  3.7
-  5.13
-  3.12.9 (main, Mar 17 2025, 23:25:08) [GCC 12.2.0]
-  11.0.0 (Pillow)

rohnsha0:
fixes #6987

### DIFF
--- a/packages/volto/src/components/theme/Register/Register.jsx
+++ b/packages/volto/src/components/theme/Register/Register.jsx
@@ -83,6 +83,7 @@ const Register = () => {
     const { fullname, email } = data;
     dispatch(
       createUser({
+        username: email,
         fullname: fullname,
         email: email,
         sendPasswordReset: true,


davisagli (changes requested):
This is not the correct way to solve this.

The Plone backend supports 2 ways of creating usernames depending on the use_email_as_login setting.

If it's false, then the user must choose their own username when they register. We have to add a field for this input.

If it's true, then the user cannot choose their own username, and it is not accepted by the API for self registration. The backend will automatically use the email as the username.

rohnsha0:
@davisagli i was not sure, why i was unable to fetch the registry records... but i have pushed whatever i got from res online!

PR #6991, opened 2025-04-14
Laggy UI response while editing - Test: try Jotai instead of Redux.

sneridagh:
It's been some time that I wanted to take a look at it, because in dev, the performance problems are quite noticeable, also, we have reports that in old machines it's unbearable... Luckily, it seems that in production builds, it's not that bad...

Anyways, I think we should be doing something in a non-performant way in the Redux side. Also, the fact that on every key press, all the store gets replaced with that big object, does not help...
So I did a quick try, and I replaced Redux with a Jotai atom and the results are quite good:

Before ( I was typing non-stop all the time...):

https://github.com/user-attachments/assets/53cc1ea3-edbd-4e7d-91de-698342747e7b

After:

https://github.com/user-attachments/assets/cca00476-36a3-47a8-b7b2-a75c4a0b6235

So, I'd try to find a way to improve the situation using Redux, because anything else would be a breaking change. Maybe reconsider the solution for 19.

@robgietema could you please take a look?

### DIFF
--- a/packages/volto/src/components/manage/Form/Form.jsx
+++ b/packages/volto/src/components/manage/Form/Form.jsx
@@ -33,7 +33,7 @@ import cloneDeep from 'lodash/cloneDeep';
 import xor from 'lodash/xor';
 import isBoolean from 'lodash/isBoolean';
 import PropTypes from 'prop-types';
-import React, { Component } from 'react';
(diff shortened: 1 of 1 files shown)


davisagli:
Closing, since @sneridagh said it didn't make a huge improvement. But maybe we should take another look at how to improve this.

PR #7062, opened 2025-05-11
Improve the vite plugin by adding a relative to absolute path imports

sneridagh:
This only works on registered add-ons.
This is useful in shadows, and makes unnecessary adding the full path in imports every time we want a component to be shadow-able.


sneridagh:
@pnicolli @tiberiuichim I only doubt if we should do some AST parsing in here, just to be sure that it's fast... however, it seems that Vite is superfast already like this, so... Maybe as ToDo for the future.

packages/registry/vite-plugin.js
```diff
@@ -5,6 +5,48 @@ import { createThemeAddonsLoader } from '@plone/registry/create-theme-loader';
...
+      const match = entries.find(([, pkgInfo]) =>
+        normalizedId.startsWith(pkgInfo.basePath),
+      );
+
+      if (!match) return;
+
+      const [, pkgInfo] = match;
+
+      const transformedCode = code.replace(
+        /from\s+['"]((?:\.\.?)\/[^'"]+)['"]/g,
```
pnicolli, line 29: This catches the following cases:

- `from '../test'`
- `from './test'`

And doesn't catch the following ones which are valid syntax:

- `from '.'`
- `from '..'`

We could change it to

```suggestion
        /from\s+['"]((?:\.\.?)\/?[^'"]*)['"]/g,
```

This would fix the issue but would also catch
- `from './'`
- `from '../'`

I think it could be a good change even if not perfect

sneridagh: @pnicolli added suggestion and tests.

pnicolli (approved):

PR #7084, opened 2025-05-16
Add isClearable prop to SelectWidget component

### LINKED ISSUE #7083: Allow configuring if SelectWidget is clearable
Currently, the `SelectWidget` component always allows the user to clear the selected value (`isClearable: true` by default). However, in some use cases, we may want to prevent users from clearing the selection, for example, when a value is required or when clearing does not make sense for the field.

**Why?**
- Provides more flexibility for form requirements.
- Prevents users from clearing required or critical fields.
- Makes the widget more reusable in different scenarios.

alexandreIFB:
This PR adds the `isClearable` prop to the `SelectWidget` component, allowing configuration of whether the select input can be cleared by the user. By default, the select remains clearable for backward compatibility, but now it can be set to `false` when needed. This makes the widget more flexible and suitable for different form requirements.

Closes #7083

### DIFF
--- a/packages/volto/news/7083.feature
+++ b/packages/volto/news/7083.feature
@@ -1,0 +1,1 @@
+add isClearable prop to `SelectWidget`, allowing configuration of whether the select input can be cleared by the user. @alexandreIFB
--- a/packages/volto/news/7083.feature
+++ b/packages/volto/news/7083.feature
@@ -1,1 +1,1 @@
-add isClearable prop to `SelectWidget`, allowing configuration of whether the select input can be cleared by the user. @alexandreIFB
+Add isClearable prop to `SelectWidget`, allowing configuration of whether the select input can be cleared by the user. @alexandreIFB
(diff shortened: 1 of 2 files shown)


wesleybl:
@alexandreIFB Always put "Close #xxx" in the description, so when the merge is done, the issue will be automatically closed.

wesleybl (approved):
LGTM

packages/volto/news/7083.feature
```diff
@@ -0,0 +1 @@
+add isClearable prop to `SelectWidget`, allowing configuration of whether the select input can be cleared by the user. @alexandreIFB
```
wesleybl, line 1:
```suggestion
Add isClearable prop to `SelectWidget`, allowing configuration of whether the select input can be cleared by the user. @alexandreIFB
```

stevepiercy:
You may ignore the failed changelog verifier.

@alexandreIFB @wesleybl do either of you have merge permission? I think it is safe to do so, but maybe the changelog verifier blocks you?

PR #7099, opened 2025-05-21
Fix ImageWidget to support DX fields and Relation Choice images (preview_image_link)

sneridagh:
Turns out that it was not working for content types fields.
image

image

I also added support for Relation Choices field (preview_image_link) fields.


sneridagh:
@davisagli ready too.

pnicolli (approved):

PR #7109, opened 2025-05-24
fix/removed etag generation for html response

### LINKED ISSUE #7094: Express server adds an etag header on served HTML
**Describe the bug**

This is the default behavior of express. The problem is that it is generated based on the sha1 of HTML body, which is always different. The main cause is the serialized `window.__data` which, on every request, it is different because:

- it's an object and the ordering of its keys is not consistent
- it includes the router redux state, which generates a random unique key every time

**To Reproduce**

Curl any Volto generated HTML page, look for the etag header. Make another similar request. Notice that the two etag headers are different.

**Expected behavior**

The etag should be the same if there hasn't been changes to the page.

**Screenshots**

Image

- Volto Version [e.g. 17.xx]

**Additional context**

This makes it impossible to benefit from caching of Volto's generated HTML

Abhishek-17h:
> [!CAUTION]

-----

-----

Closes #7094

### DIFF
--- a/packages/volto/news/7048.feature
+++ b/packages/volto/news/7048.feature
@@ -1,0 +1,1 @@
+Use translated schema title in Helmet for `Add` component page. @alexandreIFB
--- a/packages/volto/news/7096.bugfix
+++ b/packages/volto/news/7096.bugfix
@@ -1,0 +1,1 @@
+fix(fetchContent): correctly handle undefined blocksType in async fetchContent @nileshgulia1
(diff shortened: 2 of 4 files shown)


Abhishek-17h:
@tiberiuichim Instead of completely disabling estag , we’ve now disabled it only for HTML responses, since it wasn’t effective due to the dynamic nature of the HTML content.
I attempted a complete solution by trying to generate a consistent ETag, but the ordering of object keys in window.__data is not reliable and I didn't found any way to make it possible after trying multiple option.
Even though we can make redux router key static ,as it is not required in SSR of page but it is not sufficient to make estag generation consistent.

Issue #7108, opened 2025-05-23
🚨 Critical compatibility break in ObjectListWidget after Volto upgrade: site becomes unusable if schemaName does not exist


alexandreIFB:
After upgrading Volto from version 18.12.1 to 18.22.0, we identified a change that broke compatibility in the `ObjectListWidget` component when using the `schemaName` prop. If the schema utility is not registered in the registry, the following error occurs:

```
TypeError: Cannot read properties of undefined (reading 'undefined')
_Config.getUtility
node_modules/@plone/registry/dist/index.js:324
```

**How to reproduce:**
1. Use `ObjectListWidget` with a `schemaName` prop that is not registered in the Volto registry.
2. Render the component.

**Expected behavior:**  
The component should gracefully fallback to the provided `schema` prop or handle the missing utility safely, without breaking the application.

**Actual behavior:**  
The code tries to access `.method` of an undefined object, causing a runtime error and making the site unusable.

**Relevant code:**
```jsx
const schema = schemaName
  ? config.getUtility({
      type: 'schema',
      name: schemaName,
    })?.method || props.schema
  : props.schema;
```

**Suggested fix:**  
Add a safe check to ensure that `config.getUtility` returns a valid object before accessing `.method`, using optional chaining (`?.`) or a more robust fallback.

**Additional context:**  
It is a compatibility break introduced between versions 18.12.1 and 18.22.0.

PR #7113, opened 2025-05-26
Fix getUtility when called with a utility type that hasn't been registered

davisagli:
Fixes https://github.com/plone/volto/issues/7108

@sneridagh Do you think this is the right way to handle it?

Instead we could raise a more explicit error that the utility hasn't been registered. That might make the problem more obvious if the developer misspelled the utility type. But it would mean the calling code has to explicitly catch the error if it's okay for the utility to be missing (like this case).

We could also add queryUtility like we have in the backend (getUtility raises if missing, queryUtility returns an empty object)

### DIFF
--- a/packages/registry/news/7108.bugfix
+++ b/packages/registry/news/7108.bugfix
@@ -1,0 +1,1 @@
+`getUtility` returns an empty object if called with a utility type that has not been registered. @davisagli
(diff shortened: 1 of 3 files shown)


sneridagh (approved):
In other places we also returned empty instead of raising anything... I guess it's ok sticking to that. We could at some point change to being more explicit, or as you said rely on an alternate method for that `query...` but I never felt in the past that we needed it at all...

PR #7143, opened 2025-05-29
Added cmsui @search loader

pnicolli:
Still missing tests and a better path query handling


deodorhunter (approved):
LGTM, using it successfully in object widget branch

Issue #6670, opened 2025-02-05
[Seven] - `BodyClass` helper replacement


sneridagh:
> [!IMPORTANT]
> If you are not a member of the Volto Team or Developers Team in the Plone GitHub organization, then do not work on or comment on this issue.

The `BodyClass` helper was built using the old React library `react-side-effect` in order to get it's set of features.

This library does not work anymore in latest SSR implementations, and we need to find a replacement, so we can't have the `body` class names as a by product of the rendering of a component.

I assume that we could find a similar way to declare body classes using a `@plone/registry` utility, then in the root route assemble and apply them.

davisagli:
I would check how RR7 handles [links](https://reactrouter.com/start/framework/route-module#links) and [meta](https://reactrouter.com/start/framework/route-module#meta) and see if we can do something similar.

sneridagh:
@davisagli yeah, I've thought about that, but they run per route not per component.

davisagli:
@sneridagh We have to check but I guess that can cover 90% of what we use it for.

PR #7157, opened 2025-05-30
[Seven] sitemap

ksuess:
Closes https://github.com/issues/assigned?issue=plone%7Cvolto%7C6695 #6670 

----
📚 Documentation preview 📚: https://volto--7157.org.readthedocs.build/

----
📚 Documentation preview 📚: https://plone-registry--7157.org.readthedocs.build/

### DIFF
--- a/packages/publicui/routes/index.tsx
+++ b/packages/publicui/routes/index.tsx
@@ -111,7 +111,9 @@ export default function Index() {
                   location={location}
                 />
               </header>
-              <Outlet />
+              <div className="content-area">
+                <Outlet />
+              </div>
               <footer id="footer">
                 <SlotRenderer
                   name="footer"


ksuess:
I moved the slot renderers for header and footer from 'App' to the 'publicui' layout. With this change the header and footer is on all publicui routes, not only on 'content'.

packages/layout/components/App/App.tsx
```diff
@@ -18,9 +18,6 @@ const App = (props: AppProps) => {
         <SlotRenderer name="header" content={content} location={location} />
       </header>
       <SlotRenderer name="main" content={content} location={location} />
-      <footer id="footer">
```
sneridagh: I'd bring back the footer in here, since it's valuable for the `App` component itself, in case somebody wants to use it anyways. It serve its purpose in Storybook story too. To make it more obvious: we won't be using it in Seven, since the Route outlet needs to be rendered in the middle (content).

sneridagh:
@pnicolli @danalvrz @ksuess merging this one as well!! Please check if there's something wrong even after merging.

Issue #6657, opened 2025-02-05
[CMSUI] - Control Panels


sneridagh:
This is the placeholder issue for implementing the control panels.
You can find one sub-issue for each of them.

PR #7161, opened 2025-05-31
Seven controlpanel overview and schema driven control panels

ksuess:
parent issue:
controlpanels: #6657 

Plone_Site-2

Plone_Site


ksuess:
The icons will be added later. I'd rather tackle more urgent matters.

ksuess:
I'll pause here and continue when the widgets are ready.

davisagli:
This might be too ambitious an idea to bring up, and I'm not sure how to implement it, but it could be nice if it was possible to search the control panel to find a setting (similar to System Settings in macOS).

PR #7173, opened 2025-06-05
Add instructions for navigating and viewing diffs in PULL_REQUEST_TEM…

stevepiercy:
…PLATE.md


stevepiercy:
A change log is not needed, as this is an internal change that shouldn't clutter the Volto change log.

PR #7175, opened 2025-06-05
add default if no filename is supplied in FileView

### LINKED ISSUE #7174: File not downloadable if filename is Empty
**Describe the bug**
I encountered an issue while working on a migration, where we had File contenttypes with no real filename for the actual pdf stored. This leads to the download link not showing up on the view for the file content type. Therefore the file was not properly downloadable.

Sure, this can be easily fixed during the migration as well, by using something else as filename, but I think a fallback in the frontend might not hurt and could also cover some other edge cases. 

So I suggest implementing falling back to something like just "File" for the filename in the File View if no filename is supplied.

kittauri:
Add a default if there is an empty filename string.

Closes #7174

### DIFF
--- a/packages/volto/locales/ca/LC_MESSAGES/volto.po
+++ b/packages/volto/locales/ca/LC_MESSAGES/volto.po
@@ -1202,6 +1202,11 @@ msgstr "Document"
 msgid "Download Event"
 msgstr ""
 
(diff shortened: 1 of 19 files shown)


davisagli (approved):
Thanks, looks good to me

PR #7177, opened 2025-06-06
Contents: Add a pill to indicate working copies

davisagli:
Screenshot 2025-06-06 at 9 48 42 AM


packages/volto/src/components/manage/Contents/ContentsItem.jsx
```diff
@@ -178,6 +178,14 @@ export const ContentsItemComponent = ({
                   <FormattedMessage id="Scheduled" defaultMessage="Scheduled" />
                 </Button>
               )}
+            {item.is_working_copy && (
+              <Button className="button-margin working-copy" size="mini">
+                <FormattedMessage
+                  id="Working copy"
+                  defaultMessage="Working copy"
+                />
+              </Button>
```
ksuess, line 187: button or label: I would say this, and the two buttons preceding, are circular labels. Not buttons, as there is no action to take.
See https://react.semantic-ui.com/elements/label/#variations-circular

ksuess (changes requested):

sneridagh:
@davisagli @ksuess  didn't we have one for the expired content? I guess both should be consistent.

packages/volto/src/components/manage/Contents/ContentsItem.jsx
davisagli: I would prefer that too but I was following the pattern of the existing indicators for expired and scheduled content. We can fix it in Seven but I won't change the existing markup here.

ksuess: I understand your argument.

ksuess (approved):

PR #7178, opened 2025-06-06
Make sure make test runs vitest

davisagli:


packages/volto/package.json
```diff
@@ -42,9 +42,8 @@
     "build:types": "tsc --project tsconfig.declarations.json",
     "vitest": "vitest",
     "coverage": "vitest run --coverage",
-    "test": "razzle test --maxWorkers=50%",
-    "test:ci": "CI=true NODE_ICU_DATA=node_modules/full-icu razzle test",
-    "test:husky": "CI=true yarn test --bail --findRelatedTests",
+    "test": "vitest",
```
wesleybl, line 44: Couldn't the vitest script be removed?

wesleybl:
This can't be done in branch 18, right?

davisagli:
This is not ready yet, it depends on updating the cookieplone-templates first

wesleybl:
Even if `cookieplone-templates` were updated, it would still break already created addons.

davisagli:
@wesleybl Right, because existing addons run the `test` command from `@plone/volto`. But this PR is for Volto 19.

wesleybl:
@davisagli Will we have one configuration for Volto 18 and another for Volto 19? Or will cookieplone remove support for Volto 18?

davisagli:
@wesleybl My plan is to update the cookieplone-templates to use vitest for new frontend add-ons in both Volto 18 and 19.

wesleybl:
@davisagli But with this PR, `pnpm --filter @plone/volto test` will not work on both versions. Or this change will have to be made on Volto 18 as well.

davisagli:
@wesleybl The updated frontend addon does not use `pnpm --filter @plone/volto test` to run the tests.

davisagli:
Draft of changes for cookieplone-templates: https://github.com/plone/cookieplone-templates/pull/246

packages/volto/package.json
wesleybl: I prefer not to rely on vitest at the root. See: https://github.com/plone/volto/pull/7274

wesleybl (changes requested):

packages/volto/package.json
wesleybl: oops this is not the root

wesleybl (changes requested):
It is necessary to resolve the conflict.

wesleybl:
[Volto Cookieplone](https://github.com/plone/volto/actions/runs/15495664426/job/43631655245?pr=7178#logs) failed because addons use newer versions. Volto needs to be updated. See: https://github.com/plone/volto/pull/7291

packages/volto/package.json
wesleybl: > Couldn't the vitest script be removed?

Perhaps the `vitest` script should be kept, so that `make test` in root and make it work the same way on Volto 19 and 18?

Although keeping:

"test": "vitest",
"vitest": "vitest",

on Volto 19 is not necessary.

wesleybl (approved):
LGTM.

I resolved the conflict. Now `make test` runs `vitest`.

packages/volto/package.json
wesleybl: I removed the `vitest` script.

wesleybl:
@davisagli Does this still need plone/cookieplone-templates#246 ?

wesleybl:
@davisagli We can change the `Makefile` instead of `package.json`, to stop depending on https://github.com/plone/cookieplone-templates/pull/246.

davisagli:
This was done in https://github.com/plone/volto/pull/7542

PR #7188, opened 2025-06-10
Adopted SCSS plugin and silence the deprecation warnings.

sneridagh:


packages/volto/news/7188.feature
```diff
@@ -0,0 +1 @@
+Adopt Razzke SCSS plugin and silence the deprecation warnings. @sneridagh
```
davisagli:
```suggestion
Adopt SCSS plugin from Razzle and silence the deprecation warnings. @sneridagh
```

pnicolli:
I'm not sure. We're not using scss in core iirc, so why is the silencing needed here? I thought it should be something that's up to devs using scss in their projects. Let's discuss this if you have a use case I can't see.

Example of how we easily customize it for ourselves:

https://github.com/RedTurtle/design-comuni-plone-theme/blob/main/razzle.extend.js#L12

sneridagh:
@pnicolli to be honest I am not 100% sure because of the same reason, but some people complained about it already. 

/cc @fredvd 

the other option would be to document it and point to the solution. (We do something similar in VLT too).

fredvd:
I chatted with @sneridagh about this last week, because I went down a rabbit hole when using a theme add'on that does use SASS. (VLT) The add'on supresses the deprecation warnings using the razzle.extend.js, but in the frontend project add'on  will very likely also use / extend the VLT  SASS. 

Then suddenly the deprecation warnings start showing up again, and if you don't know the current issues around dart-sass, you might get tempted to start fixing the deprecation errors, which will never work. 

We couldn't really figure out 'where' I should have picked up this information: in the core volto upgrade guide for Volto 18, or the volto-light-theme upgrade guide.  The thing is: I didn't upgrade Volto itself, nor did I change my sass.  I only updated the cookieplone scaffold from a pre pnpm version generated January 2024 to cookieplone scaffold 2025. 


---
The current Volto documentation is a bit ambigious as well if this should be part of core, see https://6.docs.plone.org/volto/theming/using-third-party-themes.html#use-sass-loader. 

There's a warning at the sass-loader section: 
"Removed in version Volto: 18
This section is no longer required since Volto 18."

has the support loading sass at all been removed, or do I no longer have to register the sass loader separately?

sneridagh:
@pnicolli @davisagli @fredvd @plone/volto-team I am so over the deprecation warnings, every now and then, they just add another. Sorry I can't bear with the `sass` maintainers. I just cannot understand their point in doing this. When people have projects that really want to migrate to 2.x they will manage the migration on the spot (if that ever happen), not months or years before in their current projects. In the meanwhile leave us alone, guys!!

So, yeah, at the same time that we are working in the razzle fork, I am resurrecting this.

packages/volto/webpack-plugins/webpack-scss-plugin.js
pnicolli, line 1: I would add a comment with a link to the source that this was taken from, as to clarify we're not adding anything new here

sneridagh:
ok, https://github.com/plone/cookieplone-templates/pull/310 will fail until this is merged, and vice-versa :) 
So, we merge that one, then this one.

PR #7196, opened 2025-06-13
Unify implementations, remove custom BreadcrumbsPrimitive, since it's…

sneridagh:
WIP

- Removed `BreadcrumbsPrimitive` customization, since I've found a workaround for what we wanted to accomplish
- Unified implementation CSS/Tailwind

Left to do:
- Slot for breadcrumbs


packages/components/news/7196.breaking
```diff
@@ -0,0 +1 @@
+- Unify `Breadcrumbs` component implementations, remove custom BreadcrumbsPrimitive, since it's no longer needed. @sneridagh
```
stevepiercy: This is tricky grammar. I removed the last comma to associate the final clause with the preceding statement. Otherwise it implies the unification is no longer needed.

```suggestion
- Unify `Breadcrumbs` component implementations, while removing the custom `BreadcrumbsPrimitive` implementation since it's no longer needed. @sneridagh
```

packages/components/src/components/Breadcrumbs/Breadcrumb.stories.tsx
```diff
@@ -38,11 +95,61 @@ export const Default: Story = {
...
     items: [
       { '@id': '/folder', title: 'Folder' },
+      { '@id': '/folder/folderB', title: 'Folder with long name' },
+      {
+        '@id': '/folder/folderB/folderC',
+        title: 'Folder with long name and a bit more',
+      },
+      {
+        '@id': '/folder/folderB/folderC/folderD',
+        title: 'Folder with long name even more long',
```
stevepiercy: This one should be longer than the previous one, not the same length.

```suggestion
        title: 'Folder with the longest folder name ever, please make it stop, aaaaaaaahhhhhh!',
```

stevepiercy (changes requested):

packages/components/src/components/Breadcrumbs/Breadcrumb.stories.tsx
stevepiercy: This is still not "even more long", but the same length as the previous item.

sneridagh: image

It is more longer already. Also, does it _really_ matters so I have to go and spend time amending it? People will already get it, right? If the thing allowed me to apply suggestion, but it does not.

stevepiercy: I don't know if it matters. Maybe it has different behavior at different break points when you have a super long name. Right now you can't test a super long name because the lengths of the strings are the same.

sneridagh: @stevepiercy Good point. The implementation has no breakpoints, no nothing, it just does this, always. Making it breakpoint dependant or dependent on the container width, it's not responsability of this component itself (since it's a primitive) but of the actual integration (and how his parent container choose to behave).

stevepiercy: Yeah, and I also just thought of translations, and what would it look like with German words, such as `Neunmilliardeneinhundertzweiundneunzigmillionensechshunderteinunddreißigtausendsiebenhundertsiebzigfache`.

sneridagh: @stevepiercy FYI, this is just one flavour of the breadcrumbs implementation, not the main one, in case that you need it. We are going away of the "one component does it all" way of doing things. Also subject to continue improving it over time.

sneridagh:
@pnicolli I've implemented the Breadcrumbs slot too. I think it's ready now.

Issue #7163, opened 2025-06-01
Handle code samples in Volto and Seven


stevepiercy:
Code samples in Volto might not be linted or correct. We need to ensure they are both linted and functional.

---

Yeah, both the original and modified code samples don't pass linting. Here's what I did.

1. Check out the Volto repo.
2. Run the following command to install it and its tooling from the repo's root.
   ```shell
   make install
   ```
3. Create a scratch file with the snippet, such as `docs/source/blocks/editcomponent.jsx`.
   ```jsx
   import { SidebarPortal } from '@plone/volto/components';

   const Edit = (props) => {
     const { selected } = props;
     return (
       {/* [....] */}
       <SidebarPortal selected={selected}>
         {/* [....] */}
       </SidebarPortal>
       {/* [....] */}
     )
   };
   ```
4. Run the following command.
   ```shell
   pnpm lint docs/source/blocks/editcomponent.jsx
   ```

   ```console
   /Users/stevepiercy/projects/Plone/documentation/submodules/volto/docs/source/blocks/test.jsx
     7:19  error  Parsing error: ')' expected

   ✖ 1 problem (1 error, 0 warnings)

    ELIFECYCLE  Command failed with exit code 1.
   ```

You can also run other lint commands.

```shell
pnpm prettier docs/source/blocks/test.jsx
pnpm stylelint docs/source/blocks/test.jsx
```

In conclusion, when neither the original nor modified code sample passes quality control, then we shouldn't change it.

---

Just for giggles, I let PyCharm's AI fix the code. It yielded the following code.

```jsx
import { SidebarPortal } from '@plone/volto/components';

const Edit = (props) => {
  const { selected } = props;
  return (
    <>
      {/* [....] */}
      <SidebarPortal selected={selected}>
        {/* [....] */}
      </SidebarPortal>
      {/* [....] */}
    </>
  )
};
```

I then ran it through Volto's linters.

```jsx
import { SidebarPortal } from '@plone/volto/components';

const Edit = (props) => {
  const { selected } = props;
  return (
    <>
      {/* [....] */}
      <SidebarPortal selected={selected}>{/* [....] */}</SidebarPortal>
      {/* [....] */}
    </>
  );
};
```

---

For @sneridagh, code samples that don't pass quality control has been a problem in Volto docs that we've ignored for too long. Here's one way to address it.

1. Create a `.jsx` file with the same name as the `.md` file, such as `docs/source/blocks/editcomponent.jsx`.
[...]

PR #7198, opened 2025-06-15
refactor(docs): consolidate example code samples

ranjan103:
Addresses #7163.

----
📚 Documentation preview 📚: https://volto--7198.org.readthedocs.build/

### DIFF
--- a/docs/source/_examples/blocks/basic/block-examples.jsx
+++ b/docs/source/_examples/blocks/basic/block-examples.jsx
@@ -1,0 +1,383 @@
+/* Block-related examples and constants */
+
+// ===== Constants =====
+// Content example with object browser fields
+export const contentExample = {
+  '@id': 'page-1',
+  related_pages: [],
+  image: { url: "" },
+  link: { href: "" }
+};
+
+// Blocks schema example with object browser widget
+export const blocksSchemaExample = {
+  title: 'Item',
+  fieldsets: [
+    {
+      id: 'default',
+      title: 'Default',
+      fields: ['href'],
+    },
+  ],
+  properties: {
+    href: {
+      title: 'title',
+      widget: 'object_browser',
+      mode: 'link',
+      selectedItemAttrs: ['Title', 'Description'],
+    },
+  }
+};
+
+// Iframe schema example
+export const iframeSchema = {
+  title: 'Embed external content',
+  fieldsets: [
+    {
+      id: 'default',
+      title: 'Default',
+      fields: [
+        'url',
+        'align',
+        'privacy_statement',
+        'privacy_cookie_key',
+        'enabled',
+      ],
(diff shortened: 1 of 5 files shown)


stevepiercy:
@ranjan103 always discuss whether you should perform work before you proceed.

Although I personally would love to have code samples in docs actually tested—similar to what we do in the Python world—and this is an interesting approach, I don't know if the @plone/volto-team is interested in creating or maintaining such an effort.

davisagli (commented):
I agree this looks like an interesting approach. It has a small downside that it makes it harder to read the examples when looking at the docs source, but usually it's better to read them in rendered form anyway. I'm open to accepting this.

stevepiercy:
This PR creates new scripts that I think are unnecessary. We can run all three linters according to the Volto configuration, not just a custom implementation of one of them. See https://6.docs.plone.org/volto/contributing/linting.html and the proposal in the original issue #7163.

To try it out using all three linters, check out this branch, and run linters on the files.

```shell
pnpm lint "docs/source/_examples/**/*.jsx"
pnpm prettier "docs/source/_examples/**/*.jsx"
pnpm stylelint "docs/source/_examples/**/*.jsx"
```

From there, it would be trivial to add a Make command to run any or all of the linters, and therefore automate in CI. The only question I have is whether these commands when run from the root of the repo will pick up the linter configuration inside `packages/volto`, or if there are any other things to consider.

Setting aside the custom scripts, I see value in linting code examples in documentation, and this is a good demonstration of files to test it out on.

Ultimately, it's up to the @plone/volto-team to discuss if they want to put any effort into this approach going forward, or to create a bunch of first-timer issues to change the existing code examples in docs.

sneridagh:
@stevepiercy @davisagli This approach does not allow to write code excerpts, and it will break most of the times because the imports would be incomplete, and ESlint won't be able to get the path for the real build. Let me ask you something? Do we have something like this in place for Python code in the Plone docs? I think we are overreaching in here and we are aiming for something that would make me _don't_ write any piece of code in docs, and ultimatelly, don't want to write docs at all.

stevepiercy:
@sneridagh you've completely changed your enthusiasm from https://github.com/plone/volto/issues/7163.

In the original issue, I mentioned:

> Ultimately I'd love to run something similar to doctests and manuel for Python in documentation code samples, but for JS and JSX.

We do that now in `plone.api`.

I got this from a fellow documentarian that looks like a promising way to run linters on JS and JSX code samples inside the MyST file, not as separate files. Just set up our Volto linters, then use a Python subprocess to run the lint commands.

https://github.com/kai687/sphinxawesome-codelinter

sneridagh:
@stevepiercy I think my enthusiasm continues being the same skeptical one :)

PR #7201, opened 2025-06-16
[Seven] Object browser widget

deodorhunter:
This Draft PR represents the work done at Beethoven Sprint for Object Browser widget Seven integration. 
It's a WIP: still needs implementation and logic porting from old volto version, to be tested with a real backend that uses patternOptions for fields like Related Items, and a Seven version that allows content saving.

Note: a weird behavior was encountered when trying to reuse loaders: if you import a loader inside another loader, whole app breaks and page autoreloads, error is "Process is not defined", so it likely stems from trying to use server side code in a client side environment, maybe from plone/client?

### DIFF
--- a/apps/seven/news/7201.feature
+++ b/apps/seven/news/7201.feature
@@ -1,0 +1,1 @@
+Added a new base route for resetting react router fecthers. @deodorhunter
--- a/packages/client/news/7201.bugfix
+++ b/packages/client/news/7201.bugfix
@@ -1,0 +1,1 @@
+Fixed related items payload schema. @deodorhunter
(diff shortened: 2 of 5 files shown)


deodorhunter:
During Salamina sprint, main features were completed and accessibility is almost complete. Here's a list of known issues and todos

Todos:
- make it usable in plate and blocks via correct contracts (still to be implemented)
- styles can be a whole lot more polished and adherent to quanta styling and wireframes
- focus handling of widget in forms/correct screen reader communication of this complex widget

Known Issues:
- some search/navigation flakiness to be smoothed out
- better translation messages have to be implemented
- grid styles do not display the unselectable status correctly
- grid styles do not display item review_state like list view

PR #7202, opened 2025-06-16
Prevent crash when selection points to deleted node

tomschall:
// ⚠️ Fix for "Cannot find a descendant at path [...]" error
// In certain cases (e.g., after merging blocks via Backspace),
// `editor.selection` may temporarily point to a path that no longer exists.
// This causes `Editor.nodes()` to throw an exception.
// Wrapping it in try/catch prevents crashes and hides the inline toolbar gracefully.
// Reproducible scenario: text block with bullet list → new text block → write a word → go to first position of letters → Backspace → crash.


davisagli:
It looks like this was fixed in https://github.com/plone/volto/pull/7203

PR #7203, opened 2025-06-17
prevent crash when selection points to deleted node

tomschall:
// ⚠️ Fix for "Cannot find a descendant at path [...]" error
// In certain cases (e.g., after merging blocks via Backspace),
// editor.selection may temporarily point to a path that no longer exists.
// This causes Editor.nodes() to throw an exception.
// Wrapping it in try/catch prevents crashes and hides the inline toolbar gracefully.
// Reproducible scenario: text block with bullet list → new text block → write a word → go to first position of letters → Backspace → crash.


packages/volto-slate/src/utils/blocks.js
```diff
@@ -10,6 +10,7 @@ import includes from 'lodash/includes';
 import some from 'lodash/some';
 import first from 'lodash/first';
 import { makeEditor } from './editor';
+import { safeEditorNodes } from './safe-nodes.js';
```
sneridagh: Nitpicking here, but we can remove the suffix here, right?

sneridagh: (and above)

tomschall: Yeah sure, i removed the suffix.

PR #7205, opened 2025-06-18
[NOT MAIN] Feature/fix subitemns messages logic

alexandreIFB:
Before the code update, the subitems deletion message was always displayed when `containedItemsToDelete > 0`, regardless of broken references. After the update, this behavior was lost because the state responsible for counting subitems (`containedItemsToDelete`) was being reset together with the broken references state (`brokenReferences`). This meant that when there were no broken references, the subitems count was also set to zero, and the subitems message did not appear even when it should.

The applied fix ensures that the subitems count is always maintained correctly, regardless of broken references. Now, the message is shown whenever there are subitems to be deleted, just like before the update.

Additionally, I added the corresponding translations.

You can also copy these changes directly from my branch (link below).


ksuess:
Thank you! You are very attentive!

PR #7208, opened 2025-06-22
Fallback for className if listing in grid has only one variation.

### LINKED ISSUE #7206: Fallback for className if listing in grid has only one variation.
- Configure grid block with only one non-default variation with id not 'default', for example 'foo'
- Create grid block
- Create listing inside grid block
- See that class is 'block listing default' instead of 'block listing foo'

ksuess:
closes #7206

### DIFF
--- a/packages/volto/news/7206.bugfix
+++ b/packages/volto/news/7206.bugfix
@@ -1,0 +1,1 @@
+Fallback for className if listing in grid has only one variation. @ksuess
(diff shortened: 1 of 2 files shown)


ksuess:
I close this one, as the edit view cannot be forced with an acceptable effort to show up like the view view, as the variation id is not injected.
This is due to the fact that *withVariationSchemaEnhancer* does not include the variation field if there is only one variation. That is OK and should not be changed. But it has the downside, that the variation id is not available for the packages/volto/src/components/manage/Blocks/Block/Edit.jsx . This is the component that injects the variation id via props.data.variation.

As a workaround I provide either a second variation or override all .block.gridBlock .block.listing depending on my conditions.

PR #7212, opened 2025-06-23
Enable showRelatedItems by default in Volto 19

### LINKED ISSUE #6549: In Volto 19, change the showRelatedItems setting default to true
(Follow-up on https://github.com/plone/volto/pull/4381, but it's a breaking change so has to wait for the next major version)

davisagli:
Fixes #6549

----
📚 Documentation preview 📚: https://volto--7212.org.readthedocs.build/

### DIFF
--- a/docs/source/configuration/settings-reference.md
+++ b/docs/source/configuration/settings-reference.md
@@ -482,9 +482,13 @@ cssLayers
     ```
 
 showRelatedItems
-    If true, the `RelatedItems` component will show items from the `relatedItems` field. Default: false.
+    If true, the `RelatedItems` component will show items from the `relatedItems` field. Default: true.
 
     ```{versionadded} 18.5.0
+    ```
+
+    ```{versionchanged} 19
+    In Volto 19, the default was changed to `true`.
     ```
 
 showTags
--- a/docs/source/configuration/settings-reference.md
+++ b/docs/source/configuration/settings-reference.md
@@ -482,12 +482,13 @@ cssLayers
     ```
 
 showRelatedItems
-    If true, the `RelatedItems` component will show items from the `relatedItems` field. Default: true.
+    If `true`, the `RelatedItems` component will show items from the `relatedItems` field.
+    Default: `true`.
 
-    ```{versionadded} 18.5.0
+    ```{versionadded} Volto 18.5.0
     ```
 
-    ```{versionchanged} 19
+    ```{versionchanged} Volto 19
     In Volto 19, the default was changed to `true`.
     ```
 
--- a/packages/components/.readthedocs.yaml
+++ b/packages/components/.readthedocs.yaml
@@ -1,0 +1,32 @@
+# .readthedocs.yaml
+# Read the Docs configuration file
+# See https://docs.readthedocs.com/platform/stable/config-file/v2.html for details
+
+# Required
+version: 2
+
+# Set the OS, Python version and other tools you might need
+build:
+  os: ubuntu-24.04
+  tools:
+    python: '3.13'
+    nodejs: '22'
+  commands:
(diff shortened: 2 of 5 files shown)


docs/source/configuration/settings-reference.md
```diff
@@ -482,11 +482,15 @@ cssLayers
     ```
 
 showRelatedItems
-    If true, the `RelatedItems` component will show items from the `relatedItems` field. Default: false.
+    If true, the `RelatedItems` component will show items from the `relatedItems` field. Default: true.
```
stevepiercy:
```suggestion
    If `true`, the `RelatedItems` component will show items from the `relatedItems` field.
    Default: `true`.
```

docs/source/configuration/settings-reference.md
```diff
@@ -482,11 +482,15 @@ cssLayers
...
 
     ```{versionadded} 18.5.0
     ```
 
+    ```{versionchanged} 19
```
stevepiercy: We need to explicitly specify the product with its version when it gets included in the main documentation, otherwise it's easy to overlook the implicit context.

```suggestion
    ```{versionadded} Volto 18.5.0
    ```

    ```{versionchanged} Volto 19
```

sneridagh (approved):

stevepiercy:
@sneridagh the two RTD failures are strange. Why is the RTD project `volto` looking for `packages/layout/.readthedocs.yaml` and `packages/components/.readthedocs.yaml`? The former package doesn't even exist in `main`! The change I made was only to ensure that the `seven` branch would get pulled in PRs against the `seven` branch to do a git diff. Nothing more. I wonder if this is related to the three new RTD projects you set up for components. I had nothing to do with their configuration.

davisagli:
@stevepiercy I added configuration to skip the RTD builds for those packages on the main branch, and @sneridagh separately disabled them from the RTD side.

packages/components/.readthedocs.yaml
```diff
@@ -0,0 +1,32 @@
...
+      then
+        exit 183;
+      fi
+    # Install any other extra dependencies to build the docs
+    - npm i -g corepack@latest
+    - corepack enable
+    - corepack prepare pnpm@* --activate
+    - asdf reshim nodejs
+    - make install
+    - (cd packages/layout && pnpm build-storybook -o ${READTHEDOCS_OUTPUT}/html)
```
stevepiercy, line 32: packages/layout, not packages/components?

davisagli: @stevepiercy Gah, copy and paste error. It doesn't actually matter here, because the condition above prevents this from running, but I'll fix it.

davisagli: Fixed in https://github.com/plone/volto/commit/6a4b0bf7df9f748c5d9d1de177e53737fbe04a8f

stevepiercy: FWIW, I'm still not clear wtf @sneridagh wants for docs and storybooks. `¯\_(ツ)_/¯` I'll wait until it's made clear to me.

ksuess:
@davisagli This is a breaking change. Would you please add a note to the upgrade guide docs/source/upgrade-guide/index.md.
For projects with a custom related items component, this setting `showRelatedItems` needs to be set to false.
To avoid surprises, popping up an unexpected component , a short note in the upgrade guide would be nice.

sneridagh:
@ksuess I just stumbled upon this, I just checked and no upgrade guide entry was added. Probably we missed your message.

@davisagli could you please take care of inserting a small note?

davisagli:
@ksuess @sneridagh https://github.com/plone/volto/pull/7430

PR #7215, opened 2025-06-24
Remove styles related to volto-slate-metadata-mentions from volto-slate #7181

### LINKED ISSUE #7181: Remove styles related to volto-slate-metadata-mentions from volto-slate
**Describe the bug**

Remove styles related to `volto-slate-metadata-mentions` from `volto-slate` package since `volto-slate-metadata-mentions` is a separate add-on now. A bit of cleanup is useful.

So, if there is no opposition, delete these:

https://github.com/plone/volto/blob/92cf6dadee2dd6e0eaa9635421357b7c2495d727/packages/volto-slate/src/widgets/style.css#L6-L16

if this and inline styles in React files are enough for `volto-slate-metadata-mentions`:

https://github.com/eea/volto-slate-metadata-mentions/blob/master/src/plugins/Mentions/less/editor.less

**To Reproduce**

1. Go to the first link above
2. See that there are tree rulesets in CSS that should not be there

**Expected behavior**

Just the 2 remaining rulesets should be left in place, because they are directly using `.slate_wysiwyg_box` and `.slate.error` CSS selectors.

**After the changes, the `volto-slate-metadata-mentions` add-on should work well, and be tested in the browser.**

**Software**
- Volto Version - `main` branch tip

shashank0470:
"Fixes #7181 - Removed CSS rules for metadata mentions as they belong in the separate volto-slate-metadata-mentions package

### DIFF
--- a/packages/volto-slate/src/widgets/style.css
+++ b/packages/volto-slate/src/widgets/style.css
@@ -3,18 +3,6 @@
   border-bottom: 1px solid #c7d5d8;
 }
 
-.metadata.mention.slate.widget {
-  display: inline;
-}
-
-.metadata.mention.slate.widget > :first-child {
-  display: inline;
-}
-
-.metadata.mention.slate.widget > :last-child {
-  display: inline;
-}
-
 .slate.error {
   color: #f00;
   white-space: normal;


shashank0470:
Hi! 

Just a quick note — I’ve already signed the Contributor Agreement and received the invitation to join the Plone GitHub organization. I understand that reviews for first-time PRs are temporarily paused until the Plone 7 release, but I just wanted to confirm if there's anything else I need to do from my side.

Thanks a lot for your time and for maintaining this project!

PR #7216, opened 2025-06-25
Testbuildsstorybooks

sneridagh:
----
📚 Documentation preview 📚: https://volto--7216.org.readthedocs.build/

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -12,6 +12,8 @@ build:
     python: "3.12"
     nodejs: "20"
   commands:
+    # First fetch the changes from the non-primary branch.
+    #
     # Cancel building pull requests when there aren't changes in the docs directory or YAML file.
     # You can add any other files or directories that you'd like here as well,
     # like your docs requirements file, or other files that will change your docs build.
@@ -19,6 +21,7 @@ build:
     # If there are no changes (git diff exits with 0) we force the command to return with 183.
     # This is a special exit code on Read the Docs that will cancel the build immediately.
     - |
+      git fetch origin seven
       if [ "$READTHEDOCS_VERSION_TYPE" = "external" ] && git diff --quiet origin/seven -- docs/ .readthedocs.yaml requirements-docs.txt \
           packages/components/.storybook/ packages/volto/.storybook/;
       then


stevepiercy:
@sneridagh it turns out that we missed a fatal error. I got the tip from https://gitlab.com/sosy-lab/benchmarking/sv-benchmarks/-/issues/1311 and the solution in its last referenced commit. It was mistakenly returning nothing instead of purposefully returning nothing, then proceeding to build the docs every freaking time for any change. In fact, [what's happening is documented in RTD](https://docs.readthedocs.com/platform/stable/build-customization.html#unshallow-git-clone).

Now it correctly detects a change in the file and since it doesn't return exit code 183, the RTD preview build continues to run `make docs-rtd-pr-preview`.

In any case, PR preview builds are not triggered by CI, but by webhooks.

https://docs.readthedocs.com/platform/stable/guides/pull-requests.html

> To enable pull request builds for your project, your Read the Docs account needs to be connected to an account with a supported Git provider, with a webhook configured to send information on pull requests for your repository.

That's just the build, nothing more. The build process looks for `.readthedocs.yaml` for configuration and optionally custom commands to run, which is what we do.

If you want more than a build trigger, specifically to have links added to the PR description, then you need to configure CI.

> - In order to have pull request build links automatically added to your pull requests, you must configure an automation to accomplish this with your Git provider. For example, see [these instructions](https://github.com/readthedocs/actions/blob/v1/preview/README.md) to configure with GitHub Actions.

I hope that clarifies the process.

To summarize, you get one and only one trigger to start a PR build per branch. You can then configure CI to populate the PR with a link to that one build.

That said, there is a long discussion going on in RTD's issue tracker to [add specific links to changed documentation files to the description](https://github.com/readthedocs/readthedocs.org/issues/11780). This might do some of what you want, but ultimately I don't see any way to automatically generate a link to an isolated, not aggregated, Storybook.

Assuming that only an aggregate Storybook build is possible, how would you like to proceed?

sneridagh:
@stevepiercy ok, I'd say that let's push your fix, then move from there. My attempt was not working because the workflow does not exist in main yet, I guess... but I cannot spend more time on this matter. Let's postpone it.

PR #7225, opened 2025-06-27
Seven separate styles

pnicolli:


packages/layout/components/Breadcrumbs.tsx
```diff
@@ -20,7 +20,7 @@ const Breadcrumbs = (props: SlotComponentProps) => {
   const breacrumbs = [rootItem, ...(items || [])];
 
   return (
-    <Container as="nav" width="default" className="py-4">
+    <Container as="nav" layout className="breadcrumbs">
```
sneridagh: We need to unify containers API implementations.

packages/layout/components/App/App.module.css
```diff
@@ -0,0 +1,13 @@
+@layer custom {
```
sneridagh, line 1: I'm convinced that we need to add an extra layer for "default" Seven theming... unsure about the name, as usual.

packages/layout/components/Main.tsx
```diff
@@ -11,11 +10,11 @@ const Main = (props: MainProps) => {
   const { content, location } = props;
 
   return (
-    <Container className="content-area">
+    <div className="content-area">
```
sneridagh, line 13: It has to be a container (for the block widths managements with a container query).

packages/publicui/routes/content.tsx
```diff
@@ -11,5 +11,11 @@ export default function Content() {
   }
   const { content } = contentData;
 
-  return <App content={content} location={location}></App>;
+  return (
```
sneridagh, line 14: We will adequate it in Katja's PR, right?

sneridagh (approved):
LGTM, a few comments! Almost ready to merge.

packages/layout/components/Breadcrumbs.tsx
sneridagh: Done.

packages/layout/components/Main.tsx
pnicolli: Let me tag @danalvrz that had a different opition about this and maybe can chip in.
Also, it should be a block responsibility to add the container wherever needed inside the block view rendering, I would say. Maybe I am not understanding the use case you have in mind here.

packages/layout/components/App/App.module.css
pnicolli: Maybe we should... idk, maybe let's just call it "plone"?

packages/publicui/routes/content.tsx
pnicolli: Yes

packages/layout/components/Main.tsx
danalvrz: @pnicolli @sneridagh the idea would be to organize the styles into three layers: (1) containers (`<header>`, `<footer>`, `.content-area`) that provide container query contexts (no constraints), (2) section-specific wrappers (`.header-wrapper`, `.footer-wrapper`) that handle both spacing and layout for their sections (gutter, flex-direction, etc), (3) and self-spacing blocks (.block) for the content area. So that developers can understand where styles belong without much guesswork. The same pattern could apply for the `.breadcrumbs` being a "pure" container for the query context, and a `.breadcrumbs-wrapper` to set space limits and layout styles. Ideally the container also handles the `background` attribute, but this might be slightly more complex when considering elements like `.pre-footer` and `.post-footer` having different backgrounds. But we could have independent wrappers for each element handle the `background` there. 

It could be even better to have a `.container-wrapper` that uses CSS props as utilities for its styles (max-width, padding, layout, etc.) and then can be reused for the header, footer, breadcrumbs, etc.

packages/layout/README.md
```diff
@@ -4,6 +4,66 @@
 
 This package provides default structural layout elements for Plone 7 and Plone's frontend modular architecture.
 
+## Layout Layers
```
sneridagh, line 7: @danalvrz This has to be updated and it moved to normal documentation.

packages/layout/components/App/App.module.css
sneridagh: +1 for me, at least for now...

@danalvrz ?

packages/layout/components/Header/Header.module.css
```diff
@@ -0,0 +1,16 @@
+@layer custom {
+  /* .header-wrapper {
+    display: flex;
+    flex-direction: row;
+    align-items: baseline;
+    justify-content: space-between;
+    padding-top: 25px;
+    padding-bottom: 35px;
+  } */
```
sneridagh, line 9: @danalvrz this has to go or goes into Agave?

packages/layout/components/Logo/Logo.module.css
```diff
@@ -0,0 +1,6 @@
+@layer custom {
```
sneridagh, line 1: Change the layer if we decide so.

packages/layout/components/Main.tsx
sneridagh: +1, resolved, needs docs.

packages/layout/components/MainFooter/MainFooter.module.css
```diff
@@ -0,0 +1,13 @@
+@layer custom {
```
sneridagh, line 1: Change layer if decided so.

packages/layout/components/Navigation/Navigation.module.css
```diff
@@ -0,0 +1,24 @@
+@layer custom {
```
sneridagh, line 1: Layer change.

packages/layout/components/SectionWrapper/SectionWrapper.stories.tsx
```diff
@@ -0,0 +1,62 @@
+import React from 'react';
```
sneridagh, line 1: Needs docs.

packages/layout/components/SectionWrapper/SectionWrapper.tsx
```diff
@@ -0,0 +1,49 @@
+import React from 'react';
```
sneridagh, line 1: Needs docs.

packages/layout/styles/main.css
```diff
@@ -1,84 +1,126 @@
...
-  height: 100vh;
-  margin: 0;
-  grid-template-areas:
-    'header'
-    'content'
-    'footer';
-  grid-template-columns: 1fr;
-  grid-template-rows: auto 1fr auto;
-}
+@layer custom {
```
sneridagh, line 1: change layer if decided so.

sneridagh (commented):
I left some comments.

apps/seven/registry.config.ts
```diff
@@ -4,6 +4,7 @@ const addons = [
   '@plone/layout',
   '@plone/cmsui',
   '@plone/publicui',
+  '@plone/agave',
```
sneridagh, line 7: Not 100% sure that Agave has to be default... We can consider this in the future though.

packages/layout/components/App/App.module.css
sneridagh: Let's do it in a subsequent PR, I don't want to hold this more.

sneridagh:
I merged it, conscious that still needs docs, polish it and the layer change.

PR #7228, opened 2025-06-30
[Seven] Block editor

sneridagh:
First wireframe implementation, focusing on performance.

https://github.com/user-attachments/assets/32ac81e8-9fbe-41c8-9cc8-7e8f1cca4345


packages/blocks/RenderBlocks/RenderBlocks.tsx
```diff
@@ -44,7 +44,7 @@ const RenderBlocks = (props: RenderBlocksProps) => {
         const Block = blocksConfig[blockType]?.view || DefaultBlockView;
 
         return Block ? (
-          <BlockWrapper key={block} {...props} block={block}>
+          <BlockWrapper key={block} {...props} data={blockData} block={block}>
```
pnicolli: I'm not sure we should spread props here, could be that passing just the needed ones is fine or even better for optimization

packages/cmsui/helpers/atoms.ts
```diff
@@ -1,6 +1,7 @@
-import { useCallback } from 'react';
+import { useCallback, useMemo } from 'react';
```
pnicolli: useMemo looks unnecessary

packages/cmsui/routes/atoms.ts
```diff
@@ -0,0 +1,12 @@
+import { atom, createStore } from 'jotai';
+import { atomFamily } from 'jotai/utils';
+import { focusAtom } from 'jotai-optics';
+import type { Content } from '@plone/types';
+
+export const store = createStore();
+
+export const formAtom = atom<Content>({} as Content);
```
pnicolli, line 8: Maybe `Partial<Content>`? Haven't tried really, unsure if my suggestion makes sense, have you tried by any chance?

packages/cmsui/routes/edit.tsx
```diff
@@ -83,7 +83,8 @@ export default function Edit() {
     onSubmit: async ({ value }) => {
       // @ts-expect-error: For some reason, the type of value is not inferred correctly
       // as `useLoaderData` turns every "unknown" type into `undefined`
```
pnicolli: This is probably throwing an error and/or needs to be updated since you changed the next code line

packages/publicui/routes/index.tsx
```diff
@@ -75,13 +76,15 @@ export default function Index() {
...
       </head>
       <body>
-        <div role="navigation" aria-label="Toolbar" id="toolbar" />
-        <div id="main">
-          <RACRouterProvider navigate={navigate}>
-            <Outlet />
-          </RACRouterProvider>
-        </div>
-        <div role="complementary" aria-label="Sidebar" id="sidebar" />
+        <PluggablesProvider>
```
pnicolli: Do we need this in publicui? It's a little scary :)

packages/cmsui/routes/atoms.ts
sneridagh: Yeah, it complains :( Like this works well... could be that in some places we will not have the complete thing, but it will be the case all the time.

packages/cmsui/helpers/atoms.ts
sneridagh: removed.

packages/blocks/RenderBlocks/RenderBlocks.tsx
sneridagh: removed, we might need the pathname for some blocks (or not, let's see).

packages/cmsui/routes/edit.tsx
sneridagh: Updated, casting it.

packages/publicui/routes/index.tsx
sneridagh: removed!

sneridagh:
@pnicolli should be good to go.

packages/blocks/RenderBlocks/RenderBlocks.tsx
sneridagh: Done.

PR #7230, opened 2025-07-02
feat: mode 'display' in form field

giuliaghisini:
if in schema field the attribute 'mode' is 'display', 
make the field only viewable in form.

### DIFF
--- a/packages/volto/src/components/manage/Form/Field.jsx
+++ b/packages/volto/src/components/manage/Form/Field.jsx
@@ -8,7 +8,7 @@ import PropTypes from 'prop-types';
 import { injectIntl } from 'react-intl';
 import config from '@plone/volto/registry';
 import { injectLazyLibs } from '@plone/volto/helpers/Loadable/Loadable';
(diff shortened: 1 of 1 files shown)


packages/volto/src/components/manage/Form/Field.jsx
```diff
@@ -172,6 +174,19 @@ const UnconnectedField = (props, { intl }) => {
     return null;
   }
 
+  if (props.mode === MODE_DISPLAY) {
+    let field = {
+      ...props,
+      //widget: getWidget(f, contentSchema?.properties[f]), //import { getWidget } from '@plone/volto/helpers/Widget/utils';
+    };
+    const DisplayWidget = config.widgets?.views?.getWidget(field);
```
davisagli, line 182: We should make this consistent with how DisplayView gets the widget (https://github.com/plone/volto/blob/main/packages/volto/src/components/theme/View/DefaultView.jsx#L81)

davisagli (changes requested):

PR #7234, opened 2025-07-03
Contents view: Label 'None' for deselecting needs to be different fro…

### LINKED ISSUE #7233: Contents view: german label 'None' for deselecting is wrong and needs to be different from  'None' of a not available value in the table
before

Image

after
Image

ksuess:
…m 'None' of a not available value in the table.

fix #7233


erral:
Does this happen also in 18.x?

ksuess:
> Does this happen also in 18.x?

yes, will do a backport.

PR #7237, opened 2025-07-08
Sneridagh seven separate styles

sneridagh:
I talked this evening with @danalvrz . Just PoC'ing things in here.

The premise is: ok to add the inner-container pattern in main layout elements to unify concepts. My take is that why we don't come with a component that does the syntactic sugar for us. Then the next step was, on naming. If it's the same pattern: then leave the naming as it is in BMv3, because, at the end all is a "block" positioned in the screen using the same pattern. 

By default: no width, but setting the container. The width will be set by the theme. Downside: without theme, the "baseline" theme will look akward... Don't know if it's what we want. We could overcome that by providing a bunch of lines of CSS for defining widths by default, so everybody building a theme can import them.

So eventually:
- We could inject "width" and other styling properties eventually into them from "somewhere" that is defined in the future: thus, fallbacking to CSS property `--block-width`.
- In anticipation to the EdW PoC of "footer and slots with blocks inside, configurable", and RT inline "blocks field". The pattern is the same, we are not reinventing anything.


danalvrz (approved):
@sneridagh @pnicolli I do like having the same pattern in the header/footer as we do in the content. I tried this branch locally with agave and it works well. The unified block infrastructure means developers/editors only need to learn one pattern, no matter if they're changing a header, a content, or a footer element (all are blocks). I think it makes the system more predictable, while still providing the flexibility to handle different needs through variants and modifiers at both the block or the section level.

packages/layout/components/SectionWrapper/SectionWrapper.tsx
```diff
@@ -0,0 +1,44 @@
...
+) => {
+  const {
+    as: Component = 'div',
+    children,
+    className,
+    section,
+    width,
+    ...rest
+  } = props;
+  const baseClass = section ? `${section}-block` : 'block';
```
pnicolli: I am not sure about the default `block` class. I understand we want to treat every section as a block, but block also has a different semantic nowadays so we have to make a bold and powerful move in order to change that semantic.

packages/layout/components/SectionWrapper/SectionWrapper.tsx
```diff
@@ -0,0 +1,44 @@
...
+  const sectionClasses = cx(baseClass, className, {
+    [width || '']: true,
+  });
```
pnicolli: It's probably enough to add `width` to the list of classes like so:


```suggestion
  const sectionClasses = cx(baseClass, className, width);
```

packages/layout/components/SectionWrapper/SectionWrapper.tsx
danalvrz: There might be some benefit in having similar names for components that behave similarly in several ways, but I agree on improving the naming for sure. I was also thinking that having the semantic tag configurable in the `BlockWrapper` could allow better semantic tagging in the content.

packages/layout/components/SectionWrapper/SectionWrapper.tsx
sneridagh: done

sneridagh:
@danalvrz @pnicolli after yesterday's talk, I've updated it.
- I kept the Component, might be of some help with the goal of unify things quickly.
- I swapped `block` for `section`. I tried `slot` but then I thought it might be overloaded with true slot naming, etc... and might become confusing. So I kept section. 
- Yes, it looks a lot as the blocks structure, but I'd keep it separated for now, with the advantage that the pattern is still the same. We have always the option to unify things later.

Now it starts to make much more sense to me. I'd merge and iterate over it.
Waiting for your ACK.

packages/layout/styles/main.css
```diff
@@ -7,11 +14,113 @@
...
+      }
+
+      &.narrow {
+        --block-width: var(--narrow-container-width);
+      }
+    }
+  }
+
+  .content-area {
+    [class*='block'] {
```
pnicolli:
```suggestion
    .block {
```

I would just go with this

pnicolli (changes requested):

pnicolli:
Left just a small comment on something I would iron out right away, then I'm ok with merging and iterating

Issue #7219, opened 2025-06-26
Volto makes two requests for images in SSR


wesleybl:
**Describe the bug**
When a request is made via SSR, Volto makes two requests for the images. Is this supposed to be the case? This is bad for performance. Especially when we use a different URL in `RAZZLE_INTERNAL_API_PATH`. The SSR request uses the URL in `RAZZLE_INTERNAL_API_PATH` and the client request uses the portal URL. So there are two requests for the same image.

**To Reproduce**

1. Create a page with an image block.
2. Open the browser's development tool.
3. Make an SSR request for the page with the image block.
4. See in the development tool that two requests are made for the same image.

**Expected behavior**
Only one request should be made.

**Software (please complete the following information):**

- Browser: Chrome
- Volto Version 18
- Plone Version 6.1
- Plone REST API Version 9

**Additional context**
Can be viewed at: https://demo.plone.org

Perhaps the `img` tag is different in SSR, causing a new request.

davisagli:
I was able to reproduce this in Firefox, but I had to disable the cache in the dev tools.

The first request shows "lazy-imageset" as the initiator, and the second request shows "img" as the initiator, and that it loaded from the cache (even though I disabled the cache? weird...)

I'm not sure what's going on yet. Maybe it has something to do with using `srcset`.

<img width="1512" alt="Image" src="https://github.com/user-attachments/assets/0d271dd0-88ed-4722-bd7e-4d95a3827bb0" />

wesleybl:
> I was able to reproduce this in Firefox, but I had to disable the cache in the dev tools.

@davisagli now I realized that I had the cache disabled when I tested. With the cache enabled, only one request is made.

I think that, as you said, the double request when the cache is disabled is caused by `srcset`. During hydration the browser must try to select a new image (even if it is the same) because of `srcset`, and there is nothing in the cache. So it makes a new request.

Thanks!

wesleybl:
Anyway, there is still one question: should `RAZZLE_INTERNAL_API_PATH` go to the SSR html? Shouldn't it be used only internally? But I think that is another issue.

wesleybl:
> should RAZZLE_INTERNAL_API_PATH go to the SSR html?

Apparently it is correct:

https://github.com/plone/volto/blob/0228f66c564ca52d3222f787c0567ce3d67fa8e7/packages/volto/src/start-client.jsx#L49-L56

[...]

PR #7238, opened 2025-07-09
Fixed corner case when coming from SSR in `edit` route, where the teasers where overwriten with bogus URLs after save if `INTERNAL_API_PATH` is set.

sneridagh:
Applied a bulk `flattenToAppURL` to all the initial static store resultant from SSR initial request in the `Html` component. 
This will happen only to the first request, avoiding the leak of bogus URLs with INTERNAL_API_PATH on it.

### DIFF
--- a/packages/volto/news/7238.bugfix
+++ b/packages/volto/news/7238.bugfix
@@ -1,0 +1,1 @@
+Fixed corner case when coming from SSR in edit route, where the teasers where overwriten with bogus URLs after save if INTERNAL_API_PATH is set. @sneridagh


wesleybl:
@sneridagh I also saw a `RAZZLE_INTERNAL_API_PATH` leak. See: https://github.com/plone/volto/issues/7219

sneridagh:
@wesleybl yeah, I hope this will solve it too and no more leaks happen in there. We are testing it in production, after the feedback, I will merge this.

sneridagh:
@plone/volto-team it seems that the tests are positive (tested in 18.x), so I will merge it in main.

PR #7242, opened 2025-07-09
Pass on onClick to Link component in UniversalLink

fredvd:
Regression since Volto 18.19.0


wesleybl:
@fredvd you need to make a PR for main too.

JeffersonBledsoe (approved):
Nothing inherently in-accessible about using an onclick with a link, but it depends what you do in that event handler.

nileshgulia1 (approved):

sneridagh:
@fredvd can you create the forwardport for `main`?

PR #7248, opened 2025-07-15
Fixed corner case when coming from SSR in `edit` route, where the teasers where overwriten with bogus URLs after save if `INTERNAL_API_PATH` is set. (#7238)

sneridagh:


wesleybl:
@sneridagh @davisagli Can you please merge this here? Hope this fixes it https://github.com/plone/volto/issues/7219#issuecomment-3014379562

davisagli:
@sneridagh @wesleybl Won't this also potentially mess with the actual content? i.e. if I create a page with blocks where I explain how to access the site via the API, this would replace the full domain in the content with `/`, no? It feels like a very blunt tool and I'm worried about the potential for unanticipated side effects.

sneridagh:
@davisagli FHNW have this fix since July and they have not found any issue. I don't get the use case you are bringing. Could you be more specific? 

This only fixes a very old known issue we had since forever, but since the hydration "fixed" it on load time, it never hit us. However, FHNW found this corner case where it does matter. It only affects in the serialized store content loaded in the first SSR load. Now it does fix it, with the proper values.

This is fixed in main since August and we had no complaints too. Also, in Seven we are using this approach, so the client has always a "clean" data, with no server names on it whatsoever.

If they were to have any issues, then it's blocks or places where they are doing things wrong, and instead of using only the path, rely on the whole server URL for functioning, which is wrong, and should be fixed.

davisagli:
@sneridagh Imagine I'm adding a page to my site to explain how to access it via the API. I'll add a text block that says something like "Connect to the API at https://mysite.com/++api++/"

With this change, won't it be changed to say only "Connect to the API at /++api++/" ?

My point is that there can be copies of the API path that we don't actually want to mangle. It's not a common use case, but IMO it's an important requirement for a CMS to not mangle the content.

sneridagh:
@davisagli I see your point, although as you said it's a very narrow use case, anyways, I would really like to get this settled asap. Do you know what this is fixing, right? You have a page with teasers, call edit from SSR, save, all teasers images are gone. So this is a real issue, that involves information loss and we need to fix it asap.

So, since this is the path that we took in Seven, we need an unified fix in both.

Do you have a proposal on how to fix this? You showed me some backend only PoC in Mexico, right? As said, ideally that works in both Volto and Seven. So, we have the response, who has the URL wrong, typically with the backend:8080 URLs, we need to flatten it on the server (frontend) so the client does not need to flatten them anymore, or directly, the backend do not respond with any server URL, only paths which would be the desired option.

davisagli:
> I see your point, although as you said it's a very narrow use case, anyways, I would really like to get this settled asap. Do you know what this is fixing, right? You have a page with teasers, call edit from SSR, save, all teasers images are gone. So this is a real issue, that involves information loss and we need to fix it asap.

@sneridagh I was also pointing out a possible information loss that would happen because of this solution. So let's please slow down enough to make sure we're solving it in a way where it will actually be settled and not just cause a different problem.

I think a fix for this has to be a bit more careful, walk through the content structure, and make a distinction between data that is text entered by the user and data which is backend URLs. i.e. only process the same fields that we do in the resolveuid transform.

(The thing I was experimenting with in Mexico would change how Zope generates absolute_url so that we only have absolute paths from the start. I think that would solve this without requiring the client to do replacements, but it would be a backward incompatible change so nothing we can use with existing Volto releases.)

There's a piece of this that I don't understand: Why doesn't the edit component fetch the content from the API during hydration the way the view component does? Then we would have the same data in the store that we do for non-SSR loads of the edit route, and we wouldn't need this special handling for cleaning up what's in the serialized store. (This suggestion is for Volto only, not Seven where the loaders always run on the server.)

sneridagh:
> There's a piece of this that I don't understand: Why doesn't the edit component fetch the content from the API during hydration the way the view component does? Then we would have the same data in the store that we do for non-SSR loads of the edit route, and we wouldn't need this special handling for cleaning up what's in the serialized store. (This suggestion is for Volto only, not Seven where the loaders always run on the server.)

The way that the add/edit view are designed (coming from the very origin Plone React), simply does not work like the "view" route. It takes for granted that the information will be already on the store, then it does not make the request again. This is good in the sense that it does issue another request, but it's bad in this issue. I already considered and tried this course of action when trying to fix this issue, but I was unsuccessful. Maybe you can give it a try.

I agree to explore other options, do you have time to look at them?

wesleybl:
It was replaced by: https://github.com/plone/volto/pull/7317

PR #7257, opened 2025-07-26
Fix registerWidget, added tests

sneridagh:
@pnicolli @deodorhunter I've found an bug in `registerWidget` trying to register several widgets, they would overwrite each other :P 

Please take a look. The resultant typings are a bit meh, I know. If you want/know how to improve it in a nice way, please do.

I've added the ability to register more than one widget at once too... just in case.


deodorhunter:
@sneridagh Sprinkled some ts magic to remove the need for heavy casting, let me know if you have questions :)

sneridagh:
@deodorhunter Thanks for the pinch of magic in there! Yes to all! :) 

@pnicolli I think now this is ready!

PR #7269, opened 2025-08-12
Improve loading of context for rendering NotFound view

davisagli:
Rendering a NotFound view is contextual and can depend on data from content expanders. So, let's do a full content load for the closest content item that is visible.

This depends on https://github.com/plone/plone.rest/pull/187 in order to find the best content path to load, but it falls back to what it did previously (loading the site root or language root folder).


davisagli:
@sneridagh Let's discuss this once you are back.

sneridagh:
@davisagli can you write a changelog entry for it? Thanks!

davisagli:
This way caused too much of a flash while the data loaded in a separate request. We ended up with a different solution in https://github.com/kitconcept/volto-light-theme/pull/853

PR #7274, opened 2025-08-23
Remove vitest dependencies from root to prevent confusion

wesleybl:
Having vitest as a root dependency allows `pnpm vitest` to execute, but the package tests are not configured to run from the monorepo root. This caused confusion as the command would appear to work but fail to properly execute tests from individual packages.

Tests should be run per package using:
- pnpm --filter @plone/volto test
- pnpm --filter @plone/components test
- pnpm --filter @plone/client test

Each package maintains its own vitest configuration and dependencies.

### DIFF
--- a/packages/volto/news/7274.internal
+++ b/packages/volto/news/7274.internal
@@ -1,0 +1,1 @@
+Remove vitest dependencies from root to prevent confusion. @wesleybl


wesleybl:
I resolved the conflict here.

PR #7288, opened 2025-08-29
Synchronizes versions of Volto dependencies with other packages in the monorepo

wesleybl:
Backport of #7184 to Volto 18


packages/volto/package.json
```diff
@@ -215,7 +215,7 @@
     "process": "^0.11.10",
     "promise-file-reader": "1.0.2",
     "prop-types": "15.7.2",
-    "query-string": "7.1.0",
+    "query-string": "^9.0.0",
```
davisagli: This is potentially breaking, if there are add-ons which depend on the old version without explicitly declaring it (pnpm requires them to declare it, but Volto 18 still supports yarn)

wesleybl: @davisagli If someone has a problem, they can declare the version in their package. Would that still be a breaking?

This solves an issue where addon tests might use the wrong version. See: https://github.com/plone/cookieplone-templates/issues/236

nileshgulia1: @wesleybl If I pin  "query-string": "^9.0.0", I get this error in volto. 
```
/Users/nileshgulia/Documents/volto/packages/volto/build/server.js:88269
module.exports = require("query-string");
                 ^
Error [ERR_REQUIRE_ESM]: require() of ES Module /Users/nileshgulia/Documents/volto/node_modules/.pnpm/query-string@9.1.1/node_modules/query-string/index.js from /Users/nileshgulia/Documents/volto/packages/volto/build/server.js not supported.
Instead change the require of index.js in /Users/nileshgulia/Documents/volto/packages/volto/build/server.js to a dynamic import() which is available in all CommonJS modules.
    at query-string (/Users/nileshgulia/Documents/volto/packages/volto/build/server.js:88269:18)
    at webpack_require 
    
    ```

wesleybl: @nileshgulia1 Pinning alone doesn't work. You need this update too: https://github.com/plone/volto/pull/7288/files#diff-ef0b8236651359f0c9a218af2fda51ee6e5a913ebf5af39e24ed4656cd9aaf98

See that the tests are working on this branch here.

sneridagh: @nileshgulia1 in fact, I experienced this in the past, but I was surprised that this is gone. Which Node version are you using? Maybe it's that. Please note, that we no longer test 18.

@wesleybl remember that I told you that this package is "special"? It's a Sindre Sorhus package, this means that he's not supporting anything else but ESM packaging, thus, "old" things like Volto has to be specially careful. However, the problem does not show up...

@nileshgulia1 can you confirm?

nileshgulia1: The issue was with older node patch versions. Works in latest Node v20 and v22.

wesleybl: @nileshgulia1 from what i understand, Volto 18 only supports Node 20 and 22. Not older versions.

nileshgulia1: @wesleybl yes I'm using Node 20, but the issue was merely with some older minor version, 20.x. which doesn't support common js require. Now I upgraded to v20.14.0. Everything works well.

wesleybl: > yes I'm using Node 20, but the issue was merely with some older minor version, 20.x. which doesn't support common js require

@nileshgulia1 @davisagli @sneridagh can this be considered a breaking change? Volto 18 is not tested with Node 20.0.0, for example.

wesleybl: @davisagli So we can't do this on Volto 18?

wesleybl: @davisagli @sneridagh can we then make [@plone/client](https://github.com/plone/volto/blob/04aded248a4ce1f8c8ccdf2853ccb4dd1e5733d5/packages/client/package.json#L62) use `query-string` 7 in Volto 18?

davisagli: @wesleybl Sorry I haven't answered here -- but it's because I don't really understand the situation and I'm not confident whether or not it is safe to update this package.

wesleybl: @davisagli Ok. I'll keep the `query-string` version in Volto 18. The issue https://github.com/plone/cookieplone-templates/issues/236 can be resolved by pinning `query-string` to the project. The downside is that we'll have to do this project-wide, or the tests will use the wrong version.

wesleybl: @davisagli done.

wesleybl:
@sneridagh @davisagli @nileshgulia1 I switched to using `query-string` 7 in `@plone/client`. I don't think this will be a problem because `@plone/client` is a "legacy alpha" in both the `main` and `18.x.x` branches. See:

https://github.com/plone/volto/blob/main/packages/client/README.md?plain=1#L17

wesleybl:
@sneridagh @davisagli @nileshgulia1 Can you take a look here please? I'm no longer updating `query-string` to version 9.

sneridagh (approved):
LGTM

wesleybl:
@ericof Could you please take a look at https://plone.org.br/?

It's not working.

ericof:
@fredvd, if you are still online, could you look into it? Maybe it is a domain issue and I will have to contact the Python community

fredvd:
@ericof The domain does not resolve anymore, the stack/site is running fine. I’m afraid its the domain registration

ericof:
@wesleybl thanks for the heads up. Our friends at @pythonbrasil acted quickly

PR #7299, opened 2025-09-01
Update package versions in pnpm-lock.yaml

wesleybl:
Backport of #7291 to Volto 18


wesleybl:
@sneridagh @davisagli I fixed the conflict that is occurring here.

davisagli:
I'm not certain how to evaluate if this is safe to backport, so I'm leaving it for @sneridagh

wesleybl:
@davisagli @sneridagh I think it's safe. In fact, I think we should do it. Reasons:

- Updating dependencies is always good and reduces the use of vulnerable packages.
- These versions are only for Volto development and testing. They won't affect Volto in production.
- All tests are passing.
- These are not updates to versions with breaking changes.
- Cookieplone doesn't have a `pnpm-lock.yaml`. So a new project will use new versions. Therefore, there isn't a single project that uses the same Volto versions. The versions used in Volto tests are outdated compared to the Volto projects. By updating here, Volto tests would use versions closer to the real projects.

davisagli (approved):
Well, I'm not that worried about the changes in vite-ssr, since it's just a proof of concept that was already removed on the main branch. (@sneridagh, should we remove it on the 18.x.x branch too so we don't have to do maintenance like this?) Let's go ahead and merge this.

PR #7304, opened 2025-09-02
Fixed random failure in Cypress 'Index Text Block' test

wesleybl:
Creates an alias to fetch the element for each command. This makes the test less prone to failure when the element is re-rendered. We also ensure that the element is enabled before entering text.

This is fixes:

```javascript
  1) Block Indexing Tests
       Index Text Block:
     CypressError: `cy.type()` failed because it targeted a disabled element.

The element typed into was:

  > <input aria-label="Search" name="SearchableText" autocomplete="off" placeholder="Search Site" title="Search" type="text" value="">

Ensure the element does not have an attribute named `disabled` before typing into it.
```

See: https://github.com/plone/volto/actions/runs/17390688337/job/49364259873?pr=7302#step:4:504


wesleybl:
@davisagli I think it helps because of this: https://docs.cypress.io/app/core-concepts/variables-and-aliases#Stale-Elements

davisagli (approved):
@wesleybl Okay, that sounds hopeful.

wesleybl:
@davisagli I'm thinking to upgrade Cypress version to see if these failures improve. Cypress 15 was released, but it removed support for Node 20. We'd have to upgrade to Cypress 14.

Is it possible to remove Node 20 support in Volto 19 and add support for Node 24?

davisagli:
@wesleybl I think that's possible since Volto 19 is still in alpha stage, and it's probably a good idea. What do you think, @sneridagh?

Mostly it means updating the versions that we test in the CI configuration, and the engines listed in package.json (plus some documentation in the upgrade guide)

PR #7305, opened 2025-09-02
a11y - Content folder search results

Wagner3UB:
Update folder content search input to announce result count for screen readers on search or input change.

Screenshot 2025-09-02 at 08 56 37

Video:
https://github.com/user-attachments/assets/e35e5a7d-84e7-4301-a877-0c60093a45ac

### DIFF
--- a/packages/volto/locales/af/LC_MESSAGES/volto.po
+++ b/packages/volto/locales/af/LC_MESSAGES/volto.po
@@ -2578,6 +2578,11 @@ msgstr ""
 msgid "Number of active objects"
 msgstr ""
 
+#. Default: "Number of results"
+#: components/manage/Contents/Contents
+msgid "Number of results"
+msgstr ""
+
 #. Default: "Object Size"
 #: components/manage/Contents/Contents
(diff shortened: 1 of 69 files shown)


stevepiercy (approved):
I'd prefer "Result count" and `resultCount` for brevity and to avoid the awkward wording of "Number of results: 1", but the current wording is fine, too.

Wagner3UB:
> I'd prefer "Result count" and `resultCount` for brevity and to avoid the awkward wording of "Number of results: 1", but the current wording is fine, too.

Thank you Steve, I'll change it, no problem at all!

stevepiercy (approved):
LGTM. Thank you!

davisagli:
Can you make a PR for `main` also? I really prefer to only backport fixes after they've been merged to main, to make sure that we don't accidentally forget to port the fix to main.

JeffersonBledsoe (approved):
Having the loading states interrupt things while your typing isn't particularly nice, but this PR still moves us in the right direction! LGTM and we can improve the loading state later

PR #7311, opened 2025-09-03
Add type=button to more widget buttons

davisagli:
Similar to https://github.com/plone/volto/pull/6791


davisagli:
@pnicolli Does this look okay to you? You did an earlier set of fixes similar to this.

PR #7313, opened 2025-09-05
Convert RenderUsers component from class to functional with hooks

wesleybl:
Replace class component with functional component using React hooks.
Remove HOCs in favor of hooks.

Ref #4460


packages/volto/src/components/manage/Controlpanels/Users/RenderUsers.jsx
```diff
@@ -13,199 +14,174 @@ import Toast from '@plone/volto/components/manage/Toast/Toast';
...
+  const {
+    user: propsUser,
+    listUsers,
+    updateUser: updateUserRole,
+    isUserManager,
+    roles,
+    inheritedRole,
+    userschema,
+    onDelete,
+  } = props;
```
davisagli, line 41: These should all be included in RenderUsers.propTypes below.

packages/volto/src/components/manage/Controlpanels/Users/RenderUsers.jsx
```diff
@@ -13,199 +14,174 @@ import Toast from '@plone/volto/components/manage/Toast/Toast';
...
+      toast.success(
         <Toast
           success
-          title={this.props.intl.formatMessage(messages.success)}
-          content={this.props.intl.formatMessage(messages.updateUserSuccess)}
+          title={intl.formatMessage(messages.success)}
+          content={intl.formatMessage(messages.updateUserSuccess)}
         />,
       );
     }
```
davisagli: I find it simpler to use the promise returned from dispatching an action. e.g.:
```jsx
dispatch(updateUser(userId, userData))
  .then((result) => {
    // handle success here
  }).catch((error) => {
    // handle error here
  });
```

Then there's no need for keeping track of the previous updateRequest just to detect when it finished.

packages/volto/src/components/manage/Controlpanels/Users/RenderUsers.jsx
wesleybl: Done.

packages/volto/src/components/manage/Controlpanels/Users/RenderUsers.jsx
wesleybl: Done.

davisagli (approved):

PR #7316, opened 2025-09-05
Allow relative paths in object browser

sneridagh:
Initial work on this. 
@davisagli @pnicolli  I had a problem when writing the test for this one...

Since we are using a `path.query` in there, and we are using a `RAZZLE_API_PATH=127.0.0.1:55001/plone` for the backend to point to the robot server, the path queries are always wrong (because Plone catalogs `/plone/whatever` instead of `/whatever`...

So I changed it to use `RAZZLE_DEV_PROXY_API_PATH=127.0.0.1:55001/plone` and now works like a charm... I guess we never re-evaluated if that was ok or wrong...

I am pushing it as it is, let's see if the tests pass. But if they do, I propose to change all the acceptance tests to use it.


sneridagh:
@davisagli @pnicolli This seems to work well, and solves the problem with catalog... What do you think? I'd say, we change this is in core, then if all goes well, we change it in cookieplone as default.

davisagli:
@sneridagh I need to study it a bit before I approve.

davisagli (changes requested):
@sneridagh I think I'm missing the point of the change.

I thought the goal was to allow storing a relative path so that we can set up a footer link that always goes to a form relative to the current content item. But this looks like it is still always resolving the input to an absolute URL before it saves. So the link will end up with the target that was at the relative path at the time of editing, rather than the target at the time of viewing.

(Note: even if we actually store the relative path, I am not sure what the resolveuid transforms will do with it on the backend.)

packages/volto/news/7316.internal
```diff
@@ -0,0 +1 @@
+Change `RAZZLE_API_PATH` with `RAZZLE_DEV_PROXY_API_PATH` for all acceptance tests. @sneridagh
```
davisagli, line 1: And for this, I think we want to set RAZZLE_INTERNAL_API_PATH rather than RAZZLE_DEV_PROXY_API_PATH. We want the acceptance tests to be as similar as possible to a normal production deployment in seamless mode, and we want this URL to be used both as the target of the dev proxy and the target where SSR will fetch from the API.

I see at least one place where devProxyToApiPath is not used in production mode: https://github.com/plone/volto/blob/main/packages/volto/src/helpers/Api/APIResourceWithAuth.js#L24

davisagli:
This approach doesn't work, since the URLs are still converted to absolute. We probably need backend support for this.

PR #7317, opened 2025-09-05
Refactor fix for #7238

sneridagh:
Fixed corner case when coming from SSR in edit route, where the teasers where overwriten with bogus URLs after save if INTERNAL_API_PATH is set. @sneridagh


wesleybl:
@sneridagh this will only fix `/edit`, correct? `INTERNAL_API_PATH` will still appear in image requests. But that's another problem.

davisagli:
@wesleybl What do you mean about image requests? I tested by starting Volto with RAZZLE_INTERNAL_API_PATH=http://127.0.0.1:8080/Plone so that I can easily see when it was used instead of localhost:3000. I loaded a page that has an image using server-side rendering. I don't see 127.0.0.1 in the server-rendered src for the image, and I don't see any requests to 127.0.0.1 in the network tab. Maybe I didn't understand the scenario.

Basically the approach here is: for the edit route, make sure the content is always fetched via a separate API request on the client side during hydration, to make sure it's the same as when the edit route is loaded via the React router. We already did this for the content view route (https://github.com/plone/volto/blob/main/packages/volto/src/components/theme/View/View.jsx#L124).

This can still leave the internal API path in other parts of the redux store that was hydrated. That's not ideal, but hard to avoid (the previous solution that we're replacing here was too eager and would also replace the internal API path if it was entered by an editor as literal text).

Perhaps we could set the internal API path using a Zope virtual host path that will make the backend generate public URLs instead of internal ones. As a quick test, this seems to work if I start Volto with RAZZLE_INTERNAL_API_PATH=http://127.0.0.1:8080/VirtualHostBase/http/localhost:3000/Plone/VirtualHostRoot -- then if I look at the serialized Redux store in the SSR-rendered HTML, 127.0.0.1 doesn't appear anywhere other than the RAZZLE_INTERNAL_API_PATH value.

But that has some challenges too:
- It's fine for Zope, but other backends like Nick don't support virtual host paths
- It means the Volto server needs to be able to get the public URL of the current request when it constructs the path for the API call.

wesleybl:
> What do you mean about image requests? I tested by starting Volto with RAZZLE_INTERNAL_API_PATH=http://127.0.0.1:8080/Plone so that I can easily see when it was used instead of localhost:3000. I loaded a page that has an image using server-side rendering. I don't see 127.0.0.1 in the server-rendered src for the image, and I don't see any requests to 127.0.0.1 in the network tab. Maybe I didn't understand the scenario.

@davisagli See:https://github.com/plone/volto/issues/7219

I couldn't simulate this locally. I think it only occurs when the api URL is fetched from the `Host` header. See my comments:

https://github.com/plone/volto/issues/7219#issuecomment-3014379562

https://github.com/plone/volto/issues/7219#issuecomment-3014407520

davisagli:
@wesleybl I see. I can imagine that this can happen if the SSR HTML has an image with a src that was generated using a URL that was fetched from the RAZZLE_INTERNAL_API_PATH. Then when you're accessing the site over a network connection, the browser starts to load the SSR src before it finishes fetching the content from the API on the client side and re-rendering with the correct API path. But when you access the site on localhost, the content is fetched quickly enough that the browser does not start downloading the image from the src in the SSR HTML.

But, I don't see how the SSR HTML would include the internal API path in the image src. That should not happen, because we flatten the URL before using it as the image src (https://github.com/plone/volto/blob/main/packages/volto/src/components/theme/Image/Image.jsx#L51) and flattenToAppURL includes replacing the internalApiPath: https://github.com/plone/volto/blob/main/packages/volto/src/helpers/Url/Url.js#L105

I think I need a more clear set of steps to reproduce the problem. Maybe it would be possible to create a simple docker-compose stack that shows it, to make sure I have the same configuration?

wesleybl:
> I can imagine that this can happen if the SSR HTML has an image with a src that was generated using a URL that was fetched from the RAZZLE_INTERNAL_API_PATH

@davisagli That's exactly what happens. The img tag src contains `RAZZLE_INTERNAL_API_PATH` in the SSR HTML.

> I think I need a more clear set of steps to reproduce the problem. Maybe it would be possible to create a simple docker-compose stack that shows it, to make sure I have the same configuration?

To simulate this, we'd need a webserver container in front of Plone and Volto, sending a `Host` header. Don't we already have this somewhere? I think there's a Cypress test that uses a webserver.

davisagli:
> > I think I need a more clear set of steps to reproduce the problem. Maybe it would be possible to create a simple docker-compose stack that shows it, to make sure I have the same configuration?

> To simulate this, we'd need a webserver container in front of Plone and Volto, sending a Host header. Don't we already have this somewhere? I think there's a Cypress test that uses a webserver.

@wesleybl We can try using the local docker-compose stack that's included in the cookieplone project template: https://github.com/plone/cookieplone-templates/blob/main/templates/projects/monorepo/%7B%7B%20cookiecutter.__folder_name%20%7D%7D/docker-compose.yml

wesleybl:
@davisagli I don't understand `traefik`. But basically what it has to do is send a `Host` header to Volto, and Volto starts with a different `RAZZLE_INTERNAL_API_PATH`.

wesleybl:
@davisagli I tested with:

https://github.com/plone/cookieplone-templates/blob/main/templates/projects/monorepo/%7B%7B%20cookiecutter.__folder_name%20%7D%7D/docker-compose.yml

but the problem didn't occur. It must be something specific to my environment. I need to investigate further here. Sorry for the noise.

sneridagh:
Related to: #7248

sneridagh:
@davisagli should we merge this?

davisagli:
@sneridagh I'd say so. Like I said in https://github.com/plone/volto/pull/7317#issuecomment-3259496038, it's not a perfect solution, because we can still end up with the internal API path in other parts of the store, or in the `content` part of the store on other SSR routes. But it solves the known issue we have now on the edit route, and that one is important to fix sooner rather than later since it can lead to saving bad data.

sneridagh:
@davisagli @wesleybl uh I thought we merged and backported this... 
Should we?

wesleybl (approved):
LGTM

davisagli:
@sneridagh Yes, I already said I think we should in my previous comment. I was waiting for you to make the final call and hit the button.

davisagli (approved):

PR #7342, opened 2025-09-11
chore: pt-br missing translations

Wagner3UB:
PT-BR missing translations

**Will be concluded once i18n is implemented on Volto 7.**


pnicolli:
We can start adding a pt-br folder in seven. See the `locales` folder in the cmsui and other packages, adding the pt-br folder in all the packages is enough to add the new language and can be done already.

pnicolli:
Let's close and re-create pt-br on the new frontend repo once everything is migrated

PR #7355, opened 2025-09-12
[Seven] Blocks editor

pnicolli:


packages/blocks/Text/Edit.tsx
pnicolli, line 1: This is a placeholder implementation I had before Plate was added, it's not used at this time and will be removed before merging

packages/cmsui/index.ts
```diff
@@ -33,13 +33,6 @@ export default function install(config: ConfigType) {
         type: 'prefix',
         path: 'login',
         children: [
-          {
```
sneridagh, line 36: Are indexes not needed?

packages/cmsui/components/Sidebar/Sidebar.tsx
```diff
@@ -1,9 +1,9 @@
-import { Fragment } from 'react';
+import { useTranslation } from 'react-i18next';
 import { atom, useAtom } from 'jotai';
 import { tv } from 'tailwind-variants';
 import { Pluggable } from '@plone/layout/components/Pluggable';
 
-export const sidebarAtom = atom(true);
+export const sidebarAtom = atom(false);
```
sneridagh, line 6: Original Quanta is by default closed, but @tisto already complained about this fact a bunch of times :) I have to abide that it feels strange to me after all these years. Soooo, let's leave it open then for now!

sneridagh (approved):
LGTM!

packages/cmsui/index.ts
pnicolli: Afaik they're not, in this case, so this looks much cleaner

PR #7357, opened 2025-09-12
Cleanup and update shadowing information.

sneridagh:
Follow up to: https://github.com/plone/documentation/pull/1971

----
📚 Documentation preview 📚: https://volto--7357.org.readthedocs.build/

### DIFF
--- a/.github/ISSUE_TEMPLATE/3_PLIP.md
+++ b/.github/ISSUE_TEMPLATE/3_PLIP.md
@@ -1,6 +1,6 @@
 ---
 name: '🚀 PLIP'
-about: 'A Plone Improvement Proposal (PLIP) is a change to a Plone package that would affect everyone who uses that package.'
+about: 'A Plone Improvement Proposal (PLIP) is a larger change to Plone, usually affecting multiple packages, and which goes through a formal process.'
 title: ''
 labels: '03 type: feature (plip)'
 type: 'PLIP'
@@ -18,8 +18,6 @@ projects: 'plone/47'
 <!--
 
 Read https://6.docs.plone.org/contributing/core/plips.html first!
-
-Set "03 type: feature: plip" as label.
 
 Mention the @plone/volto-team when the PLIP is information complete!
 
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -56,20 +56,59 @@ pipx ensurepath
 
 ## Running the release process
 
-These are the commands to make a Volto release:
+There are two phases to make a Volto release.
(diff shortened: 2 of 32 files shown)


docs/source/conceptual-guides/index.md
```diff
@@ -16,4 +16,5 @@ This section of the documentation contains conceptual guides for various aspects
 :maxdepth: 2
 
 add-ons
+shadowing
```
stevepiercy: This is not a conceptual guide, but actually more of a how-to guide. Why is it here instead of under Development, which appears to be a collection of how-to guides?

docs/source/conceptual-guides/shadowing.md
```diff
@@ -0,0 +1,113 @@
...
+```
+
+```{tip}
+Use absolute imports in your shadowed modules to avoid resolution issues.
+```
+
+
+### Customize add-ons components or modules
+
+Assume the component you want to shadow is located in Volto core at {file}`node_modules/@kitconcept/volto-light-theme/src/components/Header/Header.tsx`.
```
erral: This is not Volto core, it should say "is located in an add-on" or something like that.

erral (changes requested):

docs/source/conceptual-guides/shadowing.md
```diff
@@ -0,0 +1,113 @@
...
+```{warning}
+If you update a dependency or add-on that contains shadowed components, always verify whether the shadowed component's structure, API, or dependencies have changed.
+
+Failure to adapt your shadowed code to the new version may break your application silently or cause unexpected behavior.
+```
+
+
+## Clarification: only registered add-ons can be shadowed
+
+Shadowing only works with *registered add-ons* that are explicitly declared as such in {file}`registry.config.ts`.
```
erral: This works for add-ons loaded in the `addons` key in the `package.json` file right? 

I think we should either clarify that here, or link to the "how to install an add-on" part of our docs.

sneridagh: Yeah, that's what I wanted to mean with "registered add-ons". I think we've used that terminology in places already. I amended it, and added a link.

docs/source/conceptual-guides/shadowing.md
sneridagh: Done.

docs/source/conceptual-guides/index.md
sneridagh: Moved.

erral (approved):

PR #7312, opened 2025-09-04
Search block: fix 'Clear filters' overriding the base query. [main]

mauritsvanrees:
This is a forward port from my PR #7272 (18.x.x) to the main branch.  All commits squashed into one. This fixes issue #7271 for the main branch.

PR #7362, opened 2025-09-13
Update pt_BR translation

ericof:


packages/volto/news/7271.bugfix
wesleybl, line 1: I've learned that this works. See: https://github.com/plone/volto/pull/7312#discussion_r2324014385

wesleybl (approved):
LGTM

PR #7363, opened 2025-09-13
Replace img tags with Image component to enable future src prefixing

wesleybl:
- Replace all img tag usages with Image component
- Add fetchpriority="auto" to keep the default value, since Image always adds fetchpriority
- Keep img tags only for static SVG placeholders with ESLint exceptions
- Add ESLint rule to prevent future img tag usage outside Image component
- This change prepares the codebase for future image src URL prefixing functionality

The Image component will centralize image handling and enable consistent URL transformations across the application when the prefixing feature is implemented.

This is part of #4290 and can be done beforehand, as it has no side effects.

Ref #6897


mamico (approved):
LGTM

packages/volto/.eslintrc
```diff
@@ -37,6 +37,13 @@
     "react-hooks/exhaustive-deps": "warn",
     "react/react-in-jsx-scope": "off",
     "jsx-a11y/label-has-associated-control": "off",
+    "no-restricted-syntax": [
```
davisagli, line 40: Let's use a more specific name for this rule, please. Maybe `volto-image-component`

wesleybl:
> I'm not a big fan of needing to add fetchpriority="auto" everywhere. Maybe we should change the Image component so that it does not always set fetchpriority="high" when loading="eager", and only set it for the specific use cases where this is important.

@davisagli I agree with you, but that's not the purpose of this PR. The objective here is to remove the use of the `img` tag. I don't want to mix things up here. We can discuss that in a separate issue.

To keep the code clean, I'll remove `fetchpriority="auto"` here.

packages/volto/.eslintrc
wesleybl: This is a default `eslint` rule, I can't change its identifier. See:

https://eslint.org/docs/latest/rules/no-restricted-syntax

davisagli: Oh yes, I see

packages/volto-slate/src/editor/plugins/Image/render.jsx
```diff
@@ -16,7 +17,7 @@ export const ImageElement = (props) => {
   return (
     <span {...attributes} style={{ display: 'inline-block' }}>
       {children}
-      <img alt="" src={element.url} style={style} />
+      <Image alt="" src={element.url} style={style} fetchpriority="auto" />
```
sneridagh: @wesleybl This one still has `fetchpriority="auto"`

sneridagh:
@pnicolli @wesleybl why the `Image` component sets the prio to high?

pnicolli:
> why the `Image` component sets the prio to high?

@sneridagh @davisagli 
It is in the latest image handling best practices I have studied: https://www.builder.io/blog/fast-images

The gist is: if an image can be loaded lazily it will be through loading=lazy, otherwise it means it's found above the fold in the critical part of the page and needs a higher fetch prio. This reduces the baseline suggested implementations to:
- loading eager + fetch prio high
- loading lazy + decoding async

These give really good web vitals scores and I would recommend not changing the Image component unless we can find more info. The option to lower the priority if really needed is there but I didn't find any use case yet. Maybe for some block, but honestly I would keep a strong opinionated default.

packages/volto/.eslintrc
```diff
@@ -37,6 +37,13 @@
...
+      "error",
+      {
+        "selector": "JSXElement[openingElement.name.name='img']",
+        "message": "Use the Image component from '@plone/volto/components/theme/Image/Image' instead of <img> tag. For SVG placeholders, add an eslint-disable comment."
```
pnicolli: I would probably avoid adding the SVG placeholders recommendation, you can actually use the Image components for svgs as well, f I remember correctly.

packages/volto-slate/src/editor/plugins/Image/render.jsx
wesleybl: Removed.

packages/volto/.eslintrc
wesleybl: Now I see that the Image component works with SVG. I changed it.

PR #7365, opened 2025-09-15
Fixed the load of the default layers ensuring they are on top, using a React 19 managed `<link>`

sneridagh:


packages/cmsui/components/Toolbar/Toolbar.tsx
```diff
@@ -17,6 +20,11 @@ const Toolbar = () => {
       <div className="fixed inset-y-0 z-10 my-4 flex flex-col items-center gap-4">
         <Pluggable name="toolbar-top" />
         <Pluggable name="toolbar-bottom" />
+        <Plug pluggable="toolbar-bottom" id="button-cancel">
+          <Button aria-label="Cancel" size="L" onPress={() => navigate('/')}>
+            <Close />
+          </Button>
```
pnicolli, line 26: For perfect a11y, we should really find a way to make these buttons a proper link (i.e. `<a>`) while still having the right styles. The button does not convey the meaning that the browser will navigate elsewhere.

As an alternative, the label could explain more about what's happening, but a link would be preferable.

sneridagh: I'll take note. https://react-spectrum.adobe.com/react-aria/Button.html#link-buttons
We need a `Link` that looks like a button. Should we do it as a variant of the `Link` component?

pnicolli: Yeah I think so, it could be beneficial for these situations.

PR #7367, opened 2025-09-17
Update 3_PLIP.md

stevepiercy:
Remove redundant instruction which is now done automatically under `labels`

See also https://github.com/plone/Products.CMFPlone/pull/4221

I don't think a change log is necessary for this one and would add noise, but if I must, I will do so.


stevepiercy:
> Approved. But we can additionally sync with the CMFPlone template once that has its final review.

@davisagli this PR was merged prematurely, and is not synched to CMFPlone, which should have been the first one merged. This one's missing the link.

davisagli:
@stevepiercy I don't know what you're talking about. I merged the CMFPlone one first, then found out there's a 200 character limit, so I removed the link to fix it, then updated this one to match.

Issue #7263, opened 2025-08-06
Pressing Delete at the end of a text-block duplicates the next text-block


pbauer:
When you have two text-blocks (A and B) and press Delete at the end of A the content of B is appended to A but B is still there. 

**Expected behavior**
The second block should disappear.

You can see that on https://volto.demo.plone.org and https://demo.plone.org

Ajaypanchal4326:
Hi, I was able to reproduce this issue on [volto.demo.plone.org](https://volto.demo.plone.org/?utm_source=chatgpt.com)
.
**Steps I followed:**
1. Created two text blocks (A = "hello world", B = "second block").
2. Placed cursor at the end of A.
3. Pressed Delete.

**Observed behavior**: Content of B was appended to A, but Block B was still present.
**Expected behavior**: Block B should be removed ~~after merging~~. Edited. DO NOT MERGE CONTENT.

**Video of the bug:**

https://github.com/user-attachments/assets/34cc2e39-262f-4434-80f4-5ad9c1acd95b

I’d like to work on fixing this issue. Could you please assign it to me?

davisagli:
@Ajaypanchal4326 No, we don't assign issues. Please read https://6.docs.plone.org/contributing/first-time.html before trying to contribute.

Ajaypanchal4326:
@davisagli Thank you for clarifying. 
I’ll go through the contribution guide you shared and make sure I follow the right process before working on this issue.

aryan7081:
I'm working on this issue.

aryan7081:
@JeffersonBledsoe @plone/volto-team I need your confirmation here. I can think of two expected behaviours when both blocks contain some text. One is that the second block simply gets deleted. The second behaviour, which is the better solution, is that the content of the second block is merged into the first one and the second block gets deleted.

stevepiercy:
@aryan7081 I don't think you understand the issue and the expected behavior as reported in the original issue. Please read the initial issue report again. Merging of content is not wanted or expected. I also edited the incorrect "expected behavior" from a non-Volto Team member to prevent confusion.

aryan7081:
Okay, thanks for clarifying @stevepiercy . I got confused earlier, but now it's clear. I'll proceed accordingly.

davisagli:
@stevepiercy I disagree with the expected behavior in the issue description. I would expect the behavior to depend on what the next block is:
[...]

Issue #2, opened 2017-04-18
Error: Cannot find module 'react-jsonschema-form'


tisto:
1) Checkout repo
2) Run "yarn install"
3) Run "yarn start":

```$ yarn start
yarn start v0.19.1
$ NODE_ENV=production npm-run-all run:prod:server
yarn run v0.19.1
$ better-npm-run run_prod_server
running better-npm-run in /Users/timo/workspace/plone/plone-react
Executing script: run_prod_server

to be executed: ./node_modules/.bin/babel-node ./src/start-server-prod.js


Error: Cannot find module 'react-jsonschema-form'
    at Function.Module._resolveFilename (module.js:469:15)
    at Function.Module._load (module.js:417:25)
    at Module.require (module.js:497:17)
    at require (internal/module.js:20:19)
    at Object.module.exports.Object.defineProperty.value (/Users/timo/workspace/plone/plone-react/static/dist/server.js:6754:19)
    at __webpack_require__ (/Users/timo/workspace/plone/plone-react/static/dist/server.js:519:30)
    at fn (/Users/timo/workspace/plone/plone-react/static/dist/server.js:50:20)
    at Object.<anonymous> (/Users/timo/workspace/plone/plone-react/static/dist/server.js:6621:29)
    at __webpack_require__ (/Users/timo/workspace/plone/plone-react/static/dist/server.js:519:30)
    at fn (/Users/timo/workspace/plone/plone-react/static/dist/server.js:50:20)
    at Object.module.exports.Object.defineProperty.value (/Users/timo/workspace/plone/plone-react/static/dist/server.js:1591:15)
    at __webpack_require__ (/Users/timo/workspace/plone/plone-react/static/dist/server.js:519:30)
    at fn (/Users/timo/workspace/plone/plone-react/static/dist/server.js:50:20)
    at Object.<anonymous> (/Users/timo/workspace/plone/plone-react/static/dist/server.js:1425:21)
    at __webpack_require__ (/Users/timo/workspace/plone/plone-react/static/dist/server.js:519:30)
    at fn (/Users/timo/workspace/plone/plone-react/static/dist/server.js:50:20)

error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
ERROR: "run:prod:server" exited with 1.
error Command failed with exit code 1.
Timos-MBP:plone-react timo$```

robgietema:
[...]

PR #7373, opened 2025-09-20
Fixed issue of merging block contents upon pressing backspace at start of second block with some content

aryan7081:
Ref #7263

Inside the `joinWithPreviousBlock` function of `packages/volto-slate/src/blocks/Text/keyboard/joinBlocks.js`, pressing Backspace (Delete on mac) at the start of a block incorrectly merged the previous block's content into the current block, instead of merging the current block into previous one.

This happened because the implementation relied on `mergeSlateWithBlockBackward(editor, otherBlock)` which altered the current block's editor slate causing the wrong merge direction.

This fix removes the dependency and explicitly appends the current block's children (`editor.children`) into the previous block's value, then deletes the current block.


wesleybl (changes requested):
This adds a new line to the previous paragraph. The expectation here is that the second paragraph will be added to the end of the previous paragraph, without adding a new line.

aryan7081:
@wesleybl Thanks for pointing this out. It is fixed in the latest commit. Now merging happens Inline without adding a new line.

aryan7081:
Fixed, the Cursor is now placed before the first character of the second paragraph.

aryan7081:
@wesleybl PR updated: 
- Pressing Enter at the merge point now correctly splits into a new block. 
- Fixed linting issues. 
- Added the required changelog entry, but Git/Towncrier is not detecting the file. Could you advise on what might be missing?

davisagli:
@aryan7081 You also need a changelog file in packages/volto/news since you added the cypress test there. Use the extension .internal for that one since the change there is not a user-facing change.

aryan7081:
@davisagli Thanks for the guidance! I’ve added the internal changelog for the Cypress test.

packages/volto-slate/news/7373.bugfix
```diff
@@ -0,0 +1 @@
+Fix Backspace at start of a text block: merge current block into previous inline (no extra newline), delete the current block, and place the caret before the first character of the merged content. Also handle Enter immediately after such inline merge by splitting back into two blocks. (@aryan7081)
```
wesleybl: We generally do not put the author's name in parentheses.

aryan7081:
@wesleybl I have removed the author's name from the changelog entry and fixed the issue of pressing backspace twice to remove the remaining blocks.

packages/volto-slate/news/7373.bugfix
```diff
@@ -0,0 +1 @@
+Fix Backspace at start of a text block: merge current block into previous inline (no extra newline), delete the current block, and place the caret before the first character of the merged content. Also handle Enter immediately after such inline merge by splitting back into two blocks.
```
wesleybl: Please add your username in the note.

packages/volto-slate/src/blocks/Text/keyboard/joinBlocks.js
```diff
@@ -76,21 +75,63 @@ export function joinWithPreviousBlock({ editor, event }, intl) {
 
   // Else the editor contains characters, so we merge the current block's
   // `editor` with the block before, `otherBlock`.
-  const cursor = mergeSlateWithBlockBackward(editor, otherBlock);
```
wesleybl, line 79: Why weren't the changes made to `mergeSlateWithBlockBackward`? Have you considered whether the code there is needed? Even if it isn't needed, the implementation can be done there to make the code more modular and separate. `mergeSlateWithBlockBackward` is only used here.

packages/volto/news/7373.internal
```diff
@@ -0,0 +1 @@
+Added Cypress test for backspace behavior in slate blocks.
```
wesleybl: Please add your username in the note.

aryan7081:
@wesleybl Thanks, I investigated and found that `mergeSlateWithBlockBackward` wasn't being used anywhere, so I fixed the bug in that function. Now, `joinWithPreviousBlock` uses it instead of duplicating custom logic. Also, I have added the username to the log files.

packages/volto-slate/src/blocks/Text/keyboard/index.js
```diff
@@ -10,6 +10,7 @@ export * from './traverseBlocks';
 export * from './unwrapEmptyString';
 export * from './slashMenu';
 export * from './cancelEsc';
+export * from './splitAtSeam';
```
wesleybl: Don't use barrel import. Remove this line and import the full path.

aryan7081:
@wesleybl I have replaced barrel import with direct module import.

wesleybl (approved):
LGTM

aryan7081:
@wesleybl May I know the reason why you have edited the pr description from Fixes `#7263` to `Ref #7623`? See https://6.docs.plone.org/contributing/first-time.html#create-a-pull-request-from-your-fork item https://github.com/plone/volto/issues/2. Thank you!

wesleybl:
@aryan7081 It's because this PR doesn't fix #7263. They are similar but not the same situations.

aryan7081:
@wesleybl Okay, got it. Thanks for the clarification.

wesleybl:
@aryan7081 There appears to be an intermittent error in the `AuthToken.test.js` test unrelated to your PR:

https://github.com/plone/volto/actions/runs/18220108302/job/51877958704?pr=7373#step:5:1022

Would you like to investigate this and create a new PR?

wesleybl:
@aryan7081 Thanks!

Could you please make a backport to Volto 18?

Could you also try a fix for #7263? It looks similar to this one.

aryan7081:
> @aryan7081 There appears to be an intermittent error in the `AuthToken.test.js` test unrelated to your PR:
> 
> https://github.com/plone/volto/actions/runs/18220108302/job/51877958704?pr=7373#step:5:1022
> 
> Would you like to investigate this and create a new PR?

@wesleybl Sure, I'll look into it and open a PR.

aryan7081:
> @aryan7081 Thanks!
> 
> Could you please make a backport to Volto 18?
> 
> Could you also try a fix for #7263? It looks similar to this one.

@wesleybl Okay, I'll prepare a backport to Volto 18 and also look into #7263 
Thanks and Regards

davisagli (commented):
I finally had a chance to test this and it works well. Thank you for working on it!

I tried a different but similar use case: pressing Delete at the end of a text block that precedes another text block. This adds the text from the next block to the current one (expected) but does not delete the next block (so you end up with duplicate text). Is it possible to adapt your work here to handle merging the blocks in that case as well? (Edit: I see that is the issue that is already discussed in https://github.com/plone/volto/issues/7263)

aryan7081:
@davisagli Working on it.
Thanks and regards

PR #7376, opened 2025-09-20
Add sphinx-reredirects to docs requirements-docs.txt to support moved…

stevepiercy:
… files.

This also syncs up with `seven` branch, so when building docs locally, this requirement is no longer missing.

### DIFF
--- a/requirements-docs.txt
+++ b/requirements-docs.txt
@@ -5,6 +5,7 @@ plone-sphinx-theme<2
 sphinx-autobuild
 sphinx-copybutton
 sphinx-examples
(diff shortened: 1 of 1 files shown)


stevepiercy:
Great, now npmjs.com is using Cloudflare to block requests. I'm testing out a new linkchecker in https://github.com/collective/awesome-volto/pull/42.

stevepiercy:
I'm going to merge this now, since it's a trivial change, and I'll deal with the linkcheck failure separately. It looks like Linkspector might work.

PR #7378, opened 2025-09-21
Replace readme link checker of awesome_bot with Lychee

stevepiercy:
See https://github.com/collective/awesome-volto/pull/42.

Will back and forward port to `18.x.x` and `seven` after this is merged.

PR #7382, opened 2025-09-22
Minor cleanup of name and excluded URLs for ease of spotting the name…

stevepiercy:
… and maintenance of URLs.

This PR pulls in a couple of improvements I added while forward- and back-porting #7378. It's now in synch with both `18.x.x` and `seven` branches.


stevepiercy:
Let's skip the change log to avoid clutter.

PR #7389, opened 2025-09-23
Replace img tags with Image component to enable future src prefixing

wesleybl:
Backport of #7363 to Volto 18.

### DIFF
--- a/packages/volto/.eslintrc
+++ b/packages/volto/.eslintrc
@@ -37,6 +37,13 @@
     "react-hooks/exhaustive-deps": "warn",
     "react/react-in-jsx-scope": "off",
     "jsx-a11y/label-has-associated-control": "off",
+    "no-restricted-syntax": [
+      "error",
+      {
+        "selector": "JSXElement[openingElement.name.name='img']",
+        "message": "Use the Image component from '@plone/volto/components/theme/Image/Image' instead of <img> tag."
+      }
+    ]
   },
   "settings": {
     "import/resolver": {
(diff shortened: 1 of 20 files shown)


packages/volto/.eslintrc
```diff
@@ -37,6 +37,13 @@
     "react-hooks/exhaustive-deps": "warn",
     "react/react-in-jsx-scope": "off",
     "jsx-a11y/label-has-associated-control": "off",
+    "no-restricted-syntax": [
```
sneridagh, line 40: @plone/volto-team we have to be more careful in the future with this things.

This is a breaking change, and should not be backported.  I totally overlooked it. I will take care and re-release.

@wesleybl We will need a breaking change notice written for 19, and modify the changelog everywhere (including GH release).

sneridagh: https://github.com/plone/volto/pull/7401

sneridagh: merged.

wesleybl: > and modify the changelog everywhere (including GH release).

@sneridagh Are you saying [this file](https://github.com/plone/volto/blob/main/packages/volto/news/7363.feature) should be renamed? I don't think this change itself is a breaking change. What's breaking is ESLint validation, correct?

sneridagh: @wesleybl It is, in your projects as soon you update to this 19 version, linting will break. Thus, you have to make it noticeable and add it to the upgrade guide, so people can expect it, and can act upon accordingly (eg. without having to guess how to fix it). People on 19 alpha already expects breaking changes in the alpha phase. Someone in 18, they do not.

sneridagh: @wesleybl I just saw your PR! It's perfect, thanks!

PR #7393, opened 2025-09-23
Plate september update

sneridagh:
Last improvements and advancements in `@plone/plate`:

- Update all libraries to latest.
- New `LegacyLinkPlugin` that supports the legacy `link` data structure used in `slate` block
- Removed (for now) a registry-based pluggable config system (initial, temptative one), hardcoded for now in wait for the RSC powered editor. Now it would be possible to configure using a shadow.
- Rendering using `PlateStatic` means, so it uses Plate internals (which are pluggable)
- Drop all things old slate
- More "ordered" files layout
- Reusable PlateEditor and PlateView from the outside, so you can taylor your own editor and renderer (eg. outside text block use case.

ToDo:
- [x] Initial dedicated Storybook
- [ ] Cypress tests


sneridagh:
@pnicolli this is almost ready. Same functionality as the initial implementation, no bloated config registry, configuration loaded imperatively for now. When we have at hand RSCs, we will load it via one. Nevertheless, one could override the default config via a plain shadow on plate config preset.

Please take a look (or, I know, make a faith leap) :) I will add a cypress test for edit view. It would be great to have it merged ASAP, so no more merge conflicts.

PR #7401, opened 2025-09-26
Revert #7389 ESlint rule definition

sneridagh:
Since it was a breaking change.

I added it to `.eslintrc.core.js` so it's still enforced in core.

### DIFF
--- a/packages/volto/.eslintrc
+++ b/packages/volto/.eslintrc
@@ -37,13 +37,6 @@
     "react-hooks/exhaustive-deps": "warn",
     "react/react-in-jsx-scope": "off",
     "jsx-a11y/label-has-associated-control": "off",
-    "no-restricted-syntax": [
-      "error",
-      {
-        "selector": "JSXElement[openingElement.name.name='img']",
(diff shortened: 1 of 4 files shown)


wesleybl:
@sneridagh Got it. Thank you.

sneridagh:
@wesleybl no worries, all good, but we really have to think carefully what we backport and the consequences of it. Thanks!

pnicolli:
Honestly, I would have just changed it from error to warning and called it a deprecation warning. I liked having it in here helping with transitioning to volto 19.

sneridagh:
@pnicolli but a warning breaks in our config anyways... :/ Maybe we have to relax and allow warnings in the future? 
My only concern is that people doesn't find that after an update which supposed not to be breaking, all of a sudden their builds fail.

pnicolli:
@sneridagh yeah I get your point, it makes sense. I don't want builds to break of course, it would just need to be a warning. I also would not change the policy where warnings are breaking in CI. Honestly, I don't have an idea right now, I'll see if I can come up with something in the future, let's leave this as it is.

PR #7407, opened 2025-09-27
Sync core packages seven->main branch Sept2025

sneridagh:
Ok, this update is hardcore to review, I know :) So I made a summary:

I made some housecleaning in the `main` branch.
I backported and sync'ed all new things that are not breaking or Seven specifics to `main` present in `seven` branch. In fact there's (still) no such things, and we can keep the same (alpha) versions, specially for the core packages:

- `@plone/registry`
- `@plone/types`

The rest are synced for convenience and have "alpha versions" all together and keep the Plone's modular package story in `main` so it's visible as a showcase (specially the apps in `apps` folder as a demo).

- `@plone/client`
- `@plone/components`

Also, potentially, someone would need the last components in their projects, by pulling them directly from the workspaces. Although it's recomended not to and pull from released versions always.

The changelog is huge (it seems we indeed worked in Seven this year) I went through it and I was quite careful on each of them. The rest is job of the CI, if it's green, it's green.


stevepiercy:
https://github.com/plone/volto/commit/3018ce56273534c71f3074908e437174f9b4abc5 LGTM.

Issue #7124, opened 2025-05-27
Controlpanel for showing which blocks are used where


jnptk:
> [!IMPORTANT]
> If you are not a member of the Volto Team or Developers Team in the Plone GitHub organization, then do not work on or comment on this issue.

tisto:
@davisagli @jnptk let's talk about this sometime next week.

jnptk:
The Blocks Control Panel is essentially finished!

Features implemented:
- Overview of all available blocks across the site
- Click on any block to see detailed usage information:
  - Where the block is used (which pages/content objects)
  - How often it appears on each object

Screenshots showing the current implementation:

<img width="1627" height="1030" alt="Image" src="https://github.com/user-attachments/assets/ba1944fb-3a76-4c65-8907-ffb120ad5ef9" />

_List of all blocks & their total occurrence on the site_

<img width="1627" height="1030" alt="Image" src="https://github.com/user-attachments/assets/19d6e930-8ccc-4310-9bd0-daa9035695d1" />

_List of a blocks occurrences_

Status:

- Control panel UI completed
- Extended the block_types indexer with a counter
- Wrote an endpoint to expose the index to volto
- Block listing view
- Block detail view with usage statistics

Remaining work:

- Endpoint documentation
- Update `TestBlockTypesIndexer`
- Decide where this control panel should be put

@davisagli @tisto let's talk about this

tisto:
@davisagli I agree that those "content control panels" belong to a different category and that we should reflect this in the UI. However, I would suggest not binding this question to this PLIP and rather solving this as a separate matter at a later point, once we have a meaningful amount of those "content control panels". Here is a list we came up with at Beethoven Sprint:

![Image](https://github.com/user-attachments/assets/e6122aa4-19a7-4389-94c5-ab19f7152591)

Shall we create an "Editor Tools" PLIP? What do you think?

davisagli:
@tisto That sounds good to me. This means that for now, the link to the blocks control panel will only be available to users who can access Site Setup. Is that okay?

tisto:
@davisagli yeah. That would be ok for now.

wesleybl:
@jnptk Can this be closed?

davisagli:
Yes, this was added in Plone 6.2 & Volto 19

PR #7418, opened 2025-09-29
Add block types controlpanel

jnptk:
[plone/volto#7124](https://github.com/plone/volto/issues/7124)

pr in [plone.volto#197](https://github.com/plone/plone.volto/pull/197)

### DIFF
--- a/packages/volto/src/components/manage/Controlpanels/BlockType.tsx
+++ b/packages/volto/src/components/manage/Controlpanels/BlockType.tsx
@@ -89,6 +89,7 @@ const BlockTypeControlpanel = (props: RouteProps) => {
         </h1>
         {blockTypes.items?.length > 0 ? (
           <Table
+            className="react-aria-Table cmsui-table"
             columns={[
               {
                 id: 'title',
--- a/packages/volto/src/components/manage/Controlpanels/BlockTypes.tsx
+++ b/packages/volto/src/components/manage/Controlpanels/BlockTypes.tsx
@@ -83,6 +83,7 @@ const BlockTypesControlpanel = (props: RouteProps) => {
           <FormattedMessage id="Block Types" defaultMessage="Block Types" />
         </h1>
         <Table
+          className="react-aria-Table cmsui-table"
           columns={[
             {
               id: 'blockType',
--- a/packages/volto/theme/themes/pastanaga/extras/block-types.less
+++ b/packages/volto/theme/themes/pastanaga/extras/block-types.less
@@ -1,10 +1,7 @@
-.controlpanel-block-types,
-.controlpanel-block-type {
-  .react-aria-Table {
-    --plone-table-cell-border-bottom: 1px solid var(--quanta-smoke);
-    --plone-table-header-border-bottom: 2px solid var(--quanta-silver);
-    border-radius: 0;
-  }
+.react-aria-Table.cmsui-table {
+  --plone-table-cell-border-bottom: 1px solid var(--quanta-smoke);
+  --plone-table-header-border-bottom: 2px solid var(--quanta-silver);
+  border-radius: 0;
 
   .react-aria-Row {
     --radius-top: 0px;


packages/volto/src/components/index.js
```diff
@@ -98,6 +98,7 @@ export {
   UpgradeControlPanel,
   ModerateComments,
   DatabaseInformation,
+  BlockTypesControlpanel,
```
davisagli: Don't add this. We are trying to stop using barrel imports, not add to them. (It should be imported from the actual file where the component is, not from an index.)

packages/volto/theme/themes/pastanaga/extras/block-types.less
```diff
@@ -0,0 +1,56 @@
...
+    &:first-child {
+      width: 100%;
+    }
+
+    &:last-child {
+      width: 1%;
+      text-align: right;
+    }
+  }
+}
```
davisagli, line 16: We should use the existing `ui table` classes, not add new styles just for this page. That's a good way to make sure the appearance isn't consistent with other tables.

packages/volto/src/components/manage/Controlpanels/Controlpanels.jsx
davisagli, line 1: The test snapshots for this component need to be updated, since the markup now includes the new control panel.

packages/volto/src/components/manage/Controlpanels/index.tsx
```diff
@@ -182,3 +182,13 @@ export const DatabaseInformation = loadable(
...
 );
+
+// BLOCKTYPES CONTROLPANEL
+
+export const BlockTypesControlpanel = loadable(
+  () => import('@plone/volto/components/manage/Controlpanels/BlockTypes'),
+);
+
+export const BlockTypeControlpanel = loadable(
+  () => import('@plone/volto/components/manage/Controlpanels/BlockType'),
```
davisagli: Don't add these. No new barrel imports.

packages/volto/src/reducers/index.js
```diff
@@ -53,6 +53,7 @@ import upgrade from './upgrade/upgrade';
 import userschema from './userschema/userschema';
 import site from './site/site';
 import navroot from './navroot/navroot';
+import blockTypes from './blockTypes/blockTypes';
```
davisagli, line 56: This should be ./blocktypes/blocktypes. Some filesystems are case sensitive.

packages/volto/src/components/manage/Controlpanels/BlockTypes.tsx
```diff
@@ -0,0 +1,113 @@
...
+  const isClient = useClient();
+  const blocks = [];
+
+  useEffect(() => {
+    dispatch(getBlockTypes());
+  }, [dispatch]);
+
+  for (const block of Object.values(blocksConfig)) {
+    blocks.push(block);
+  }
```
davisagli: The block types seem to be in random order. We should sort them by title or # of occurences (probably title is better, since the user is likely to be looking for a specific block)

packages/volto/src/components/manage/Controlpanels/BlockTypes.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+  if (blockTypes.loading) {
+    return <div>Loading...</div>;
```
davisagli: Needs i18n

packages/volto/src/components/manage/Controlpanels/BlockTypes.tsx
```diff
@@ -0,0 +1,113 @@
...
+        id="page-block_types"
+        className="ui container controlpanel-block_types"
+      >
+        <h1>
+          <FormattedMessage id="block-types" defaultMessage="Block Types" />
+        </h1>
+        <table className="table">
+          <thead className="table-header">
+            <tr className="table-row">
+              <th className="table-heading">
```
davisagli: These class names table-header, table-row, table-heading are redundant with the semantic information from the tags thead, tr, th

packages/volto/src/components/manage/Controlpanels/BlockTypes.tsx
```diff
@@ -0,0 +1,113 @@
...
+                <FormattedMessage id="Type" defaultMessage="Type" />
```
davisagli: I would say Block Type here to avoid confusion with Content Type

packages/volto/src/components/manage/Controlpanels/BlockType.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+  useEffect(() => {
+    dispatch(getBlockTypes(id));
+  }, [dispatch, id]);
+
+  const onChangeSearch = (value: string) => {
+    dispatch(getBlockTypes(id, value));
+  };
+
+  const debouncedSearch = debounce(onChangeSearch, 600);
```
davisagli: The search is not working well. If I pause typing, the page reloads and the search box is empty. I assumed it would filter down the titles I can see on the page, but actually it does a catalog query by path, so it doesn't return anything unless the user types a path starting with /Plone, which is very non-obvious.

I would just remove the search. The results are already all on the page and the user can use their browser's find feature.

(The search would be more important if we had paginated results. But then it should search titles, not paths. Or it should make it obvious that it is a path search, and take care of adding the site root path automatically.)

packages/volto/src/components/manage/Controlpanels/BlockType.tsx
```diff
@@ -0,0 +1,147 @@
...
+
+  if (blockTypes.loading) {
+    return <div>Loading...</div>;
```
davisagli: Needs i18n

packages/volto/src/components/manage/Controlpanels/BlockType.tsx
```diff
@@ -0,0 +1,147 @@
...
+      <div
+        id="page-block_type"
+        className="ui container controlpanel-block_type"
+      >
+        <h1>
+          <FormattedMessage
+            id="block-type"
+            defaultMessage="{title}"
+            values={{ title: block.title }}
+          />
```
davisagli: Defining a message that only includes a variable is useless (the only thing we want to translate here is the variable, and that won't be translated).

Instead of this, do `{intl.formatMessage({ id: block.title, defaultMessage: block.title })}` to use the existing messages that are defined for block titles.

(similar to here: https://github.com/plone/volto/blob/main/packages/volto/src/components/manage/BlockChooser/BlockChooser.jsx#L148)

packages/volto/src/components/manage/Controlpanels/BlockType.tsx
```diff
@@ -0,0 +1,147 @@
...
+                    defaultMessage="Occurrence"
+                  />
+                </th>
+              </tr>
+            </thead>
+            <tbody className="table-body">
+              {blockTypes.items.map((item) => (
+                <tr key={item['@id']} className="table-row">
+                  <td className="table-cell">
+                    <a href={item['@id']}>{item.title}</a>
```
davisagli: Add a space before the path:
```suggestion
                    <a href={item['@id']}>{item.title}</a>{' '}
```

packages/volto/src/components/manage/Controlpanels/BlockType.tsx
```diff
@@ -0,0 +1,147 @@
...
+                  <td className="table-cell">{item.count}</td>
+                </tr>
+              ))}
+            </tbody>
+          </table>
+        ) : (
+          <FormattedMessage
+            id="no-blocks-found"
+            defaultMessage="No items found for type: {type}."
+            values={{ type: id }}
```
davisagli: Translate the block title like I said above, then pass the translated title in as the type here.

davisagli (changes requested):

packages/volto/src/reducers/index.js
jnptk: Git didn't track the rename of this file (which I did at some point to be consistent in naming) idk why but should be fixed now

packages/volto/src/components/manage/Controlpanels/BlockTypes.tsx
```diff
@@ -0,0 +1,131 @@
...
+  return (
+    blockTypes.loaded && (
+      <div
+        id="page-block_types"
+        className="ui container controlpanel-block_types"
+      >
+        <h1>
+          <FormattedMessage id="Block Types" defaultMessage="Block Types" />
+        </h1>
+        <Table
```
davisagli, line 85: @sneridagh We used the @plone/components table here, but it isn't styled very well for this use case (minimum width, no grid, too much padding). I think we need a variation of the table that works well for CMSUI routes.

Screenshot 2026-02-13 at 2 35 40 PM

packages/volto/src/components/manage/Controlpanels/BlockTypes.tsx
```diff
@@ -0,0 +1,131 @@
...
+    }))
+    .sort((a, b) => (a.title === b.title ? 0 : a.title > b.title ? 1 : -1));
+
+  useEffect(() => {
+    dispatch(getBlockTypes());
+  }, [dispatch]);
+
+  if (blockTypes.loading) {
+    return <div>{intl.formatMessage(messages.loading)}&hellip;</div>;
+  }
```
davisagli, line 70: @sneridagh We need a loading spinner in `@plone/components`

sneridagh: PR: https://github.com/plone/volto/pull/7918

packages/volto/src/components/manage/Controlpanels/BlockTypes.tsx
sneridagh: @davisagli I will take a look tomorrow and add some nice styling for tables in Volto control panels in Volto itself.

sneridagh:
@davisagli @jnptk I merged the new Spinner component in Volto 19a26.

sneridagh:
@davisagli @jnptk 
image

sneridagh:
@davisagli what's the status of the plone.volto PR? I thought that it was merged already.

packages/volto/theme/themes/pastanaga/extras/block-types.less
```diff
@@ -0,0 +1,19 @@
+.controlpanel-block-types,
+.controlpanel-block-type {
```
davisagli: @sneridagh We should aim for consistent table styles in all CMSUI, not one-off styles for a particular control panel. Should we replace this with:
```suggestion
.cms-ui {
```

?

davisagli (commented):
@sneridagh I was waiting to merge https://github.com/plone/plone.volto/pull/197 until we had a chance to test it with this. I think we can go ahead and merge it and make a new alpha release now.

packages/volto/theme/themes/pastanaga/extras/block-types.less
sneridagh: @davisagli this will apply to the "theme" tables as well in `/edit` and `/add` too. The `.cmsui` is set at `body` level. I would be more surgical on this.

But it's a good playground for the debate that we should have when applying the new `@plone/components` policies and how we do theming in the times to come.

davisagli: @sneridagh Oh, good point.

That makes sense to be more surgical but I would like to do something that is semantic rather than applying styles to one specific control panel. It should be possible to use the same table in another control panel without changing the stylesheet.

davisagli:
I released plone.volto 6.0.0a2 with the backend for this.

packages/volto/theme/themes/pastanaga/extras/block-types.less
sneridagh: @davisagli I suggest that we bind it to `.section-controlpanel table` then. What do you think?

davisagli: @sneridagh My first reaction: that sounds ok.

My second reaction: it's still styling based on location rather than semantics. It wouldn't be possible to reuse the styles in another place where we want the same table appearance, like one of the content views (sharing, aliases, etc)

I would suggest we apply these styles to tables with a particular class, such as `table.cmsui-grid`.

tisto:
@davisagli @sneridagh if we want this to be shipped with Volto 19 final and Plone 6.2 we need to push this soon. I mentioned this in the state of plone keynote, so I would love to see my promise on the stage to become a reality. :)

sneridagh:
@jnptk @davisagli what's left to do here? I guess that make Volto CI use the new release (just until the next 6.2a is out) then merge it?

davisagli:
@sneridagh @tisto This is ready but was sitting here waiting until there is a new Plone release that includes plone.volto 6.0.0a2. We can also temporarily add plone.volto 6.0.0a2 as an override, and we should probably do that.

packages/volto/theme/themes/pastanaga/extras/block-types.less
davisagli: @sneridagh I changed it to use a `.cmsui-table` class in https://github.com/plone/volto/pull/7418/commits/fe6e0bf11539f0e6bbcef001fc80ecca470ba336

sneridagh:
@davisagli 🎉 thanks!!!

tisto:
@davisagli @sneridagh @jnptk awesome! Can we ship this somewhere? @jnptk can you pls create a screencast of that feature?

davisagli:
@tisto We'll need a release first...so maybe once @sneridagh is back.

PR #7427, opened 2025-10-01
Fix logo retrieval from the site endpoint data in the root loader.

sneridagh:


packages/layout/components/Logo/Logo.tsx
```diff
@@ -6,16 +6,20 @@ import { messages } from '../messages';
...
 const Logo = (props: SlotComponentProps) => {
   const { content } = props;
+  const rootData = useRouteLoaderData<RootLoader>('root');
+
   const intl: (id: string) => string = config.getUtility({
     name: 'translation',
     type: 'factory',
   }).method;
 
   const navRootPath = content['@components'].navroot?.navroot?.['@id'] || '/';
```
pnicolli, line 21: Maybe we should make it a habit to get the content from the root loader instead of the props? Not necessarily now, but something that should be discussed imho

sneridagh: Yes, could recommend that by default. Since this is a slot, and you get it for free...

PR #7435, opened 2025-10-02
chore(@plone/scripts): upgrade execa to avoid security vulnerability

nileshgulia1:
execa <2.0.0 has a critical security vulnerability in docker. According to docker scout:

```
   1C     0H     0M     0L  execa 0.6.3
pkg:npm/execa@0.6.3

Dockerfile (22:22)
COPY --from=builder /app/ /app/

    ✗ CRITICAL GMS-2020-2 [OWASP Top Ten 2017 Category A9 - Using Components with Known Vulnerabilities]
      https://scout.docker.com/v/GMS-2020-2
      Affected range : <2.0.0                                        
      Fixed version  : 2.0.0                                         
      CVSS Score     : 9.8                                           
      CVSS Vector    : CVSS:3.0/AV:N/AC:L/PR:N/UI:N/S:U/C:H/I:H/A:H  
      
```
      
      
I updated it to the latest version with a small API change. Right now, execa 1.0.0 is still being installed through @testing-library/jest-dom and razzle (which isn’t maintained anymore). execa@1.0.0 is also a direct dependency of "sane" (v4.x, which again comes from majorly "jest"). This issue is fixed in sane@5.0.1, so it’s worth considering an upgrade/override.


packages/scripts/package.json
```diff
@@ -46,7 +46,7 @@
     "chalk": "4",
     "commander": "8.2.0",
     "comment-json": "^4.2.3",
-    "execa": "0.6.3",
```
nileshgulia1, line 49: @sneridagh Any idea why we are using very old version, bundlesize?

wesleybl:
> Right now, execa 1.0.0 is still being installed through @testing-library/jest-dom and razzle (which isn’t maintained anymore). execa@1.0.0 is also a direct dependency of "sane" (v4.x, which again comes from majorly "jest"). This issue is fixed in sane@5.0.1, so it’s worth considering an upgrade/override.

I agree with that.

In fact, I'm wondering if it's worth creating a new "@volto/razzle" package to update these packages. But maybe that wouldn't be worth it because these issues will be fixed in Seven.

sneridagh:
@wesleybl we could explore the idea of such a package. We could fork it and see what happens. What concerns me is that the number of packages to fork would explode quickly.

sneridagh:
@nileshgulia1 changelog, please. I'm ok with this if it works well.

wesleybl:
> we could explore the idea of such a package. We could fork it and see what happens. What concerns me is that the number of packages to fork would explode quickly.

@sneridagh I'll try to make a PR for this. At least so we can update the packages with vulnerabilities.

davisagli:
> we could explore the idea of such a package. We could fork it and see what happens. What concerns me is that the number of packages to fork would explode quickly.

I think it makes sense to explore for razzle, which is not really maintained any more. It will still be some time before we have a final version of Seven, and this would give us a way to make fixes or upgrade dependencies when necessary.

nileshgulia1:
> @nileshgulia1 changelog, please. I'm ok with this if it works well.

@sneridagh I tested running the lockhook.js script manually works like normal.

PR #7443, opened 2025-10-03
[Seven] Added registry server loader feature

sneridagh:
Ok, I could not help it. I had this idea (a subsequent PR which uses this is coming):

What if I want configuration that only is loaded in the server, because it's releated to loaders? What if we offload the config to not be in the bundle, but backend powered, loaded in loaders. So we free the client config from it. Still to be seen, but what if we can offload there all the component registry, serve it only when it's needed via RSCs?

And, regarding the next PR. What if my add-on wants to add things to the root loader? as in, gimme some extra info from the backend? eg. a slot that gets some extra data.

I think also that if we succeed with this, we can use it for separate PublicUI config from CMSUI config (eg. blocks).

----
📚 Documentation preview 📚: https://volto--7443.org.readthedocs.build/


docs/conceptual-guides/add-on-loader.md
```diff
@@ -47,6 +47,28 @@ Don't modify it or add your own styles in it.
...
+If your add-on does not need to load any server-side configuration, you can omit this file.
+```
+
+This loader is a JavaScript file and it is placed in the root of your application.
+By default, it's called {file}`registry.loader.server.js`.
+`@plone/registry` will create or overwrite {file}`registry.loader.server.js` in the root of your app whenever you run the bundler.
+
+```{important}
+The file {file}`registry.loader.server.js` is generated and maintained by `@plone/registry`.
+Don't modify it or add your own styles in it.
```
pnicolli:
```suggestion
Don't modify it or add your own code in it.
```

Maybe copy-paste?

pnicolli (approved):

Issue #5951, opened 2024-04-09
You can't add a user if email login is set in security


djay:
**Describe the bug**
You can't add a user if email login is set in security settings

**To Reproduce**

1. Set email login in control panel > security
2. Add user
3. if put the email in the user name field you will get an error saying it needs to be left blank. If you leave it blank you get an error that it's required.

**Expected behavior**
No login name field is displayed  when adding a user and a user can be added

**Screenshots**

**Software (please complete the following information):**

- Volto Version 16
- Plone Version 6
- Plone REST API Version [e.g. 7.0.1]

**Additional context**

djay:
might get fixed as part of #5952 ?

djay:
A quick fix might be to remove the clientside validation on the username field if there is currently no easy way to know the security settings when that form is rendered?

davisagli:
The form already fetches the usergroup control panel settings. I haven't looked at the details but I would guess it's not hard to make it also fetch the security settings.

wesleybl:
> if you leave it blank you get an error that it's required.

This shouldn't happen see: https://github.com/plone/plone.restapi/blob/5cf1418e928f143ef1b42fdc3e4ed72a3126e9dd/src/plone/restapi/services/users/add.py#L69-L71

What version of `plone.restapi` are you using?

JeffersonBledsoe:
It appears that this is mostly a Volto 16 issue, as in Volto 17 we use the userschema to display the fields where the username field isn't required so can be skipped over. I've included screenshots below showing the difference in an out-the-box Volto experience

Add user form in Volto 16
<img width="605" alt="" src="https://github.com/plone/volto/assets/30210785/0e4dfebd-82a3-49b7-b65c-bf6453e3fab6">
Add user form in Volto 17
<img width="639" alt="" src="https://github.com/plone/volto/assets/30210785/669931cf-589a-4c46-8002-721aa4a756c5">

However, it still isn't ideal that this field is display. I did look into adding the `use_email_as_login` option to the userschema, but I don't this makes sense as the userschema can be used elsewhere and the option isn't really linked to users. I think it would make sense to include the `use_email_as_login` as part of the `@site` endpoint as it's more of a global setting which could make sense to be used elsewhere (for example, in the Login form as mentioned in the issue description). This would also be a suitable place for the other public security settings like `enable_self_reg`.

@djay @davisagli Any thoughts on this?

djay:
still a bug

PR #7450, opened 2025-10-04
Convert UsersControlpanel component from class to functional with hooks

wesleybl:
- Convert UsersControlpanel from class to functional component using React hooks (useState, useEffect, useCallback).
- Optimize useCallback usage and add optional chaining for safer prop access.
- Uses hooks instead of compose.

Ref https://github.com/plone/volto/issues/4460


wesleybl:
@davisagli @sneridagh @ksuess When you have time, can you please test this?

packages/volto/src/components/manage/Controlpanels/Users/UsersControlpanel.jsx
```diff
@@ -430,359 +431,303 @@ class UsersControlpanel extends Component {
...
+  if (userschema?.loaded) {
+    isUserManager = isManager(user);
+    adduserschema = JSON.parse(JSON.stringify(userschema?.userschema));
+
+    // Add custom form fields to the schema
+    adduserschema.properties.username = {
+      title: intl.formatMessage(messages.addUserFormUsernameTitle),
+      type: 'string',
+      description: intl.formatMessage(messages.addUserFormUsernameDescription),
+    };
```
ksuess, line 496: Field `username` should not be added if `loginUsingEmail` is true.

packages/volto/src/components/manage/Controlpanels/Users/UsersControlpanel.jsx
```diff
@@ -297,131 +307,122 @@ class UsersControlpanel extends Component {
...
+          <Toast
+            error
+            title={intl.formatMessage(messages.error)}
+            content={intl.formatMessage(
+              messages.addUserFormPasswordAndSendPasswordTogetherNotAllowed,
+            )}
+          />,
+        );
+      } else {
+        if (userGroups && userGroups.length > 0) addUserToGroup(data);
```
ksuess, line 328: I think `addUserToGroup` should happen after user is created successfully.

ksuess (changes requested):
Modal "add user"
- field `username` only if not "Use email address as login name"
- field `username` should be required if not "Use email address as login name*

On creating a new user
- group membership is not set

Nice to have: close modal "add user" on esc and click outside

packages/volto/src/components/manage/Controlpanels/Users/UsersControlpanel.jsx
wesleybl: This is a problem that occurs in Volto today. See:

https://github.com/plone/volto/issues/5951

In this PR I'm only concerned with migrating to a functional component. This already changes a lot of things. Fixing a bug in mine would make the PR more confusing. This can be fixed in another PR, after the component has been migrated.

packages/volto/src/components/manage/Controlpanels/Users/UsersControlpanel.jsx
wesleybl: Nothing is sent to the backend until the form is validated. No attributes are set until the user is created. I followed the current method:

https://github.com/plone/volto/blob/main/packages/volto/src/components/manage/Controlpanels/Users/UsersControlpanel.jsx#L317

If necessary, we can change this later.

wesleybl:
> On creating a new user

> group membership is not set

@ksuess I couldn't simulate that. I was able to add the user to a group during user creation. Could you please test it again?

wesleybl:
> field username should be required if not "Use email address as login name*

@ksuess It's being requested. Could you please test it again?

wesleybl:
> Nice to have: close modal "add user" on esc and click outside

@ksuess This is an improvement. It can be done after this PR.

ksuess:
> > On creating a new user
> 
> > group membership is not set
> 
> @ksuess I couldn't simulate that. I was able to add the user to a group during user creation. Could you please test it again?

If "Use email address as login name" is true, then the group is not set. This is because 
PATCH http://localhost:3000/++api++/@groups/Reviewers is called with payload {"users":{"undefined":true}}

ksuess:
> > field username should be required if not "Use email address as login name*
> 
> @ksuess It's being requested. Could you please test it again?

yes it's required, you are right. i was misleaded because there is no red circle to mark the field visually as required.

ksuess:
I'm fine with not fixing anything here, but focussing on the conversion to a functional component. Therefore I approve.

PR #7454, opened 2025-10-05
Skip link check for lanku.eus

davisagli:
I think this is the most appropriate workaround since the website is functional when accessed from somewhere other than GitHub.

### DIFF
--- a/.github/workflows/readme-link-check.yml
+++ b/.github/workflows/readme-link-check.yml
@@ -17,7 +17,7 @@ jobs:
       with:
         fail: true
         args: >-
-          --timeout 10
+          --timeout 5
           --max-retries 2
           --exclude https://www.npmjs.com/
           --exclude http://localhost:8080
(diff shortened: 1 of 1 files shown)


stevepiercy:
What happens if you increase lychee's timeout from 5 to 10 seconds? We haven't tried that yet.

davisagli:
I don't think we should increase our timeout everywhere because one site is responding slowly, and it's quite likely it's blocking the connection altogether, since the site loads reasonably quickly for me.

stevepiercy:
I'd still like to try it before excluding @erral's site. Lychee is blazing fast compared to previous link checkers, so maybe the timeout increase won't matter. The only checks that finish faster are change log and Plone Contributor Agreement.

davisagli:
@stevepiercy No, it still times out with a 10-second check. I'd like to prioritize merging this so the builds aren't red, and then if there are other things you want to try, it can happen on a branch.

davisagli:
Now a different site is not responding: https://www.intk.com/en

@stevepiercy Maybe we should run this check on a schedule once a day instead of on every push?

stevepiercy (approved):
It was worth a try. Thanks!

That other site was temporary. Let's merge it.

davisagli:
@stevepiercy I cherry-picked the change to the 18.x.x branch also: https://github.com/plone/volto/commit/4cff069176dcb204fcdd10856b2ca71e4766b1cb

PR #7469, opened 2025-10-08
Update to latest RAC, some styling fixes here and there. 

sneridagh:
Added ObjectBrowserWidget story, TagGroup and ListBox. Added TokenSelect from RAC examples.

@deodorhunter please, take a look at the story, totally AI generated, but I wanted to have something in there. Also for catching regressions (the TagsGroup component had a thinggy and I wanted to be sure that I was not breaking anything).

I think this is ready for now, the components still need some love, though.

----
📚 Documentation preview 📚: https://volto--7469.org.readthedocs.build/


deodorhunter:
Sure thing, I'll have a look at this in the weekend :)

sneridagh:
@deodorhunter @rboixaderg @pnicolli Question: Why did we create an `icon` variant? I can't remember why did we decide it. I worked in supporting it without having to define anything eg. the styling detecting the svg inside.

Problem is, that we cannot have combinations of them eg. icon primary, etc... TBH, I don't know if that's a problem or not. Also, do you remember where did we copied the style from? it's different thant the Quanta ones.

sneridagh:
@deodorhunter @pnicolli merging this now, I'd like to have it as baseline for other things.

PR #7474, opened 2025-10-09
[Seven] Added Brain to image component item types

pnicolli:
I forgot to add the Brain type in the last PR :(

----
📚 Documentation preview 📚: https://volto--7474.org.readthedocs.build/

----
📚 Documentation preview 📚: https://plone-registry--7474.org.readthedocs.build/

### DIFF
--- a/packages/layout/components/Image/Image.tsx
+++ b/packages/layout/components/Image/Image.tsx
@@ -5,6 +5,7 @@ import type {
   ContainedItem,
   Content,
   RelatedItem,
+  Brain,
 } from '@plone/types';
 
 function removeObjectIdFromURL(basePath: string, scale: string) {
@@ -31,7 +32,7 @@ export function flattenScales(path: string, image: any) {
 }
 
 export interface ImageProps extends ImgHTMLAttributes<HTMLImageElement> {
-  item?: Content | ContainedItem | RelatedItem;
+  item?: Content | Brain | ContainedItem | RelatedItem;
   imageField?: string;
   src?: string;
   alt: string;
--- a/packages/layout/news/+moreimagetypes.bugfix
+++ b/packages/layout/news/+moreimagetypes.bugfix
@@ -1,0 +1,1 @@
+Added Brain to image component item types. @pnicolli


pnicolli:
> We don't have anything called a Brain in the frontend. We have the Summary serialization of a content item, which the REST API creates from a brain.

Catalog search results and summary serializations (e.g. related items) go through different serializers in plone.restapi afaik.
We have two separate types in `@plone/types` today that refer to these two separate situations, and the Image component typing is currently missing the former.
I understand these are two topics that are not fully dependant on each other (what is serialized and the types we have defined in the front-end), but I think the current state of the types resembles what is actually serialized from plone.restapi, or at least this is my understading.

davisagli:
> Catalog search results and summary serializations (e.g. related items) go through different serializers in plone.restapi afaik.

They are both using a summary serializer, that is an adapter to ISerializeToJsonSummary:

https://github.com/plone/plone.restapi/blob/main/src/plone/restapi/serializer/catalog.py#L62

https://github.com/plone/plone.restapi/blob/main/src/plone/restapi/serializer/relationfield.py#L23

In both cases that uses DefaultJSONSummarySerializer (https://github.com/plone/plone.restapi/blob/main/src/plone/restapi/serializer/summary.py), but it in turn adapts its context to IContentListingObject

For catalog search results we're adapting a brain so that gets CatalogContentListingObject (https://github.com/plone/plone.app.contentlisting/blob/master/src/plone/app/contentlisting/catalog.py)

For related items we're adapting a full object so that gets RealContentListingObject (https://github.com/plone/plone.app.contentlisting/blob/master/src/plone/app/contentlisting/realobject.py)

What is the actual difference that you're trying to represent in the types here? The intent is that the summary serialization format should be consistent, so if it's not, that might be a bug that we should look at in the REST API, instead of updating the types to pretend the inconsistency is correct. But I need more info in order to give specific advice.

pnicolli:
Thank you for the thorough explanation. My backend-fu was not good enough to use the proper terms, basically what I meant is that I get different data if I get something from a catalog search result and from a related item.

I just went to a plain Volto 18 site, created and Event with a preview image and a News Item. In the News Item I created a teaser block targeting the Event and also added the Event as related item in the News Item. Now if I look at the data I receive for the block and for the relatedItems field, they are different. Is this a different scenario and I didn't realize that?

I also understand my experience is mostly on slightly older restapi and volto versions unfortunately so things might have gotten better and I just don't know :) I'm fine with changing `@plone/types` eventually if we can define a single baseline type for whatever is serialized from the backend.

sneridagh:
@davisagli I think the change is sound and depicts exactly the mess that we have now in the Images component usage in Plone/Volto, which we have to fix in the near future. It's not the first time we talk about it.

I guess that your only concern is about the naming, right?

sneridagh:
@davisagli Since it seems your concern is about the current `@plone/types` naming, I'll merge this for now, because it only completes the typings for this component.

I talked with @pnicolli the other day that we have to review that naming anyways, I also hinted that maybe a name with "serialize" in it would be better, however, I'm not entirely sure that the types have to note that, since everything has been serialized somehow at some point :) So, calling that by its name maybe is not that bad. And, in Plone world, that is a brain, right? Granted, a "conditional" Partial<Brain> in most of cases (some does not).

pnicolli:
We should definitely talk about the restapi responses during the conference next week if we can, I have several ideas and a good amount of confusion in mind 😂

davisagli:
> I just went to a plain Volto 18 site, created and Event with a preview image and a News Item. In the News Item I created a teaser block targeting the Event and also added the Event as related item in the News Item. Now if I look at the data I receive for the block and for the relatedItems field, they are different. Is this a different scenario and I didn't realize that?

I just tried this. This appears to be a difference not in the data returned from the backend, but in how Volto is filtering it. They both use the object browser widget, so both of them get data from the same place (summary serialization of a brain returned in the @querystring-search endpoint). The related items field does not specify selectedItemAttrs for the widget, so it stores all of the fields returned (https://github.com/plone/volto/blob/main/packages/volto/src/components/manage/Widgets/ObjectBrowserWidget.jsx#L310). The teaser block specifies a limited set of selectedItemAttrs to include (https://github.com/plone/volto/blob/main/packages/volto/src/components/manage/Widgets/ObjectBrowserWidget.jsx#L310) so the widget only returns those (https://github.com/plone/volto/blob/main/packages/volto/src/components/manage/Widgets/ObjectBrowserWidget.jsx#L195).

pnicolli:
Interesting, I didn't investigate further at this time. I remember bigger discrepancies, but I cannot promise it was because of other client customizations and because it was a previous version of restapi and volto.

PR #7475, opened 2025-10-09
[Seven] Added PoC for content types views

pnicolli:
The filter by CT name is definitely not final but it works for development purposes at this time, I believe. I think eventually we should have restapi just tell us whether the behavior is enabled, or create a custom view, or something more robust than guessing if we should render blocks.

### DIFF
--- a/packages/layout/slots/ContentArea.tsx
+++ b/packages/layout/slots/ContentArea.tsx
@@ -1,16 +1,16 @@
 import RenderBlocks from '../blocks/RenderBlocks';
 import type { SlotComponentProps } from './SlotRenderer';
+import { hasBlocksData } from '@plone/helpers';
 import config from '@plone/registry';
 
 const ContentArea = (props: SlotComponentProps) => {
   const { content } = props;
-  const hasBlocksData = !!content.blocks && !!content.blocks_layout;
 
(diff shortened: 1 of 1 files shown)


packages/layout/slots/ContentArea.tsx
```diff
@@ -4,16 +4,32 @@ import config from '@plone/registry';
...
-        blocksConfig={config.blocks.blocksConfig}
-        pathname="/"
-      />
-    </>
-  );
+  if (
+    ['Document', 'Plone Site', 'LRF', 'News Item', 'Event'].includes(
+      content['@type'],
+    ) &&
+    hasBlocksData
```
pnicolli: Maybe checking the blocks data doesn't make sense?

sneridagh: I think it does, and in Volto it's checked like this... but it's done per view, for supporting default views for content types with no blocks.

BTW, there's already a helper in `@plone/helpers` for it. Check it out.

sneridagh:
@pnicolli I added the use of the `@plone/helpers` helper.

pnicolli:
Thanks for completing this, all good to me!

PR #7477, opened 2025-10-10
Update slate to latest

sneridagh:


packages/volto-slate/src/editor/SlateEditor.jsx
```diff
@@ -193,14 +193,32 @@ class SlateEditor extends Component {
...
       this.props.onUpdate(editor);
     }
   }
 
+  scheduleFocus(editor) {
+    if (!editor) return;
+
+    const schedule =
+      typeof window !== 'undefined' && window.requestAnimationFrame
+        ? window.requestAnimationFrame.bind(window)
```
wesleybl: Use `(fn) => window.requestAnimationFrame(fn)` to avoid creating a new function with `bind`.

packages/volto-slate/src/editor/SlateEditor.jsx
```diff
@@ -193,14 +193,32 @@ class SlateEditor extends Component {
...
+        : (fn) => setTimeout(fn, 0);
+
+    schedule(() => {
+      if (this.isUnmounted) return;
+      try {
+        ReactEditor.focus(editor);
+      } catch (e) {
```
wesleybl: The `e` can be removed since it is not used.

packages/volto-slate/src/editor/SlateEditor.jsx
sneridagh: Done!

packages/volto-slate/src/editor/SlateEditor.jsx
sneridagh: Done!

PR #7479, opened 2025-10-10
Fixed corner case where RAZZLE_INTERNAL_API_PATH is set and id is still undefined on first SSR load. (#7478)

sneridagh:


davisagli:
I cherry-picked this already.

PR #7482, opened 2025-10-10
Missed update slate libraries in volto package itself

sneridagh:


packages/volto/package.json
```diff
@@ -259,9 +259,10 @@
     "semantic-ui-less": "2.4.1",
     "semantic-ui-react": "2.1.5",
     "serialize-javascript": "3.1.0",
-    "slate": "0.100.0",
-    "slate-hyperscript": "0.100.0",
-    "slate-react": "0.98.4",
+    "slate": "0.118.1",
+    "slate-dom": "0.118.1",
```
wesleybl: It seems that `slate-dom` is not imported in Volto. So it shouldn't be here.

sneridagh:
@wesleybl done!

PR #7481, opened 2025-10-10
Internationalizes help for the `Group Name` field.

wesleybl:

PR #7483, opened 2025-10-10
Internationalizes help for the `Group Name` field

wesleybl:
Backport of #7481 to Volto 18


wesleybl:
Why does Voto 18 have more languages ​​than Voto 19? Isn't it missing in Voto 19?

davisagli:
@wesleybl I think that's expected, due to a recent project to add more translations with the help of LLMs. We'll sync the translations once Volto 19 is closer to the final release.

PR #7268, opened 2025-08-08
Fix InlineForm to always pass error as array to prevent prop type errors

alexandreIFB:
This PR fixes an issue where the `InlineForm` component would pass the `error` prop as an object instead of an array, causing prop type errors and warnings in widgets that use `FormFieldWrapper`, such as Checkbox.

**How to test:**
1. Add a Teaser block to the page.
2. Check the browser console and confirm that prop type errors related to the error prop are no longer present.

Closes #7267

PR #7484, opened 2025-10-10
Fix InlineForm to always pass error as array to prevent prop type errors

wesleybl:
Backport of #7268 to Volto 18


wesleybl:
@alexandreIFB FYI

PR #7486, opened 2025-10-11
Avoid including always the original size in the Image component.

sneridagh:
@davisagli lots of things to improve in the `Image` component, but please let's have this in already. I'd like to have a wow effect if someone is checking lighthouse today in Seven. At least, with the things that are already there!

As first step, for now it reads the sizes config from the site, then compares them with the image sizes, and if the image sizes are not complete, the original image is included.

By all means we need to add the sizes too whenever necessary too and the switch for the use case that we really want the original in there or include a big enough size too for those use cases (maybe also on demand).

With this in place lighthouse throws:

image

🎉 

PD: Some other quick fixes included in this one.

### DIFF
--- a/packages/layout/components/Image/Image.tsx
+++ b/packages/layout/components/Image/Image.tsx
@@ -40,7 +40,6 @@ export interface ImageProps extends ImgHTMLAttributes<HTMLImageElement> {
   alt: string;
   loading?: 'eager' | 'lazy';
   responsive?: boolean;
-  sizes?: string;
 }
 
 export default function Image(props: ImageProps) {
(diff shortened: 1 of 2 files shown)


davisagli:
Okay, but I think we'll also need to:
- Add a default scale size which is large enough to serve full-width images on a large high-DPI monitor
- Fix the image field serializer in plone.restapi to include the same scales as the code which creates image_scales for the catalog (so that we aren't generating scales larger than the original).
- Always set `sizes` appropriately when we use the Image component (so that we don't end up with the largest scale when a smaller one would be sufficient)

pnicolli:
Fully agree, let's make sure we create three issues for these items where appropriate, unless there is work in progress already.

sneridagh:
@davisagli +1000 to all!

Issue #2458, opened 2021-05-17
Search component seems to work "differently" than Plone's classic one


sneridagh:
**Describe the bug**
Following the thread of a search problem I hinted a problem with the "sort_order" in the search component, and I realised that it really does not behave like the Plone one.

Also, the `Search` component should be refactored to match the server side improvements made in #2264. Since most of the code is not required anymore because of that.

**To Reproduce**
This Cypress test fixture works different in Plone than in Volto:
https://github.com/plone/volto/blob/33727ceddc67fa562cbfbd69b00e5acaba9e667d/cypress/tests/core/search.js#L91

Also, the results are different.

**Expected behavior**
Both results are consistent.

davisagli:
This issue is too vague to be useful for knowing what needs to be fixed. I think it's the problem we fixed in https://github.com/plone/plone.restapi/pull/1954

PR #7489, opened 2025-10-11
Updated Cypress test for search date sorting to verify newest-first ordering

### LINKED ISSUE #7452: Search result  (Date newest first) is not sort right
in the search result page, Search result  (Date newest first) is not sort right

Volto 18.27.3
Plone 6.1.3

aryan7081:
Ref #7452

Inside the `doSearch` function of `packages/volto/src/components/theme/Search/Search.jsx`, date sorting was incorrectly using ascending order (oldest first) by default instead of descending order (newest first). This happened because the implementation didn't set a default sort_order for date fields and the `searchContent` function in `packages/volto/src/actions/search/search.js` used incompatible parameter names with the Plone REST API.
The fix adds default descending sort order logic in `doSearch` when sort_on is present but sort_order is missing, and converts sort_order parameter to sort_reverse in the `searchContent `function for Plone REST API compatibility, ensuring search results show newest content first by default.


aryan7081:
@wesleybl  @davisagli , can you please help me pass the Acceptance test? I have tried exhaustively from my end, but I am unable to resolve the issue.
Thanks and Regards

packages/volto/src/actions/search/search.js
```diff
@@ -41,6 +41,11 @@ export function searchContent(url, options, subrequest = null) {
               // Adds the wildcard to the SearchableText param
               item[1] = `${item[1]}*`;
             }
+            // Convert sort_order to sort_reverse for Plone compatibility
+            if (item[0] === 'sort_order') {
+              item[0] = 'sort_reverse';
+              item[1] = item[1] === 'descending' ? '1' : '0';
+            }
```
wesleybl: The `searchContent` action is also used to sort `/contents` (folder icon in the left bar). With this change, sorting there stopped working.

wesleybl (changes requested):

packages/volto/src/actions/search/search.js
wesleybl: In fact, it is used in several places:

https://github.com/search?q=repo%3Aplone%2Fvolto%20searchContent&type=code

aryan7081:
@wesleybl Thanks. I have updated the `searchContent` logic to convert `sort_order`,  `sort_reverse` only when `sort_on `is `effective` (search page) prevent breaking contents and other search functionality.

packages/volto/src/actions/search/search.js
```diff
@@ -41,6 +41,17 @@ export function searchContent(url, options, subrequest = null) {
               // Adds the wildcard to the SearchableText param
               item[1] = `${item[1]}*`;
             }
+            // Convert sort_order to sort_reverse for Plone compatibility
+            // Only convert for search page date fields, not folder contents date fields
+            if (item[0] === 'sort_order' && item[1] !== undefined) {
```
wesleybl: Isn't there a way to fix this in `Search.jsx` itself and already pass the fixed options?

aryan7081: @wesleybl Moved the conversion to `Search.jsx` with a `convertSortOptions` helper function. Now it only affects the search page and doesn't impact folder contents or other `searchContent `uses. 
Thanks for the suggestion

packages/volto/src/components/theme/Search/Search.jsx
```diff
@@ -25,6 +25,19 @@ import Icon from '@plone/volto/components/theme/Icon/Icon';
 import paginationLeftSVG from '@plone/volto/icons/left-key.svg';
 import paginationRightSVG from '@plone/volto/icons/right-key.svg';
 
+/**
+ * Convert sort_order to sort_reverse for Plone compatibility (search page only)
+ * @param {Object} options Search options
+ * @returns {Object} Converted options
+ */
+const convertSortOptions = (options) => {
```
wesleybl: It's good practice not to change the function's parameters. You can create a copy:

```javascript
const converted = { ...options };
```

packages/volto/src/components/theme/Search/Search.jsx
```diff
@@ -112,6 +125,15 @@ class Search extends Component {
...
     this.setState({ currentPage: 1 });
     options['use_site_search_settings'] = 1;
+
+    // Ensure default sort order for date fields
+    if (options.sort_on && !options.sort_order) {
+      options.sort_order = 'descending';
+    }
+
+    // Convert sort options for Plone compatibility
+    convertSortOptions(options);
```
wesleybl: Now we have clear URL options and search options. You can create two variables, `urlOptions` and `searchOptions`. And create `searchOptions` based on `urlOptions`:

```javascript
const searchOptions = {
  use_site_search_settings: 1,
  b_size: this.defaultPageSize,
  ...urlOptions,
};
```

Here you can only pass `searchOptions`

```javascript
const apiOptions = convertSortOptions(searchOptions);
```

packages/volto/src/components/theme/Search/Search.jsx
```diff
@@ -123,6 +145,9 @@ class Search extends Component {
     let options = qs.parse(this.props.history.location.search);
     options['use_site_search_settings'] = 1;
 
+    // Convert sort options for Plone compatibility
+    convertSortOptions(options);
```
wesleybl: Create the two variables here as well.

packages/volto/src/components/theme/Search/Search.jsx
```diff
@@ -136,9 +161,24 @@ class Search extends Component {
...
     let options = qs.parse(this.props.history.location.search);
-    options.sort_on = event.target.name;
-    options.sort_order = sort_order || 'ascending';
-    if (event.target.name === 'relevance') {
+    const sortOn = event.currentTarget?.name || event.target?.name;
+
+    if (
+      sortOn === 'effective' &&
+      options.sort_on === 'effective' &&
+      (options.sort_order === 'descending' || options.sort_reverse === '1')
```
wesleybl: The URL never has `sort_reverse=1`. So this check can be removed.

packages/volto/src/components/theme/Search/Search.jsx
```diff
@@ -358,13 +398,15 @@ export default compose(
...
       key: 'search',
-      promise: ({ location, store: { dispatch } }) =>
-        dispatch(
-          searchContent('', {
-            ...qs.parse(location.search),
-            use_site_search_settings: 1,
-          }),
-        ),
+      promise: ({ location, store: { dispatch } }) => {
+        const options = qs.parse(location.search);
```
wesleybl: Create `urlOptions` and `searchOptions` here as well.

wesleybl (changes requested):

wesleybl:
@aryan7081 When I do:

http://localhost:8080/Plone/++api++/@search?sort_on=effective&sort_order=descending

sorting works. So why doesn't it work with the `searchContent` action?

wesleybl:
@aryan7081 The original request is:

http://localhost:3000/++api++/@search?b_size=25&sort_on=effective&sort_order=reverse&use_site_search_settings=1

This doesn't work. But when I remove `use_site_search_settings=1`, it works correctly. So we need to investigate `use_site_search_settings=1`.

aryan7081:
@wesleybl Sorry for the delay. Made `convertSortOptions` immutable by creating a copy, separated `URL/search/API options` across methods, and removed the unnecessary `sort_reverse` URL check. The `use_site_search_settings=1` issue is due to Plone's legacy search expecting `sort_reverse` instead of `sort_order` for the effective field.
Thanks for the review

wesleybl:
> The use_site_search_settings=1 issue is due to Plone's legacy search expecting sort_reverse instead of sort_order for the effective field.

@aryan7081 I'm wondering if the fix here couldn't be simply removing `use_site_search_settings=1`.

Maybe @davisagli can help.

aryan7081:
@wesleybl We could remove it, but `use_site_search_settings=1` enables site-specific search configurations. The conversion preserves that functionality while fixing the sorting.

wesleybl (approved):
LGTM

wesleybl:
@sneridagh I think this PR fix ​​#2458. Please check.

wesleybl:
@sneridagh I think we have this problem with Volto's sorting, because it uses `effective` and not `Date`, like in Classic Plone. Since the search control panel only sorts by `Date` and not by `effective`, the problem occurs in Volto.

Why was it changed to `effective` in Volto? I think it makes more sense than `Date`. But shouldn't it follow the Classic standard? I think if it were to continue with `effective`, perhaps it would be necessary to add sorting by `effective` to the search control panel (maybe some extra change would be needed, like in the plone.restapi)?

wesleybl (changes requested):
Perhaps this needs to be fixed in the backend. The `use_site_search_settings` parameter forces the search to use the sorting configuration defined in the search control panel. But if we have the option to select a sorting on the search screen, it shouldn't force the sorting from the control panel.

Related to: https://github.com/plone/plone.restapi/pull/1954

aryan7081:
@wesleybl Now we only send `use_site_search_settings=1` when the user hasn't explicitly selected a sort, so the control panel doesn't override the user's choice.
The backend PR (plone/plone.restapi#1954) fixes the handler `removing sort_order` when `use_site_search_settings=1` is present. Once that's merged, the sorting test should pass.
Regarding effective vs Date: this is a separate architectural question. The current fix ensures user selections are respected regardless of the field used.

aryan7081:
@wesleybl Sorry for tagging, I just need clarification on my last commit. Based on your comment about `this needs to be fixed in the backend.....` I updated the frontend so that `use_site_search_settings=1` is only sent when the user has not explicitly selected a sort option. I would like to confirm whether this matches your expectations, or if I misunderstood and should revert the last commit changes and keep the fix backend only.

wesleybl:
@aryan7081 My doubt is whether, after merging https://github.com/plone/plone.restapi/pull/1954, this PR will still be necessary.

aryan7081:
@wesleybl Thanks for the clarification. I pulled https://github.com/plone/plone.restapi/pull/1954 on my local and tested against the main branch of volto. Sorting works as expected without any changes in volto repo. 

https://github.com/user-attachments/assets/055db31d-0b3e-4d0e-9033-8da6b2da883c

So, this PR may not be necessary, but for an architecture and design perspective it would be good to wait for @davisagli's POV.

davisagli:
@aryan7081 I think it's best that we use the fix in https://github.com/plone/plone.restapi/pull/1954, because `use_site_search_settings=1` also has another effect (applying the filter by content type that is configured in the search control panel).

Thanks a lot for your effort here, even though we didn't end up merging the pull request, your testing and analysis helped move this along. Actually, it would be good if we still update the cypress test. Could you update this PR to include only that part, or make a new PR? (Merging it will probably have to wait until there is a new plone.restapi release and Volto is updated to use it.)

aryan7081:
> I think it's best that we use the fix in https://github.com/plone/plone.restapi/pull/1954, because use_site_search_settings=1 also has another effect (applying the filter by content type that is configured in the search control panel).

@davisagli That makes sense. Thanks for the clarification. I have updated the PR to include only the Cypress test for sorting functionality. This will pass once there is a new plone.restapi release and Volto is updated to use it

packages/volto/news/7489.bugfix
```diff
@@ -0,0 +1 @@
+Update Cypress test for search date sorting to verify newest-first ordering. @aryan7081
```
wesleybl: Please remove this file.

wesleybl (changes requested):

packages/volto/news/7489.bugfix
aryan7081: Done.

wesleybl (approved):

Issue #5619, opened 2024-01-11
a11y: add aria-required on required fields


giuliaghisini:

Aarav238:
@giuliaghisini  Can you please help me with that issue , I thought issue is regarding `aria-required` and `aria-described` in normal html tags , but this issue is something else , so please help me toh get there.

giuliaghisini:
> @giuliaghisini Can you please help me with that issue , I thought issue is regarding `aria-required` and `aria-described` in normal html tags , but this issue is something else , so please help me to get there.

When you are editing a content-type, for exaple, and you have some required fields in form, they doesn't have aria-required="true" attribute if it's a required field, and in case of some validation error, the error message isn't linked to the field, so the field has to have aria-described-by="id_of_the_wrapper_error_field_description" to enable screen reader to read error message for that field. 

This attributes are required for accessibility of the interface.

Aarav238:
Thanks @giuliaghisini  I will work on it.

praveen-rikhari:
I guess this is still not worked on? can i pick this one @giuliaghisini

stevepiercy:
@praveen-rikhari please read and follow [First-time contributors](https://6.docs.plone.org/contributing/first-time.html), especially [Things not to do](https://6.docs.plone.org/contributing/first-time.html#things-not-to-do), [Contributing to Plone](https://6.docs.plone.org/contributing/index.html), and [Contributing to Volto](https://6.docs.plone.org/contributing/volto.html).

giuliaghisini:
@Aarav238 have you done something about this? or is still an open issue?

Aarav238:
I will fix ths. Its been a long time , want to contribute to this project again.

Wagner3UB:
https://github.com/plone/volto/pull/7494

PR #7494, opened 2025-10-13
Required cms fields

Wagner3UB:
Fixing required fields inside the CMS UI / Editor – A11Y

- aria-required – works together with the visible validation (red dot).
- aria-invalid – works with the current field validation and the visible feedback.
- aria-label – sometimes needed to complement accessibility requirements.

### Components changes:

**1. FileWidget && RegistryImageWidget**
- `<label>` changed to a `<button> ` to a correct semantic.
- Added a label when required
- Added a correct identification for a field with required props

**IMPORTANT ABOUT THE FileWidget && RegistryImageWidget:**
It is not correct to use aria-required on a button. In addition, the entire select field has a poor A11Y structure, there isn’t a single element that correctly indicates that this field is a select/input field. To properly fix the issue, the entire component would need to be replaced with an accessible version, but this is beyond the scope of this PR.

**2. DatetimeWidget - react-dates and rc-time-picker components**
- These components do not natively accept ARIA attributes, meaning it is not possible to add the necessary properties via props to make them accessible when they are required.
- It was necessary to perform a scan when these components are rendered, looking for fields that can be modified by users and marking them as required.
- Unfortunately, the code ended up not being small due to the need to externally search for and apply these attributes to the elements.

**3. FormFieldWrapper**
- This component renders most of the text/text-like fields.
- When the component is of type Input and has the required prop set to true, the corresponding ARIA attributes are added to the element.

**4. UrlWidget**
- Recreated the linkInvalid variable to correctly identify whether a link is valid or not.
- Added the necessary ARIA attributes to the input.

**IMPORTANT ABOUT THE UrlWidget:**
The current Volto 19 validation for these links is incorrect. For example, it accepts a single letter as a valid link. If you enter a single "x", the validation considers it valid and allows the save.

### DIFF
--- a/packages/volto/news/7494.bugfix
+++ b/packages/volto/news/7494.bugfix
@@ -1,1 +1,2 @@
-Required fields - TODO @Wagner3UB
+- Improved A11Y in CMS UI by adding proper ARIA attributes (aria-required, aria-invalid, aria-label) across key form components and adjusting semantics. @Wagner3UB
+- Applied workarounds for 'DatetimeWidget' limitations and noted structural a11y issues in 'FileWidget' and incorrect URL validation. @Wagner3UB


Wagner3UB:
@sneridagh @tisto @stevepiercy Folks, I have a question about a button layout.

Issue location: 
https://volto.demo.plone.org/add?type=File

Code: https://github.com/plone/volto/blob/6a2f3649d409177be840b19429def9e843dda7a7/packages/volto/src/components/manage/Widgets/FileWidget.jsx#L204

Question: 
The "Choose a file" button isn't a real button, it's just a simple `<label>`, which is wrong. My question is if we can replace it with an actual `<button>` and use the default styling from our UI Button CSS. Would it be a problem to change it from a rounded blue button to a gray one?

Old:
Screenshot 2025-10-14 at 08 16 24

New:
Screenshot 2025-10-14 at 08 24 35

packages/volto/src/components/manage/Widgets/__snapshots__/RegistryImageWidget.test.jsx.snap
```diff
@@ -39,11 +39,12 @@ exports[`RegistryImageWidget > renders a file widget component with value 1`] =
               id="field-my-field-image"
               src="http://localhost:3000/@@site-logo/logo.cab945d8.svg"
             />
-            <label
-              class="label-file-widget-input"
+            <button
+              aria-required="false"
```
giuliaghisini, line 43: aria-required sul button?

Wagner3UB: @giuliaghisini The problem is a little more complicated than that. That select is built with a [Dropzone component](https://github.com/plone/volto/blob/481149f8b72c65e519744efd889c31f41fa41fc8/packages/volto/src/components/manage/Widgets/RegistryImageWidget.jsx#L121) which creates a group of divs containing a button styled as a label and an invisible input.

In my opinion it is completely wrong. It works as is but it is not semantic and feels made up.

I agree with you, it's not right to put an aria-required on a button, but it's the only way I found to add that requirement and make some sense without changing the whole Dropzone.

Here I made a screen video of the original and my version:
https://github.com/user-attachments/assets/5891c1e3-c77a-4299-b5ba-6adf58b06d18

packages/volto/news/7494.bugfix
```diff
@@ -0,0 +1 @@
+Required fields - TODO @Wagner3UB
```
stevepiercy: This looks to be incomplete.

Wagner3UB: @stevepiercy Let me know if what I did is okay or if there's a better way to do it. Thank you.

Wagner3UB: @stevepiercy I will close this PR and split it into three parts to make it easier for the community to review and work on. So my question above still stands, but for future reference.

Wagner3UB:
PR closed because it was too large and included changes to three different widgets, making it difficult for the community to review.

I will open three separate PRs, which will be referenced in the comments.

Issue #5118, opened 2023-08-23
a11y - expand/shrink toolbar - cms ui


Wagner3UB:
**Describe the bug**
The button called "toolbar-handler", locate inside the toolbar, have a fixed `aria-label="Shrink toolbar"` even having an expand/shrink behavior.

**To Reproduce**

1. Go, using a screen reader and shift + tab keys, to the button that controls the toolbar
2. Press enter, the toolbar will shrink
3. Another press, the toolbar will expand.

**Expected behavior**
The screen-reader needs to know when the toolbar is open/closed and must have a properly aria-label (or something related) for each behavior and read it each time.

**Screenshots**
<img width="683" alt="Schermata 2023-08-23 alle 16 36 09" src="https://github.com/plone/volto/assets/60133113/77a4f1b8-1c2e-4de3-ba2d-8b2e50e44fe8">

**Software (please complete the following information):**

- Volto Version 17.0.0-alpha.19
- Plone Version 6.0.4

**Additional context**
There is some question to answer about this button: 
This button needs to be accessible by a screen-reader? 
Why a user using a screen reader need to shrink/expand this toolbar?

nsundriyal62:
i want to work on this issue.
should i directly generate a pr or you gonna assign this to me?

stevepiercy:
@nsundriyal62 please read and follow [Contributing to Plone](https://6.docs.plone.org/contributing/index.html) and [First-time contributors](https://6.docs.plone.org/contributing/first-time.html).

Noobham:
should i just double the opacity ?

or change color on shrink.

Wagner3UB:
@Noobham If possible, only increase the contrast. 

But first you need to follow https://6.docs.plone.org/contributing/index.html and https://6.docs.plone.org/contributing/first-time.html, you must sign the Plone Contributor Agreement if you haven't done so already

Noobham:
@Wagner3UB I have raised PR #5431 .
Do i have to do anything more with this issue?

Thanks ...

PR #7500, opened 2025-10-15
Toolbar expand/collapse button a11y

Wagner3UB:
https://github.com/plone/volto/issues/5118

- Fixed the low contrast of the toolbar expand/collapse button and added a hover effect.
- Added an additional dynamic text for the expand button to properly indicate both open and closed states.

https://github.com/user-attachments/assets/a792576e-8d02-491b-b8b7-c51bf04340cf


Wagner3UB:
> The label changes look good, if you could provide a screenshot of the style changes it would be awesome. Thanks!

@pnicolli Done

packages/volto/locales/it/LC_MESSAGES/volto.po
```diff
@@ -1491,6 +1491,11 @@ msgstr "Url alternativi esistenti per questo elemento"
 msgid "Expand sidebar"
 msgstr "Espandi la sidebar"
 
+#. Default: "Expand toolbar"
+#: components/manage/Toolbar/Toolbar
+msgid "Expand toolbar"
+msgstr "Espandi la barra laterale"
```
pnicolli:
```suggestion
msgstr "Espandi la barra degli strumenti"
```

This feels more specific, but I have no strong opinion about it.

Wagner3UB:
@pnicolli Test passed

pnicolli (dismissed):
I think it's probably ok, I'm not fully convinced about having the visually-hidden class nested in there in the css code, it could as well be global but that could also cause breaking changes. I would love a couple more eyes from @plone/volto-team on this regard please.

JeffersonBledsoe:
@Wagner3UB We should probably set 'aria-controls' on the button too and have its ID set as the toolbar to make it clear that the thing that is expanded isn't the button itself.

wesleybl:
> it could as well be global but that could also cause breaking changes.

@pnicolli I don't understand why you think this would be a breaking change. There's no `visually-hidden` class in Volto. I think since it's global, it could be reused. But isn't there already a class that does this?

There is also this PR #6356

pnicolli:
@JeffersonBledsoe actually I think you are right, nice catch!
@wesleybl adding any global class that can be a breaking change for users. If users have a different implementation of the same class we could break it. I'm ok with adding this since we are in v19 alpha but I would document this in the upgrade guide. Good pointing out that another PR exists, let's resurrect that, I'll leave a comment/review there!

Wagner3UB:
@JeffersonBledsoe Thank you! I will manage to have that too.

@pnicolli @wesleybl The global insertion of this class was already discussed within the A11Y team, and we all agreed it was an excellent idea. I overlapped Jeff’s PR even though I was aware it already existed and was nearly complete, since I had been in contact with him and he mentioned he was having some difficulties deciding the best approach and location to define the variable in order to cover all possible cases.

Wagner3UB:
@JeffersonBledsoe @pnicolli  I added the aria-controls even though it may serve only a semantic purpose and not really as a functional instrument. The way I had implemented it before, using aria-live with translations to announce the states, mentioning the toolbar, and including aria-expanded, was already sufficient according to several sources I found. Still, I added it anyway to make sure accessibility testing tools would not flag its absence.

pnicolli:
> Still, I added it anyway to make sure accessibility testing tools would not flag its absence.

Makes perfect sense to me 👍

Let's maybe try to merge the other PR so we can then use the global class here.

packages/volto/src/components/manage/Toolbar/Toolbar.jsx
```diff
@@ -455,6 +459,7 @@ class Toolbar extends Component {
             </div>
           </div>
           <div
+            id="toolbar"
```
pnicolli: I just realized the `toolbar` id already exists in the page: https://github.com/plone/volto/blob/main/packages/volto/src/helpers/Html/Html.jsx#L195

```suggestion
            id="toolbar-body"
```

packages/volto/src/components/manage/Toolbar/Toolbar.jsx
```diff
@@ -616,15 +621,20 @@ class Toolbar extends Component {
...
-                  messages.shrinkToolbar,
-                )}
                 className={cx('toolbar-handler-button', {
                   [this.props.content?.review_state]:
                     this.props.content?.review_state,
                 })}
                 onClick={this.handleShrink}
-              />
+                aria-expanded={expanded}
+                aria-controls="toolbar"
```
pnicolli:
```suggestion
                aria-controls="toolbar-body"
```

pnicolli (changes requested):
Small but important detail I just noticed

pnicolli:
@sneridagh interesting fact, unrelated to the PR itself, I think I see an issue here: tests are green but it's a fake info. Tests should break, after the second to last commit, the one that I made to change the `id`, snapshots should break but then I merged main into this one and tests are green because it is not running all the tests, I think it is actually skipping those unit tests. In other words, merging this would break the tests on `main`. If what I understand is true, I think we should run all tests on CI.

Wagner3UB:
> LGTM.
> 
> @Wagner3UB One question, the one in the sidebar remains the same, right? I can see it in the video but I can't remember if they share CSS or not. Otherwise it might interfere with the sidebar content. It is pending that we find a better solution for that as well. sigh!

@sneridagh Thank you! Yes, the content didn’t change. And yes, unfortunately, we still need to find a better solution for the buttons...

I will merge this PR today, ok?. cc:/ @pnicolli

PR #7501, opened 2025-10-15
[Seven] React Server Components

pnicolli:
This is a work in progress PR to start working on an RSC implementation of Seven. It won't work for now, we need react-router to be able to reveal the rsc entry point in order to be able to inject i18n, or maybe we can do something of our own, it needs to be experimented with.
I think it's still valuable to track work on RSCs here to keep the needed changes in check and maybe start optimizing/updating the components that need the `use client` directive.

Useful link: https://reactrouter.com/how-to/react-server-components#server-component-routes


mtoepfl:
Hey, I hope you all enjoy the conference. Unfortunately I cannot participate this year. Thank you for your great work on Seven. It will be a big thing!

I don't know if this is the right place, but it's a technical thing like i18n or header-tags one should consider when designing the architecture:

--------

[Schema.org](https://schema.org/) in SEO AI is crucial for making content understandable to AI by using a standardized vocabulary for structured data. This helps AI-powered search engines understand content's context, categorize it accurately, and increase the likelihood of appearing in AI-generated summaries and enhanced search results. By adding structured data, websites can improve visibility, traffic, and user experience, becoming a foundational element for future-proof SEO.

--------

I do not know if it's not a problem with react-helmet or the new React feature to add additional tags to the head-tag.

We are implementing this just now in our custom frontend based in Plone/Zope. It's demanded to add this additional script-tags on the page- and one the block-level. This json-data can be generated by the block-component e.g. for an FAQ or any other listing of events, offers, facilities, jobs, amenities and so on. And you should be able to deduplicate the added entities on page-level by @id-attribute.

Perhaps a thing you should have in mind when you make your next design desisions for SSR. But I'm sure they will be as right as the others you did already for Seven! As I said, it will be a great thing in my opinion. 🎉🎉🎉

pnicolli:
Closing this old implementation. RR7 has added most of the things that we were missing in the meantime, but it's probably better to start this from scratch and leave this old implementation as documentation for when this will be reimplemented.

PR #7509, opened 2025-10-17
Drop support to Node 20

wesleybl:
Part of #7508

----
📚 Documentation preview 📚: https://volto--7509.org.readthedocs.build/

### DIFF
--- a/packages/volto/cypress/tests/core/volto-slate/02-block-slate.js
+++ b/packages/volto/cypress/tests/core/volto-slate/02-block-slate.js
@@ -3,8 +3,9 @@ import { slateBeforeEach } from '../../../support/helpers';
 describe('Block Tests', () => {
   beforeEach(slateBeforeEach);
 
-  it('should create 4 slate blocks, first 3 with mouse, the last with an Enter in the third block', () => {
-    cy.getSlate().focus().click().type('Hello Slate World').type('{enter}');
+  it('should save typed content in the Slate editor', () => {
+    cy.getSlateEditorAndType('Hello Slate World');
+    cy.getSlateEditorAndType('{enter}');
 
     // Save
     cy.toolbarSave();
--- a/packages/volto/news/7507.internal
+++ b/packages/volto/news/7507.internal
@@ -1,0 +1,1 @@
+Fix error and rename test to "should save typed content in the Slate editor". @wesleybl


stevepiercy:
Volto 19 is still in alpha, so we can drop Node.js 20.

Is there any reason not to add support for Node.js 24, technical or otherwise?

packages/volto/news/7509.breaking
```diff
@@ -0,0 +1 @@
+Drop support to Node 20. @wesleybl
```
stevepiercy:
```suggestion
Drop support of Node.js 20. @wesleybl
```

docs/source/upgrade-guide/index.md
```diff
@@ -32,6 +32,21 @@ It is usually better and quicker to move your items into new locations and copy
 
 ## Upgrading to Volto 19.x.x
 
+(nodejs-20-removed-label)=
+
+### Node.js 20 is no longer supported
```
```suggestion
### Removed support of Node.js 20
```

docs/source/upgrade-guide/index.md
```diff
@@ -32,6 +32,21 @@ It is usually better and quicker to move your items into new locations and copy
...
+```{versionremoved} Volto 19
```
stevepiercy: This will be the next alpha version.

````suggestion
```{versionremoved} Volto 19.0.0-alpha.7
````

File: docs/source/upgrade-guide/index.md [hunk 4]
```diff
@@ -32,6 +32,21 @@ It is usually better and quicker to move your items into new locations and copy
...
+```
+
+Volto 19 removes official support and CI testing for Node.js 20. While the codebase may still work on Node.js 20 in some cases, it is no longer guaranteed or tested.
```
```suggestion
Support and CI testing for Node.js 20 was removed in Volto 19.0.0-alpha.7.
While the code base may still work on Node.js 20 in some cases, it is no longer guaranteed or tested.
```

docs/source/upgrade-guide/index.md
```diff
@@ -32,6 +32,21 @@ It is usually better and quicker to move your items into new locations and copy
...
+
+Recommended action for projects:
```
```suggestion
You should take the following actions for your Volto 19 projects.
```

docs/source/upgrade-guide/index.md
```diff
@@ -32,6 +32,21 @@ It is usually better and quicker to move your items into new locations and copy
...
+
+- Upgrade your development and CI environments to a supported Node.js LTS version (for example, Node.js 22). 
+- Update any Docker images, build agents, and project documentation that reference Node.js 20.
```
```suggestion
-   Upgrade your development and CI environments to a supported Node.js LTS version, such as Node.js 22.
-   Update any Docker images, build agents, and project documentation that reference Node.js 20.
```

docs/source/upgrade-guide/index.md
```diff
@@ -32,6 +32,21 @@ It is usually better and quicker to move your items into new locations and copy
...
+
+If you cannot upgrade immediately, you may continue to run Volto 19 on Node.js 20 at your own risk, but be aware that issues specific to Node.js 20 will not be fixed in the Volto core CI or releases.
```
```suggestion
If you can't upgrade immediately, you may continue to run Volto 19 on Node.js 20 at your own risk, but be aware that issues specific to Node.js 20 will not be fixed in the Volto core CI or releases.
```

packages/volto/package.json
```diff
@@ -175,7 +175,7 @@
     "not dead"
   ],
   "engines": {
-    "node": "^20 || ^22"
+    "node": "^22"
```
davisagli: I discussed this with @pnicolli and @sneridagh at the conference sprint. We will stop testing on node 20 and remove official support for it, but we won't remove it from the engines because @pnicolli has a deployment environment that needs node 20. (This is similar to the status of Node 14 in the Volto 16 branch.)
```suggestion
    "node": "^20 || ^22"
```

davisagli:
> Is there any reason not to add support for Node.js 24, technical or otherwise?

We want to add support for it. Mostly that means trying it in a PR and seeing if the tests pass. Let's do that in a separate PR so it doesn't delay this one.

davisagli (approved):

packages/volto/package.json
wesleybl: @davisagli I think this will cause confusion. I don't see a problem with removing Node 20. By default, pnpm doesn't block installation if the Node version isn't compatible with the engines.

davisagli: @wesleybl I tested it and you're right: pnpm only logs a warning and doesn't abort.

```
core/packages/volto                      |  WARN  Unsupported engine: wanted: {"node":"^22"} (current: {"node":"v20.10.0","pnpm":"9.1.1"})
```

I agree with removing 20 then. @pnicolli?

pnicolli: Didn't know that, most of our sites are a little older and use yarn so I hadn't tested that. I'm ok with removing it from here then, I think it's safe to assume that volto 19 sites will need to use pnpm for now.

sneridagh: @pnicolli yeah, the idea was to drop the yarn build in 19.

davisagli: Indeed, we documented that a long time ago: https://github.com/plone/volto/blob/main/docs/source/upgrade-guide/index.md?plain=1#L197

davisagli: Done in https://github.com/plone/volto/pull/7517/commits/f90a995edc5741cc848219be33ae5e621ddb7018

PR #7513, opened 2025-10-18
Image content type view

### LINKED ISSUE #6706: [Seven Content Area Views] - Image
> [!IMPORTANT]
> If you are not a member of the Volto Team or Developers Team in the Plone GitHub organization, then do not work on or comment on this issue.

The Image content type view.

**Link to the parent PLIP**

https://github.com/plone/volto/issues/6638

ebrehault:
Notes:

- Components styles are not imported in the basicui styles at the moment, consequently `<Container>` is not remdered properly.

- @pnicolli commented in the PR that we must use `<Link>` instead of `<a>` to render the download link.
But for some reason,  `<Link>` triggers the streaming endpoint (the `.data` url), so the image does not render, we get a bunch of random text.
So I have decided to open the image in a new tab, it fixes the problem and on the functional point of view, it makes sense to open the downloaded image in another tab.

-----

-----

Closes #6706

-----

----
📚 Documentation preview 📚: https://plone-registry--7512.org.readthedocs.build/

----
📚 Documentation preview 📚: https://volto--7512.org.readthedocs.build/

### DIFF
--- a/packages/layout/views/ImageView.tsx
+++ b/packages/layout/views/ImageView.tsx
@@ -4,7 +4,6 @@ import type { RootLoader } from 'seven/app/root';
 import { useTranslation } from 'react-i18next';
 import prettybytes from 'pretty-bytes';
 import Image from '../components/Image/Image';
-import type { ContainedItem } from '@plone/types';
 
 export default function ImageView() {
   const { t } = useTranslation();
@@ -15,17 +14,6 @@ export default function ImageView() {
   }
 
   const { content } = rootData;
-  const imageItem: ContainedItem | undefined = content.image?.download
-    ? {
-        ...content,
-        image_field: 'image',
-        image_scales: {
-          image: {
-            ...content.image,
-          },
-        },
-      }
-    : undefined;
 
   return (
     <Container width="default">
@@ -34,7 +22,7 @@ export default function ImageView() {
       {content.image?.download ? (
         <Link href={content.image.download} download={content.image.filename}>
           <figure>
-            <Image src={content.image.download} item={imageItem} alt="" />
+            <Image item={content} alt="" />
             <figcaption>
               {t('layout.contenttypes.image.size')}
               {prettybytes(content.image.size)}


package.json
```diff
@@ -56,5 +56,9 @@
       "lightningcss-cli",
       "unrs-resolver"
     ]
+  },
+  "dependencies": {
+    "layout": "^2.2.0",
+    "pretty-bytes": "^7.1.0"
```
pnicolli: `layout` looks like it's there by mistake
`pretty-bytes` should be installed in the packages it's using it and not globally, that means it should be installed via `pnpm --filter @plone/layout add pretty-bytes` and removed from here

packages/layout/views/ImageView.tsx
```diff
@@ -0,0 +1,39 @@
...
+  }
+
+  const { content } = rootData;
+
+  return (
+    <Container>
+      <h1 className="documentFirstHeading">{content.title}</h1>
+      <p className="documentDescription">{content.description}</p>
+      {content.image?.download ? (
+        <a href={flattenToAppURL(content.image.download)}>
```
pnicolli: `flattenToAppURL` is actually not needed anymore (I should have mentioned it during the introduction to Seven, my bad). I recommend using the Link component from `@plone/components`, that will take care of client-side navigation which improves perf a little bit (not as much as in the past, but still).


```suggestion
        <Link href={content.image.download}>
```

packages/layout/views/ImageView.tsx
```diff
@@ -0,0 +1,39 @@
...
+          <figure>
+            <img src={content.image.download} alt="" />
```
pnicolli: Use the Image component here from `import Image from '@plone/layout/components/Image/Image`
You can use like so:
```
<Image item={content.image} alt="" />
```

davisagli: @pnicolli Wesley added linting to check for this on the main branch, we should copy it to `seven`. https://github.com/plone/volto/pull/7363/files#diff-9608aa597cd41c814187b8ee7ec35faf0ef27dbc7b390ce71a1ecdb393d37102R40

ebrehault:
@pnicolli the PR is fixed

packages/layout/views/ImageView.tsx
```diff
@@ -0,0 +1,39 @@
...
+  const { content } = rootData;
+
+  return (
+    <Container width="default">
+      <h1 className="documentFirstHeading">{content.title}</h1>
+      <p className="documentDescription">{content.description}</p>
+      {content.image?.download ? (
+        <Link href={content.image.download} download={content.image.filename}>
+          <figure>
+            <Image src={content.image.download} alt="" />
```
sneridagh: @ebrehault instead of `src` we have to use `item` prop from the `Image` component, so we get all the sizes in the `srcset` property. Can you change that? 

https://plone-layout.readthedocs.io/?path=/story/image--plone-image

ebrehault: @sneridagh 
Done.
I pass the `item` attribute now. For the record, `Image` still needs an `src` attr anyway, if not, it renders nothing (`item` is only use for scales, not sure if that's a bug…).
Also, I had to make a small fix in `@plone/types`, I hope that's fine.

pnicolli: It could be a bug, yes, it should accept either an src or an item, preferably the item, but not both. I'll test it on the `seven` branch.

pnicolli: Checked and yes it could be a bug, will work on it today.

pnicolli: I tested the Image component, I have noticed in a couple situation a mismatch in the HMR behavior for server rendering and client rendering, meaning after saving in the editor the server was rendering the old code and the client the new one until I restarted. This is definitely not on us but also not very good, let's keep an eye on framework patched @sneridagh 

Once I realized this, I could actually try to just use the Image component how it should be used and since I did and it worked I just pushed the change. I think this is good to merge now.

PR #7526, opened 2025-10-20
[Seven] fixed layout package styles

pnicolli:
Import `@plone/components` styles where the components are used (that is `@plone/layout`). If integrators remove `@plone/layout` then they should provide these same styles or any other that are needed, but this is a super advanced use case that I'm not sure will really happen.

This means these styles will be always loaded but they are also super thin at this time and afaict scoped in a good way.


sneridagh:
@pnicolli probably a glitch, now it's green. Merge?

PR #7531, opened 2025-10-21
Refactor `LanguageSelector` into TypeScript. Modernize `tsconfig.json` settings in core.

sneridagh:


packages/volto/news/7531.2.internal
davisagli, line 1: Refactoring is a bugfix or internal, not a feature. It doesn't make any new functionality available to users.

sneridagh: Done!

PR #7532, opened 2025-10-21
Re-enable scripts pnpm build. Remove `why`.

sneridagh:
As we talked yesteday, re-enabling these ones.


packages/volto/package-why.json
```diff
@@ -1,34 +0,0 @@
...
-    "stylelint:patches": "Description for `npm run stylelint:patches` command",
-    "patches": "Description for `npm run patches` command",
-    "dry-release": "Description for `npm run dry-release` command",
-    "release": "Description for `npm run release` command",
-    "release-major-alpha": "Description for `npm run release-major-alpha` command",
-    "release-alpha": "Description for `npm run release-alpha` command",
-    "storybook": "Description for `npm run storybook` command",
-    "build-storybook": "Description for `npm run build-storybook` command"
-  }
-}
```
wesleybl, line 34: Wouldn't it be better to update the scripts description instead of removing the file?

sneridagh: We never really used it, or updated the comments, and it's not an official package, meaning is something we put on our own. We encouraging in the documentation properly to use `Makefile` convenience commands now, and it's well documented. So I think having them is redundant (and also pushes for a build script).

wesleybl: Ok.

wesleybl:
I think we'll now have to expire the GitHub action cache if package.json changes. See:

https://github.com/plone/cookieplone-templates/pull/292#pullrequestreview-3357980414

But this can be done in another PR.

PR #7543, opened 2025-10-23
Update CreateTranslation Component

rboixaderg:
This pull request refactors how the translation object is managed in the `CreateTranslation` component to improve state handling and consistency. The main change is switching from local component state to using Redux state via `useSelector`.

@davisagli We need to get the translationObject from the store instead of directly from the response, because in the content reducer we adapt this object, for example by flattening some attributes. In Guillotina, we have some attributes inside an object instead of at the top level of the object.

### DIFF
--- a/packages/volto/news/7543.bugfix
+++ b/packages/volto/news/7543.bugfix
@@ -1,0 +1,1 @@
+Get translationObject from the store instead of directly from the response in CreateTranslation component. @rboixaderg
--- a/packages/volto/news/7543.feature
+++ b/packages/volto/news/7543.feature
@@ -1,1 +1,0 @@
-Get translationObject from the store instead of directly from the response in CreateTranslation component. @rboixaderg


packages/volto/src/components/manage/Multilingual/CreateTranslation.jsx
```diff
@@ -46,15 +45,16 @@ const CreateTranslation = (props) => {
 
   return (
     translationLocation &&
-    translationObject && (
+    translationObject.data &&
+    translationObject.loaded > 0 && (
```
pnicolli, line 49: `loaded` is a boolean value usually and I believe in this case as well. Even if `true > 0` is true in JavaScript, I don't think it's ideal to have it checked this way

pnicolli:
Left a comment even if it was merged already. This looks like a good improvement for sure, but I also noticed there are other calls that don't use redux to read the fetched data in that same component, is there a reason why those were not fixed as well?

davisagli:
@pnicolli We specifically needed the getContent action to be processed by the reducer in order to do some transformations of the static_behaviors from guillotina. It doesn't look like that's necessary for the other actions.

PR #7545, opened 2025-10-24
use_prerelease_versions=Yes in Cookieplone job

wesleybl:
For now, there are no differences between the prerelease and the current version. But we will have:

https://github.com/plone/cookieplone-templates/pull/292

### DIFF
--- a/packages/volto/news/7545.internal
+++ b/packages/volto/news/7545.internal
@@ -1,0 +1,1 @@
+`use_prerelease_versions=Yes` in Cookieplone job. @wesleybl


wesleybl:
> Can we use a matrix and test both stable and prerelease?

@davisagli This job tests the current branch (Volto 19). So it doesn't make sense here. The 18.x.x branch will test the current version (it doesn't have use_prerelease_versions=Yes).

wesleybl:
> Maybe we could calculate use_prerelease_versions based on whether there is an alpha

@davisagli I had thought the main branch would always be the alpha version. But when the main branch becomes version 19, we would have to remember to remove that. So what you said makes sense.

davisagli:
@wesleybl On the other hand: cookieplone with use_prerelease_versions=Yes will pick the latest stable release if there are no newer alpha releases. So it should be okay to always use use_prerelease_versions=Yes here on main.

wesleybl:
@davisagli The problem would arise when the main branch is Volto 20 (alpha), and the 19.x.x branch is created. We'd have to remember to remove use_prerelease_versions=Yes on the 19.x.x branch.

But then again, we'd still have the 19.x.x and 18.x.x branches. So it's best to use the Volto version, not `use_prerelease_versions`.

wesleybl:
Would there be a problem testing Volto 18 with a Volto 19 template? Will cookieplone work with 3 versions (20, 19, 18)? I'm already thinking about it ahead of time, lol.

sneridagh:
@wesleybl @davisagli I just saw this one, after I created https://github.com/plone/volto/pull/7548 😩 

I am reading and using the current Volto version in the `package.json`, which I think is good enough. We could even port it to 18, I guess. What do you think?

davisagli:
@sneridagh Let's go with #7548. You did there what we were discussing here anyway. :)

PR #7547, opened 2025-10-25
Fix reset password error handling

suhaibfida:
image

### DIFF
--- a/packages/volto/news/7533.bugfix
+++ b/packages/volto/news/7533.bugfix
@@ -1,5 +1,1 @@
-[bugfix]
-title = "Fix reset password error handling"
-module = "volto"
-author = "suhaibfida"
-issue = 7533
+Fixed the reset password page to display "not found" error message when an invalid user attempts to reset their password. @suhaibfida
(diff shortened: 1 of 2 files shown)


packages/volto/news/7533.bugfix
aryan7081, line 1: @suhaibfida We generally don't write a change log entry in this format. Also use this format for the author's name `@github_username`. See: https://6.docs.plone.org/contributing/index.html#contributing-change-log-label

aryan7081:
@suhaibfida Thanks for your contribution! Could you please add Cypress acceptance tests for this change? Volto requires acceptance tests to ensure quality. See: https://6.docs.plone.org/volto/contributing/acceptance-tests.html

suhaibfida:
I found the solution for the issue.
The problem was with the error handling logic.

✅ Updated logic:

const errmsg = this.props.error?.response?.body?.error || null;
error = {this.state.error || errmsg || this.props.error};

This safely handles nested error responses using optional chaining.

I wasn’t able to complete the test case for it, so I’m closing this PR for now.
This is the solution I reached — feel free to build upon it in future updates.

suhaibfida:
@aryan7081, hey, can you help me write tests for this? I tried various things, but couldn’t figure out how to write tests with the forgot password token, or how to get a valid token here. New to testing. I really appreciate any help you can provide.

PR #7548, opened 2025-10-25
Force cookieplone CI job to use 19

sneridagh:


.github/workflows/cookieplone.yml
```diff
@@ -27,7 +34,7 @@ jobs:
       - name: Generate Cookieplone-based frontend addon
         run: |
           pipx install cookieplone
-          pipx run --no-cache cookieplone frontend_addon --no-input
+          pipx run --no-cache cookieplone frontend_addon --no-input __version_plone_volto=${{ steps.vars.outputs.volto-version }}
```
davisagli: @sneridagh Shouldn't we set `volto_version` instead of `__version_plone_volto`? Because the node version is derived from `volto_version`: https://github.com/plone/cookieplone-templates/blob/main/templates/add-ons/frontend/cookiecutter.json#L22

sneridagh: @davisagli I pushed the suggestion, however, didn't we have to change it in plone-frontend because that was problematic? (can't remember why).

sneridagh: FTR, it seems it worked well in CI.

.github/workflows/cookieplone.yml
```diff
@@ -18,6 +18,13 @@ jobs:
       project-directory: volto-add-on
     steps:
       - uses: actions/checkout@v4
+
+      - name: Compute several vars needed for the CI
+        id: vars
+        run: |
+          python3 -c 'import json; data = json.load(open("./packages/volto/package.json")); print("volto-version=" + (data.get("version") or "19.0.0-alpha.0"))' >> $GITHUB_OUTPUT
```
wesleybl: I think it's better to give an error if the version is not found, as it can mask errors and the fallback may become outdated.

```suggestion
          python3 -c 'import json; data = json.load(open("./packages/volto/package.json")); print("volto-version=" + data["version"])' >> $GITHUB_OUTPUT
```

wesleybl (changes requested):

.github/workflows/cookieplone.yml
wesleybl: Another option would be to move this step to after Node configuration and do it with JavaScript. It's simpler, and we can accommodate developers who don't understand Python:

```bash
node -e "const v = require('./packages/volto/package.json').version || process.exit(1); console.log('volto-version=' + v)" >> $GITHUB_OUTPUT
```

sneridagh: @wesleybl your suggestion makes sense, I commited it.

wesleybl:
@davisagli @sneridagh I think it's now possible to remove the Jest settings from Volto 19.

Maybe @Abhishek-17h  can help?

sneridagh:
@wesleybl It is on my list, but if anyone wants to help...

Abhishek-17h:
> @davisagli @sneridagh I think it's now possible to remove the Jest settings from Volto 19.
> 
> Maybe @Abhishek-17h can help?

@wesleybl @sneridagh Sure, I’d be happy to help!

Abhishek-17h:
@wesleybl @sneridagh When I have to do this work and against which branch? Should we remove all traces of jest from volto which also includes support for add-ons testing?

PR #7550, opened 2025-10-26
Header slots

### LINKED ISSUE #6673: [CMSUI Slots] - Header
Sub-slots:

- Logo
- Sections
- Breadcrumbs

arybakov05:
Closes plone/aurora#59


arybakov05:
@sneridagh were the _anontools_, _search widget_ and _site actions_ slots supposed to be sub-slots of the header tools (with the same slot id) or is it fine to have them as individual slots? Also, I'm not sure if I should've used CSS modules or not, I just followed the way it was done in the _Header_ component.

packages/layout/slots/HeaderTools/AnonymousTools.tsx
```diff
@@ -0,0 +1,30 @@
...
+import { Link } from '@plone/components';
+import config from '@plone/registry';
+import styles from './Tools.module.css';
+import { flattenToAppURL } from '@plone/helpers';
+
+const AnonymousTools = (props: SlotComponentProps) => {
+  const { content } = props;
+  const { t } = useTranslation();
+
+  const returnUrl = flattenToAppURL(content['@id']);
```
pnicolli: I think we don't need the `flattenToAppURL` call here, all urls from `content` should be already flattened (this is something new we introduced in Seven)

packages/layout/config/slots.ts
```diff
@@ -36,16 +39,34 @@ export default function install(config: ConfigType) {
...
-    slot: 'headertools',
+    slot: 'headerTools',
     component: HeaderTools,
   });
 
+  config.registerSlotComponent({
+    name: 'Tools',
+    slot: 'siteActions',
+    component: SiteActions,
+  });
```
pnicolli, line 53: I'm not sure about this, I don't think I've seen this slot in the figma UI and we didn't have it in Volto iirc. @sneridagh what do you think, should we keep it? In case we keep it, it should be named UserAction or something similar and it should get the actual user actions from `content['@components'].user`. I feel like the implementation of this menu is something we should leave to integrators for their specific project use case.

packages/layout/slots/Header/Header.module.css
```diff
@@ -9,8 +9,6 @@
   } */
 
   .header-tools {
-    display: flex;
-    flex: 1 1 0;
-    flex-direction: row-reverse;
+    align-items: center;
```
pnicolli, line 46: This does not make sense without `display: flex`, maybe it's leftover?

packages/layout/slots/HeaderTools/SearchWidget.tsx
```diff
@@ -0,0 +1,53 @@
...
+
+    setSearchQuery('');
+    e.preventDefault();
+  };
+
+  return (
+    <Form
+      className={clsx(styles['searchwidget'], 'searchwidget')}
+      onSubmit={onSubmit}
+    >
```
pnicolli, line 21: Try using standard html form behavior, except you use the react-router Form component. You will have the same result, but with cleaner code and it will work even for users that have js disabled.

This means doing something like

```
<Form
  className={clsx(styles['searchwidget'], 'searchwidget')}
  action="/search"
  method="GET"
>
  {pathname?.length > 0 && (
    <input type="hidden" name="path" value={encodeURIComponent(pathname)} />
  )}
```

and removing the onChange from the Input.

pnicolli (changes requested):

packages/layout/config/slots.ts
arybakov05: @sneridagh listed the site actions as a sub slot in issue #6673.  I do think it's good to have a slot for actions for authenticated (in this case the site actions) and unauthenticated users (the anonymous tools).

pnicolli: Ah, yeah, I forgot he mentioned it there. Let's clarify with @sneridagh what he means, because historically "site actions" were a completely different thing and not the user actions.

packages/layout/slots/Header/Header.module.css
arybakov05: Inside `@plone/layout/styles/header.css` in line 24, the third child of the header, which turns out to be the header tools, receives `display: inline-flex`. The `align-items: center` therefore vertically aligns the items inside these header tools. Without it, the "Log Out" link would be sitting at the top of the container. Should I move the styles over to the `Header.module.css`?

sneridagh:
> @sneridagh were the _anontools_, _search widget_ and _site actions_ slots supposed to be sub-slots of the header tools (with the same slot id) or is it fine to have them as individual slots? Also, I'm not sure if I should've used CSS modules or not, I just followed the way it was done in the _Header_ component.

The idea always was to model what we had back in the day with viewlets and viewlets managers, and the idea that you can nest them. The do not have to share the name or id, but they are meant to be semantic in the naming. So we have zones in the header (and in other slots) that we named back in the day, but we can name them now. Take a look at the classic viewlet managers to get the point:

https://classic.demo.plone.org/en/manage-viewlets

image

As long as they cover the "areas" left (logo), center (sections, menu) then right (actions, search widget, anontools language switcher?), is enough, because all are extensible and layered (stackable) by default. So we are good.

packages/layout/config/slots.ts
sneridagh: @arybakov05 yes, I listed it in there, but I'd re-word it since the feedback that we had the last year(s). The site actions refer to the "old" site actions, as in unique site actions. We don't have (yet) a way to modify them TTW in Seven (we might in the future), nor in Volto today. So we fallbacked to have it as behaviors to be able to have actions configurable in a subsite fashion. 

Summary: I would remove it for now. We can get them back in the near future, by adding another component to  the slot at any time.

packages/layout/config/slots.ts
```diff
@@ -36,16 +39,34 @@ export default function install(config: ConfigType) {
...
+  // Header Tools
   config.registerSlotComponent({
     name: 'Tools',
-    slot: 'headertools',
+    slot: 'headerTools',
     component: HeaderTools,
   });
 
+  config.registerSlotComponent({
+    name: 'Tools',
```
sneridagh: Not sure about calling hem all Tools. The name of the slot component should be different... at least how I see it.

packages/layout/slots/Header/Header.tsx
```diff
@@ -26,12 +26,7 @@ const Header = (props: HeaderProps) => {
       <SlotRenderer name="navigation" content={content} location={location} />
       <div className={clsx(styles['header-tools'], 'header-tools')}>
         <SlotRenderer
-          name="headertools"
-          content={content}
-          location={location}
-        />
-        <SlotRenderer
-          name="language-switcher"
+          name="headerTools"
```
sneridagh, line 29: I'm ok on doing this, and move it inside the tools.

packages/layout/slots/Header/Header.module.css
pnicolli: Yeah I just had a look and realized that we have header styles both in `@plone/layout/styles/header.css` and in `@plone/layout/slots/Header/Header.module.css`. I would move all the styles related to the header slot in `Header.module.css`, and leave the `#header` styles in `header.css` since that id is hardcoded.

sneridagh: yes, let's unify things.

sneridagh:
> [!important]
> Hello! Thanks for contributing to Plone Aurora! FYI, since June 5th, Plone Aurora lives in its own repository, thus, the `seven` branch does not accept any merges. If you want to merge your PR, please follow instructions: https://github.com/plone/volto/issues/8318
> Sorry for the inconvenience!

arybakov05:
PR moved to Aurora repository.

PR #7551, opened 2025-10-27
Toolbar button behavior a11y

Wagner3UB:
Important: Must read to understand why this PR is different from the V19 one https://github.com/plone/volto/issues/5118

v19 PR: https://github.com/plone/volto/pull/7500/

- Fixed the low contrast of the toolbar expand/collapse button and added a hover effect.
- Added an additional dynamic text for the expand button to properly indicate both open and closed states.

update 19/11/2025

https://github.com/user-attachments/assets/05fd6648-297a-42b5-8fbf-09a89b19dc33

----
📚 Documentation preview 📚: https://volto--7551.org.readthedocs.build/

### DIFF
--- a/packages/volto/src/components/manage/Toolbar/__snapshots__/Toolbar.test.jsx.snap
+++ b/packages/volto/src/components/manage/Toolbar/__snapshots__/Toolbar.test.jsx.snap
@@ -17,7 +17,7 @@ exports[`Toolbar > renders the Toolbar component 1`] = `
   </div>,
   <div
     className="toolbar expanded"
-    id="toolbar"
+    id="toolbar-body"
   >
     <div
       className="toolbar-body"
@@ -183,7 +183,7 @@ exports[`Toolbar > renders the Toolbar component 1`] = `
       className="toolbar-handler"
     >
       <button
-        aria-controls="toolbar"
+        aria-controls="toolbar-body"
         aria-expanded={true}
         className="toolbar-handler-button"
         onClick={[Function]}
@@ -220,7 +220,7 @@ exports[`Toolbar > renders the Toolbar component with lock 1`] = `
   </div>,
   <div
     className="toolbar expanded"
-    id="toolbar"
+    id="toolbar-body"
   >
     <div
       className="toolbar-body"
@@ -386,7 +386,7 @@ exports[`Toolbar > renders the Toolbar component with lock 1`] = `
       className="toolbar-handler"
     >
       <button
-        aria-controls="toolbar"
+        aria-controls="toolbar-body"
         aria-expanded={true}
         className="toolbar-handler-button"
         onClick={[Function]}


packages/volto/src/components/manage/Toolbar/Toolbar.jsx
```diff
@@ -625,7 +630,15 @@ class Toolbar extends Component {
...
                 onClick={this.handleShrink}
-              />
+                aria-expanded={expanded}
+                aria-controls="toolbar"
+              >
+                <span aria-live="assertive" className="visually-hidden">
+                  {expanded
+                    ? this.props.intl.formatMessage(messages.shrinkToolbar)
+                    : this.props.intl.formatMessage(messages.expandToolbar)}
+                </span>
```
wesleybl, line 637: Shouldn't the button's `aria-label` be the one that needs to change? It became different proposed in the issue.

packages/volto/theme/themes/pastanaga/extras/toolbar.less
```diff
@@ -252,13 +252,31 @@ body:not(.has-sidebar):not(.has-sidebar-collapsed) {
...
+        }
+
         &::before {
           position: relative;
           display: block;
           width: 100%;
           height: 4px;
           background-color: red;
           content: '';
+          opacity: 0.65;
```
wesleybl, line 270: This is not being applied. The color is, for example, `.private:before`.

packages/volto/theme/themes/pastanaga/extras/toolbar.less
```diff
@@ -252,13 +252,31 @@ body:not(.has-sidebar):not(.has-sidebar-collapsed) {
         cursor: pointer;
         transition: opacity 0.3s;
 
+        .visually-hidden {
```
wesleybl, line 251: I think it's better to keep this class in a separate file and import it into the components. That way we can reuse it. Otherwise, we'll have to repeat the class everywhere we use it.

pnicolli: If I understand your suggestion correctly, you end up with a file with a global class and as soon as you import it you have it globally everywhere, meaning we would get back to the point of having a breaking change because we add a global class.

wesleybl: @pnicolli You're right. Do you know if Volto uses CSS modules?

wesleybl: @pnicolli I searched the code and didn't see any indication that the core uses CSS modules.

Perhaps it's worth creating a component that receives another component and applies the style inline? The CSS isn't that large. Maybe we'll need this CSS elsewhere.

davisagli: I think this is fine as is. It's a temporary solution for the 18.x branch until we can use the global class in 19.x. So I would avoid over-engineering it.

wesleybl: @davisagli ok.

packages/volto/theme/themes/pastanaga/extras/toolbar.less
Wagner3UB: Thank you @wesleybl . The opacity issue was caused by a style defined before this one. I'll tag the right place. I also found the same mistake in the main branch and fixed it here as well.

Here: https://github.com/plone/volto/blob/dbef1210597f693e0ae25671e6078e6e2f824462/packages/volto/theme/themes/pastanaga/extras/toolbar.less#L132-L134

packages/volto/src/components/manage/Toolbar/Toolbar.jsx
Wagner3UB: We chose this approach to minimize the risk of misdirection. This way, we can be sure that the 'expand' and 'shrink' labels are never associated with the button itself, and that the label is always read by the screen reader through aria-live.

But it can be done, and I can confirm that it also works when placing the expanded check inside the aria-label.

Wagner3UB:
@pnicolli @wesleybl I updated the screen recording in the description with the latest changes.

packages/volto/src/components/manage/Toolbar/Toolbar.jsx
```diff
@@ -455,6 +459,7 @@ class Toolbar extends Component {
             </div>
           </div>
           <div
+            id="toolbar"
```
pnicolli:
```suggestion
            id="toolbar-body"
```

Just realized that `toolbar` already exists https://github.com/plone/volto/blob/main/packages/volto/src/helpers/Html/Html.jsx#L195

packages/volto/src/components/manage/Toolbar/Toolbar.jsx
```diff
@@ -617,15 +622,20 @@ class Toolbar extends Component {
...
-                  messages.shrinkToolbar,
-                )}
                 className={cx('toolbar-handler-button', {
                   [this.props.content?.review_state]:
                     this.props.content?.review_state,
                 })}
                 onClick={this.handleShrink}
-              />
+                aria-expanded={expanded}
+                aria-controls="toolbar"
```
pnicolli:
```suggestion
                aria-controls="toolbar-body"
```

PR #7557, opened 2025-10-28
fix: html slate widget state management

### LINKED ISSUE #7556: Slate - Cursor moving to end of text at each keystroke in HtmlSlateWidget
**Describe the bug**
The cursor jumps at the end of the text at each keystroke when there's an external link

**To Reproduce**

This widget seems not to be in use in the latest versions of Plone, at least in the demo, but when the HtmlSlateWidget is used, the bug occurs in the editor.
If there is an external link, if the cursor is placed before said link, the cursor moves at the end of the text at each keystroke, making it impossible to write anything without having to replace the cursor after each letter.

**Expected behavior**
The cursor should not move at the end of the text at each character being typed in

**Screenshots**

Image

- OS: iOS
- Browser: Chrome
- Volto Version 17.22.3 (but supposedly on 18 too, since the widget is the same)
- Plone Version 6.0.14
- Plone REST API Version 9.15.2
- 
**Additional context**
The fix is necessary because this widget is the default widget assigned to richtext and it is used in an array of other addons such as RedTurtle/volto-editablefooter and possibily many others

sabrina-bongiovanni:
> [!CAUTION]

-----

- [ ] I signed and returned the [Plone Contributor Agreement](https://plone.org/foundation/contributors-agreement), and received and accepted an invitation to join a team in the Plone GitHub organization.
- [ ] I verified there aren't other open [pull requests](https://github.com/plone/volto/pulls) for the same change.
- [ ] I followed the guidelines in [Contributing to Volto](https://6.docs.plone.org/volto/contributing/index.html).
- [ ] I successfully ran [code linting checks](https://6.docs.plone.org/volto/contributing/linting.html) on my changes locally.
- [ ] I successfully ran [unit tests](https://6.docs.plone.org/volto/contributing/testing.html) on my changes locally.
- [ ] I included a [change log entry](https://6.docs.plone.org/contributing/index.html#contributing-change-log-label) in my commits.

-----

Closes #7556

-----

---
Issue:
The cursor jumps at the end of the text at each keystroke when there's an external link

Fix:
Using useState instead of useMemo

const valueFromHtml = React.useMemo(() => fromHtml(value), [value, fromHtml]);
This runs every time the parent passes a new value prop.

Every keystroke:

User types → handleChange → onChange(id, toHtml(newValue))
Parent form/state updates the HTML value → passes new [value] prop down
useMemo sees [value] changed → recomputes fromHtml(value) → creates a brand-new Slate JSON tree
So useMemo recalculates a new Slate tree every render and it loses track of the selection → cursor jumps to the end.

useState only runs that initializer once, on mount, [slateValue] is stored locally in the component.
setSlateValue(newValue) updates the same slateValue array, just mutating it through React.
React re-renders, but the objects inside the Slate tree are mostly the same — Slate sees the same node references.
Because of that, Slate does not reset selection, so cursor stays.

**Internal vs External links**
**Internal links**: isInternalURL(url) → often converted to relative paths like /some-page or /folder/item.
Slate sees the with href="/..." as a normal inline element.

**External links:** !isInternalURL(url) → full URL like https://example.com/.

In edit mode, your gets attributes for marking it as “external”
Browsers treat as a “special” editable node. When typing before it, the browser moves the caret to the end of the link automatically


pnicolli (approved):

davisagli (commented):
Using React state within the component means that it will no longer update if the value changes from outside (for example if a new value is fetched from the backend). I'm not sure if that's a problem or not.

sabrina-bongiovanni:
> Using React state within the component means that it will no longer update if the value changes from outside (for example if a new value is fetched from the backend). I'm not sure if that's a problem or not.

That’s technically correct — using React state means the editor won’t automatically update if new data comes in from the backend after the initial mount. However, in our current use cases this doesn’t present an actual issue.

Here’s why:

- The widget initializes once with the data fetched from the backend.
- While editing, all updates happen locally inside the editor and are handled through onChange.
- When you save, the updated content is submitted to the backend.
- On page refresh or form reload, the widget mounts again with the new backend data — so everything stays in sync.

In other words, the editor doesn’t need to react to live backend updates during editing, since its lifecycle already ensures data consistency on save/load.

That said, it’s a valid point — if anyone has a use case where the editor should reflect external data updates in real time (e.g., collaborative editing or live refresh scenarios), it would be useful to explore that further.

sabrina-bongiovanni:
- need to check the tests
- need to get some more input on previous comments

tiberiuichim (approved):

packages/volto-slate/src/widgets/HtmlSlateWidget.jsx
```diff
@@ -104,16 +104,30 @@ const HtmlSlateWidget = (props) => {
     [editor],
   );
 
-  const valueFromHtml = React.useMemo(() => {
-    return fromHtml(value);
-  }, [value, fromHtml]);
+  // const valueFromHtml = React.useMemo(() => {
```
avoinea, line 107: @sabrina-bongiovanni Can we get rid of this comments?

packages/volto-slate/src/widgets/HtmlSlateWidget.jsx
```diff
@@ -104,16 +104,30 @@ const HtmlSlateWidget = (props) => {
...
+  // that caused the internal Slate tree to re-render after each keystroke
+
+  // React does not replace slateValue unless you call setSlateValue() yourself.
+  // With useState Slate gets the same object reference between renders.
+  // As you type, only internal Slate ops change the document — React doesn’t rebuild it from HTML.
+
+  // handleChange does call setSlateValue but:
+  // - The object identity of slateValue remains stable within Slate’s internal lifecycle
+  // - React re-renders, but Slate’s Editable sees the same node tree reference,
+  //   so it doesn’t reset the selection.
```
avoinea, line 123: Also these ones. I suppose if it's a fix, we don't need explanations in the code.

avoinea (changes requested):

avoinea:
@sabrina-bongiovanni Check also cypress.

PR #7573, opened 2025-10-31
fix ImageWidget inside non folderish items

jnptk:


sneridagh:
@jnptk backport to 18 please!

PR #7588, opened 2025-11-03
fix(css-modules): replace literal [local] placeholder in production .…

### LINKED ISSUE #7587: Production LESS CSS Modules emit literal [local] in class names
In production builds (Volto with css-loader + LESS), classes generated for .module.less files appear with the literal placeholder [local] instead of the actual local class name. Example observed:
```
Carousel-module__[local]___KTvpt
```
Root cause: there is a custom getLocalIdent in volto/webpack-plugins/webpack-less-plugin.js. When you define getLocalIdent, css-loader stops post‑processing the template (localIdentName). The custom function calls interpolateName, which does not know the [local] token, so it remains unchanged. In development (dev=true) this getLocalIdent is not applied, hiding the bug.

To Reproduce  
1. In a Volto project using the original webpack-less-plugin.js, create Example.module.less:
   ```less
   .container {
     color: red;
   }
   ```
2. Import it in a React component:
   ```jsx
   import styles from './Example.module.less';
   <div className={styles.container}>Test</div>
   ```
3. Run a production build:
   ```
   pnpm --filter @plone/volto build
   ```
4. Inspect the final HTML or CSS (build/public/static/css/*.css) and locate the generated class:
   ```
   Example-module__[local]___a1B2C
   ```
5. Start in dev mode (pnpm --filter @plone/volto start) and confirm the class name is correct (no literal [local]).

Expected behavior  
The placeholder [local] should be replaced by the original local class name:
```
Example-module__container___a1B2C
```
Preserving the pattern defined in localIdentName: '[name]__[local]___[hash:base64:5]'.

Additional context  
Original config snippet:
```js
modules: {
  auto: true,
  localIdentName: '[name]__[local]___[hash:base64:5]',
  getLocalIdent: getLocalIdent,
}
```
Custom function:
```js
function getLocalIdent(loaderContext, localIdentName, localName, options) {
  const relativeResourcePath = /* ... */;
  options.content = `${options.hashPrefix}${relativeResourcePath}\x00${localName}`;
  return interpolateName(loaderContext, localIdentName, options);
}
```
interpolateName replaces [hash], [name], [path], etc., but not [local]. Without css-loader’s internal post-processing, [local] stays literal.

Why it went unnoticed: dev environment doesn’t apply getLocalIdent; more .module.less components were added recently; production class names weren’t inspected before.

Impact:
- Unexpected class names with literal tokens.
- Harder debugging and substring-based global overrides.
- Potential inconsistency where selectors assumed expansion of [local].

Fix options:
1. Patch getLocalIdent to manually replace [local]:
   ```js
   function getLocalIdent(loaderContext, localIdentName, localName, options) {
     const relativeResourcePath = normalizePath(
       path.relative(options.context, loaderContext.resourcePath),
     );
     options.content = `${options.hashPrefix}${relativeResourcePath}\x00${localName}`;
     const finalLocalIdentName = localIdentName.includes('[local]')
       ? localIdentName.replace(/\[local]/g, localName)
       : localIdentName;
     return interpolateName(loaderContext, finalLocalIdentName, options);
   }
   ```

alexandreIFB:
### Summary
Fix CSS class name generation for `.module.less` files in production builds: the literal `[local]` placeholder was not expanded, producing names like `Component-module__[local]___Abc12`. It is now correctly replaced, yielding `Component-module__container___Abc12`.

### Context / Problem
In production (`dev=false`), the `webpack-less-plugin` provides a custom `getLocalIdent`. When `getLocalIdent` is defined, `css-loader` skips its usual post-processing of `localIdentName`. Our implementation called `interpolateName` directly, which doesn’t know the `[local]` token, leaving it untouched.

Observed:
```
Carousel-module__[local]___KTvpt
```
Expected:
```
Carousel-module__slide___KTvpt
```

It went unnoticed because in development mode (`dev=true`) the custom `getLocalIdent` isn't applied.

### Reproduction Steps
1. Create `Example.module.less`:
   ```less
   .container {
     color: red;
   }
   ```
2. Import in a React component:
   ```jsx
   import styles from './Example.module.less';
   <div className={styles.container}>Test</div>
   ```
3. Run a production build:
   ```
   pnpm --filter @plone/volto build
   ```
4. Inspect generated CSS and find a class with literal `[local]`.
5. In dev mode (`pnpm --filter @plone/volto start`) the name is correct.

### Implemented Solution
In webpack-less-plugin.js, inside `getLocalIdent`:
- Manually replace all `[local]` occurrences with `localName` before calling `interpolateName`.

### Risks / Impact
- Production class names for `.module.less` change (now correct). Any global overrides relying on the buggy literal `[local]` will stop working (expected).
- Hash stability preserved (hash content basis unchanged).

Closes #7587

### DIFF
--- a/packages/volto/news/7587.bugfix
+++ b/packages/volto/news/7587.bugfix
@@ -1,0 +1,1 @@
+Fixes a bug where `[local]` remained literal in class names generated for `.module.less` files in production builds due to a custom `getLocalIdent`; the token is now replaced before calling `interpolateName`.
--- a/packages/volto/webpack-plugins/webpack-less-plugin.js
+++ b/packages/volto/webpack-plugins/webpack-less-plugin.js
@@ -19,7 +19,13 @@ function getLocalIdent(loaderContext, localIdentName, localName, options) {
   // eslint-disable-next-line no-param-reassign
   options.content = `${options.hashPrefix}${relativeResourcePath}\x00${localName}`;
 
-  return interpolateName(loaderContext, localIdentName, options);
+  // css-loader does not post-process localIdentName when a custom getLocalIdent is defined.
+  // Therefore we must manually replace the [local] token before calling interpolateName.
+  const finalLocalIdentName = localIdentName.includes('[local]')
+    ? localIdentName.replace(/\[local\]/g, localName)
+    : localIdentName;
+
+  return interpolateName(loaderContext, finalLocalIdentName, options);
 }
 
 const hasPostCssConfig = () => {


wesleybl:
I can confirm this problem.

I can also confirm that updating to the latest version of `css-loader` (7.1.2) and removing the customization in `getLocalIdent` resolves the issue.

However, version 7 brings significant changes that break compatibility:

- `import styles from './Example.module.less';` does not work. It has to be: `import * as styles from './Example.module.less';`

- Using base64 in the value of a CSS property does not work:

```CSS
border-image: url('data:image/svg+xml;base64,abc...');
```

But there is a workaround, see: https://github.com/webpack/css-loader/issues/1367

Perhaps this fix can help for Volto18. But we need to check if we can update the `css-loader` version in Volto 19.

AarishMansur:
Hi @alexandreIFB  I noticed there hasn't been any recent activity here. Are you still planning to continue this PR, or would it be okay if I picked this up and opened a new PR to help move it forward?

PR #7589, opened 2025-11-04
Enable ESlint detection of '.only' in unit and cypress tests

wesleybl:
Backport of #7567 to Volto 18


sneridagh:
@davisagli yes, I think we should treat it as a bugfix.

PR #7603, opened 2025-11-08
Ebrehault listing block

ebrehault:
Naive implementation.

Comments:

- The call to collect the items is done directly in `apps/seven/app/root.tsx`, not sure if that's accurate
- I have changed the `BlocksFormData` type as followed:
```ts
export type BlocksFormData = ListingBlockFormData | DummyBlockFormData;
```
That way we can have a specific typing for each type of blocks instead of having one generic interface with a very unhelpful `[x: string]: unknown;`.
`DummyBlockFormData` is there for now as the other specific block types are not existing yet, but it is meant to be removed once everything is properly typed.

----
📚 Documentation preview 📚: https://plone-registry--7603.org.readthedocs.build/

----
📚 Documentation preview 📚: https://volto--7603.org.readthedocs.build/

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,8 +35,8 @@
   },
   "devDependencies": {
     "concurrently": "^8.2.2",
-    "husky": "^9.1.7",
-    "lint-staged": "^15.5.2",
+    "husky": "^9.0.11",
+    "lint-staged": "^15.2.2",
     "tailwindcss": "catalog:",
     "tsconfig": "workspace:*",
     "typescript": "catalog:"
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -17,10 +17,10 @@ catalogs:
       version: 7.12.0
     '@tailwindcss/vite':
       specifier: ^4.1.4
-      version: 4.1.17
+      version: 4.1.13
     '@tanstack/react-query':
       specifier: ^5.59.0
-      version: 5.90.10
+      version: 5.87.4
     '@testing-library/react':
       specifier: ^16.3.0
       version: 16.3.0
@@ -35,10 +35,10 @@ catalogs:
       version: 19.2.3
     '@vitejs/plugin-react':
       specifier: ^5.0.2
-      version: 5.1.1
+      version: 5.0.2
     '@vitest/coverage-v8':
       specifier: ^4.0.0
-      version: 4.0.13
+      version: 4.0.12
     cypress:
       specifier: 15.7.0
       version: 15.7.0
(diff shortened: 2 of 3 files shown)


davisagli:
I changed the base branch to `seven` instead of `main`.

ebrehault:
> I changed the base branch to `seven` instead of `main`.

oops, sorry

apps/seven/app/root.tsx
```diff
@@ -52,6 +52,18 @@ export async function loader({ params, request }: Route.LoaderArgs) {
       cli.getSite(),
     ]);
 
+    const listingBlocks: { id: string; block: ListingBlockFormData }[] =
```
sneridagh: yes, this was the idea, however in the future we should have a generic "visitor" that executes specific extra calls if needed. Using an utilitity for that then call them all and execute them like the `rootLoaderDataUtilities`.

ebrehault: @sneridagh ok, done :)

I was not very comfortable using a very loose typing for the utility method (`(...args: any[]) => any`), so I have created a more precise typing, and default to the original as `DummyUtilityMethod`.

That way, you can do things like:
```
config.getUtilities<ContentSubRequestUtilityMethod>({
  type: 'rootContentSubRequest',
});
```
and you know for sure what you get.
That's safer.

Ideally, we should bind the typing with the utility type, like when calling `getUtilities` with `type: 'rootContentSubRequest'` then we should mandatorily obtain some `ContentSubRequestUtilityMethod`, but that would change the `setUtility` and `getUtilities` signatures, so maybe that should be done in another PR.

sneridagh: That's great! We were wondering if we could infer that that automatically in utilities, but this is sufficient, for sure.
Let's do this in another PR, yep.

ebrehault:
I think my PR is ready.

Last questions:
- not sure why `packages/plate/output.css` got modified, I did commit it, but not sure I had to
- when adding catalog dependencies in the different packages with `pnpm add react-i18next@catalog: --filter @plone/blocks`, I saw `pnpm-workspace.yaml` was modified automatically. My understanding it should not, because catalog is managed in `catalog.json`. What is the proper pnpm command?

ebrehault:
@sneridagh rebased according changes from #7652 
Shall I merge it?

ebrehault:
@pnicolli @sneridagh as this PR was there for a while (and as rebasing is sometimes a bit painful), I have just merged it.
I hope it's fine, if not, I'll revert it.

packages/types/package.json
```diff
@@ -49,8 +49,12 @@
   "devDependencies": {
     "@types/react": "catalog:",
     "@types/react-dom": "catalog:",
+    "react-router": "catalog:",
     "release-it": "catalog:",
     "tsconfig": "workspace:*",
     "typescript": "catalog:"
+  },
+  "dependencies": {
+    "@plone/client": "workspace:*"
```
sneridagh, line 58: @ebrehault I know it sucks but, unfortunately, we can't do this... :( `@plone/types` can't rely on any other package.

I just stumbled upon a cyclic import when getting back to this.

I don't know how we can overcome this, in an elegant way, but we managed so far to avoid it.

/cc @pnicolli

sneridagh:
@ebrehault also, sorry for being so late to take a look at this :(

packages/blocks/index.ts
```diff
@@ -1,9 +1,11 @@
 import type { ConfigType } from '@plone/registry';
+import type { ListingBlockFormData, LoaderUtilityArgs } from '@plone/types';
```
sneridagh, line 2: @ebrehault I would treat `LoaderUtilityArgs` as a Seven artifact. In fact, it's unlikely that anybody could use it outside it (thus, we remove the necessity of being in `@plone/types`), then we can remove the hard dependency from there.

packages/types/src/config/Utilities.d.ts
```diff
@@ -1,3 +1,7 @@
+import { Content } from '../content';
+import type { Params } from 'react-router';
```
sneridagh, line 2: Also, `@plone/types` can't rely on `react-router` 7, otherwise, Volto would break :/

ebrehault:
I have reverted it, I'll work on the fixes in a new PR :)

sneridagh:
@ebrehault I was working this morning on it, I will push what I have (this is previous to the revert) then you can rework it as you'd like! Thanks!

PR #7605, opened 2025-11-09
Update to React 18.3.1 and use RenderToPipeableStream in the SSR part

sneridagh:
I've been investigating lately the use of Suspense in Volto itself. Since we are not using `RenderToPipeableStream` in the SSR part, our build is not fully compatible with Suspense. This addresses that and upgrades to latest React 18.3.1. I am updating `@loadable/component` to latest version too. It seems that the three of them have not yet fully support for working together, since `@loadable/component` is not maintained anymore, so there's a small drawback.

If using Suspense and React.lazy together (sometimes, not always since it's a timing issue) you get this error:

```
react-dom.development.js:20701 Uncaught Error: The server did not finish this Suspense boundary: The server used "renderToString" which does not support Suspense. If you intended for this Suspense boundary to render the fallback content on the server consider throwing an Error somewhere within the Suspense boundary. If you intended to have the server wait for the suspended component please switch to "renderToPipeableStream" which supports Suspense on the server
    at updateDehydratedSuspenseComponent (react-dom.development.js:20701:1)
    at updateSuspenseComponent (react-dom.development.js:20401:1)
    at beginWork (react-dom.development.js:21663:1)
    at beginWork$1 (react-dom.development.js:27465:1)
    at performUnitOfWork (react-dom.development.js:26596:1)
    at workLoopSync (react-dom.development.js:26505:1)
    at renderRootSync (react-dom.development.js:26473:1)
    at performConcurrentWorkOnRoot (react-dom.development.js:25777:1)
    at workLoop (scheduler.development.js:266:1)
    at flushWork (scheduler.development.js:239:1)
    at MessagePort.performWorkUntilDeadline (scheduler.development.js:533:1)
```

Because how loadable components behave... On the other side, the SSR renders well (eg. the suspense boundary is observed, and it does not render until it's resolved), and the response returned in SSR mode (eg. no JS involved, SEO mode).

BUT this (misleading) error is shown, ONLY sometimes... 

Anyways, I would not use React.lazy in core, continuing using loadables instead, but state that Volto support Suspense in add-ons and projects, if needed. So, I would push for this one. 

Also, it would be needed to update the add-ons... UNFORTUNATELY, all are pinned to use as peer 18.2.0 especifically.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -732,9 +732,6 @@ importers:
       '@fiverr/afterbuild-webpack-plugin':
         specifier: ^1.0.0
         version: 1.0.0
-      '@jest/globals':
-        specifier: ^29.7.0
-        version: 29.7.0
       '@loadable/babel-plugin':
         specifier: 5.16.1
         version: 5.16.1(@babel/core@7.28.5)
@@ -801,9 +798,6 @@ importers:
       '@types/history':
         specifier: ^4.7.11
         version: 4.7.11
-      '@types/jest':
-        specifier: ^29.5.8
-        version: 29.5.14
       '@types/loadable__component':
         specifier: ^5.13.10
         version: 5.13.10
@@ -13314,6 +13308,7 @@ snapshots:
     dependencies:
       '@babel/core': 7.28.5
       '@babel/helper-plugin-utils': 7.27.1
+    optional: true
 
   '@babel/plugin-syntax-bigint@7.8.3(@babel/core@7.28.5)':
     dependencies:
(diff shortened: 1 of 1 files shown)


sneridagh:
Ok, acceptance tests looks like a christmas tree 🎄 😰

wesleybl:
@sneridagh I tried to merge them but it didn't work. So I undid it. I'll try again later.

sneridagh:
@wesleybl this was a test that I did in case that it was easy and straightforward, however I no longer think it's relevant, or useful... What I wanted to achieve it seems that cannot be done in 18... Nevertheless, feel free to continue with it if you want to.

wesleybl:
@sneridagh So I'll at least make a PR to update the React version.

sneridagh:
@wesleybl I think that's not a good move either... we will win very little, and lots of hassle. Latest 18 only brings a layer of compat to migrate easily to 19, which we will never do. Also, turns out that the add-ons template has the version pinned (because 18.2.1 was meant to be the last 18 one...). This will mean that we will have to force to all projects and add-ons to fix their peer dependencies which will be terrible for real nothing in return. So I'd leave this and live with it.

PR #7615, opened 2025-11-12
Improve upgrade guide19

sneridagh:
@davisagli @wesleybl trying to complete the steps, but this for sure will be needed.
I will update the Dockerfile ones in cookieplone.

----
📚 Documentation preview 📚: https://volto--7615.org.readthedocs.build/

### DIFF
--- a/package.json
+++ b/package.json
@@ -53,8 +53,7 @@
       "jest-resolve@26.6.2": "patches/jest-resolve@26.6.2.patch"
     },
     "ignoredBuiltDependencies": [
-      "@parcel/watcher",
-      "@swc/core"
+      "@parcel/watcher"
     ],
     "onlyBuiltDependencies": [
       "@tailwindcss/oxide",


docs/source/upgrade-guide/index.md
```diff
@@ -48,7 +48,64 @@ You should take the following actions for your Volto 19 projects.
 
 If you can't upgrade immediately, you may continue to run Volto 19 on Node.js 20 at your own risk, but be aware that issues specific to Node.js 20 will not be fixed in the Volto core CI or releases.
 
-(19-removed-support-for-loading-configuration-from-project-label)=
+### Changes to the project frontend folder boilerplate
```
wesleybl: We already had this: https://github.com/plone/volto/blob/main/docs/source/upgrade-guide/index.md#pnpm-10-no-longer-runs-lifecycle-scripts

Perhaps this could be removed?

docs/source/upgrade-guide/index.md
```diff
@@ -48,7 +48,64 @@ You should take the following actions for your Volto 19 projects.
...
+in {file}`frontend/package.json` update:
+
+```diff
+  "pnpm": {
+    "overrides": {
+      "@pmmmwh/react-refresh-webpack-plugin": "^0.5.15",
+      "react-refresh": "^0.14.2",
+    },
++   "ignoredBuiltDependencies": [
++      "@parcel/watcher"
```
wesleybl, line 70: In Volto, `@swc/core` is also ignored. Perhaps it needs to be added to cookieplone.

docs/source/upgrade-guide/index.md
stevepiercy: Or merged with it? It does appear to be somewhat redundant.

wesleybl: I think if we're going to keep it, we'd have to move to this step.

sneridagh: @wesleybl I think that is better to merge, and explain well what is this feature doing.

docs/source/upgrade-guide/index.md
sneridagh: I think that's because of Next.js build, but not sure. In all the projects with 19 that I setup these last days, in none it appeared.

wesleybl: If it's not in the ignore list, pnpm won't execute the build, but will give a warning saying it ignored it. Not having it there doesn't cause a problem if you don't need it. It's more to prevent pnpm from saying it ignored it. This would only apply to projects that use the same configuration as the Volto Core.

docs/source/upgrade-guide/index.md
wesleybl: But I think these are related things and should be done in the same step. If you want to do this in another PR, that's fine.

docs/source/upgrade-guide/index.md
wesleybl: I did a test here removing it from the Volto Core and it didn't give any warnings. So it's not being used. The best thing to do here is to remove it from Volto.

docs/source/upgrade-guide/index.md
sneridagh: I just merged them.

docs/source/upgrade-guide/index.md
sneridagh: @wesleybl pushed, I removed it, and indeed it did not throw any error.

PR #7623, opened 2025-11-15
Fix : UI crash in Reset Password form when username/password is invalid #7533

### LINKED ISSUE #7533: Reset Password UI crash on invalid username
**Describe the bug**
Reset Password UI crash on invalid username

**To Reproduce**

1. Go to https://demo.plone.org/passwordreset/b78200e82a05411e9e20857bd9aa9d49
2. type fake username and password, eg. 'user.foo', 'password.bar'
3. Click on the continue arrow
5. See error

**Expected behavior**
A message that says the username is invalid, or one that explains that if the username exists, the user will receive an email with instructions (to avoid user enumeration).

**Screenshots**

Image

- Volto 18.27.3
- Plone 6.1.3
- plone.restapi 9.15.3

SumitSingh69:
Fixes #7533

Before (when there was no error handling for invalid username/password) : 

503770663-fc2f60f7-c417-4596-b824-33374340d689

After (proper error handling) :
Screenshot 2025-11-15 at 5 46 31 PM

Screenshot 2025-11-15 at 5 45 59 PM

Quick gist of what I did : 

The Reset Password form was crashing when the username was invalid or when the backend returned an error.
This happened because UNSAFE_componentWillReceiveProps(nextProps) only handled the successful submission case.
In failure cases, no error message was set, leaving error = null, which caused the UI to break when accessed.

🔧 What I changed

Added error handling inside UNSAFE_componentWillReceiveProps(nextProps) for unsuccessful form submissions.

Added both:

Personalized error messages (e.g., invalid username, bad request, etc.)

A fallback generic error message for unexpected cases.

Ensures error is always a string, preventing UI crashes.

✅ Result

The form now behaves correctly:

Success → loads normally

Failure → shows the appropriate error message

No more error is null UI crashes

Let me know , if we need addons in this @stevepiercy @mamico


SumitSingh69:
i have accepted the contributor agreement, Kindly review this pull request .

packages/volto/src/components/theme/PasswordReset/PasswordReset.jsx
```diff
@@ -167,6 +192,14 @@ class PasswordReset extends Component {
...
    */
   onSubmit(data) {
+    if (data.password.length < 8 || data.passwordRepeat.length < 8) {
+      this.setState({
+        error: {
+          message: this.props.intl.formatMessage(messages.passwordNotValid),
+        },
+      });
+      return;
+    }
```
mamico, line 199: In my opinion, this is out of scope for this PR, and this type of validation should be handled in the backend rather than here.”

packages/volto/src/components/theme/PasswordReset/PasswordReset.jsx
```diff
@@ -157,6 +161,27 @@ class PasswordReset extends Component {
...
+
+      if (status === 404) {
+        message = 'Username is invalid.';
+      } else if (status === 400) {
+        message = 'Bad request.';
+      } else if (status === 401) {
+        message = 'Unauthorized.';
+      } else if (status === 500) {
+        message = 'Server error.';
+      }
```
mamico: I believe those messages will end up untranslated. Is that correct?

packages/volto/src/components/theme/PasswordReset/PasswordReset.jsx
SumitSingh69: Agreed,  just wanted to make sure if password is less than 8 characters , then the form should not submit . Let me know if you want me to remove this check .

packages/volto/src/components/theme/PasswordReset/PasswordReset.jsx
SumitSingh69: ahh yes they aren't translated . I will update this PR and let you know.

SumitSingh69: I have added the translations .

packages/volto/locales/ca/LC_MESSAGES/volto.po
```diff
@@ -496,6 +496,11 @@ msgstr ""
 msgid "Back"
 msgstr "Enrere"
 
+#. Default: "Bad request"
```
stevepiercy: It would be more helpful to tell the user what do after getting this error message than just "Bad request." Should they contact their website administrator? Make a sacrifice to the gods? Cry out in anguish? All the above?

```suggestion
#. Default: "Bad request."
```

packages/volto/locales/ca/LC_MESSAGES/volto.po
```diff
@@ -3544,6 +3559,11 @@ msgstr ""
 msgid "Some relations are broken. Please fix."
 msgstr ""
 
+#. Default: "Something went wrong"
```
stevepiercy: Not user friendly. See above comment. This can be better.

```suggestion
#. Default: "Something went wrong."
```

packages/volto/locales/volto.pot
```diff
@@ -492,6 +492,11 @@ msgstr ""
 msgid "Back"
 msgstr ""
 
+#. Default: "Bad request"
```
stevepiercy: It would be more helpful to tell the user what do after getting this error message than only "Bad request." Should they contact their website administrator? Make a sacrifice to the gods? Cry out in anguish? All the above?

```suggestion
#. Default: "Bad request. Please contact the website admin {site_admin}."
```

packages/volto/locales/volto.pot
```diff
@@ -3383,6 +3393,11 @@ msgstr ""
 msgid "Server Error"
 msgstr ""
 
+#. Default: "Server error"
```
stevepiercy: Not user friendly. See above comment. This can be better.

```suggestion
#. Default: "Server error. Please contact the website admin {site_admin}."
```

packages/volto/locales/volto.pot
```diff
@@ -4123,6 +4143,11 @@ msgstr ""
 msgid "Unauthorized"
 msgstr ""
 
+#. Default: "Unauthorized"
+#: components/theme/PasswordReset/PasswordReset
+msgid "Unauthorized."
+msgstr ""
+
```
stevepiercy: I'd reuse the existing message.

`msgid "You are trying to access a protected resource, please {login} first."`

```suggestion
```

packages/volto/locales/volto.pot
```diff
@@ -3540,6 +3555,11 @@ msgstr ""
 msgid "Some relations are broken. Please fix."
 msgstr ""
 
+#. Default: "Something went wrong"
+#: components/theme/PasswordReset/PasswordReset
+msgid "Something went wrong."
+msgstr ""
+
```
stevepiercy: Reuse:

`msgid "Sorry, something went wrong with your request"`

```suggestion
```

packages/volto/locales/volto.pot
```diff
@@ -4300,6 +4325,11 @@ msgstr ""
 msgid "Username"
 msgstr ""
 
+#. Default: "Username is invalid"
```
stevepiercy:
```suggestion
#. Default: "Username is invalid."
```

packages/volto/locales/volto.pot
```diff
@@ -3540,6 +3555,11 @@ msgstr ""
 msgid "Some relations are broken. Please fix."
 msgstr ""
 
+#. Default: "Something went wrong"
+#: components/theme/PasswordReset/PasswordReset
+msgid "Something went wrong."
```
stevepiercy:
```suggestion
#. Default: "Something went wrong. Please contact the website admin {site_admin}."
#: components/theme/PasswordReset/PasswordReset
msgid "Something went wrong. Please contact the website admin {site_admin}."
```

stevepiercy (changes requested):

packages/volto/locales/volto.pot
SumitSingh69: Updated the message so users know what to do — no divine intervention needed haha .

packages/volto/locales/volto.pot
SumitSingh69: Done

packages/volto/locales/volto.pot
SumitSingh69: Done

packages/volto/locales/volto.pot
SumitSingh69: updated with a better user friendly message.

packages/volto/locales/volto.pot
SumitSingh69: done

stevepiercy (approved):
Messages look good. Needs a final technical review.

Does this PR need new test cases?

SumitSingh69:
Hi @mamico  @plone/volto-team ,
it’s been about two weeks since I opened this PR. Could you please take a look when you have a moment?

davisagli (changes requested):
This isn't consistent with how we do error handling for other forms in Volto that make a call to a server-side API. In general the API is supposed to return a meaningful, translated error message, and the frontend should just display it and not have a lot of logic for validation and translation.

That way:
- We don't have to include all these messages in the frontend locale files.
- We have consistent messaging between the API and UI
- We handle the fact that it's possible to configure a different password policy in the backend which has additional restrictions or a different password length limit.

SumitSingh69:
@davisagli I get it, what you are talking about , you mean directly showing the error messages coming from api to the user right . Like in the below contact-form 

Screenshot 2025-12-15 at 7 38 38 AM

I can do that but  the reset-password api is returning **null**  when there is an **invalid username** , that is the reason why the UI was breaking as there was no error handling to the null response in the frontend (that is the reason why i had to add my own user-friendly error messages) . Kindly , let me know what should I do next . Here is a image sample for your reference.
Screenshot 2025-12-15 at 7 48 25 AM

davisagli:
@SumitSingh69 What you did is okay for the case where the username is invalid and the API returns a 404 response, since the API is not returning a message in that case (this comes from https://github.com/plone/plone.restapi/blob/main/src/plone/restapi/services/users/add.py#L276).

But in other cases (especially 400 and 403 responses), the API does return a specific message and we should display it.

SumitSingh69:
alright @davisagli , just to let you know @stevepiercy wanted to have the errorMessages more user friendly like instead of saying "bad request" , he wanted the message should be "We couldn’t process your request. Please check your input and try again.". But , as per your comment , I will proceed to remove all other errorMessages made by me except the 404 error message, and simply display the API error messages in all the other cases.

SumitSingh69:
As per the changes , 
All the error status except 404 will have their own error messages coming from the API 
For example:
Screenshot 2025-12-15 at 7 36 32 PM
And for 404 (null response) status , We are the showing this error message 
Screenshot 2025-12-15 at 8 01 34 PM

stevepiercy:
> alright @davisagli , just to let you know @stevepiercy wanted to have the errorMessages more user friendly like instead of saying "bad request" , he wanted the message should be "We couldn’t process your request. Please check your input and try again.".

@davisagli is this something that can be handled in plone.restapi? Currently the response is user-hostile.

davisagli:
@stevepiercy We need to take a step back here.

If this form is used by a logged in user resetting their own password, we should show them the current username (to avoid mistakes) but not allow them to edit it. Then it won't be possible to get a 404 from the API here.

For logged out users, for security reasons we have to make sure this API can't be used to find out whether a username exists in the system. So we have to pretend the password reset was successful even if it was not. Yeah, that's bad UX, but security takes priority here.

We also have to make sure there is a way for an admin user to request a password reset to be sent to another user. I don't remember if we have this already in Volto or not, or whether it uses this form vs something else.

Long story short: I need to come back to this when I have more time to take a holistic look at everything.

SumitSingh69:
@davisagli Let me know , if you need any assistance from my side or any changes.

PR #7625, opened 2025-11-17
7397 invalid login credentials 401 issue

aryan7081:
Main Fix: #7649 

When login fails with 401 Unauthorised, the loader continues spinning and no error message is displayed to the user.
This happened because the hydration check in `middleware/api.js` was blocking `LOGIN_FAIL` dispatch for all actions during hydration, not just `GET_CONTENT`. Since `isHydrating` is `true` on initial page load and `hasExistingError` is `undefined` for static routes like `/login`, the error handler returned early without dispatching `LOGIN_FAIL`, leaving the loader in a perpetual loading state.

This fix restricts the hydration check to only apply to `GET_CONTENT` actions, allowing user-initiated actions like `LOGIN` to always dispatch their `_FAIL` actions. Additionally, `LOGIN` is added to `actions_raising_api_errors` config to enable proper 401 error handling.


davisagli:
@aryan7081 Thanks for the analysis and fix. However we like to merge fixes first to the `main` branch (Volto 19) and then backport to other branches if necessary. Could you please rebase this on main (or open a different PR)?

aryan7081:
@davisagli Raised #7649 against main. I'll backport after that merge. Thanks!

aryan7081:
Closing this as fixed here https://github.com/plone/volto/pull/7737

PR #7643, opened 2025-11-20
fix snapshots changed due to Jest remove pr

Abhishek-17h:


Abhishek-17h:
@sneridagh Do we need changelog for this?

davisagli:
@Abhishek-17h Thank you. I think we can skip the changelog since it's a minor internal change.

PR #7646, opened 2025-11-21
Fixed production build add prettier as direct dependency, added tests. 

sneridagh:
Acceptance tests now convert into production build after building too to make sure we test that too.


davisagli (commented):
@sneridagh I really don't understand this. We run `pnpm install --prod` but only after Volto (`pnpm start:prod`) has exited?

wesleybl:
Based on the description, it seems it is wrong:

> Acceptance tests now convert into production build after building too to make sure we test that too.

sneridagh:
@davisagli @wesleybl I will make a PR today.

davisagli:
@sneridagh https://github.com/plone/volto/pull/7664

Issue #7132, opened 2025-05-28
Unable to mirror/archive a Volto site into static HTML/CSS/JS, also affecting Wayback-machine


polyester:
**Describe the bug**

Using standard tools like wget, httrack, and others does not create a working static copy of a Volto site. The copy will keep trying to make calls to the 'backend', which is not online. This also affects common services like the [Wayback Machine](https://web.archive.org) 

This makes it hard to archive sites that no longer need updating, but where it would be nice to keep the contents for posterity. There are many other usecases for preserving the content of websites, such as LLM training etc.

Currently one tool that can create a WACZ file of a Volto site is [Browsertrix Crawler](https://crawler.docs.browsertrix.com), as it is using a full browser to actually render the results and then store this. The resulting WACZ is then 'playable' using Replayweb webservice or desktop app, but this is far from ideal, as tools for extracting WACZ to HTML are mostly non-functional.

**To Reproduce**

1. using Wget as example, issue the command `wget --recursive --page-requisites --adjust-extension --span-hosts --convert-links --restrict-file-names=windows --domains demo.plone.org --no-parent https://demo.plone.org`
2. open the resulting scrape in a browser
3. you will see a short glimpse of the front page, followed by an error stating "The backend is not responding, due to a server timeout or a connection problem of your device. Please check your connection and try again."

**Expected behavior**

Volto websites in their SSR make it possible for archiving tools to create a functional, if static, copy of the website.

**Screenshots**

![Image](https://github.com/user-attachments/assets/b28a3e52-40a3-439e-934e-c6dd9022a0cb)

**Software (please complete the following information):**

Discovered in trying to archive various older Plone conference sites, the problem appears both in older Volto sites and current demo.plone.org

I am not sure if this should be qualified as "bug" or "feature request".

polyester:
There's one use case where that would presumably not help: where you want to 'archive' a website at it's original URL, e.g. year.ploneconf.org sites for older years. You would want to serve the static copy at its original URL to maintain search engine rankings, etc. 

> The markup for a Volto page has embedded the publicURL where it was served, e.g. from demo.plone.org:
[...]

PR #7649, opened 2025-11-21
Fix login failure handling during hydration

### LINKED ISSUE #7397: Wrong login credentials not showing errors on 401 error
**Describe the bug**
When trying to login with wrong credentials, and backend responds with 401 - unauthorized, no errors are shown and loader on submit login doesn't disappear. 

**To Reproduce**

1. Go to '/login.'
2. Insert wrong credentials
3. Press login button
4. See error: loader is still loading and no error are displayed

**Expected behavior**
Display error and remove laoder on submit button. 

- OS: MAc
- Browser: Chrom
- Volto Version [e.g. 18.25.0]
- Plone Version [e.g. 6.1.0]
- Plone REST API Version [e.g. 9.1.0]

**Additional context**
I think the problem is in middleware/api.js when handling error (line ~310).
In that moment, isHydrating is true, because it becomes false only when you have a positive response, and `hasExistingError` consider only `content` action, but here we are in a static view (/login) not related to any content,  so it returns and not handles errors. 

Another problem i think there's in this error handling, is that `config.settings.actions_raising_api_errors` doesn't contains action `LOGIN`

@sneridagh the error was introduced here 
https://github.com/plone/volto/pull/7192/files#diff-01d7617436701d779414389f02f43b75a948e31334e862f240ff8c3b6f45e82b

aryan7081:
Fixes #7397
Backport Ref #7625 
When login fails with a 401 Unauthorised error, the loader keeps spinning. This is because the hydration guard in `middleware/api.js` short-circuited every action (including `LOGIN_FAIL`) while `isHydrating` was true, so the error handler returned early before dispatching `LOGIN_FAIL`.

This patch limits the hydration guard to `GET_CONTENT` only, so user-initiated actions like `LOGIN` always dispatch their `_FAIL` actions and stop the loader. It also adds `LOGIN` to `settings.actions_raising_api_errors`, ensuring the global API error handler surfaces 401s properly.

### DIFF
--- a/.github/actions/node_env_setup/action.yml
+++ b/.github/actions/node_env_setup/action.yml
@@ -3,7 +3,7 @@ name: Set up Node.js environment
 runs:
   using: "composite"
   steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v5
 
     - name: Use Node.js ${{ inputs.node-version }}
       uses: actions/setup-node@v4
@@ -19,7 +19,7 @@ runs:
       run: |
         echo "STORE_PATH=$(pnpm store path --silent)" >> $GITHUB_ENV
 
-    - uses: actions/cache@v4
+    - uses: actions/cache@v5
       name: Setup pnpm cache
       with:
         path: ${{ env.STORE_PATH }}
@@ -29,7 +29,7 @@ runs:
 
     - name: Cache Cypress Binary
       id: cache-cypress-binary
-      uses: actions/cache@v4
+      uses: actions/cache@v5
       with:
         path: ~/.cache/Cypress
         key: binary-${{ inputs.node-version }}-${{ hashFiles('pnpm-lock.yaml', 'package.json') }}
--- a/.github/workflows/acceptance.yml
+++ b/.github/workflows/acceptance.yml
@@ -11,7 +11,7 @@ jobs:
       matrix:
         node-version: [22.x, 24.x]
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
 
       - name: Set up Node.js environment
         uses: ./.github/actions/node_env_setup
@@ -59,7 +59,7 @@ jobs:
       matrix:
         node-version: [22.x, 24.x]
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
 
       - name: Set up Node.js environment
         uses: ./.github/actions/node_env_setup
@@ -107,7 +107,7 @@ jobs:
       matrix:
         node-version: [22.x, 24.x]
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
 
       - name: Set up Node.js environment
         uses: ./.github/actions/node_env_setup
@@ -155,7 +155,7 @@ jobs:
       matrix:
         node-version: [22.x, 24.x]
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
 
       - name: Set up Node.js environment
         uses: ./.github/actions/node_env_setup
@@ -203,7 +203,7 @@ jobs:
       matrix:
         node-version: [22.x, 24.x]
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
 
       - name: Set up Node.js environment
         uses: ./.github/actions/node_env_setup
@@ -251,7 +251,7 @@ jobs:
       matrix:
         node-version: [22.x, 24.x]
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
 
       - name: Set up Node.js environment
         uses: ./.github/actions/node_env_setup
@@ -299,7 +299,7 @@ jobs:
       matrix:
         node-version: [22.x, 24.x]
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@v5
 
       - name: Set up Node.js environment
         uses: ./.github/actions/node_env_setup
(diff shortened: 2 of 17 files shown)


packages/volto/src/middleware/api.js
```diff
@@ -341,8 +341,13 @@ const apiMiddlewareFactory =
         (error) => {
           // Make sure an error during hydration
           // (for example when serving an archived page)
-          // doesn't hide the SSR content.
-          if (isHydrating && !hasExistingError) {
+          // doesn't hide the SSR content. Only suppress GET_CONTENT failures;
+          // user-initiated actions (LOGIN, etc.) must always dispatch _FAIL so
+          // loaders stop and errors surface.
+          const shouldIgnoreHydrationError =
+            isHydrating && !hasExistingError && type === GET_CONTENT;
```
wesleybl, line 348: Perhaps it's best to be more conservative here and use `type !== LOGIN`.

Do you see any other situations where not having `type === GET_CONTENT` might be detrimental?

packages/volto/src/config/index.js
```diff
@@ -98,7 +98,7 @@ let config = {
     proxyRewriteTarget: process.env.RAZZLE_PROXY_REWRITE_TARGET || undefined,
     // apiPath: process.env.RAZZLE_API_PATH || 'http://localhost:8000', // for Volto reference
     // apiPath: process.env.RAZZLE_API_PATH || 'http://localhost:8081/db/web', // for guillotina
-    actions_raising_api_errors: ['GET_CONTENT', 'UPDATE_CONTENT'],
+    actions_raising_api_errors: ['GET_CONTENT', 'UPDATE_CONTENT', 'LOGIN'],
```
wesleybl: This is not necessary. The login does not have a specific error screen.

aryan7081:
@wesleybl Thanks for the review! I have addressed both the comments. The build failure is unrelated to my changes. It's a TypeScript error in `@plone/components` package's `ColorPicker.tsx` where the `ColorPickerProps` interface has a type mismatch with `react-aria-components` `ColorPickerProps` regarding the `children` property.
Thanks!

packages/volto/src/middleware/api.js
wesleybl: I'm not sure if my suggestion is the most correct one.

@davisagli  you added this check in #7148 to make the archived Volto site accessible: #7132

But this caused this side effect. My suggestion is more conservative, but other errors of this type may occur.

Opinion?

packages/volto/news/7397.bugfix
```diff
@@ -0,0 +1 @@
+Prevent login failures during hydration from being swallowed by GET_CONTENT-only guard and treat LOGIN as an action that should surface API errors. @aryan7081
```
wesleybl: Remove: `and treat LOGIN as an action that should surface API errors`

packages/volto/news/7397.internal
```diff
@@ -0,0 +1 @@
+Add Cypress regression test ensuring login failures stop the loader and show the toast notification. @aryan7081
```
wesleybl: It is not necessary to add a note for tests, unless only the test was added.

packages/volto/src/middleware/api.js
davisagli: In general isHydrating is marked as false as soon as any API call processed by this middleware succeeds. So we only get this type of problem when the first API has an error that we actually do want to report to the user.

As is, the PR should be enough to fix the specific error with the login form that we know about and understand, so I'll approve it.

It might be possible to do a better fix by being more selective about which errors we ignore here (mostly we want to avoid destorying the SSR content if there's a connection error because there's no actual backend available...but if we called the backend and got a response that is has an error status, it's ok to process the error normally).

packages/volto/news/7397.bugfix
davisagli: Let's write a release that focuses on the impact on the user rather than the technical details of the fix (someone who cares about the details can follow the link to the PR):

```suggestion
Fixed a bug in the login form where the user could see a spinner instead of an error message after a failed login, when the login form was loaded via server-side rendering. @aryan7081
```

packages/volto/news/7397.internal
davisagli: I agree, the test is part of the overall fix. Please just remove this file.

aryan7081: I have removed this file.
Thanks for the clarification!

packages/volto/src/middleware/api.js
wesleybl: @davisagli This error might be more common than we think. It likely occurs in the password reset form. And it can occur in any custom form. It's difficult to debug. So now that we're looking at this more closely, can we find a smarter solution?

If I understand correctly, what makes the content "disappear" is the global error page, which says the backend is down. So, is it safe to test only `GET_CONTENT`? Other errors won't display this global error page.

davisagli: @wesleybl @aryan7081 I tested this some more, and yes, I think we can change this back to `&& type === GET_CONTENT`

aryan7081: @wesleybl @davisagli I have restored hydration error check to `GET_CONTENT` only. Please review.
Thanks!

packages/volto/cypress/tests/core/basic/login.js
```diff
@@ -14,4 +14,16 @@ describe('Login Tests', () => {
     cy.get('#login-form-submit').click();
     cy.get('body').should('have.class', 'has-toolbar');
   });
+
+  it('Shows an error when logging in with wrong credentials', function () {
```
wesleybl, line 18: This error only occurs in SSR. You need to reload the `/login` page.

davisagli: Adding `cy.reload();` here should do it.
```suggestion
  it('Shows an error when logging in with wrong credentials', function () {
    cy.reload();
```

wesleybl:
@aryan7081 @davisagli Thank you for addressing this complicated problem!

wesleybl:
@aryan7081 Could you please do a backport to Volto 18?

PR #7654, opened 2025-11-24
Add classname for unauthorized page

iRohitSingh:
We need a slector so that we can style the unauthorized page.

### DIFF
--- a/.github/workflows/readme-link-check.yml
+++ b/.github/workflows/readme-link-check.yml
@@ -26,5 +26,6 @@ jobs:
           --exclude https://my-server-DNS-name.tld/api
           --exclude 2021.ploneconf.org
           --exclude https://www.lanku.eus/
+          --exclude https://medium.com/
           '**/README.md'
           'PACKAGES.md'
--- a/docs/source/blocks/block-style-wrapper.md
+++ b/docs/source/blocks/block-style-wrapper.md
@@ -130,7 +130,7 @@ const BlockView = (props) => (
 ```
 
 ```{seealso}
-[CSS custom properties basics](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_cascading_variables/Using_CSS_custom_properties)
+[CSS custom properties basics](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Cascading_variables/Using_custom_properties)
 ```
 
 ### Nested custom CSS properties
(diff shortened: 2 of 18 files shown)


davisagli:
This should be `view-unauthorized` for consistency with how other classes are assigned based on the view. And we need to make sure that the page does NOT have the `view-viewview` class.

Currently View.jsx gets the class from displayName for normal content views but not error views (https://github.com/plone/volto/blob/main/packages/volto/src/components/theme/View/View.jsx#L242). We should add that for error views as well (https://github.com/plone/volto/blob/main/packages/volto/src/components/theme/View/View.jsx#L223) for consistency.

sneridagh:
@iRohitSingh please follow David's suggestion.

iRohitSingh:
@sneridagh @davisagli Done.

packages/volto/src/components/theme/Unauthorized/Unauthorized.jsx
```diff
@@ -5,13 +5,15 @@ import { useSelector } from 'react-redux';
...
 import { getBaseUrl } from '@plone/volto/helpers/Url/Url';
+import BodyClass from '@plone/volto/helpers/BodyClass/BodyClass';
 
 const Unauthorized = () => {
   const error_message = useSelector((state) => state.apierror?.message);
   let location = useLocation();
 
   return (
     <Container className="view-wrapper">
+      <BodyClass className="view-unauthorized" />
```
davisagli, line 16: You don't need this, since you set displayName below.

iRohitSingh: @davisagli I also thing the same But for edit componet i need it. 

exact case

When you try to access the private edit page without loggedin user.

code:
https://github.com/plone/volto/blob/main/packages/volto/src/components/manage/Edit/Edit.jsx#L418

iFlameing:
@sneridagh waiting for your review.

sneridagh:
@iRohitSingh @iFlameing please check:

for /controlpanel
and accessing an unauthenticated/unauthorized resource, have the same styling.

iFlameing:
> @iRohitSingh @iFlameing please check:
> 
> for /controlpanel and accessing an unauthenticated/unauthorized resource, have the same styling.

@sneridagh this pr is related to adding className so that we can select the selector in VLT and style the page.

wesleybl:
@iRohitSingh Could you please backport this to Vote 18?

PR #7655, opened 2025-11-24
Only include the original image in the Image component if the image does not have all the scales present. Fixed and update tests.

sneridagh:
Follow up: #7486 (Seven PR):

- [x] Add a default scale size which is large enough to serve full-width images on a large high-DPI monitor -> https://github.com/plone/plone.volto/pull/205
- [ ] Fix the image field serializer in plone.restapi to include the same scales as the code which creates image_scales for the catalog (so that we aren't generating scales larger than the original). -> https://github.com/plone/plone.restapi/pull/1992
- [x] Always set sizes appropriately when we use the Image component (so that we don't end up with the largest scale when a smaller one would be sufficient)

### DIFF
--- a/docs/source/upgrade-guide/index.md
+++ b/docs/source/upgrade-guide/index.md
@@ -364,6 +364,19 @@ To support horizontal scrolling of tables on small viewports, the table block is
 In the past, the table block was a direct child of the content area.
 If you had custom styles for the table block that assumed it was a direct child of the content area, then adjust your CSS styles accordingly.
 
+### The image component now includes the original image only if necessary
+```{versionadded} Volto 19.0.0-alpha.29
+```
+
+The `Image` component has been optimized to include the original image URL only when necessary.
+Now it is only included if the image does not have all the defined scales present, which could happen if the image uploaded originally is smaller than the defined scales.
+In other scenarios where all the scales are present, including the original image could lead the browser to choose it over the scaled versions, impacting performance.
+This happened especially in high-density resolution screens where the largest scale available was not enough for the browser to pick a scaled version.
+
+This is a breaking change for projects that relied on the original image always being present, for example, in those projects where the original image was always included for large displays, such as televisions or wide-screen displays.
+A pair of additional scales were added to cover those use cases, enough to cover the highest density screens at the largest common resolutions.
+Additionally, if your project relied on the original image to always be present, then you need to either add an additional scale to cover your use case, run the upgrade steps defined in `plone.volto>=6.0.0a0`, or, in Plone 6.2, to use the new image scales named `2k` and `4k`.
+
 (upgrading-to-volto-18-x-x)=
 
 ## Upgrading to Volto 18.x.x
--- a/packages/volto/news/7655.bugfix
+++ b/packages/volto/news/7655.bugfix
@@ -1,1 +1,1 @@
-Only include the original image in the Image component if the image does not have all the scales present. Fixed and update tests. @sneridagh
+Only include the original image in the `Image` component if the image does not have all the scales present. @sneridagh


docs/source/upgrade-guide/index.md
```diff
@@ -333,6 +333,18 @@ We pinned the version of `sass` to `1.32.0`, which is the one before they introd
...
 In case that you need a later version of `sass` in your project or add-on, you can override it in your project's {file}`package.json` file.
 
+### The image component now includes the original image only if necessary
+```{versionadded} Volto 19.0.0-alpha.18
+```
+
+The `Image` component has been optimized to include the original image URL only when necessary.
+Now it is only included if the image does not have all the defined scales present, which could happen if the image is smaller than the defined scales.
+In other scenarios where all the scales are present, including the original image forced the browser to choose it over the scaled versions, impacting performance.
+This happened specially in high-density screens where the `large` scale was not enough for the browser to pick a scaled version.
```
```suggestion
This happened especially in high-density resolution screens where the `large` scale was not enough for the browser to pick a scaled version.
```

docs/source/upgrade-guide/index.md
```diff
@@ -333,6 +333,18 @@ We pinned the version of `sass` to `1.32.0`, which is the one before they introd
...
+It is a breaking change for projects that relied on the original image always being present, for example, those projects that wanted to use the original image always in big screens (not necessarily high-density, eg TV screens, etc).
```
```suggestion

This is a breaking change for projects that relied on the original image always being present, for example, in those projects where the original image was always included for large displays, such as televisions or wide-screen displays.
```

docs/source/upgrade-guide/index.md
```diff
@@ -333,6 +333,18 @@ We pinned the version of `sass` to `1.32.0`, which is the one before they introd
...
+We are adding an additional scale to cover that use cases, enough to cover the highest density screens at the largest common resolutions.
```
```suggestion
An additional scale was added to cover those use cases, enough to cover the highest density screens at the largest common resolutions.
```

docs/source/upgrade-guide/index.md
```diff
@@ -333,6 +333,18 @@ We pinned the version of `sass` to `1.32.0`, which is the one before they introd
...
+This is a breaking change, because if your project relied on the original image being always present, you need to add an additional scale to cover your use case, or run the upgrade steps defined in `plone.volto>=6.0.0a0` or in Plone 6.2 to use the new image scale named `humongous`
```
```suggestion
Additionally, if your project relied on the original image to always be present, then you need to either add an additional scale to cover your use case, run the upgrade steps defined in `plone.volto>=6.0.0a0`, or, in Plone 6.2, to use the new image scale named `humongous`.

```

packages/volto/src/components/manage/Blocks/Teaser/DefaultBody.jsx
```diff
@@ -64,6 +64,7 @@ const TeaserDefaultTemplate = (props) => {
                       alt=""
                       loading="lazy"
                       responsive={true}
+                      sizes="(max-width: 940px) 100vw, 940px"
```
davisagli: Ideally we should use a smaller size when there are multiple columns.

I don't love hardcoding the default width, but I don't have a better suggestion. We can't use CSS properties because the whole point of `sizes` is to give the browser enough information to pick the right image before CSS has loaded.

davisagli (commented):
The presence of the original image didn't force the browser to choose it. But we set width and height to the original size and in most cases didn't set `sizes`, so the browser didn't have any way to know it could use a smaller scale.

I think setting `sizes` appropriately would allow the browser's scale selection to work better even if we did always include the original image. I have a slight preference for leaving the original image there, because if we remove it then it takes extra work to use it in the use cases where it's really what you need (i.e. full-width on a large monitor)

sneridagh:
@stevepiercy I accepted all the suggestions, but David is working in the backend side, and we might amend them a bit the next days. Thanks!

stevepiercy:
@sneridagh @davisagli please ping me again for a review when it's done cooking.

stevepiercy:
Can someone else fix the broken links? I'm swamped.

https://github.com/plone/volto/actions/runs/19663404467/job/56314471303?pr=7655

sneridagh:
@stevepiercy it seems medium added some antibot measures... we cannot do anything about it. How do we proceed with this use cases?

stevepiercy:
@sneridagh for Medium, ignore the entire domain with a regex in `conf.py`. There are existing examples. It also needs to be applied to the upstream plone/documentation `conf.py`.

For the others, fix the links that redirect.

sneridagh:
@davisagli once https://github.com/plone/plone.volto/pull/205 is merged and released, we can go for this one. We might want to amend the docs with the released p.volto version.

@stevepiercy I amended the docs to reflect the latest changes.

packages/volto/src/components/manage/Blocks/Teaser/DefaultBody.jsx
```diff
@@ -64,6 +64,7 @@ const TeaserDefaultTemplate = (props) => {
                       alt=""
                       loading="lazy"
                       responsive={true}
+                      sizes="auto, (max-width: 940px) 100vw, 940px"
```
wesleybl: It would be better if this were a configuration. Different themes may have different sizes.

erral:
I would say that for the `srcset` attribute we need to always show the original image URL with its original width set. We have been seen this also in classic and have just prepared a PR in plone.namedfile.

Our usecase was the following:

The `srcset` attribute was showing only the URLs of scales that are smaller than the original url. In our example, the user uploaded a 500px image, so the `srcset` showed the 400px one as the first scale. The `sizes` attribute was setting a 500px wide space, so the browser was rendering the 400px scale, when the original image would fit perfectly.

davisagli:
@erral This still includes the original image in the srcset unless it is enormous (width greater than the largest defined scale size, which is 4000 pixels in Plone 6.2)

packages/volto/src/components/manage/Blocks/Teaser/DefaultBody.jsx
pnicolli: I agree with Wesley here, we should use an approach like the Image block approach.

See the config and the block here:
https://github.com/plone/volto/blob/main/packages/volto/src/config/Blocks.jsx#L292
https://github.com/plone/volto/blob/main/packages/volto/src/components/manage/Blocks/Image/View.jsx#L67

pnicolli (changes requested):

packages/volto/news/7655.bugfix
```diff
@@ -0,0 +1 @@
+Only include the original image in the Image component if the image does not have all the scales present. Fixed and update tests. @sneridagh
```
stevepiercy:
```suggestion
Only include the original image in the `Image` component if the image does not have all the scales present. Fixed and updated tests. @sneridagh
```

davisagli:
@stevepiercy Good catch, the upgrade guide was accidentally lost in a bad merge. I re-added it in https://github.com/plone/volto/pull/7655/commits/ddb87a8b34b9a3aea4882ea222d6f377a461a0e6

stevepiercy (approved):

sneridagh:
Merged! Thanks to all involved!

PR #7663, opened 2025-11-28
Enhancement — Render `NotFound` when a control panel does not exist

### LINKED ISSUE #7214: Enhancement suggestion: Show NotFound when a control panel is not found
Currently, when accessing a non-existent control panel URL (e.g., `/controlpanel/does-not-exist`), the `Controlpanel` component returns an empty `<div />` if the panel is not found. This provides no visual feedback to the user about the error.

### Problem

- Users who type an incorrect URL or try to access a removed panel receive no indication that the panel does not exist.
- This can be confusing, as the screen remains blank with no error message.

### Proposal

Change the component to explicitly render the `NotFound` screen (`<NotFound />`) when the panel is not found, instead of returning an empty `<div />`.

**Benefits:**
- Provides clear feedback to the user when a panel does not exist.
- Improves the user experience and usability of the system.

VibhuGupta-dev:
## Proposed Solution
Update the component to explicitly render `<NotFound />` when no matching control panel is found.  
This makes the behavior consistent with the rest of Volto when navigating to invalid routes.

---

##  Benefits
- Provides clear feedback when a control panel does not exist  
- Improves UX and accessibility  
- Eliminates confusing blank screen behavior

---

### 🔗 Closes
Closes #7214

---

### 📋 Checklist
- [x] I signed and returned the Plone Contributor Agreement  
- [x] I verified there aren't other open pull requests for the same change  
- [x] I followed the guidelines in Contributing to Volto


wesleybl (changes requested):
The lint is failing. See:

https://6.docs.plone.org/volto/contributing/linting.html

Add an entry to the changelog. See:

https://6.docs.plone.org/contributing/index.html#change-log-entry

PR #7664, opened 2025-11-28
Fix order of installation for prod acceptance test commands

davisagli:
Fixes regression from https://github.com/plone/volto/pull/7646

PR #7665, opened 2025-11-29
Fixprodbuildin ci

sneridagh:


sneridagh:
Meh, @davisagli of course, if we have the prod build, we won't have Cypress or other build required tools in there :/

davisagli:
@sneridagh I guess you didn't see https://github.com/plone/volto/pull/7664 where I ran into a similar problem. I think we should just revert this back to how it was before, without `pnpm install --prod`

PR #7667, opened 2025-11-30
Moved hitorical amendment for the CssMinimizerPlugin to the Razzle fork

sneridagh:
and removed calc processing due to problems with some modern CSS.

### DIFF
--- a/.github/workflows/readme-link-check.yml
+++ b/.github/workflows/readme-link-check.yml
@@ -26,5 +26,6 @@ jobs:
           --exclude https://my-server-DNS-name.tld/api
           --exclude 2021.ploneconf.org
           --exclude https://www.lanku.eus/
+          --exclude https://medium.com/
           '**/README.md'
           'PACKAGES.md'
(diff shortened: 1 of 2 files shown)


packages/volto/razzle.config.js
```diff
@@ -116,26 +115,11 @@ const defaultModify = ({
...
-    // This is needed to override Razzle use of the unmaintained CleanCSS
-    // which does not have support for recently CSS features (container queries).
-    // Using the default provided (cssnano) by css-minimizer-webpack-plugin
-    // should be enough see:
-    // (https://github.com/clean-css/clean-css/discussions/1209)
     // TODO: remove this before merging the Razzle fork into Volto 19
+    // @sneridagh: Tried to remove this now that we have forked Razzle
+    // but I cannot pinpoint where this comes from, and it seems undefined
+    // always.
     delete options.webpackOptions.terserPluginOptions?.sourceMap;
```
sneridagh, line 122: @davisagli do you remember why this was needed for? I can't find the `sourceMap` setting in Razzle. Maybe @wesleybl you already removed it from the fork? I also debugged it, and the option is false both in dev and prod.

davisagli: @sneridagh I don't remember. It was added when we updated terser-webpack-plugin here: https://github.com/plone/volto/pull/4086/commits/84b85499b6876466d8814863aae4cc29e668ba31

PR #7677, opened 2025-12-05
Handle case with empty schema fieldsets in ModalForm

tedw87:
ModalForm component crashes when rendered with a schema that has an empty fieldsets array:
{
  "fieldsets": [],
  "properties": {},
  "required": []
}
ModalForm should handle empty schemas, rendering an empty form with just the submit and cancel buttons.

This issue can occur when adding a content-rule action that has no configuration form, the backend returns an empty schema for a content-rule action
For example, in the Rules configuration (/controlpanel/rules), when adding certain actions, the backend may return an empty schema causing the modal to crash instead of displaying.


wesleybl:
@tedw87 Could you please backport to the 18.x.x branch?

tedw87:
> @tedw87 Could you please backport to the 18.x.x branch?

https://github.com/plone/volto/pull/7687

PR #7692, opened 2025-12-11
fix: show only error message in toast instead of stringified JSON

Shyam-Raghuwanshi:
Now shows only error message in toast instead of stringified JSON

How to reproduce 
- go to - https://demo.plone.org/controlpanel/dexterity-types/Document/schema
- Try adding any new field using - Add field
- And fill only the required field - title
- Try to save

Before
image

After

image

- [X] I successfully ran [code linting checks](https://6.docs.plone.org/volto/contributing/linting.html) on my changes locally.
- [X] I successfully ran [unit tests](https://6.docs.plone.org/volto/contributing/testing.html) on my changes locally.
- [X] I successfully ran [acceptance tests](https://6.docs.plone.org/volto/contributing/acceptance-tests.html) on my changes locally.

### DIFF
--- a/packages/volto/src/components/theme/Login/Login.jsx
+++ b/packages/volto/src/components/theme/Login/Login.jsx
@@ -112,13 +112,11 @@ const Login = (props) => {
         toast.dismiss('loggedOut');
       }
       if (!toast.isActive('loginFailed')) {
-        const errMessage =
-          JSON.parse(error)?.error?.message || JSON.parse(error)?.message || '';
         toast.error(
           <Toast
             error
             title={intl.formatMessage(messages.loginFailed)}
-            content={errMessage}
+            content={intl.formatMessage(messages.loginFailedContent)}
           />,
           { autoClose: false, toastId: 'loginFailed' },
         );
--- a/packages/volto/src/reducers/userSession/userSession.js
+++ b/packages/volto/src/reducers/userSession/userSession.js
@@ -57,7 +57,7 @@ export default function userSession(state = initialState, action = {}) {
         login: {
           loading: false,
           loaded: false,
-          error: action.error.response.text,
+          error: action.error.response.error,
         },
       };
     case `${LOGOUT}_FAIL`:


wesleybl:
@Shyam-Raghuwanshi managed to simulate this error. Were you able to get past this error? Were you able to add a field? I filled in all the fields and it's still giving the error? Could it be a problem with the demo?

Shyam-Raghuwanshi:
Hi @wesleybl , actually, I never tried filling in all the fields. I was thinking this would work after we fill all the fields, but now I tried and found that it's a problem. We are unable to add a field and don't have any idea what the error is, but I will find out soon. If you have any idea about it, please share it with me too.

Shyam-Raghuwanshi:
Hi, @wesleybl, just fixed the problem. The problem was in the backend
Here is the PR https://github.com/plone/plone.restapi/pull/1969

davisagli:
@Shyam-Raghuwanshi No, the problem seems to be that Volto's SchemaWidget is setting incorrect factory names that don't match the available field types in the backend.

@robgietema Is it possible this is a side effect of the work you did for volto-form-block?
- Choosing the Text field is setting the factory to label_text_field but it should be Text
- There is a Full Name field available in the widget which doesn't match any field type available in a Plone content type schema

robgietema:
> @Shyam-Raghuwanshi No, the problem seems to be that Volto's SchemaWidget is setting incorrect factory names that don't match the available field types in the backend.
> 
> @robgietema Is it possible this is a side effect of the work you did for volto-form-block?
> 
> * Choosing the Text field is setting the factory to label_text_field but it should be Text
> * There is a Full Name field available in the widget which doesn't match any field type available in a Plone content type schema

That is possible but as far as I know I didn't change the factory fields and kept them the way they were. You can however now change the available fields (with the factories) in the config.

Shyam-Raghuwanshi:
Hi @davisagli, I've investigated the "Full Name" issue you mentioned. I found that:

The SchemaWidget gets field types from the backend's Fields vocabulary
"Full Name" is meant to be added only via additionalFactory prop (for volto-form-block)
It shouldn't appear in the content type schema editor by default
To fix this in Volto, should I:

Option A: Add a filterFactory prop to ContentTypeSchema to explicitly exclude "Full Name" and other form-block-only fields?
Option B: Find where "Full Name" is being incorrectly added and fix it at the source?
Could you clarify where "Full Name" is coming from in your testing?

packages/volto/src/components/manage/Controlpanels/ContentTypeSchema.jsx
```diff
@@ -169,7 +169,7 @@ class ContentTypeSchema extends Component {
           error
           title={this.props.intl.formatMessage(messages.error)}
           content={JSON.stringify(
-            nextProps.schemaRequest.put.error.response.body ||
+            nextProps.schemaRequest.put.error.response.body.message ||
```
davisagli: Be more careful in case the response isn't in the expected format:
```suggestion
            nextProps.schemaRequest.put.error.response.body?.message ||
```

packages/volto/src/components/manage/Controlpanels/ContentTypeSchema.jsx
```diff
@@ -311,7 +311,7 @@ class ContentTypeSchema extends Component {
                       />
                     </Button>
                     <Button
-                      className="cancel"
+                      className="Cancel"
```
davisagli: Don't change the class name, it's a breaking change.

```suggestion
                      className="cancel"
```

packages/volto/src/components/theme/Login/Login.jsx
davisagli, line 1: Please remove the changes in this file, it's not related to the issue described in the PR description.

packages/volto/src/reducers/userSession/userSession.js
davisagli, line 1: Same here, don't include unrelated changes.

davisagli (approved):

wesleybl:
@Shyam-Raghuwanshi Could you please do a backport to Volto 18?

PR #7693, opened 2025-12-12
Remove references to @plone/providers and @plone/helpers packages that don't exist in Volto 18 branch

wesleybl:
See: https://github.com/plone/volto/tree/18.x.x/packages

They exist in the master: https://github.com/plone/volto/tree/main/packages

### DIFF
--- a/packages/volto/razzle.config.js
+++ b/packages/volto/razzle.config.js
@@ -408,6 +408,7 @@ const defaultModify = ({
               /^@plone\/volto/,
               /^@plone\/components/,
               /^@plone\/client/,
+              /^@plone\/providers/,
             ].filter(Boolean),
           }),
         ]


wesleybl:
@stevepiercy Could you take a look here, please?

packages/volto/razzle.config.js
```diff
@@ -408,7 +408,6 @@ const defaultModify = ({
               /^@plone\/volto/,
               /^@plone\/components/,
               /^@plone\/client/,
-              /^@plone\/providers/,
```
sneridagh, line 411: I wouldn't remove this one.

PACKAGES.md
```diff
@@ -50,8 +50,6 @@ The bundle of these packages must work on both CommonJS and ECMAScript Module (E
 
 ## Utilities packages
 
--   `@plone/providers`
--   `@plone/helpers`
 -   `@plone/react-router`
```
sneridagh: react-router is not there also.

stevepiercy: I also don't see any of the Add-on packages below on lines 63-69. @sneridagh should that section be removed? Sorry, my Volto 18 is very bad.

stevepiercy: Additionally, I see packages in 18.x.x, but not documented:

- eslintconfig
- tsconfig
- volto-guillotina
- volto-testing

packages/volto/razzle.config.js
wesleybl: Is there a non-alpha version of `@plone/providers` compatible with Volto 18? Could someone be using it?

PACKAGES.md
wesleybl: > I also don't see any of the Add-on packages below on lines 63-69. @sneridagh should that section be removed? Sorry, my Volto 18 is very bad.

@stevepiercy I think this would have to be dealt with in the main first.

wesleybl: > Additionally, I see packages in 18.x.x, but not documented:
>
>- eslintconfig
>- tsconfig
>- volto-guillotina
>- volto-testing

@stevepiercy I think those are support packages, and they don't need to be listed there.

wesleybl: > react-router is not there also.

@sneridagh Removed.

packages/volto/razzle.config.js
wesleybl: In any case, this is the server-side compilation of the bundle. It would be in node_molules. Looking at the package code, I don't see the need for it to be in the bundle.

sneridagh: This setting does not do what you think it does, and it does not hurt to have it in there. If the package is there, it will do what it does, it does not, it won't.

sneridagh (approved):
LGTM, except the removal of `@plone/providers` from the razzle config. That setting is a "just in case" someone decided to use it in their build, otherwise it won't work (then it's a breaking).

packages/volto/razzle.config.js
wesleybl: I think we could remove that, and it wouldn't affect who uses `@plone/providers`. `@plone/providers` wouldn't need to be in the bundle. Its distribution already comes transpiled.

In any case, I'll reverse this so we can finish.

wesleybl: @sneridagh done.

PR #7722, opened 2025-12-18
Exclude linkedin from link checks

davisagli:
(It's returning HTTP 429 Too Many Requests responses)

### DIFF
--- a/.github/workflows/readme-link-check.yml
+++ b/.github/workflows/readme-link-check.yml
@@ -27,6 +27,7 @@ jobs:
           --exclude 2021.ploneconf.org
           --exclude https://www.lanku.eus/
           --exclude https://medium.com/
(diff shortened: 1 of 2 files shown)


stevepiercy:
I'll explore other things that lychee recommends in https://lychee.cli.rs/troubleshooting/rate-limits/. This one is more of a last resort. Locally I don't get 429 responses.

stevepiercy:
Meh. I don't want to reduce concurrency to less than the 3 linkedin.com links, nor do I want to accept 429 responses. Since the context for these links is a vendored in readme of contributors, it's not worth compromising performance or the simple cofiguration. Sledgehammer of exclude is the best option here. HULK SMASH!

stevepiercy:
@davisagli thank you!

PR #7730, opened 2025-12-18
fix: prevent blocksClipboard persistence for anonymous users

### LINKED ISSUE #6997: Empty localstorage entry created even for anonymous users
**Describe the bug**

For GDPR purposes, it would be best if no localstorage entries are created for anonymous users

**To Reproduce**

1. Go to https://demo.plone.org/
2. Open developer inspector

**Expected behavior**

Nothing is stored on the anonymous user's browser

**Screenshots**

Image

pratyush07-hub:
This PR fixes an issue where Volto was creating localStorage entries even for anonymous users.
The problem comes from the blocksClipboard reducer being configured as persistent. Even though anonymous users can’t copy or paste blocks, Volto would still create an empty localStorage entry, which isn’t really needed and can be a concern from a privacy (GDPR) point of view.
With this change, Redux state is only saved to localStorage after a user is logged in. For anonymous visitors, nothing is stored.

-----

This PR fixes the issue by:
1. **Check if the user is logged in**
   We only act when the user has a valid session (userSession.token).
2. **Save data only for logged-in users**
   The clipboard and other persistent data are stored in the browser only when someone is logged in; anonymous  visitors are skipped.
3. **Keep browsers clean and private**
   Anonymous users no longer get empty localStorage entries, improving privacy and GDPR compliance, while editors’ behavior stays the same.

**Demonstration**

[Screencast from 2026-02-05 18-05-32.webm](https://github.com/user-attachments/assets/25f7b1e3-9898-44a8-a369-548ebbba4e96)

-----

-----

Closes #6997 

-----


pratyush07-hub:
@wesleybl  The issue is closed as completed, and this PR implements the suggested fix.
I’m happy to update the implementation if you’d prefer a different approach.

pratyush07-hub:
> This is assuming that all persistentReducers should only run if the user is logged in. That's not a valid assumption. It would be ok for the blocks clipboard, since only logged in users can edit blocks. But there can be custom persistentReducers added for a particular project or addon, and we can't assume they are all for logged in users only. It would be a potentially breaking change.

@davisagli You’re right, my assumption was too broad. Would it be okay to refactor this using a more granular approach, so that `blocksClipboard` only persists for authenticated users while keeping the default behavior unchanged for everything else?

davisagli:
@pratyush07-hub Yeah -- I'm not sure how exactly to do that, but that is the goal.

It would also be ok to change the blocks clipboard so that it is implemented in a different way instead of using persistentReducers in the Redux store, if that makes it easier to implement.

pratyush07-hub:
@davisagli  I have updated the implementation to be granular as suggested `blocksClipboard` is now treated as a special "auth-only" reducer that only persists for logged-in users, while all other persistent reducers continue to work unconditionally to avoid breaking changes. I have updated the tests to verify this behavior and added a demo video in PR description.

davisagli:
I merged https://github.com/plone/volto/pull/7925 instead. Thanks for working on this.

PR #7739, opened 2025-12-20
Fix #7708: detailed viewableInBrowserObjects support and semantic-ui-…

### LINKED ISSUE #7708: `viewableInBrowserObjects` configuration not working for links in text blocks
**Describe the bug**
The [viewableInBrowserObjects](https://github.com/plone/volto/blob/ede0335834445988dd0639ab2361180251c97e4e/packages/volto/src/config/index.js#L113) configuration does not work correctly for links created in text blocks. Even with the configuration set to open files in the browser, links created in text blocks continue to generate download URLs (@@download/file) instead of display URLs (@@display-file/file), causing files to be downloaded instead of opened in the browser when the user is anonymous.

**To Reproduce**

1. Configure `viewableInBrowserObjects: ['File']` in the Volto configuration file.
2. Create a file.
3. Create a page with a text block.
4. In the text block, create a link that points to the created file.
5. Access the page as an anonymous user.
6. Click on the file link
7. Observe that the file is downloaded instead of being opened in the browser.

**Expected behavior**
When `viewableInBrowserObjects` is configured with `['File']`, links to files should open in the browser instead of being downloaded, especially when created in text blocks. The generated URL should be `@@display-file/file` instead of `@@download/file` for anonymous users.

- OS: Linux
- Browser: chrome
- Volto Version: 19
- Plone Version 6.1.3

**Additional context**
In the listing block, this setting works. It seems the setting was designed for the listing block only.

antcybersec:
Fixes #7708

I've addressed the issue where `viewableInBrowserObjects` wasn't working for links inside Slate text blocks. It turns out the link renderer was bypassing the configuration check completely. I updated the [ViewLink](cci:1://file:///d:/plonevolto/packages/volto-slate/src/editor/plugins/Link/render.jsx:7:0-31:2) component in `volto-slate` to correctly check if the user is anonymous and if the configuration allows files to be viewed in the browser. If so, it now sets the URL to `@@display-file`, so it opens properly instead of forcing a download.

While I was at it, I also fixed a build error I hit on Windows with pnpm. The webpack LESS configuration wasn't finding `semantic-ui-less` because it was looking for hardcoded paths that didn't match the pnpm symlinks on Windows. I updated it to use `require.resolve` and better regexes so the build works reliably across different setups.

-----

Closes #7708

### DIFF
--- a/packages/volto/src/components/manage/UniversalLink/UniversalLink.tsx
+++ b/packages/volto/src/components/manage/UniversalLink/UniversalLink.tsx
@@ -68,7 +68,19 @@ export function getUrl(
   children: React.ReactNode,
 ): string {
   if ('href' in props && typeof props.href === 'string') {
-    return props.href;
+    let url = props.href;
+    // If user is anonymous and viewableInBrowserObjects is configured,
+    // replace @@download with @@display-file for href-based links
+    // This handles cases where we don't have the item (content type) available
+    if (
+      !token &&
+      config.settings.viewableInBrowserObjects &&
+      config.settings.viewableInBrowserObjects.length > 0 &&
+      url.includes('@@download')
+    ) {
+      url = url.replace('@@download', '@@display-file');
+    }
+    return url;
   }
 
   if (!item || item['@id'] === '') return config.settings.publicURL;
--- a/packages/volto-slate/src/editor/plugins/Link/render.jsx
+++ b/packages/volto-slate/src/editor/plugins/Link/render.jsx
@@ -1,5 +1,4 @@
 import React from 'react';
-import { useSelector } from 'react-redux';
 import cx from 'classnames';
 import UniversalLink from '@plone/volto/components/manage/UniversalLink/UniversalLink';
 import config from '@plone/volto/registry';
@@ -7,22 +6,12 @@ import { isInternalURL, flattenToAppURL } from '@plone/volto/helpers/Url/Url';
 
 const ViewLink = ({ url, target, download, children }) => {
   const { openExternalLinkInNewTab } = config.settings;
-  const token = useSelector((state) => state.userSession?.token);
-  let href = url;
-
-  if (
-    !token &&
-    config.settings.viewableInBrowserObjects?.includes('File') &&
-    url.includes('/@@download/file')
-  ) {
-    href = url.replace('/@@download/file', '/@@display-file/file');
-  }
 
   return (
     <UniversalLink
-      href={href}
+      href={url}
       openLinkInNewTab={
-        (openExternalLinkInNewTab && !isInternalURL(href)) || target === '_blank'
+        (openExternalLinkInNewTab && !isInternalURL(url)) || target === '_blank'
       }
       download={download}
     >
--- a/packages/volto/news/7708.bugfix
+++ b/packages/volto/news/7708.bugfix
@@ -1,0 +1,1 @@
+Fix viewableInBrowserObjects configuration for links in Slate text blocks by centralizing URL rewriting logic in UniversalLink component.


davisagli (changes requested):
This is not the right way to fix this:
1. We keep the logic for rewriting URLs centralized in the UniversalLink component, so that it works the same everywhere we link to content.
2. This assumes that the content type is a File if the URL includes /@@download-file, but it is possible to have File URLs that don't include that suffix yet, and it is possible to have other content types that also use the download view. In order to apply the setting correctly, we need to actually have the content type of the target item, but it is not currently available in the block data.

wesleybl:
Although it's not the same situation, I think #7816 would solve the problem of #7708. The download URL would not be defined by the type, but by the mime type. But I don't know if the mime type would be available in the Slate block.

davisagli:
@wesleybl It's the same challenge. We don't have the content type or the mimetype here unless we change how links get serialized.

wesleybl:
@davisagli It seems that `UniversalLink` only checks `viewableInBrowserObjects` if `item` is passed, but Slate doesn't pass `item`. Could we possibly pass the item?

Maybe we shouldn't destructure the object here: https://github.com/plone/volto/blob/main/packages/volto-slate/src/editor/plugins/Link/render.jsx#L7

wesleybl:
> Maybe we shouldn't destructure the object here: https://github.com/plone/volto/blob/main/packages/volto-slate/src/editor/plugins/Link/render.jsx#L7

Hmm, this doesn't seem to be the content item. We would have to look for where we accessed the content item.

wesleybl:
> It seems that UniversalLink only checks viewableInBrowserObjects if item is passed, but Slate doesn't pass item. Could we possibly pass the item?

Actually, we don't have the corresponding item for the File here. This would have to be implemented on the backend. See:

https://github.com/plone/volto/issues/7708#issuecomment-3660552440

wesleybl:
@davisagli Could we have a configuration that replaces `@@download` with `@@display-file`, regardless of the content type?

davisagli:
I don't think we should try to replace @@download with @@display-file in the frontend where we don't have enough information.

In the backend, the /@@download part of the URL is added when serializing a file field here: https://github.com/plone/plone.restapi/blob/bc7fc79d447d625fe90696285b6b425a83a760be/src/plone/restapi/serializer/dxfields.py#L209

We should probably use /@@display-file there instead. Then when this view is actually accessed, plone.namedfile determines whether it should be downloaded or displayed using this list: https://github.com/plone/plone.namedfile/blob/main/src/plone/namedfile/browser.py#L16

wesleybl:
> Then when this view is actually accessed, plone.namedfile determines whether it should be downloaded or displayed using this list: https://github.com/plone/plone.namedfile/blob/main/src/plone/namedfile/browser.py#L16

@davisagli This list only exists in the backend. For the slate block, we could use this list. In the listing block, the definition is done by `viewableInBrowserObjects`, which is based on content type:

https://github.com/plone/volto/blob/main/packages/volto/src/components/manage/UniversalLink/UniversalLink.tsx#L101-L107

So, would `viewableInBrowserObjects` no longer be used? Currently, the listing result doesn't include the URL to be used for downloading/viewing. It would have to start including that information.

packages/volto/src/components/manage/UniversalLink/UniversalLink.tsx
```diff
@@ -68,7 +68,19 @@ export function getUrl(
...
-    return props.href;
+    let url = props.href;
+    // If user is anonymous and viewableInBrowserObjects is configured,
+    // replace @@download with @@display-file for href-based links
+    // This handles cases where we don't have the item (content type) available
+    if (
+      !token &&
+      config.settings.viewableInBrowserObjects &&
+      config.settings.viewableInBrowserObjects.length > 0 &&
+      url.includes('@@download')
```
wesleybl, line 79: @antcybersec We need another boolean setting to decide whether to replace `@@download`. This new setting should also be used when passing `item`. See:

https://github.com/plone/volto/pull/7816#issuecomment-3927841084
https://github.com/plone/volto/pull/7816#issuecomment-3931923503

PR #7751, opened 2025-12-24
fixing not able to add availabe languages issue

### LINKED ISSUE #7720: It is not possible to add a language when the primary language is English
**Describe the bug**

When the portal's default language is English, we are unable to add a new language. The following error occurs:

`The selected default language must be in the list of the field 'Available languages'`

**To Reproduce**

1. Create a portal with English as the default language.
2. Go to `Site Setup` -> `Language`.
3. In `Available languages`, add a language.
4. Save

**Expected behavior**

I hope it will be possible to save the form in the new language.

- OS: Ubuntu
- Browser: Chrome
- Volto Version: 19
- Plone Version: 6.1.3
- Plone REST API: 9.x

**Additional context**
This can be reproduced at https://volto.demo.plone.org/

Shyam-Raghuwanshi:
Closes #7720

- [X] I successfully ran [code linting checks](https://6.docs.plone.org/volto/contributing/linting.html) on my changes locally.
- [X] I successfully ran [unit tests](https://6.docs.plone.org/volto/contributing/testing.html) on my changes locally.
- [X] I successfully ran [acceptance tests](https://6.docs.plone.org/volto/contributing/acceptance-tests.html) on my changes locally.

### DIFF
--- a/packages/volto/src/actions/controlpanels/controlpanels.js
+++ b/packages/volto/src/actions/controlpanels/controlpanels.js
@@ -91,6 +91,7 @@ export function updateControlpanel(url, data) {
   return (dispatch) => {
     const normalizedData = { ...data };
     if (
+      url.includes('@controlpanels/language') &&
       normalizedData.default_language &&
       Array.isArray(normalizedData.available_languages)
     ) {
@@ -99,7 +100,6 @@ export function updateControlpanel(url, data) {
           ? normalizedData.default_language
           : normalizedData.default_language?.token ||
             normalizedData.default_language?.value;
-
       const isDefaultInAvailable = normalizedData.available_languages.some(
         (lang) => {
           const langCode =
@@ -109,9 +109,18 @@ export function updateControlpanel(url, data) {
       );
 
       if (!isDefaultInAvailable && defaultLangCode) {
+        // Preserve the existing format (string vs object) when appending
+        const firstItem = normalizedData.available_languages[0];
+        const isObjectFormat =
+          firstItem && typeof firstItem === 'object' && firstItem !== null;
+
+        const newLangEntry = isObjectFormat
+          ? { token: defaultLangCode }
+          : defaultLangCode;
+
         normalizedData.available_languages = [
           ...normalizedData.available_languages,
-          defaultLangCode,
+          newLangEntry,
         ];
       }
     }


packages/volto/src/actions/controlpanels/controlpanels.js
```diff
@@ -89,12 +89,25 @@ export function listControlpanels() {
...
 export function updateControlpanel(url, data) {
   return (dispatch) => {
+    const normalizedData = { ...data };
+    if (
+      normalizedData.default_language &&
+      Array.isArray(normalizedData.available_languages) &&
+      !normalizedData.available_languages.includes(
+        normalizedData.default_language,
+      )
+    ) {
```
aryan7081, line 97: It would be great if we could check whether the default language exists before adding to avoid duplicates. The current state will create duplicates if default_language already exists in available languages.

packages/volto/src/actions/controlpanels/controlpanels.js
```diff
@@ -89,12 +89,25 @@ export function listControlpanels() {
...
+      normalizedData.available_languages = [
+        ...normalizedData.available_languages,
+        normalizedData.default_language,
```
aryan7081: I think the backend can return languages as strings or objects `"en"` vs `{token: "en"}`. Maybe we should normalise the token before comparing to avoid edge cases.

Shyam-Raghuwanshi: Yes, you are right

packages/volto/src/actions/controlpanels/controlpanels.js
```diff
@@ -89,12 +89,39 @@ export function listControlpanels() {
  */
 export function updateControlpanel(url, data) {
   return (dispatch) => {
+    const normalizedData = { ...data };
+    if (
```
aryan7081, line 93: Also, the current logic runs for all control panels. Can you please guard it with a URL check? eg. `url.includes(@controlpanels/language)` so it only applies to the language control panel and avoids unintended side effects.

packages/volto/src/actions/controlpanels/controlpanels.js
```diff
@@ -89,12 +89,39 @@ export function listControlpanels() {
...
+      const isDefaultInAvailable = normalizedData.available_languages.some(
+        (lang) => {
+          const langCode =
+            typeof lang === 'string' ? lang : lang?.token || lang?.value;
+          return langCode === defaultLangCode;
+        },
+      );
+
+      if (!isDefaultInAvailable && defaultLangCode) {
+        normalizedData.available_languages = [
```
aryan7081, line 121: Here while adding the default language, it always pushes a string. If `available_languages` contains objects, this results in a mixed format. We need to preserve the existing format when appending the default language.

aryan7081:
@Shyam-Raghuwanshi Volto uses Cypress for acceptance testing. Can you please add that as well? Please see: https://6.docs.plone.org/volto/contributing/acceptance-tests.html

PR #7752, opened 2025-12-25
Support setting VALEFILES as an environment variable

stevepiercy:
- This allows checking a single file, which is useful when testing new terms or only the file you're editing.

Usage:

```shell
export VALEFILES=path/to/file.md && make docs-vale
```

To clear the environment variable, use the following command:

```shell
unset VALEFILES
```


stevepiercy:
nm, there's a better way.

PR #7756, opened 2025-12-29
Docs: Add deprecation warning for src/config.js (Volto 18+)

gummybearansh:
## Description
Updates the configuration registry documentation to reflect that the project configuration approach using `src/config.js` is deprecated in Volto 18+ in favor of the `config/` folder structure.

## Related Issue
Closes #7601

## Changes
- Added a `{warning}` block to `docs/source/configuration/how-to.md` regarding the deprecation.
- Added a link to the Volto Customization training for the new `config/` approach.
- Clarified that `src/config.js` is still valid for older projects.

----
📚 Documentation preview 📚: https://volto--7756.org.readthedocs.build/

PR #7758, opened 2025-12-29
Prevent VSCode from mangling the `html_meta` header of documentation …

stevepiercy:
…files in Volto.

This happens too often, and I'm tired of reminding contributors. We have this in the upstream `plone/documentation` repo.

See https://github.com/plone/volto/pull/7756/files#r2651550447

### DIFF
--- a/packages/volto/news/7758.internal
+++ b/packages/volto/news/7758.internal
@@ -1,0 +1,1 @@
+Prevent VSCode from mangling the `html_meta` header of documentation files. @stevepiercy


stevepiercy:
After merge, there was a readme link check failure. Opened a PR at https://github.com/collective/awesome-volto/pull/53 and notified 5 Eau de Web folks.

sneridagh:
@stevepiercy @davisagli we already were setting this in https://github.com/plone/volto/blob/c3e5be1db26f7865e3777cca4210a21bfd39f4b2/packages/scripts/vscodesettings.js

So every time we `make install` we not only setup this setting, but others.
I'd like to maintain it as it was. I am aware that people not doing `make install` will not apply this. Hard to control if people are not using vscode too (cursor, etc... ) 

However, I see value in not customize it, so people can have their own config (if needed), in addition/amending to the "recommended config". Would you be ok if we revert it?

stevepiercy:
@sneridagh how can we get both? I wouldn't be happy with a reversion. Can we put the config at the root of repo and remove it from `make install`?

stevepiercy:
Oh, maybe there is a way to have both:

https://github.com/microsoft/vscode/issues/40233

pnicolli:
Multi-root workspaces are not something you can set up via `make install` afaik, it's a user choice, I've been using those ever since they exist to have my own set of settings, maybe shared among projects, but afaik it can't be setup for you by a script, would be happy to know I'm wrong.
In my personal experience, the moment you check into git a vscode settings file you will stop reviewing PRs with changes to the markdown and start reviewing PRs with random colors changed in the vscode settings file.

stevepiercy:
Doesn't the description in that issue handle the situation?

> So there would be three locations for settings (for a single-folder workspace):
>
> - user settings
> - `${workspaceRoot}/.vscode/settings.json`  # includes some JS to check for existence and inclusion of the next file
> - `${workspaceRoot}/.vscode/settings.local.json`

...then add `.vscode/settings.local.json` to `.gitignore` for Volto. I don't use VSCode, so I can't try it out.

If not, there *must* be some better way to keep everyone happy.

I would think it's easier to automate and detect changes in a single settings file than dozens of documentation files, too, similar to how we detect changes in docs files to trigger a pull request preview build on RTD using a diff command.

Anyway, I'd like to try this change to see how bad the problem is with customizations to settings file. Right now, the reformatting of docs files causes me considerable headaches when I have to look through sometimes hundreds of whitespace-only changes that have nothing to do with the PR.

---

Finally it appears this has made it into Microsoft's backlog: https://github.com/microsoft/vscode/issues/282806

pnicolli:
> Doesn't the description in that issue handle the situation?
> 
> > So there would be three locations for settings (for a single-folder workspace):
> > 
> > * user settings
> > * `${workspaceRoot}/.vscode/settings.json`  # includes some JS to check for existence and inclusion of the next file
> > * `${workspaceRoot}/.vscode/settings.local.json`

Afaik this in description is the wish the OP has in that issue, not the current situation; the settings.local.json file is not a thing.

> Finally it appears this has made it into Microsoft's backlog: [microsoft/vscode#282806](https://github.com/microsoft/vscode/issues/282806)

This is awesome news!

stevepiercy:
My point was, can you use JS to check for the existence of a settings file and load it? We do this for local Makefiles, why not VSCode settings?

sneridagh:
@stevepiercy I am afraid we cannot. As you said, the feature that we are looking for, is in the community wish list for long time ago... :( 

I am ok on including this by default, it's a pity that we cannot have that feature yet T_T

wesleybl:
@stevepiercy @sneridagh I also think we shouldn't commit that file. Every time I run `make install` the file is modified and git sees that.

Those with custom configurations are also left in the lurch.

PR #7770, opened 2026-01-08
Fix toolbar button contrast by removing opacity

### LINKED ISSUE #7765: a11y - toolbar icons - lack of contrast - cms ui
Describe the bug
This issue is related to insufficient color contrast that does not pass accessibility contrast requirements. All toolbar buttons, when the user is inside the CMS UI in edit mode, have an opacity value set to 0.8. Because of this opacity, the visual contrast between the buttons and the background is reduced and falls outside the accepted accessibility standards.

To reproduce
1 Go to https://demo.plone.org/edit
2 Log in to the CMS
3 Enter edit mode so that the CMS UI toolbar is visible
4 Observe the left side toolbar
5 Notice that all toolbar buttons, not only the save or cancel buttons, have a CSS class that applies opacity 0.8

Expected behavior
All toolbar buttons should have sufficient contrast to meet accessibility guidelines, without using opacity values that reduce readability and visual clarity.

Actual behavior
Every toolbar button uses opacity 0.8, which lowers contrast and causes the buttons to fail accessibility contrast tests.

**Screenshots**

Image Image

Example:
Image

- Volto Version 18.27.3

Saniya-r-Shaikh:
### Description

Fixes an accessibility contrast issue in the CMS UI toolbar.

All toolbar buttons were using `opacity: 0.8`, which reduced contrast and caused them to fail accessibility contrast requirements.  
This change removes opacity usage so buttons retain full contrast and readability.

This aligns toolbar buttons with accessibility contrast guidelines by avoiding opacity-based contrast reduction.

### Notes

- CSS-only change
- No functional behavior changed
- Tested manually on https://demo.plone.org/edit
- Volto 18.27.3

Closes #7765


changelog.d/feature/fix-toolbar-contrast.yaml
```diff
@@ -0,0 +1,5 @@
+---
+'volto':
+  - Fix toolbar button contrast by removing opacity (#7765)
```
wesleybl, line 3: The changelog is in the wrong place and has the wrong format. See: https://6.docs.plone.org/contributing/index.html#create-a-news-item-file

packages/volto/theme/themes/pastanaga/extras/toolbar.less
```diff
@@ -201,8 +201,8 @@ body:not(.has-sidebar):not(.has-sidebar-collapsed) {
       flex-direction: row-reverse;
       align-items: center;
       padding: 0 20px;
-      opacity: 0;
-      transition: @opacity-transition;
+      opacity: 1;
```
wesleybl, line 204: I think this is the only change needed in these files. The others are not necessary.

wesleybl (changes requested):

PR #7785, opened 2026-01-12
Remove `apps` folder from main

sneridagh:
I had this in mind for a long time. They are unmaintained and not documented properly. After the injection attempt incident, I think it's better not to leave so much surface unattended.

----
📚 Documentation preview 📚: https://volto--7785.org.readthedocs.build/


pnicolli:
Fully agree

PR #7793, opened 2026-01-15
Use `es-toolkit` instead of `lodash`

### LINKED ISSUE #7789: Use `es-toolkit` as drop-in replacement of `lodash`
> [!IMPORTANT]
> If you are not a member of the Volto Team or Developers Team in the Plone GitHub organization, then do not work on or comment on this issue.

## Abstract

Move to use `es-toolkit` instead of `lodash`.

## Motivation

`lodash` is deprecated in Volto 18 (see https://github.com/plone/volto/issues/6350), and it's unmaintained and it bloats builds. Some of their utilities are already part of the ESNext specifications and needed anymore.

See https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore for more information.

## Proposal & Implementation

Modify all imports and usage from the remaining `lodash` utilities and move to use the `es-toolkit` ones.

## Risks

Add-ons and projects will have to update to use `es-toolkit` or add the `lodash` dependency locally.
It will be recommended that everybody adapts their code to not use `lodash`.

sneridagh:


davisagli (commented):
The basic idea looks all right to me. I didn't do a line by line review.

Would this be for Volto 19 or a future major version?

We should mention it in the upgrade guide, otherwise add-ons will continue to use lodash and both will end up in the bundle.

sneridagh:
@davisagli the build increased 20k, for some reason. We would need to continue with it:
- Move to use "native" es-toolkit utilities, the ones really improved.
- Move map/filter/etc... to vanilla JS counterparts.

This has to be an ongoing effort, TBH I don't see it happening for 19 given the amount of work it entails. This first step was easy, the complex is the next one.

davisagli:
I talked to @sneridagh, we'll close it since it didn't turn out to be an easy win.

PR #7803, opened 2026-01-21
External link classname is missing for mailto links

Tishasoumya-02:
Add 'mail' classname for mailto links

### DIFF
--- a/packages/volto-slate/src/editor/plugins/Link/render.jsx
+++ b/packages/volto-slate/src/editor/plugins/Link/render.jsx
@@ -1,11 +1,8 @@
+import React from 'react';
 import cx from 'classnames';
 import UniversalLink from '@plone/volto/components/manage/UniversalLink/UniversalLink';
 import config from '@plone/volto/registry';
-import {
-  isInternalURL,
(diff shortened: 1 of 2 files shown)


packages/volto/src/components/theme/Widgets/__snapshots__/EmailWidget.test.jsx.snap
```diff
@@ -12,7 +12,7 @@ exports[`EmailWidget > renders an email view widget component 1`] = `
 
 exports[`EmailWidget > renders an email view widget component with children 1`] = `
 <a
-  className="metadata email widget"
```
davisagli, line 15: @Tishasoumya-02 It's a breaking change to remove the `email` class that was there before.

I don't really understand, why is the new class necessary? Can't you style it based on the `email` class or use `[href^="mailto:"]`?

Tishasoumya-02: yes,I missed it. Will use [href^="mailto:"], so this PR is of no use now. I'm going to close it.

PR #7810, opened 2026-01-23
Upgrade to Plone 6.2a1

sneridagh:


davisagli:
@sneridagh Did you have a chance to check what is failing in the acceptance tests?

sneridagh:
@davisagli not in depth, a 404 accessing a content in the search block (??)

wesleybl:
@davisagli The failing test performs a search using `@querystring-search` passing:

```
"sort_on":"effective"
"sort_order":"descending"
```

What's different is the sorting (even though the test contents haven't been published).

This difference seems to be related to: https://github.com/plone/plone.app.querystring/pull/174

Can we consider your change correct and simply modify the test?

wesleybl:
@davisagli Well, the catalog should contain the `sort_on` index in Plone 6.1 and 6.2. So perhaps it's not your modification. Maybe it's some change in how Zope sorts indexes with the same value, since the contents haven't been published?

davisagli:
@wesleybl Probably because of https://github.com/plone/plone.restapi/pull/1954. Without this fix, the sort_order from the query was lost. Then we just have to update the test to expect the new, correct order.

wesleybl:
> Probably because of https://github.com/plone/plone.restapi/pull/1954

@davisagli It's not because of that, because it's not in any release yet. I think it's because of https://github.com/zopefoundation/Products.ZCatalog/pull/159. The test isn't very smart. It sorts by "effective," but the content isn't published. So "effective" is "None." Ideally, the test should be modified to publish at least one piece of content.

The error occurs here: https://github.com/plone/volto/blob/main/packages/volto/cypress/tests/core/blocks/blocks-search.js#L543

wesleybl:
Is it just me, or is GitHub not considering the anchor for the line in the links? lol

davisagli:
@wesleybl A sort where all the values are None would be a stable sort that keeps the existing order of the items. So it probably ends up with the order being something that is determined incidentally based on the catalog query plan. Yeah, it's possible that https://github.com/zopefoundation/Products.ZCatalog/pull/159 is the reason

wesleybl:
@davisagli @sneridagh I fixed the tests.

wesleybl:
It is necessary to update cookieplone now.

PR #7827, opened 2026-01-28
Use Playwright

sneridagh:


sneridagh:
@pnicolli I merged it, if we need to change strategy, we can revert later.

PR #6435, opened 2024-10-24
build the registration from using the userschema provided by backend

erral:
- [X] I signed and returned the [Plone Contributor Agreement](https://plone.org/foundation/contributors-agreement), and received and accepted an invitation to join a team in the Plone GitHub organization.
- [X] I verified there aren't other open [pull requests](https://github.com/plone/volto/pulls) for the same change.
- [X] I followed the guidelines in [Contributing to Volto](https://6.docs.plone.org/volto/contributing/index.html).
- [X] I succesfully ran [code linting checks](https://6.docs.plone.org/volto/contributing/linting.html) on my changes locally.
- [X] I succesfully ran [unit tests](https://6.docs.plone.org/volto/contributing/testing.html) on my changes locally.
- [ ] I succesfully ran [acceptance tests](https://6.docs.plone.org/volto/contributing/acceptance-tests.html) on my changes locally.
- [ ] If needed, I added new tests for my changes.
- [ ] If needed, I added [documentation](https://6.docs.plone.org/volto/contributing/documentation.html#narrative-documentation) for my changes, either in the Storybook or narrative documentation.
- [X] I included a [change log entry](https://6.docs.plone.org/contributing/index.html#contributing-change-log-label) in my commits.

-----

If your pull request closes an open issue, include the exact text below, immediately followed by the issue number. When your pull request gets merged, then that issue will close automatically.

Closes #6434

PR #7830, opened 2026-01-29
fix(register): include username when creating user

shivaansh0610-LUFFY:
### What this does

The registration form did not include the required `username` field
When creating a user, which caused the backend to reject the request.

This change adds `username` (derived from the email) to the payload
sent to the `createUser` action.

### Why this is needed

`plone.restapi` requires a `username` when creating users.
Without it, registration fails with a validation error.

### How to test

1. Go to the registration form
2. Submit fullname + email
3. Verify that user creation succeeds

Contributes to #6987


shivaansh0610-LUFFY:
CLA has been signed and processed. 
Happy to update anything else or rebase if needed.

davisagli (changes requested):
This is incomplete. username is not required or allowed if the "use email address as login name" setting is turned on: https://github.com/plone/plone.restapi/blob/main/src/plone/restapi/services/users/add.py#L69

erral:
Not only incomplete, I would even go further, the registration form should be built according to the userschema configuration we already have in the backend and provided by the REST API.

See: https://github.com/plone/plone.restapi/pull/1874

I even tried to add this for Volto: https://github.com/plone/volto/pull/6435 but stopped at some point. I think that if we are going to work on the registration form, we should resurrect the said PR.

shivaansh0610-LUFFY:
> Not only incomplete, I would even go further, the registration form should be built according to the userschema configuration we already have in the backend and provided by the REST API.
> 
> See: [plone/plone.restapi#1874](https://github.com/plone/plone.restapi/pull/1874)
> 
> I even tried to add this for Volto: #6435 but stopped at some point. I that if we are going to work on the registration form, we should resurrect the said PR.

Thanks for the suggestion.

I understand that ideally, the registration form should be driven by the userschema exposed by the backend.

For this PR, I focused on addressing the immediate validation issue related to `username` and `use_email_as_login`.

Would you prefer expanding this PR to implement a schema-driven registration form, or should that be handled separately?

shivaansh0610-LUFFY:
@erral 
I see that ideally, the registration form should be schema-driven from the backend.
Should this PR be limited to fixing the username/email-as-login case, and handle the schema-driven refactor separately?
I would be very thankful for your guidance.

davisagli:
@erral Thanks for the reminder about your work to have a schema-driven registration form.

I would love to have that, but @shivaansh0610-LUFFY it's also helpful to have a more focused PR that fixes the existing Register component to pay attention to the use_email_as_login setting.

Note: we probably have to first add the value of use_email_as_login to the return value from the `@site` endpoint in plone.restapi. Then the frontend can show the right thing based on how the setting is configured for a specific Plone site, instead of using a Volto config setting (which can get out of sync with the actual backend configuration).

PR #7834, opened 2026-01-29
refactor: move vscode settings setup to pnpm postinstall hook

wesleybl:
- Remove vscodesettings.js execution from make install targets
- Add .vscode/settings.json to .gitignore
- Remove .vscode/settings.json from git tracking
- Add postinstall hook to package.json to run vscodesettings.js
- Fix script to properly check if settings.json exists before reading

Therefore, when we run `make install`, `settings.json` will not appear as modified in git, and the settings will be added to the file. They will also be added when only `pnpm install` is run.

### DIFF
--- a/packages/scripts/news/7834.internal
+++ b/packages/scripts/news/7834.internal
@@ -1,1 +1,1 @@
-Fix vscodesettings.js to properly check if settings.json exists before reading. @wesleybl
+Fix the {file}`vscodesettings.js` script to properly check if {file}`.vscode/settings.json` exists before reading. @wesleybl


wesleybl:
See: https://github.com/plone/volto/pull/7758#issuecomment-3722907742

sneridagh:
@wesleybl I still haven't taken action on this one, because I'm still not sure about commiting the `.vscode/settings.json`. I'd like to bring this up in the next meeting.

My take: leave it as it was. 

If we decide otherwise, then the `postinstall` hook is unnecessary and running that script is useless, and should be removed.

wesleybl:
@sneridagh I think it's better to leave it in the script. Committing `settings.json` can cause problems in the future if a developer wants to make local configurations that other developers won't want.

I find it hard to believe that someone would commit without at least running `pnpm install`.

stevepiercy:
> I find it hard to believe that someone would commit without at least running `pnpm install`.

Can you clarify this scenario?

I often want to make a change only to Volto documentation. I don't need to run `pnpm install`, but I will still make a commit.

sneridagh:
@stevepiercy I don't see the point in narrow the fix to VSCode either... specially today that the IDE scenario is so broad... If we were to be coherent, we would have to fix the issue for all of them. Can't we fix the original issue by running a linter on the top of the PRs? Then the contributor would fix it according it to the IDE he would be using, if using VSCode, then the issue would be gone by running `pnpm install`.

stevepiercy:
> Can't we fix the original issue by running a linter on the top of the PRs?

I know of no linter that will do what is needed, preventing unintentional reformatting of `.md` files while allowing deliberate changes. You'd have to be a mind reader.

wesleybl:
@stevepiercy What exactly does formatting do that it shouldn't do? Is there a lot of it?

stevepiercy:
@wesleybl random whitespace adjustments, converting `"` to `'`, converting the list markers, and other unnecessary and unwanted changes.

The real problem, though, is not the tool, but the person irresponsibly using the tool and committing those kinds of changes.

wesleybl:
@stevepiercy I believe that 99% of these formatting issues are caused by Prettier. It's by far the most used formatter in VS Code. So I propose the following: Add the configuration in setting.json to ignore formatting in `pnpm install`. Ignore `*.md` files in [.prettierignore](https://github.com/plone/volto/blob/main/.prettierignore). This way, in most cases, we won't have formatting problems. Even if the user doesn't run `pnpm install`, if they are using Prettier (which is very likely the formatter), it won't format. We can leave it like this and monitor for other issues. This solution also has the advantage of working with other IDEs, not just VS Code.

stevepiercy:
@wesleybl let's try that. If it doesn't work out, we can always change it. It's been nice not having to deal with the noise from `.md` reformats for the past month or so, and I'd like that to continue.

wesleybl:
@stevepiercy Done!

You can run, for example:

```
npx prettier --write docs/source/blocks/editcomponent.md
```

and see that the file will not be modified. Lets go?

packages/volto/news/7834.internal
```diff
@@ -0,0 +1 @@
+Move vscode settings setup to pnpm postinstall hook. @wesleybl
```
stevepiercy: I'd like to include a mention of one of the driving purposes for this implementation.

```suggestion
Continue to prevent editors from automatically reformatting Markdown files by moving this configuration from VSCode to prettier.
Move the VSCode settings setup from `make install` to a pnpm post-install hook. @wesleybl
```

packages/scripts/news/7834.internal
```diff
@@ -0,0 +1 @@
+Fix vscodesettings.js to properly check if settings.json exists before reading. @wesleybl
```
stevepiercy:
```suggestion
Fix the {file}`vscodesettings.js` script to properly check if {file}`.vscode/settings.json` exists before reading. @wesleybl
```

stevepiercy (changes requested):
Looks good, some enhancement for the change logs to improve clarity and purpose. Thank you!

stevepiercy (approved):

PR #7835, opened 2026-01-30
A11Y - Changed the order of the Save and Cancel buttons - Sharing page - CMS UI

Wagner3UB:
To fix this issue: https://github.com/plone/volto/issues/5128

**Attention: this change does not affect the visual appearance of this page, only the order for assistive technologies.**

The order of the Save and Cancel buttons has been changed to improve accessibility. It is more intuitive for the Cancel button to come first.

Replaced "float" with "align". The elements do not need to float in this case, so it is safer to simply align them to the right.


stevepiercy (approved):

wesleybl:
@Wagner3UB Could you please do a backport for Volto 18?

PR #7839, opened 2026-01-31
Fix copying of .well-known directory from public/ in production builds

### LINKED ISSUE #7804: Fix collection of dot-file directories like .well-known in public for security.txt and other files.
plone.org issue https://github.com/plone/plone.org/pull/222

.well-known  is not gathered unless using a razzle.extend.js.  In volto 19 we can fix this with the vendorized Razzle.

Kathrina-dev:
### What I did
Fixed an issue in volto-razzle where the .well-known directory placed in public/ was excluded from production builds due to dotfile filtering during the build process.

### Why
Standard discovery endpoints (for example security.txt or assetlinks.json) rely on the .well-known directory being publicly accessible.
Previously, Volto production builds silently omitted this directory because all dotfiles were ignored, making these endpoints unavailable.

This change explicitly allows .well-known to be copied while keeping the existing behavior of ignoring other dotfiles.

### Checklist

-----

Closes #7804 

-----

### Notes for reviewers
- The change is non-breaking
- Existing dotfile exclusion behavior remains unchanged
- .well-known is handled explicitly via a separate copy rule

### DIFF
--- a/packages/volto-razzle/__tests__/createConfigAsync.test.js
+++ b/packages/volto-razzle/__tests__/createConfigAsync.test.js
@@ -1,0 +1,114 @@
+const createConfigAsync = require('../createConfigAsync');
+
+// Mock fs to simulate presence of the public directory
+jest.mock('fs-extra', () => ({
+  existsSync: () => true,
+}));
+
+// Mock 'paths' to ensure consistent test behavior on any machine
+jest.mock('../paths', () => ({
+  appPublic: '/mock/root/public',
+  appBuild: '/mock/root/build',
+  appPath: '/mock/root',
+  appNodeModules: '/mock/root/node_modules',
+  appHtml: '/mock/root/public/index.html', // Added to satisfy template check if needed
+  appAssetsManifest: '/mock/root/build/assets.json',
+}));
+
+// Mock Webpack to prevent plugins from crashing the test
+jest.mock('webpack', () => ({
+  DefinePlugin: class {},
+  HashedModuleIdsPlugin: class {},
+  optimize: {
+    AggressiveMergingPlugin: class {},
+    LimitChunkCountPlugin: class {},
+  },
+  WatchIgnorePlugin: class {},
+  HotModuleReplacementPlugin: class {},
+}));
+
+// Mock other plugins to avoid load errors
+jest.mock(
+  'mini-css-extract-plugin',
+  () =>
+    class {
+      static loader = 'css-loader';
+    },
+);
+jest.mock('webpack-manifest-plugin', () => ({
+  WebpackManifestPlugin: class {},
+}));
+jest.mock('webpackbar', () => class {});
+
+describe('createConfigAsync - CopyPlugin Logic', () => {
+  let config;
+  let patterns;
+
+  beforeAll(async () => {
+    // Generate the production configuration
+    config = await createConfigAsync(
+      'web',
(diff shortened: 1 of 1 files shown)


packages/volto-razzle/config/createConfigAsync.js
```diff
@@ -1004,16 +1004,32 @@ module.exports = (
...
                 {
                   from: paths.appPublic.replace(/\\/g, '/') + '/**/*',
                   to: paths.appBuild,
                   context: paths.appPath,
                   globOptions: {
+                    dot: false,
                     ignore: [
                       paths.appPublic.replace(/\\/g, '/') + '/index.html',
+                      // Exclude .well-known here so it can be handled explicitly below
+                      paths.appPublic.replace(/\\/g, '/') + '/.well-known/**/*',
```
stevepiercy, line 1017: Will this regular expression allow the file path of `/.well-known/funding-manifest-urls`?

Would you please add a unit test for this use case, as well as one that tests for blocking other dotfiles at the root and directories unless they already exist?

stevepiercy (changes requested):

packages/volto-razzle/config/createConfigAsync.js
Kathrina-dev: Yes, this regex allows all files inside .well-known folder to be accessible after you build Volto. 
I’ll add unit tests to cover:
- allowing `.well-known/*` files
- blocking other dotfiles at the root and in subdirectories

stevepiercy:
@Kathrina-dev thank you for adding tests.

It needs a technical review from the @plone/volto-team, and perhaps @davisagli from the original issue.

This PR relates to another recent issue for [plone.org](https://github.com/plone/plone.org/issues/222), where @erral, @fredvd, and @Fosten participated. @plone/marketing-communications might be interested, as well.

Kathrina-dev:
Alright, thanks for the info!

packages/volto-razzle/__tests__/createConfigAsync.test.js
```diff
@@ -0,0 +1,114 @@
+const createConfigAsync = require('../createConfigAsync');
```
sneridagh, line 1: @Kathrina-dev I think these tests are not running at all, nor in CI. Did you run them?

Kathrina-dev: You're right. I completely forgot about it. I had created the __test__ folder within the config folder and forgot to correct the import statement after moving the folder under Volto-razzle. Thanks for mentioning. 

I've fixed the import statement but Jest is still not running the test. I moved the test file under Volto and now it's able to detect the file. Should I  keep this test under Volto or Volto-razzle?

Kathrina-dev: @sneridagh please let me know if I should keep the test in volto or volto-razzle

wesleybl: @Kathrina-dev Ideally, you would add a Vite configuration to the package. See for example: https://github.com/plone/volto/blob/main/packages/volto/vitest.config.mjs

Kathrina-dev: okay I'll add a vitest test in volto and keep the current jest test in volto-razzle

Issue #2617, opened 2021-08-16
Support constraints for adding new content


ericof:
Plone Classic allows the editor to restrict the addable content to a Folderish object, but this is currently not supported in Volto

**Describe the solution you'd like**
- Volto add menu respects the constraints of a folderish
- Volto add menu allows users to add/manage constraints

davisagli:
Needs work in plone.restapi: https://github.com/plone/plone.restapi/issues/334

PR #7842, opened 2026-02-01
feat(Toolbar): Add expandable More… submenu for secondary addable types

aryan7081:
Ref #2617 

In the Add Content menu `packages/volto/src/components/manage/Toolbar/Types.jsx`, secondary addable types (those with `immediately_addable === false` from the API) were shown in the same flat list as main types. As a result, secondary types were always visible instead of being hidden under an expandable “More…” submenu, unlike Plone Classic.
This happened because the menu rendered all addable types in one list and did not use `immediately_addable` to split main types from secondary types or to control visibility of the secondary list.
The fix splits types into main (immediately addable) and secondary (only when “More…” is expanded), adds an expandable `More…` row that toggles visibility of the secondary list.


aryan7081:
Unrelated tests are failing in CI, so I'm leaving this as it is.

davisagli:
I don't think it makes sense to add a secondary menu based on immediately_addable without adding the full type constraints feature, which includes a UI for editors to configure which types are immediately addable. This requires work first at the API level before it can be added to Volto.

stevepiercy:
@aryan7081 also it's a Very Good Idea™ to clarify what work should be done *before* you start work on the issue. See https://6.docs.plone.org/contributing/first-time.html#work-with-github-issues for guidance.

aryan7081:
@stevepiercy Thanks for the pointer. Sorry for jumping in without clarifying the work first. I’ll clarify the work with maintainers before starting on issues like this from now on.

aryan7081:
@davisagli Thanks for the clarification. I raised the PR thinking we could add the “More…” section first (using immediately_addable from @types) and then do the backend + manage-constraints UI later. Your point of doing the full feature makes sense. 

While researching and reading the discussions so far, I've been writing docs on what work is needed in the backend. I'll share and clarify that in https://github.com/plone/plone.restapi/issues/334. It seems like a good place to discuss it.

PR #7849, opened 2026-02-03
Fix default case selection in ButtonsWidget.

iFlameing:
https://github.com/user-attachments/assets/6fe99ad5-5faa-428d-b644-2f2c7bdfea21


wesleybl:
@iFlameing Could you please do a backport to Volto 18?

Issue #5125, opened 2023-08-24
a11y - sharing page - Visible focus - cms ui


Wagner3UB:
**Describe the bug**
- All interactive items: They don't have focus or have low visible/contrasted focus, e.g., checkboxes, buttons, search field, etc.

**To Reproduce**
Navigate using the keyboard

**Expected behavior**
Make visible the border if the element have focus

**Software (please complete the following information):**
Volto Version 17.0.0-alpha.19
Plone Version 6.0.4

PR #7850, opened 2026-02-03
Lack of visible focus for interactive components

Wagner3UB:
To fix the issue https://github.com/plone/volto/issues/5125

Centralize the management of focus states to ensure visual consistency, simplify future maintenance, and resolve accessibility (A11y) issues within Semantic UI components.

### Main Changes
**Architecture:** Created the focus.less file and the .universal-focus-style() mixin. This eliminates CSS rule duplication and establishes a "single source of truth" for focus indicators.

**UX Improvement:** Global implementation of the :focus-visible pseudo-class. This ensures that focus indicators appear only during keyboard navigation, keeping the interface clean for mouse users.

### Component Fixes:

**Semantic UI Checkboxes:** Since the framework hides the original input, focus logic was applied to the label::before pseudo-element, ensuring visibility during keyboard navigation.

**Inputs and Skiplinks:** Border and contrast adjustments to meet accessibility standards.

### Color (@innerFocusShadow & @outerFocusOutline)
The choice of White (#fff) and Black (#000) follows the "Dual-Ring Strategy" for maximum accessibility:

**Universal Visibility:** Using a white inner shadow combined with a black outer outline ensures the focus indicator remains visible on any background (light, dark, or multicolored).

**WCAG Compliance:** This high-contrast pairing (black/white) guarantees we meet the minimum 3:1 contrast ratio required for UI components.

**Brand Neutrality:** Neutral colors avoid visual conflicts with the site's primary palette while providing a clear, standard functional cue for keyboard navigation.

Note
Please be aware that since these are global rules, specific component styles or legacy CSS may still override these definitions in certain contexts. Future refactoring might be needed if specific edge cases arise.

Example:
https://github.com/user-attachments/assets/20021d84-4b5c-4e58-8a39-a32db1c012a0


Wagner3UB:
@wesleybl Hey Wesley,  how it's going in Brasil?

I need to ask you about the focus when we’re inside the editor title field.

As I’m trying to modify the universal focus style to overcome some Semantic UI limitations, your Cypress test fails when it tries to assert that there is no outline. This happens because the `:focus-visible` selector has higher priority than `:focus`, so the outline ends up being visible again. https://github.com/plone/volto/blob/49a483c200e4c1d2da6c071c6dd0510df0b50630/packages/volto/cypress/tests/core/blocks/blocks-title.js#L20

My question is a bit more specific, we need a visual indicator for all interactive elements, we can’t simply hide the outline. I know it looks ugly to have a border inside an input, and I’m aware of that. Maybe our goal here should be to work on this and make it more acceptable for those who see it. But unfortunately, we can’t make it disappear.

Wagner3UB:
> @wesleybl Hey Wesley, how it's going in Brasil?
> 
> I need to ask you about the focus when we’re inside the editor title field.
> 
> As I’m trying to modify the universal focus style to overcome some Semantic UI limitations, your Cypress test fails when it tries to assert that there is no outline. This happens because the `:focus-visible` selector has higher priority than `:focus`, so the outline ends up being visible again.
> 
> https://github.com/plone/volto/blob/49a483c200e4c1d2da6c071c6dd0510df0b50630/packages/volto/cypress/tests/core/blocks/blocks-title.js#L20
> 
> My question is a bit more specific, we need a visual indicator for all interactive elements, we can’t simply hide the outline. I know it looks ugly to have a border inside an input, and I’m aware of that. Maybe our goal here should be to work on this and make it more acceptable for those who see it. But unfortunately, we can’t make it disappear.

Same thing for the files [blocks-table.js](https://github.com/plone/volto/blob/main/packages/volto/cypress/tests/core/blocks/blocks-table.js) and [blocks-slate.js](https://github.com/plone/volto/blob/main/packages/volto/cypress/tests/core/blocks/blocks-slate.js)

wesleybl:
> @wesleybl Hey Wesley, how it's going in Brasil?
> 
> I need to ask you about the focus when we’re inside the editor title field.
> 
> As I’m trying to modify the universal focus style to overcome some Semantic UI limitations, your Cypress test fails when it tries to assert that there is no outline. This happens because the `:focus-visible` selector has higher priority than `:focus`, so the outline ends up being visible again.
> 
> https://github.com/plone/volto/blob/49a483c200e4c1d2da6c071c6dd0510df0b50630/packages/volto/cypress/tests/core/blocks/blocks-title.js#L20
> 
> My question is a bit more specific, we need a visual indicator for all interactive elements, we can’t simply hide the outline. I know it looks ugly to have a border inside an input, and I’m aware of that. Maybe our goal here should be to work on this and make it more acceptable for those who see it. But unfortunately, we can’t make it disappear.


Hi @Wagner3UB!

Actually, it wasn't me who removed the border. It was @tiberiuichim  in: https://github.com/plone/volto/pull/5186

I only added the test.

I personally find the black border on inputs distracting, since the editor already adds a blue border to the focused block.

But if this is better for UX, okay. I just think that if the border is applied, it should be applied to all blocks, not just the core blocks. But it's already like that, right?

Wagner3UB:
@tiberiuichim Hey Tiberium! I’ll need your help with this one. We kind of have a “conflict of interest” here. It’s not pretty, but in my opinion, from an a11y perspective, it’s necessary, or at least important. Here is your fix: https://github.com/plone/volto/pull/5186

wesleybl:
@Wagner3UB I really find it strange to have a black border when we already have a blue border when the Slate input is selected:

image

Doesn't the blue border serve its purpose? Is there another option?

Wagner3UB:
@wesleybl I will work on this matter and will get back to you with a positioning on it.

> I personally find the black border on inputs distracting, since the editor already adds a blue border to the focused block.
> 
> But if this is better for UX, okay. I just think that if the border is applied, it should be applied to all blocks, not just the core blocks. But it's already like that, right?

Wagner3UB:
> @wesleybl I will work on this matter and will get back to you with a positioning on it.
> 
> > I personally find the black border on inputs distracting, since the editor already adds a blue border to the focused block.
> > But if this is better for UX, okay. I just think that if the border is applied, it should be applied to all blocks, not just the core blocks. But it's already like that, right?

@wesleybl It seems like I already did this change, adding some exceptions to it. Take a look if it can be solved with this: https://github.com/plone/volto/pull/7850/changes#diff-96e616bc401c4216dd3df32954f63e8c5b5ea24ec11cb3931f090c9d08b8e1a5R8

wesleybl:
In texts with the "copy anchor" functionality, a border break occurs. Is there any way to improve this?

image

packages/volto/theme/themes/pastanaga/extras/focus.less
```diff
@@ -0,0 +1,33 @@
+.universal-focus-style() {
+  box-shadow: 0 0 0 2px @innerFocusShadow !important;
```
wesleybl, line 2: Would it be possible to remove these `!important`?

packages/volto/theme/themes/pastanaga/extras/focus.less
```diff
@@ -0,0 +1,33 @@
...
+  .universal-focus-style();
+  outline-offset: 4px;
+}
+
+.ui.checkbox label::before {
+  transition: none !important;
+}
+
+// Skiplinks focus styles
+.skiplinks a:focus:not(.focus--mouse) {
```
wesleybl: Why not use `:focus-visible` here, as in the rest of the file?

packages/volto/theme/themes/default/globals/site.variables
```diff
@@ -92,6 +92,9 @@
     Focused Input
 --------------------*/
 
+@innerFocusShadow: #fff;
+@outerFocusOutline: #000;
```
wesleybl: Why was it included in `themes/default/globals/site.variables` if the theme being used is Pastanaga?

packages/volto/theme/themes/pastanaga/extras/focus.less
```diff
@@ -0,0 +1,33 @@
...
+  border-color: @outerFocusOutline !important;
+}
+
+// Semantic UI checkbox focus styles
+.ui.checkbox input:focus-visible ~ label::before {
+  .universal-focus-style();
+  outline-offset: 4px;
+}
+
+.ui.checkbox label::before {
```
wesleybl: I think we can be more specific here so as not to kill the transition in all situations:

`.ui.checkbox input:focus-visible ~ label::before`

wesleybl (changes requested):

packages/volto/theme/themes/pastanaga/extras/focus.less
Wagner3UB: The !important flag is necessary here because Semantic UI applies its own focus styles with !important on several selectors. Without it, the border-color override would be ignored, and inputs that don't receive a border by default would look unfocused. This solution adds a border to many inputs that were previously overlooked—or intentionally ignored by Semantic UI—such as those in the sidebar or the site search field.

wesleybl (approved):

Wagner3UB:
@stevepiercy Can you run your review also here buddy? Ty!

packages/volto/news/7850.bugfix
```diff
@@ -0,0 +1 @@
+Lack of visible focus for interactive components @Wagner3UB
```
stevepiercy:
```suggestion
Added consistent visible focus for interactive components to improve accessibility. @Wagner3UB
```

stevepiercy (changes requested):
Looks fine, with a tweak to the news fragment to describe what this PR actually does, as I understand it. Please correct it, if I'm wrong. Thank you!

stevepiercy (approved):

PR #7861, opened 2026-02-06
Fix canonical URL editing in URL Management

### LINKED ISSUE #6058: it's not obvious how to change the url (UX Bug)
# Problem

It is a UX bug when someone expects to be able to do something and can't work out how. Thats not an enhancement.

Currently volto offers two ways to change a url
- Edit then change "short name" in the object metadata
- Contents view and "rename"
  - even worse you need to go to the parent to rename the item you want (#6061 )

Neither are obvious to someone new to plone. We have got multiple support requests asking how to achieve this.

In addition, it's not obvious that the only way to change the the full url is by changing the parent objects short-names.

# Proposed solution

## Url Management Rename

"Url management" might be the first place someone would look and it makes sense that you can control both the current url and all the alternative urls in the same place.

So add a "change url" field onto url management dialog.
- get rid of "short name". Better to change url

```
Change URL: /path/to/my/ [ page          ]"
```

shivaansh0610-LUFFY:
This PR addresses issue #6058 by enabling editors to change the canonical (primary) URL of a content item from the URL Management view.

What was missing

The URL Management screen allowed adding and removing alternative URLs (aliases), but there was no way to update the canonical URL, even though this is part of URL management.

What this PR does

Adds a Canonical URL section to the URL Management UI

Displays the current canonical URL as read-only

Allows entering a new canonical URL

Validates that the new URL starts with /

Updates the content ID using updateContent

Redirects the user to the new canonical URL after the change

Notes

The change is limited to the frontend (Aliases.jsx)

Existing alias functionality is untouched

No additional UX behavior (e.g., success toast) was added to keep the scope aligned with the issue

Fixes #6058

### DIFF
--- a/news/6058.feature
+++ b/news/6058.feature
@@ -1,0 +1,2 @@
+Fix canonical URL editing in URL Management.
+


packages/volto/src/components/theme/Register/Register.jsx
```diff
@@ -85,6 +85,7 @@ const Register = () => {
       createUser({
         fullname: fullname,
         email: email,
+        username: email,
```
davisagli, line 88: Why is there an unrelated change to the Register component here?

packages/volto/src/components/manage/Aliases/Aliases.jsx
```diff
@@ -172,6 +178,24 @@ class Aliases extends Component {
...
+
+  handleChangeCanonicalUrl() {
+    const baseUrl = getBaseUrl(this.props.pathname);
+    const newId = this.state.newCanonicalUrl.replace(/^\//, '');
+
+    this.props.updateContent(baseUrl, {
+      id: newId,
+    });
+
+    this.props.history.push(this.state.newCanonicalUrl);
```
davisagli, line 196: In my testing:
- This navigates to the new URL before the updateContent action has finished, so the user can see a 404
- The new URL might not be the same as what the user entered (for example if it includes characters that are not allowed in a URL, or the new id was already used by another content item). So you have to wait for the result from updateContent and redirect to the `@id` it returned.

PR #7859, opened 2026-02-05
Fix: Filter invalid block IDs in getBlocks() to prevent DOM duplicates

aryan7081:
Fixes #7858

When `blocks_layout.items` contained invalid IDs (null, undefined, 'undefined'), `getBlocks()` mapped them to `blocks[undefined]`, creating duplicate DOM elements and Cypress test failures.

This happened because `getBlocks()` didn't validate block IDs before mapping.

This fix adds `isValidBlockId()` helper to filter invalid IDs in `getBlocks()`, updates `RenderBlocks` to use it, and adds unit tests.

PR #7535, opened 2025-10-21
Fix blocks_layout sync when content has empty blocks fields from migr…

aryan7081:
Fixes #2887 

When content has empty `blocks={}` and `blocks_layout={items:[]}` adding text to blocks created new block IDs in the blocks field but blocks_layout remained empty, causing view breakage.

This happened because `changeBlock` only updated the blocks field and `getOnlyFormModifiedValues` didn't include `blocks_layout` in the save operation, so the server never received the synchronized `blocks_layout` update.

This fix adds `ensureBlocksLayoutSync` helper to synchronize `blocks_layout` with `blocks`, updates `changeBlock` to use it, and modifies `getOnlyFormModifiedValues` to include `blocks_layout` when blocks is modified.

PR #7867, opened 2026-02-06
fix(search): show Effective date instead of effective in sort

### LINKED ISSUE #7866: Search block sort dropdown shows raw key "effective" instead of "Effective date" until @querystring loads
**Describe the bug**
When a search block page loads or is saved, the `@querystring` API may not have loaded yet, causing the SortOn dropdown to show raw index keys (like `"effective"`) instead of human-readable labels (like `"Effective date"`)

**To Reproduce**

1. Create a new page with a title.
2. Add a Search block.
3. Check the checkbox of `Show sorting?` in the controls section of the sidebar.
4. Select any Sort on Options from the dropdown, For eg, Effective date.
5. Save the page.
6. Now open dev tools, then go to netwrok tab and select 3G from the throttling dropdown.
7. Navigate to your page
8. Quickly click the sort dropdown before `@querystring` finishes loading.
9. Repeat steps 7 and 8 until you reproduce the issue
10. You should see a raw key like `effective` instead of `effective date`.

**Expected behavior**
The `@querystring` API should be preloaded when content with blocks is loaded, so labels are available immediately when the component renders.

**Screenshots**

https://github.com/user-attachments/assets/fdbeab24-f42a-4a24-b6c5-aa7bb22fb1af

- OS: macOS Tahoe
- Browser Chrome
- Volto Version 19.0.0-alpha.4
- Plone Version 6.1.0
- Plone REST API Version 9.1.0

aryan7081:
Fixes #7866

When the Search block sort dropdown loaded before `@querystring`, it showed the raw index key `effective` instead of the label `Effective date`.

This happened because `SortOn`showed the backend title or the key with no loading state or fallback when the title was the raw key.

This fix adds a `Loading…` state until the querystring is loaded, fallback labels for common indexes (e.g. effective → "Effective date"), and loading placeholders in the edit form so the sort dropdown never shows the raw key.

### DIFF
--- a/packages/volto/news/7866.bugfix
+++ b/packages/volto/news/7866.bugfix
@@ -1,1 +1,1 @@
-Preload @querystring when content with blocks is loaded so Search block sort dropdown shows labels (e.g. "Effective date") instead of raw keys (e.g. "effective"). Show "Loading…" in SortOn until labels are available. @aryan7081
+Preload `@querystring` when content with blocks is loaded so the Search block sort select menu shows labels, such as "Effective date", instead of raw keys, such as "effective". Show "Loading…" in `SortOn` until labels are available. @aryan7081


wesleybl:
@aryan7081 I haven't seen it running yet, but I think adding a "Loading" message will be a distraction for the user. The problematic string is in a combo box, initially hidden. When users perform a search, they are initially interested in the result. They probably won't even notice it. What do you think about leaving it as is? Am I missing something?

aryan7081:
> They probably won't even notice it. What do you think about leaving it as is? Am I missing something?

@wesleybl You're right, I'm fine leaving it as is. The problem shows up in PR #7535: the block-search Cypress test uses `cy.findByText('Effective date')` and was failing when the combo showed the raw key effective before `@querystring `loaded. Should we change the test (e.g. wait for the label or use a different matcher), or do you prefer another approach?

packages/volto/src/components/manage/Blocks/Search/components/SortOn.jsx
```diff
@@ -48,12 +52,32 @@ const SortOn = (props) => {
...
 
+  // Fallback labels when backend returns raw key (e.g. "effective") as title
+  const FALLBACK_LABELS = {
+    effective: 'Effective date',
+    created: 'Creation date',
+    modified: 'Modification date',
+    sortable_title: 'Title',
+    id: 'ID',
+    getRawCount: 'Number of items',
+  };
```
erral, line 94: If we are introducing these labels that will be shown to the end-user, they should be properly internationalized.

aryan7081: Thanks for the review! Fallback labels are now in defineMessages and use intl.formatMessage(), and the locale files are updated.

wesleybl:
> @wesleybl You're right, I'm fine leaving it as is. The problem shows up in PR >https://github.com/plone/volto/pull/7535: the block-search Cypress test uses cy.findByText('Effective date') and was >failing when the combo showed the raw key effective before @querystring loaded. Should we change the test (e.g. >wait for the label or use a different matcher), or do you prefer another approach?

@aryan7081 I think it's best to update the test, to wait for the final label.

aryan7081:
@wesleybl Okay, I'll update the test. May I know the reason for reopening this PR??
Thanks.

wesleybl:
@aryan7081 I haven't fully analyzed this PR yet. Do you think it's worth adding the fallback without the Loading?

aryan7081:
@wesleybl Yes, I think the fallback is still worth keeping without `Loading…` because it fixes the real bug: users see "Effective date" instead of the raw key "effective" when the API hasn’t loaded, or the backend returns the key as the title. It’s a small, non-intrusive change that keeps the Cypress test stable and doesn’t add any extra UI.

aryan7081:
@wesleybl also, we can fix the test for now, but in the future, it may confuse other contributors. Since we can’t see this in the UI, the bug will be hard to track down. similar to how it took a lot of effort for me to figure out why it was failing.

packages/volto/news/7866.bugfix
```diff
@@ -0,0 +1 @@
+Preload @querystring when content with blocks is loaded so Search block sort dropdown shows labels (e.g. "Effective date") instead of raw keys (e.g. "effective"). Show "Loading…" in SortOn until labels are available. @aryan7081
```
stevepiercy: I'm not clear whether y'all want to include the last sentence in the change log, based on the [discussion in the PR](https://github.com/plone/volto/pull/7867#issuecomment-3864701619).

```suggestion
Preload `@querystring` when content with blocks is loaded so the Search block sort select menu shows labels, such as "Effective date", instead of raw keys, such as "effective". Show "Loading…" in `SortOn` until labels are available. @aryan7081
```

packages/volto/src/middleware/preloadQuerystringForBlocks.js
```diff
@@ -0,0 +1,38 @@
...
+const CREATE_CONTENT_SUCCESS = `${CREATE_CONTENT}_SUCCESS`;
+const UPDATE_CONTENT_SUCCESS = `${UPDATE_CONTENT}_SUCCESS`;
+
+export default function preloadQuerystringForBlocks({ dispatch, getState }) {
+  return (next) => (action) => {
+    const result = next(action);
+    if (action.subrequest || !action.result) return result;
+
+    if (action.type === GET_CONTENT_SUCCESS) {
+      dispatch(getQuerystring());
```
wesleybl, line 24: Not all content needs `getQueryString`. I was thinking about fixing the label, but without making extra requests. Initially displaying `Effective date` instead of `effective`. But with that, we'll have to add translations, and if the translation changed on the backend side, Volto would be wrong.

Perhaps it's better to leave everything as it is?

aryan7081: @wesleybl Agreed. We can leave everything as is. I'll update the tests in https://github.com/plone/volto/pull/7535. But we need to close https://github.com/plone/volto/pull/7859 before proceeding with https://github.com/plone/volto/pull/7535. What do you think about the next actionable steps?. I would love your guidance on how you would like to move forward?.
Thanks!

wesleybl: @aryan7081 I think it's best to fix the test. I'm going to close this one.

Thanks!

PR #7872, opened 2026-02-07
Increase wait time between link check retries with Lychee

wesleybl:
When some links fail, we now wait 10 seconds between retries (--retry-wait-time 10), improving the chance of success for temporarily unstable links.

### DIFF
--- a/docs/source/theming/using-third-party-themes.md
+++ b/docs/source/theming/using-third-party-themes.md
@@ -96,7 +96,7 @@ module.exports = Object.assign({}, volto_config, {
 ```
 
 Complete example in an active project:
-https://github.com/RedTurtle/design-volto-theme/blob/master/razzle.config.js
+https://github.com/RedTurtle/design-comuni-plone-theme/blob/main/razzle.extend.js
 
 In that project, there is the Sass loader and the SVG loader, too.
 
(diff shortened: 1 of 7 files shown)


stevepiercy:
Which specific links are causing issues?

Sometimes the specific site has rate limiting, and we need to use care. Lychee also has some built-in exponential backoff for [handling rate limits](https://lychee.cli.rs/troubleshooting/rate-limits/).

stevepiercy:
Also keep in mind that increasing the value of [`--retry-wait-time`](https://lychee.cli.rs/guides/cli/#-r---retry-wait-time) from the default of `1` could result in a very long test run.

I'm OK with trying this change out, but first want to know which links it actually is intended to address.

wesleybl:
@stevepiercy I didn't record exactly which links they are. But several times the job fails, and I run it again and it works. About taking longer, that interval only occurs when there's a failure. And even if there's a failure, the other jobs take much longer. So, it wouldn't make much difference.

stevepiercy:
@wesleybl yeah, but I like keeping runs fast 🚀 !

Although runs that take more time also burn more energy, subsequent runs burn more. Let's watch the result of this change, and adjust as needed. Will merge when green. Thank you!

PR #7883, opened 2026-02-11
Fix href override in UniversalLink component.

iFlameing:


packages/volto/src/components/manage/UniversalLink/UniversalLink.tsx
```diff
@@ -122,6 +122,7 @@ const UniversalLink = React.memo(
         onClick,
         onKeyDown,
         item,
+        href,
```
sneridagh, line 125: I don't understand the fix. `href` is not used anywhere...

sneridagh: OK, clarified, the point is its absence from the `...rest`, that does the trick.

PR #7895, opened 2026-02-13
Update users controlpanel to work with new response format

jnptk:
https://github.com/plone/plone.restapi/pull/1971 introduces a new response format for the users endpoint which is more consistent with the response formats of other services in plone.restapi. **This change is for break backwards compatibility!**


packages/volto/news/+userscp.bugfix
davisagli, line 1: It's .breaking -- we changed the format of the data in the redux store, so any custom components that expected the old format need to be updated.

That's allowed, because we're still in the alphas for Volto 19. But we have to call it breaking in the release notes, so it's more obvious that custom code might need to be updated.

davisagli: That makes me think. We could make the reducer convert the new format from the API into the same format that was previously stored in the Redux state. Then components don't have to be updated.

Sorry I keep adjusting the plan.

jnptk: > That makes me think. We could make the reducer convert the new format from the API into the same format that was previously stored in the Redux state. Then components don't have to be updated.

Then we won't have `items_total` and `@id` (not that this is really important but maybe in the future?).

davisagli (approved):

sneridagh:
ok, let's merge it, but it's breaking or not? :) 

I guess that if we have to backport it, we need to handle the data conversion in the reducer, right?

davisagli:
@sneridagh We changed it to be not breaking, by converting it in the reducer to the old format.

We will probably need a similar conversion in `@plone/client`.

PR #7897, opened 2026-02-13
Reinstalled all dependencies to latest compatible versions

wesleybl:
Backport of #7871 to Volto 18.

### DIFF
--- a/packages/volto/news/7871.internal
+++ b/packages/volto/news/7871.internal
@@ -1,0 +1,1 @@
+Reinstalled all dependencies to latest compatible versions. @wesleybl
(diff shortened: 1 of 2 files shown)


sneridagh:
@wesleybl we don't do this normally for stable versions, what do you think @plone/volto-team ?

wesleybl:
@sneridagh This will not affect existing projects. Projects have their own `pnpm-lock.yaml` file. New projects will use newer versions. Updating here would be avoiding packages with vulnerabilities.

wesleybl:
This has been done before without consequences: https://github.com/plone/volto/pull/7299

sneridagh:
@wesleybl yeah I remember something like it, I only I wanted to be sure! Ok, let's do it then, if anybody else is _not_ opposed.

PR #7906, opened 2026-02-14
Update dependencies to reduce duplicate packages

davisagli:
- Update `@plone/client` and `@plone/registry` to use the same version of vitest as `@plone/volto`.
- Update http-proxy-middleware to the same version as webpack-dev-server
- Update tmp (used by `@plone/registry`) and remove it as a devDependency of volto since it's not used.

### DIFF
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -77,7 +77,7 @@
     "tsup": "^8.3.5",
     "typescript": "^5.7.3",
     "uuid": "^9.0.1",
-    "vitest": "^3.2.4",
+    "vitest": "catalog:",
     "wait-on": "^7.2.0"
   },
   "peerDependencies": {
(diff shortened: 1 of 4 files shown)


packages/volto/package.json
```diff
@@ -149,7 +149,7 @@
     "github-slugger": "1.4.0",
     "history": "4.10.1",
     "hoist-non-react-statics": "3.3.2",
-    "http-proxy-middleware": "2.0.1",
+    "http-proxy-middleware": "2.0.9",
```
sneridagh, line 152: mmm I recall this having a problem when I tried in the past. Well, if the tests pass...

sneridagh:
@davisagli We can use catalog versions in core, in fact, they are in place, but we need to start using them. I was waiting for the moment that it makes sense.

packages/volto/package.json
davisagli: I searched a bit to see if I could find a clue about what happened last time, but didn't find it.

davisagli:
@sneridagh I updated it to use the catalog for the vitest version.

PR #7909, opened 2026-02-16
Add missing i18n and fix duplicate id in Facet widget

iFlameing:


packages/volto/src/components/manage/Blocks/Search/components/SelectFacet.jsx
```diff
@@ -12,16 +13,35 @@ import {
   selectFacetValueToQuery,
 } from './base';
 
+const messages = defineMessages({
+  selectOption: {
+    id: 'Select option',
+    defaultMessage: 'Select option',
+  },
+  select: {
+    id: 'select...',
```
sneridagh: Let's use capitals, and the special character for `...` (not three dots). Also, are we sure that we don't have this literals anywhere else?

stevepiercy:
```suggestion
    id: 'select…',
```

iFlameing: done.

Issue #7899, opened 2026-02-13
RegistryImageWidget: stray '}' appended to base64 data on file upload


mpalomaki:
## Description

`RegistryImageWidget` appends a stray `}` character to the base64 data of every uploaded file due to a template literal bug. This produces invalid data URIs that browsers cannot decode, causing uploaded images (e.g. site logo) to appear blank.

## Location

`packages/volto/src/components/manage/Widgets/RegistryImageWidget.jsx`, line 102:

```javascript
onChange(id, `filenameb64:${btoa(file.name)};datab64:${fields[3]}}`);
```

The template literal `${fields[3]}}` has a double `}}` — the first `}` closes the template expression, the second `}` is a literal character that gets appended to the output string.

## Expected

```
filenameb64:TG9nby5zdmc=;datab64:PD94bWw...Zz4K
```

## Actual

```
filenameb64:TG9nby5zdmc=;datab64:PD94bWw...Zz4K}
                                               ^ invalid base64 character
```

## Impact

Any backend that constructs a data URI from the stored value produces an invalid URI:
```
data:image/svg+xml;base64,PD94bWw...Zz4K}  ← browser cannot decode
```

The site logo appears blank after upload, even though the data is otherwise correctly stored and retrieved.

## Fix

One-character change:

```diff
- onChange(id, `filenameb64:${btoa(file.name)};datab64:${fields[3]}}`);
+ onChange(id, `filenameb64:${btoa(file.name)};datab64:${fields[3]}`);
```

## Reproduction

1. Any Volto + Plone/Nick installation
2. Site Setup → Site → upload a logo image → Save
3. Inspect the stored value — it ends with `}`
4. The `@site` endpoint constructs a data URI with the trailing `}`, which browsers reject

## Versions

Confirmed present in:
- Volto 19.0.0-alpha.24
- Current `main` branch

## Related

- Nick CMS PR: https://github.com/robgietema/nick/pull/19 (includes server-side workaround that strips trailing `}` on read)

PR #7900, opened 2026-02-13
Fix RegistryImageWidget appending stray '}' to base64 data

mpalomaki:
## Summary

One-character fix. `RegistryImageWidget` appends a stray `}` to the base64 data of every uploaded file, producing invalid data URIs.

## Bug

Line 102 of `RegistryImageWidget.jsx`:

```javascript
onChange(id, `filenameb64:${btoa(file.name)};datab64:${fields[3]}}`);
//                                                              ^^ double }}
```

The template literal `${fields[3]}}` has a double `}}` — the first `}` closes the template expression, the second is a literal character appended to the string.

## Fix

```diff
- onChange(id, `filenameb64:${btoa(file.name)};datab64:${fields[3]}}`);
+ onChange(id, `filenameb64:${btoa(file.name)};datab64:${fields[3]}`);
```

## Impact

Every file uploaded through `RegistryImageWidget` (e.g. site logo via Site Setup → Site) gets a `}` appended to its base64 data. Backends constructing data URIs from the stored value produce invalid URIs that browsers cannot decode, causing images to appear blank.

## Test plan

- [ ] Upload an SVG logo via Site Setup → Site → Save
- [ ] Inspect stored value — should NOT end with `}`
- [ ] Logo should display correctly on the frontend
- [ ] Re-open Site Setup → Site — logo preview should render

Fixes #7899

Related: https://github.com/robgietema/nick/pull/19 (Nick CMS PR that includes a server-side workaround)

PR #7910, opened 2026-02-17
Fix RegistryImageWidget stray '}' in base64 data

ParthMozarkar:
Title: Fix RegistryImageWidget stray '}' appended to base64 data on file upload

Type of PR: Bugfix

Purpose of this PR:
This PR fixes an issue in `RegistryImageWidget` where a stray `}` character was appended to the base64 data of uploaded files.

Root Cause:
A double `}}` was present in a template literal. The first `}` correctly closed the template expression, while the second was interpreted as a literal character and appended to the resulting string. This caused every generated base64 data URI to include an unexpected trailing `}`.

Impact:
The malformed data URI could not be properly decoded by browsers, resulting in uploaded images (e.g., site logos) appearing blank or failing to render.

Changes:
- Removed the extra `}` from the template literal.
- Ensured generated data URIs conform to the expected `data:<type>;base64,<payload>` format.
- Verified correct rendering of uploaded images after the fix.

Fixes:
- #7899

I have tested the changes: Yes
- Confirmed uploaded images render correctly in supported browsers.
- Verified no regressions in existing file upload behavior.

### DIFF
--- a/packages/volto/news/7899.bugfix
+++ b/packages/volto/news/7899.bugfix
@@ -1,0 +1,1 @@
+Fix RegistryImageWidget stray '}' appended to base64 data on file upload. @ParthMozarkar [#7899](https://github.com/plone/volto/issues/7899)


davisagli:
There is already a fix in progress in https://github.com/plone/volto/pull/7900. Please don't work on an issue that already has work in progress.

PR #7916, opened 2026-02-17
Make Slate tables responsive horizontally

sneridagh:
Supersedes https://github.com/plone/volto/pull/5745

https://github.com/user-attachments/assets/2563e9f6-2e41-4f8f-8710-d7f5945b7763

### DIFF
--- a/docs/source/upgrade-guide/index.md
+++ b/docs/source/upgrade-guide/index.md
@@ -349,7 +349,6 @@ To support horizontal scrolling of tables on small viewports, the table block is
 </div>
 ```
 
-It follows the block model 3 style wrappers and anatomy.
 In the past, the table block was a direct child of the content area.
 If you had custom styles for the table block that assumed it was a direct child of the content area, then adjust your CSS styles accordingly.
 


davisagli (commented):
Needs changelog (breaking) and upgrade guide

docs/source/upgrade-guide/index.md
```diff
@@ -333,6 +333,26 @@ We pinned the version of `sass` to `1.32.0`, which is the one before they introd
 It is unlikely that using this version will cause problems since no real new features were added in later versions that are relevant for Volto developers.
 In case that you need a later version of `sass` in your project or add-on, you can override it in your project's {file}`package.json` file.
 
+### Table block is now wrapped with containers to support horizontal scrolling on small viewports
+```{versionadded} Volto 19.0.0-alpha.26
+```
+
+In order to support horizontal scrolling of tables on small viewports, the `table` block is now wrapped with containers that allow horizontal scrolling when the table width exceeds the viewport width.
```
```suggestion
To support horizontal scrolling of tables on small viewports, the table block is now wrapped with containers that allow horizontal scrolling when the table width exceeds the viewport width.
```

docs/source/upgrade-guide/index.md
```diff
@@ -333,6 +333,26 @@ We pinned the version of `sass` to `1.32.0`, which is the one before they introd
...
+<div className="block slate-table">
+  <div className="block-inner-container">
+    <table>
+      <!-- table content -->
+    </table>
+  </div>
+</div>
+```
+
+It follows the block model 3 style wrappers and anatomy.
```
stevepiercy: What is that? There should be a term that references a glossary entry.

docs/source/upgrade-guide/index.md
```diff
@@ -333,6 +333,26 @@ We pinned the version of `sass` to `1.32.0`, which is the one before they introd
...
+In the past, the table block was a direct child of the content area.
+Adjust your CSS styles if you had custom styles for the table block that assumed it was a direct child of the content area.
```
```suggestion
If you had custom styles for the table block that assumed it was a direct child of the content area, then adjust your CSS styles accordingly.
```

stevepiercy (changes requested):

docs/source/upgrade-guide/index.md
sneridagh: You are correct, and now I think that there is no need to introduce that (yet) to avoid confusion, so I removed the line.

stevepiercy: Less is more. :grin:

stevepiercy:
Docs look good, too.

https://volto--7916.org.readthedocs.build/7916/upgrade-guide/index.html#table-block-is-now-wrapped-with-containers-to-support-horizontal-scrolling-on-small-viewports

wesleybl:
@sneridagh @davisagli We had this complaint in Volto 18 already — specifically the issue where tables break on narrow viewports because they don’t scroll horizontally. Given that experience, my reading is that this behavior feels like a bug rather than an intentional breaking change, because the lack of responsiveness affects all users on smaller screens, whereas the proposed CSS/style breakage likely only affects a small subset of edge cases.

It seems preferable to fix the mobile/tablet UX by enabling horizontal scrolling, even if that entails some changes to table styles in Volto 18, instead of leaving tables broken on mobile for everyone.

Given that Volto 18 is still actively supported and receives bugfix backports when appropriate, can we backport this change to the 18.x branch? If it’s not straightforward to backport cleanly, it would be good to understand:

- whether this change must be treated as a breaking change for 18.x (i.e., requiring theme customizations),
- or if we can backport it as a bugfix/improvement with minimal disruption.

In other words, is this something that can be safely backported to Volto 18? Or would downstream projects need to implement customizations at the functionality level in order to achieve the same responsive table behavior?

Thanks!

sneridagh:
@wesleybl I see and understand your request, however, we have to protect the people that is using Plone in their projects. There is real threat that if we add the wrappers we might break people's CSS selectors, then, all of a sudden, apply a bugfix release, we break their tables in their sites.

As an integrator I would not accept it. A bugfix can never lead to broken sites. People rely on semantic versioning because of its predictability. People rely on us and in our well defined process. Even if we thing that the possibility is slightly and subtle and probably it won't affect anyone. 

So yes, people that wants to have this fix would have to create a customization, but they will do it consciously and aware of that they are doing, and they will amend their CSS accordingly, if required.

If it was in the CMSUI we have a policy for allowing these use cases, but given the fact that this affects the public UI, it is not acceptable.

I know this is not what you wanted to hear, but it's what release managing is all about.

PR #7918, opened 2026-02-18
[components] New spinner component

sneridagh:
image

### DIFF
--- a/.github/workflows/readme-link-check.yml
+++ b/.github/workflows/readme-link-check.yml
@@ -10,7 +10,7 @@ jobs:
   readme-link-checker:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v5
+    - uses: actions/checkout@v6
     - name: Check links in README.md with Lychee
       id: lychee
       uses: lycheeverse/lychee-action@v2


.github/workflows/readme-link-check.yml
```diff
@@ -10,7 +10,7 @@ jobs:
   readme-link-checker:
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v4
+    - uses: actions/checkout@v5
```
stevepiercy: Bump. https://github.com/actions/checkout/releases

```suggestion
    - uses: actions/checkout@v6
```

packages/components/src/components/Spinner/Spinner.quanta.tsx
```diff
@@ -0,0 +1,102 @@
...
+  if (typeof size === 'string' && !isPresetSize(size)) {
+    return size;
+  }
+  return undefined;
+}
+
+export function Spinner({
+  size = 'sm',
+  label = 'Loading',
+  showLabel = false,
```
davisagli: Do we really need a separate showLabel param? I think it would be more intuitive to just allow setting `label` to "" or null, and make that the default label.

If an actual label is going to be shown, it has to be translated into the current language (which is easiest if it has to be passed in and we don't have "Loading" as a default)

packages/components/src/components/Spinner/Spinner.tsx
```diff
@@ -0,0 +1,65 @@
...
+    return { dataSize: size };
+  }
+  if (typeof size === 'number') return { customSize: `${size}px` };
+  return { customSize: size };
+}
+
+export function Spinner({
+  size = 'sm',
+  label = 'Loading',
+  showLabel = false,
```
davisagli: Same

packages/components/src/components/Spinner/Spinner.quanta.tsx
sneridagh: @davisagli I applied the suggestion. I removed the `showLabel` in favor of `label` and added an additional `aria-label` like other RAC components does, in case you want to customize it, otherwise, defaults to "Loading". Remember that he @plone/components do not have i18n! They should be passed from above when instantiating.

davisagli: Yes, that's what I was saying too (re i18n)

sneridagh:
@davisagli I will merge it for now, and take care of the backport to 19.

sneridagh:
@davisagli Uh, my last push was rejected so you could not review the changes, I just pushed. I might wait for you then.

davisagli (approved):

PR #7868, opened 2026-02-07
Add Vitest test for .well-known handling in production build

Kathrina-dev:
This PR adds Vitest test coverage for the `.well-known` public asset handling.

- Adds a new **Vitest unit test** in `packages/volto/__tests__/createConfigAsync.test.js` to verify that:
  - dotfiles are excluded by default
  - the `.well-known` directory under `public/` is explicitly included in production builds
- Fixes a **small import path issue** in the existing Jest test in  
  `packages/volto-razzle/__tests__/createConfigAsync.test.js` so it correctly references `createConfigAsync`

-----

- [x] I signed and returned the [Plone Contributor Agreement](https://plone.org/foundation/contributors-agreement), and received and accepted an invitation to join a team in the Plone GitHub organization.
- [x] I verified there aren't other open [pull requests](https://github.com/plone/volto/pulls) for the same change.
- [x] I followed the guidelines in [Contributing to Volto](https://6.docs.plone.org/volto/contributing/index.html).
- [x] I successfully ran [code linting checks](https://6.docs.plone.org/volto/contributing/linting.html) on my changes locally.
- [x] I successfully ran [unit tests](https://6.docs.plone.org/volto/contributing/testing.html) on my changes locally.
- [x] I successfully ran [acceptance tests](https://6.docs.plone.org/volto/contributing/acceptance-tests.html) on my changes locally.
- [ ] If needed, I added new tests for my changes.
- [ ] If needed, I added [documentation](https://6.docs.plone.org/volto/contributing/documentation.html#narrative-documentation) for my changes, either in the Storybook or narrative documentation.
- [x] I included a [change log entry](https://6.docs.plone.org/contributing/index.html#contributing-change-log-label) in my commits.

-----

This PR is linked to the requirements for a Vitest test coverage requested by @sneridagh and @wesleybl in a previous PR

Linked to #7839 

-----

### Notes

- No functional changes to the build logic
- Added a Vitest test
- Fixed import path for previous Jest test
- Runs on my local machine and shows successful

image

PR #7919, opened 2026-02-19
Add changelog check for volto-razzle

davisagli:
I noticed in https://github.com/plone/volto/pull/7868 that CI did not catch that the changelog was in the wrong folder.


sneridagh:
@davisagli My initial thought was that we would never have to release any of the razzle packages (I still think we should not), and that we were not to change really anything worth a changelog. However, time has proved me wrong. Let's have it then, and maybe, have an internal (not npm) release from time to time.

sneridagh (approved):

PR #5511, opened 2023-12-08
Reorganize README, merging content into authoritative locations

stevepiercy:
Closes #5437

See also https://github.com/collective/awesome-volto/pull/17

Issue #5437, opened 2023-11-24
[docs] Reorganize content in README.md


stevepiercy:
There is marketing content in the README.md. READMEs should focus on the developer. As it is, a reader must scroll down to the end of the README to learn how to contribute. Additionally some of the content belongs in the Documentation which would reduce maintenance and redundancy. I would suggest the following reorganization.

# Volto

No change.

## Introduction

No change.

## Demo

No change.

## Create a Volto project

No change.

## Contributing to Volto

Rename to "Contributing".

Add a sentence.

`For newcomers to Volto, Plone, or open source software, you must read and follow [First-time contributors](https://6.docs.plone.org/contributing/first-time.html).`

## Documentation

No change.

## Supported Plone and Python versions

Move to https://6.docs.plone.org/volto/contributing/version-policy.html#plone-and-plone-rest-api, merge content in destination, and provide a link to its new location.

Also include this sentence:

`See the [Plone Release Schedule](https://plone.org/download/release-schedule) for details of maintenance and support.`

## Supported Node.js versions

Move content to https://6.docs.plone.org/volto/contributing/version-policy.html#node-js, and provide a link to its new location.

## Supported browsers

Move to just after https://6.docs.plone.org/volto/contributing/version-policy.html#node-js, and provide a link to its new location.

## Known good set of versions for backend packages

Move to https://6.docs.plone.org/volto/contributing/version-policy.html#plone-and-plone-rest-api, merge content in destination. Remove heading, no reference link.

## Volto in Production

Move content to [Volto Awesome](https://github.com/collective/awesome-volto). Merge content into https://github.com/collective/awesome-volto#open-source-websites-built-with-volto, avoid duplication, and create a new section "Other websites built with Volto" for the non-open source websites.

## Contributors

No change.

## License

No change.

fredvd:
>There is marketing content in the README.md. READMEs should focus on the developer. 

The main reason there is 'marketing' and generic information in this readme, as is in the README 's of Plone/ Products.CMFlone and volto repo is that it is also used to generate the homepage page on package repositories like pypi or in this case npmjs.com:   https://www.npmjs.com/package/@plone/volto

These sites are indexed and have reach and visibility from people searching for packages. And it is not only developers who see/find those homepages.

stevepiercy:
[...]

PR #7929, opened 2026-02-20
Remove link to Jobfamilie MEDICE: site is no longer a Volto site (redirects to non-Volto destination)

wesleybl:
Redirects: https://jobfamilie.medice.de/de --> https://career.medice-health-family.com/

https://career.medice-health-family.com is not Volto.

Fix: https://github.com/plone/volto/actions/runs/22220379480/job/64274050086

### DIFF
--- a/packages/volto/news/7929.documentation
+++ b/packages/volto/news/7929.documentation
@@ -1,0 +1,1 @@
+Remove link to Jobfamilie MEDICE: site is no longer a Volto site (redirects to non-Volto destination). @wesleybl


stevepiercy:
This also needs to be forward ported to https://github.com/collective/awesome-volto/ because we often copy-paste from that authoritative source.

davisagli:
Could we remove this section in the volto readme and replace it with a link to that repo? I'm tired of trying to maintain multiple copies.

stevepiercy:
@davisagli [I asked for that before](https://github.com/plone/volto/issues/5437), but [got vetoed](https://github.com/plone/volto/pull/5511#pullrequestreview-1776850702) (expand @tisto comment to see discussion).

sneridagh:
@stevepiercy @davisagli @tisto I think that since we have https://plone.org/why-plone/they-use-plone/they-use-plone-6 we could remove it from the repo, and link to that plone.org page (awesome-volto is _not_ the place). And make sure that https://plone.org/why-plone/they-use-plone/they-use-plone-6 gets updated often, of course. Do we have a procedure for doing it?

stevepiercy:
@sneridagh I'll check with @plone/marketing-communications team. I don't know if there is a specific person assigned to update plone.org on a periodic basis or whenever Awesome Volto is updated. That page was last updated on 8/16/2024.

Screenshot 2026-02-23 at 4 04 58 AM

We discussed this at Axolote 2024 with @skleinfeldt, but it appears to have fallen by the wayside.

@astridjuizi @rioksane is this something that you would take up on a regular basis?

sneridagh:
@astridjuizi @rioksane @stevepiercy It is important that this is provided from an "official" reliable source, eg. plone.org. So decision makers and other official processes (like tenders, etc) have a canonical way to check this information.

We would need in place (eg. a form in the same page) to request to be included in the list. We could easily setup a scheduled CI task in plone.org repo that checks via leechee the link integrity contained in that page too.

astridjuizi:
This falls under the "Showcase Plone better" imo 

There's no-one currently checking any of the "they use..." places, except manually when someone points something out, like in this case. @acsr is looking into automating it but automation is not in place yet. For the new plone.org, let's relook at why these lists exist, how can we move to a SSOT with automation, and where this info should be visible. 

Re "@astridjuizi @rioksane is this something that you would take up on a regular basis?" I'll ask in the Marketing team meeting tomorrow which team member wants to take on and has capacity for this right now. 

adding @fosten for input

stevepiercy:
Thanks, @astridjuizi.

For context, currently https://github.com/collective/awesome-volto/ is the authoritative source of [Websites built with Volto](https://github.com/collective/awesome-volto/?tab=readme-ov-file#websites-built-with-volto) and [Open-source websites built with Volto](https://github.com/collective/awesome-volto/?tab=readme-ov-file#open-source-websites-built-with-volto).

Both of those sections' lists get copy-pasted into the two Volto `README.md`s on a regular basis. They were supposed to get copy-pasted into plone.org, but that stopped happening. Each section has a blurb stating this, and the blurb will also need to be modified according to the new set up.

I'll move this comment under the plone.org repo as a new issue, so that the discussion can continue there and not get lost in this repo.

acsr:
@astridjuizi @stevepiercy @rioksane et al: I am working here on the crawler code using crawlee.python : https://gitlab.evenios.org/plone-public/plone-isvolto-crawler . Maybe we can do something during the sprint on that. It can be forked to github and we can use github actions to automate it or do we have better infrastructure for reoccuring tasks? We should use some supervisior level monitoring outside of github to monitor e.g. github.

PR #7932, opened 2026-02-20
Remove link to Jobfamilie MEDICE: site is no longer a Volto site (redirects to non-Volto destination)

wesleybl:
Backport of  #7929 to Volto 18.


davisagli:
Maybe we should just remove this entire section on old branches so we don't have to maintain it?

stevepiercy:
We need to maintain it because of https://www.npmjs.com/package/@plone/volto?activeTab=versions

stevepiercy:
We also need to update the authoritative source at https://github.com/collective/awesome-volto/ from which these entries are copy-pasted on a regular basis.

PR #7936, opened 2026-02-22
Add file size upload limit validator.

robgietema:
> [!CAUTION]

-----

-----

Closes #

-----

### DIFF
--- a/packages/volto/locales/af/LC_MESSAGES/volto.po
+++ b/packages/volto/locales/af/LC_MESSAGES/volto.po
@@ -4668,6 +4668,11 @@ msgstr ""
 msgid "delete"
 msgstr ""
 
+#. Default: "delete {type} block"
+#: components/manage/Blocks/Block/EditBlockWrapper
+msgid "delete_block"
+msgstr ""
+
 #. Default: "You are using {browsername} {browserversion} which is deprecated by its vendor. That means that it does not get security updates and it is not ready for current modern web features, which deteriorates the user experience. Please upgrade to a modern browser."
 #: components/theme/OutdatedBrowser/OutdatedBrowser
 msgid "deprecated_browser_notice_message"
@@ -4691,6 +4696,11 @@ msgstr ""
 #. Default: "Draft"
 #: components/manage/Contents/ContentsItem
 msgid "draft"
+msgstr ""
+
+#. Default: "drag {type} block"
+#: components/manage/Blocks/Block/EditBlockWrapper
+msgid "drag_block"
 msgstr ""
 
 #. Default: "Input must be valid email (something@domain.com)"
--- a/packages/volto/locales/ar/LC_MESSAGES/volto.po
+++ b/packages/volto/locales/ar/LC_MESSAGES/volto.po
@@ -4668,6 +4668,11 @@ msgstr ""
 msgid "delete"
 msgstr ""
 
+#. Default: "delete {type} block"
+#: components/manage/Blocks/Block/EditBlockWrapper
+msgid "delete_block"
+msgstr ""
+
 #. Default: "You are using {browsername} {browserversion} which is deprecated by its vendor. That means that it does not get security updates and it is not ready for current modern web features, which deteriorates the user experience. Please upgrade to a modern browser."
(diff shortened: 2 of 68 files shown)


robgietema:
> We already have validateFileUploadSize which is used in a number of places (
> 
> https://github.com/plone/volto/blob/4f930a06470b5ddaed6509af92417ddd04f10023/packages/volto/src/helpers/FormValidation/FormValidation.jsx#L337
> 
> )
> Please update or replace that instead of just adding an additional validator that isn't used everywhere and doesn't honor the existing default maxFileUploadSize setting.
> 
> (I do like the idea of supporting a per-field size limit)

These are 2 different issues; the validator I build is checking the limit of a single field which is set by the schema (or the editor in the case of volto-form-block). This will give the user feedback before submitting the form (inline validation). The validation you linked checks the total amount of data send of all the files combined and is not on a per field basis. The per field check can not (and should not) be done in this validator since it does not have access to the needed data and this also breaks the way we handle form validation.

davisagli:
@robgietema @sneridagh From what I see, validateFileUploadSize is NOT checking the total size of all files. It's checking one file at at time against config.settings.maxFileUploadSize. And it also does the validation as soon as a file is selected or dropped, not when the form is submitted.

Unless I'm missing something, I see a way to adjust validateFileUploadSize to also handle the per-field size limit:
1. add an optional size arg to validateFileUploadSize. if provided, check it instead of config.settings.maxFileUploadSize
2. pass size to validateFileUploadSize from the places where we do have a schema (i.e. widgets)

That way we keep the validation code in one place and it's a bit clearer how we apply both limits together.

That's my preference, but I also see that the new validator is not really in conflict with the existing one (if we have both, the lower limit applies), and I guess I'm okay with adding it if you really prefer it that way.

In that case there are still a few details to take care of:
1. Submit this first to the main branch, then backport.
2. Documentation for the new `size` setting on schema fields.
3. Change the fallback size limit to use config.settings.maxFileUploadSize instead of being hardcoded to 7000000

robgietema:
1. I made a PR to main also: https://github.com/plone/volto/pull/7960
2. Agree, I've added the validator to the list in the docs
3. I've updated the fallback to use the config setting

packages/volto/src/helpers/MessageLabels/MessageLabels.js
```diff
@@ -25,6 +25,10 @@ export const messages = defineMessages({
     id: 'Maximum value is {len}.',
     defaultMessage: 'Maximum value is {len}.',
   },
+  maxSize: {
+    id: 'Maximum size is {maxSize} bytes, file is {size} bytes.',
```
stevepiercy: Although slightly more verbose, I think a few more words improves clarity.

I also wonder if we should prompt the user what to do, for example, "Either make the file size smaller, break it into smaller files, or contact your site administrator." What do you think?

```suggestion
    id: 'Maximum file size is {maxSize} bytes, but the uploaded file is {size} bytes.',
```

davisagli: Generally "break it into smaller files" is not an option, since this is feedback for a single file field.

packages/volto/src/config/validation.ts
```diff
@@ -33,6 +34,13 @@ const registerValidators = (config: ConfigType) => {
     method: maxLengthValidator,
   });
 
+  config.registerUtility({
+    name: 'size',
+    type: 'validator',
+    dependencies: { fieldType: 'object' },
```
davisagli, line 40: Does this mean it will run for all fields with a JSON-schema type of `object`, including objects that are not files, and then break when the value doesn't have a `size`?

robgietema: No

sneridagh: I created a test for that, and I checked, the JSONFields are not serialized as `object` but as `dict` :/

sneridagh (approved):

packages/volto/src/config/validation.ts
davisagli: I see: it does run, but it passes validation because value.size > maxSize evaluates to false when value.size is undefined

PR #7946, opened 2026-02-25
Handle create-user 500 responses where body contains 'message' instead of 'error'

wesleybl:
When the server returns HTTP 500 during user creation (for example due to an SMTP MailHost error), the response body may not include an 'error' key and instead contains a top-level 'message'. This caused a runtime TypeError when trying to read 'response.body.error.message'.

This change makes the client tolerant of that shape by extracting 'response.body.message' when 'response.body.error' is absent, falling back to other error fields. It prevents reading properties of undefined and ensures a readable error string is shown to the user.

### DIFF
--- a/packages/volto/src/components/manage/Controlpanels/Users/UsersControlpanel.jsx
+++ b/packages/volto/src/components/manage/Controlpanels/Users/UsersControlpanel.jsx
@@ -25,6 +25,7 @@ import { Link } from 'react-router-dom';
 import Helmet from '@plone/volto/helpers/Helmet/Helmet';
 import { messages } from '@plone/volto/helpers/MessageLabels/MessageLabels';
 import { isManager, canAssignGroup } from '@plone/volto/helpers/User/User';
+import { getErrorMessage } from '@plone/volto/helpers/Utils/Utils';
 import clearSVG from '@plone/volto/icons/clear.svg';
 import addUserSvg from '@plone/volto/icons/add-user.svg';
 import saveSVG from '@plone/volto/icons/save.svg';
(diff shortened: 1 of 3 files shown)


packages/volto/src/components/manage/Controlpanels/Users/UsersControlpanel.jsx
```diff
@@ -300,6 +300,23 @@ const UsersControlpanel = () => {
...
+    const respBody = error.response?.body;
+    if (respBody?.error?.message) return respBody.error.message;
+    if (respBody?.message) return respBody.message;
+    if (error.message) return error.message;
+    try {
+      return JSON.stringify(error);
+    } catch (e) {
+      return String(error);
+    }
+  };
```
davisagli: This looks like a helper that could be useful throughout Volto, not only in the users controlpanel.

davisagli (approved):
Thanks.

I think the root of the problem here is that the `@users` endpoint was using a non-standard way of serializing errors (https://github.com/plone/plone.restapi/blob/main/src/plone/restapi/services/users/add.py#L291) -- which uses "error" -- instead of raising an exception and letting it get serialized by the exception view in plone.rest (https://github.com/plone/plone.rest/blob/main/src/plone/rest/errors.py) -- which uses "message"

We should update the users service to raise exceptions the standard way. But in order to do that smoothly we have to accept both ways here, so this is good.

I'm making a suggestion to move this into helpers so we can use it from other places in Volto too. I'll let you decide whether to do that or merge it as is.

packages/volto/src/components/manage/Controlpanels/Users/UsersControlpanel.jsx
wesleybl: @davisagli done.

davisagli (approved):

PR #7950, opened 2026-02-26
fix(richTextWidget): fixed the invalid nesting of data in richtextwid…

nileshgulia1:
…get view

When used with a metadata block, the widget view renders nothing due to invalid nesting of `<p>` inside `<p>` tags. However I see it working in storybook but fails in our projects.


sneridagh:
@nileshgulia1 shouldn't we fix it then in the metadata block? Also this is a breaking if we make it so.

nileshgulia1:
@sneridagh The metadata block just maps the correct widget view. We're not sanitizing input here: https://github.com/eea/volto-metadata-block/blob/a731a9ec97b975ed90870924222e1998f4bb93c1/src/components/manage/Blocks/Metadata/View.jsx#L25

even if we do, since we are amending html inside a `<p>` which contains a `<p>` as its a richtext, the resulting html becomes invalid. Browsers should destrucutre that in many cases but It seems to have issues with current state.

davisagli:
@sneridagh Well, it's the right time to break it. It makes sense to me that the rich text widget should not be wrapped with a `p`, since it could also contain one.

davisagli (approved):

PR #7966, opened 2026-03-05
Remove tsconfig root

wesleybl:
Remove the `include` and `exclude` from `packages/tsconfig/base.json`, since the root directory extends from it and we don't have `packages/tsconfig/src`.

Add the `include` and `exclude` to the root tsconfig file, with the correct paths.

This fixes the following error that was displayed by VS Code in the root tsconfig:

```
The error message "No inputs were found in config file '/home/user/git/volto/tsconfig.json'. Specified 'include' paths were '["packages/tsconfig/src"]' and 'exclude' paths were '["packages/tsconfig/node_modules","packages/tsconfig/build","packages/tsconfig/public","packages/tsconfig/coverage"]'
```


davisagli:
I'll leave this one for @sneridagh to take a look, because I don't know much about Typescript or how we are using it.

wesleybl:
@sneridagh Could you take a look here, please?

wesleybl:
@plone/volto-team Could someone please take a look at this?

davisagli:
@wesleybl I think this probably makes sense, but I can't reproduce the error you saw in VSCode

wesleybl:
@davisagli It seems that the new versions of VS Code only show errors in `tsconfig.json` if it is "active". Create a `test.ts` file in the repository root and it should show the error.

In any case, the error exists. The current configuration doesn't make sense.

wesleybl:
@davisagli Looking again, I realized that all packages that use TypeScript already have their own `tsconfig.json` file. So the `tsconfig.json` file in root could be simple. Just a marker indicating that the project is TypeScript. It wouldn't need include and it wouldn't need to extend `tsconfig/base.json`. I'll implement that later.

sneridagh:
@wesleybl I don't get the error you mentioned, and removing things from the base.json, will have consequences on all packages.

wesleybl:
> I don't get the error you mentioned

@sneridagh See my comment: https://github.com/plone/volto/pull/7966#issuecomment-4270505777

> I would remove completely tsconfig.json from the root. In Seven, we do not have one.

I think we can have a `tsconfig.json` file in the root directory, with minimal configuration, just to indicate to TypeScript that we have TypeScript projects here and to serve as a reference for linters and IDEs. It can also serve as a starting point for reusable configurations in the future, and for package path configurations.

I think we should also add one to the cookieplone templates.

>  My intention with packages/tsconfig was to have a workable shared config. We do use it in Seven, but that has been never the case in Volto. We could even remove it.

I don't see it being used here or  Seven. So we can definitely remove it. Generally, reusable configurations are in the root tsconfig, not in a package.

wesleybl:
@davisagli @sneridagh done.

wesleybl:
@sneridagh @davisagli VSCode also complains about `tsconfig.json` with:

```json
"files": []
```

So I decided to remove it. I searched in other monorepos and it's actually common practice not to have tsconfig.json in the root directory. For example:

https://github.com/facebook/react

PR #7967, opened 2026-03-05
Fix null port in virtual hosting path passed to backend by devproxy

davisagli:
I think this is why we're getting bad URLs from the backend on https://demo.plone.org/next/


sneridagh:
@davisagli we left it because then we know that something is wrong configured in the reverse proxy, but I'm ok on exploring it if it improves things.

PR #7970, opened 2026-03-06
razzle.config.js: Fix defaultPlugins missing name property to enable addon SCSS plugin replacement

wesleybl:
Backport of #7908 to Volto 18.

### DIFF
--- a/packages/volto/razzle.config.js
+++ b/packages/volto/razzle.config.js
@@ -420,11 +420,26 @@ const defaultModify = ({
 const addonExtenders = registry.getAddonExtenders().map((m) => require(m));
 
 const defaultPlugins = [
-  { object: require('./webpack-plugins/webpack-less-plugin')({ registry }) },
-  { object: require('./webpack-plugins/webpack-svg-plugin') },
-  { object: require('./webpack-plugins/webpack-bundle-analyze-plugin') },
-  { object: require('./jest-extender-plugin') },
-  'scss',
+  {
+    name: 'less',
+    object: require('./webpack-plugins/webpack-less-plugin')({ registry }),
+  },
+  {
+    name: 'svg',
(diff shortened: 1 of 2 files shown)


packages/volto/razzle.config.js
```diff
@@ -420,11 +420,26 @@ const defaultModify = ({
...
+    object: require('./webpack-plugins/webpack-bundle-analyze-plugin'),
+  },
+  {
+    name: 'jest-extender',
+    object: require('./jest-extender-plugin'),
+  },
+  {
+    name: 'scss',
+    object: require('razzle-plugin-scss'),
+  },
```
wesleybl, line 442: When the string 'scss' is used, the plugin used is `razzle-plugin-scss`. See:

https://github.com/jaredpalmer/razzle/blob/master/packages/razzle/config/loadPlugins.js#L41

It makes it explicit so that the `name` can be added.

davisagli: @wesleybl Doesn't that return earlier because `object` is set? https://github.com/jaredpalmer/razzle/blob/master/packages/razzle/config/loadPlugins.js#L26

wesleybl: @davisagli I replaced the string `scss` with `{ name: 'scss', object: require('razzle-plugin-scss')}`. When using the string `scss`, we won't have the `object` attribute, and it will bypass the `if` statement you pointed out and use the `razzle-plugin-scss` plugin.

What I meant is that even changing it to `{ name: 'scss', object: require('razzle-plugin-scss')}`, the final result will be the same. The same plugin will be used in both situations. It's a transparent modification.

davisagli: @wesleybl Yep, I was reading the code wrong.

davisagli (approved):

PR #7552, opened 2025-10-27
refactor: visually-hidden class

Wagner3UB:
Move .visually-hidden class to main.less and remove unused utils.less and references

Motivation:
The utils.less file has not been imported by main.less since this PR (https://github.com/plone/volto/issues/1265), about 5 years ago. As a result, the .visually-hidden class had no effect.

What was done:
1. Removed the utils.less file and all references to it.
2. Moved the .visually-hidden class to the main.less file, placing it in the most appropriate section for a generic utility class.

Reference: https://github.com/plone/volto/pull/6356

PR #7978, opened 2026-03-09
feat: add visually-hidden class

Wagner3UB:
This PR adds a visually-hidden class which is useful in situations where you want to provide additional information to Assistive Technology and an aria-label may not be appropriate.

Backport from: 
https://github.com/plone/volto/pull/6356
https://github.com/plone/volto/pull/7552


davisagli:
The conclusion of the discussion on https://github.com/plone/volto/pull/6356#issuecomment-3444888562 was that this is a breaking change that cannot be backported to Volto 18.

PR #7980, opened 2026-03-09
DatetimeWidget - react-dates and rc-time-picker components

Wagner3UB:

- These components do not natively accept ARIA attributes, meaning it is not possible to add the necessary properties via props to make them accessible when they are required.
- It was necessary to perform a scan when these components are rendered, looking for fields that can be modified by users and marking them as required.
- Unfortunately, the code ended up not being small due to the need to externally search for and apply these attributes to the elements.


packages/volto/news/7980.bugfix
```diff
@@ -0,0 +1 @@
+Requirement aria attributes added to date and time inputs on 'DatetimeWidget' - @Wagner3UB
```
stevepiercy: It's an acronym.

```suggestion
Added required ARIA attributes to date and time inputs on `DatetimeWidget`. @Wagner3UB
```

stevepiercy (changes requested):
Needs a technical review.

packages/volto/src/components/manage/Widgets/DatetimeWidget.jsx
```diff
@@ -165,68 +162,128 @@ const DatetimeWidgetComponent = (props) => {
...
+        if (timeInput) timeInput.setAttribute('aria-required', 'true');
+      }
+    }
+
+    // Apply immediately
+    applyAria();
+
+    // Observe DOM changes since rc-time-picker and react-dates can recreate inputs
+    const observer = new MutationObserver(() => applyAria());
+    observer.observe(document.body, { childList: true, subtree: true });
```
wesleybl: Passing `document.body` here causes `applyAria` to be called every time the DOM is modified. It's better to create a `useRef` to monitor only the widget container.

packages/volto/src/components/manage/Widgets/DatetimeWidget.jsx
```diff
@@ -165,68 +162,128 @@ const DatetimeWidgetComponent = (props) => {
...
+
+    // Cleanup on unmount
+    return () => observer.disconnect();
+  }, [props.required, id, isDateOnly]);
```
wesleybl: We should use `renderWidget` inside `useEffect` and return if it's false. When `renderWidget` is false, we don't need to do anything with `useEffect`.

packages/volto/src/components/manage/Widgets/DatetimeWidget.jsx
Wagner3UB: Great, thank you! I did the same for other parts of the code too.

packages/volto/src/components/manage/Widgets/DatetimeWidget.jsx
Wagner3UB: Make total sense, I'll do the changes.

wesleybl:
@Wagner3UB In `SingleDatePicker` you can pass the `required` prop. And according to the HTML attribute mapping documentation for accessibility attributes, the `required` attribute is mapped to `aria-required`:

https://www.w3.org/TR/html-aam-1.0/#html-attribute-state-and-property-mappings

So wouldn't simply using `required` be enough?

Wagner3UB:
> @Wagner3UB In `SingleDatePicker` you can pass the `required` prop. And according to the HTML attribute mapping documentation for accessibility attributes, the `required` attribute is mapped to `aria-required`:
> 
> https://www.w3.org/TR/html-aam-1.0/#html-attribute-state-and-property-mappings
> 
> So wouldn't simply using `required` be enough?

@wesleybl Hey! Thanks for showing me, you're right. I'm so used to working with components that don't support accessibility props that I didn't even consider it could be a native feature. I've also made a new change, I removed the aria-invalid attributes. They aren't really necessary since these fields are pre-filled 100% of the time. However, I kept the required attribute, as alerting the user that the field is mandatory is still useful, even when it's pre-compiled.

packages/volto/src/components/manage/Widgets/DatetimeWidget.jsx
```diff
@@ -165,68 +166,97 @@ const DatetimeWidgetComponent = (props) => {
   const datetime = getInternalValue();
   const isDateOnly = getDateOnly();
 
+  // aria-required for the time input (rc-time-picker is lazy-loaded,
+  // so MutationObserver is needed to catch when it mounts its input)
```
wesleybl, line 170: I think it's worth adding to this comment why it was done this way. Something like:

`rc-time-picker does not support aria-required via props, so we apply it imperatively.`

packages/volto/src/components/manage/Widgets/DatetimeWidget.jsx
```diff
@@ -165,68 +166,97 @@ const DatetimeWidgetComponent = (props) => {
...
+  useEffect(() => {
+    if (!renderWidget || isDateOnly) return;
+
+    function applyTimeAria() {
+      const input = timeInputRef.current?.querySelector('input');
```
wesleybl, line 179: Since this can break without a visual indication, I think it's a good idea to add a test at:

https://github.com/plone/volto/blob/main/packages/volto/src/components/manage/Widgets/DatetimeWidget.test.jsx

Just copy the existing test and pass `required true` to the component.

Wagner3UB: @wesleybl Nice catch, thank you!

packages/volto/src/components/manage/Widgets/DatetimeWidget.test.jsx
```diff
@@ -66,3 +66,32 @@ test('datetime widget converts UTC date and adapts to local datetime', async ()
...
+  );
+
+  // Wait for the lazy-loaded TimePicker to be mounted in the DOM
+  await waitFor(() => screen.getByPlaceholderText('Time'));
+
+  // The rc-time-picker doesn't support aria-required natively,
+  // so we verify if our MutationObserver/useEffect successfully injected it.
+  const timeInput = container.querySelector('.time-input input');
+
+  expect(timeInput).toHaveAttribute('aria-required', 'true');
```
wesleybl, line 122: You should also test the `required` attribute in the `SingleDatePicker` input.

wesleybl (approved):

PR #7982, opened 2026-03-09
FileWidget - a11y

Wagner3UB:
- kept the label inside the Dropzone area with `aria-hidden="true"` and `stopPropagation` on click.
- Added a label when required
- Added a correct identification for a field with required props

IMPORTANT ABOUT THE FileWidget && RegistryImageWidget:
It is not correct to use aria-required on a button. In addition, the entire select field has a poor a11y structure, there isn’t a single element that correctly indicates that this field is a select/input field. To properly fix the issue, the entire component would need to be replaced with an accessible version, but this is beyond the scope of this PR.


packages/volto/src/components/manage/Widgets/FileWidget.jsx
```diff
@@ -205,18 +205,24 @@ const FileWidget = (props) => {
...
+                  : null
+              }
+            >
               {value
                 ? intl.formatMessage(messages.replaceFile)
                 : intl.formatMessage(messages.addNewFile)}
-              <span className="visually-hidden">
-                {intl.formatMessage(messages.dragAndDropActionA11y)}
-              </span>
-            </label>
```
giuliaghisini, line 215: this span visually-hidden are we sure we want to delete it?

packages/volto/src/components/manage/Widgets/FileWidget.jsx
```diff
@@ -205,18 +205,24 @@ const FileWidget = (props) => {
               </div>
             )}
 
-            <label className="label-file-widget-input">
+            <Button
+              className="label-file-widget-input"
+              tabIndex={-1}
+              aria-label={
+                props.required
+                  ? `${value ? intl.formatMessage(messages.replaceFile) : intl.formatMessage(messages.addNewFile)}. (${intl.formatMessage(messages.requiredField)})`
```
giuliaghisini: i think this condition is not correct. 
If field is required you write "Replace file" also if nothing is set. 
If field is not required you always write "Add a new file" also if file is set.

giuliaghisini (changes requested):

packages/volto/src/components/manage/Widgets/FileWidget.jsx
Wagner3UB: Fair point. It was an attempt to apply aria-label in another spot, but I see it’s required here. Thanks for the heads-up.

packages/volto/src/components/manage/Widgets/FileWidget.jsx
```diff
@@ -205,30 +227,41 @@ const FileWidget = (props) => {
               </div>
             )}
 
-            <label className="label-file-widget-input">
+            <Button
+              className="label-file-widget-input"
+              tabIndex={-1}
```
giuliaghisini: i'm not sure button with tabindex=-1 is correct here. Maybe label was correct and only needed to be associated with input with attribute `for`?

giuliaghisini (changes requested):

packages/volto/src/components/manage/Widgets/FileWidget.jsx
Wagner3UB: Good point on using a label, I've switched to it. However, using htmlFor without extra handling caused the file dialog to open twice on click (once from the label activating the input, once from the Dropzone container). To avoid that, I kept the label inside the Dropzone area with aria-hidden="true" and stopPropagation on click. This makes it a visual-only affordance, while keyboard access is handled exclusively by the parent div[role="button"], avoiding the duplicate path that the original tabIndex={-1} was meant to prevent.

packages/volto/src/components/manage/Widgets/FileWidget.jsx
```diff
@@ -171,6 +179,16 @@ const FileWidget = (props) => {
     reader.readAsDataURL(files[0]);
   };
 
+  const statusTextA11y = [
+    intl.formatMessage(messages.dragAndDropActionA11y), // Interaction instructions
+    props.required &&
+      `${props.title}: ${intl.formatMessage(messages.requiredField)}`, // Required field status
```
wesleybl: The destination label for `aria-labelledby` is already the title ("File"). So there's no need to repeat `props.title` here. Otherwise, the reader will say "File" twice. And this string here is already long (at least in Portuguese).

packages/volto/src/components/manage/Widgets/FileWidget.jsx
```diff
@@ -171,6 +179,16 @@ const FileWidget = (props) => {
...
+    props.error?.length && props.error.join(' '), // Validation error messages
```
wesleybl: Errors should be placed in a separate element with `role="alert"`. Otherwise, the reader is "blind" if an error occurs dynamically.

wesleybl: To avoid a rushed reading experience, they should be joined using `join('. ')`.

packages/volto/src/components/manage/Widgets/FileWidget.jsx
```diff
@@ -171,6 +179,16 @@ const FileWidget = (props) => {
...
+    value?.filename, // Current file name if a file is uploaded
+  ]
+    .filter(Boolean)
+    .join(' ');
```
wesleybl: `join('. ')`.

packages/volto/src/components/manage/Widgets/FileWidget.jsx
```diff
@@ -60,8 +60,16 @@ const messages = defineMessages({
...
-    defaultMessage: 'Press Enter to browse files from your computer.',
+    id: 'File upload area. Press Enter to open the file browser',
+    defaultMessage: 'File upload area. Press Enter to open the file browser',
+  },
+  requiredField: {
+    id: 'This field is required.',
+    defaultMessage: 'This field is required.',
+  },
+  downloadFile: {
+    id: 'field.file.downloadFile',
```
wesleybl: It's best to use the default as `id`, which is how it's used in most Volto translations.

wesleybl (changes requested):

packages/volto/src/components/manage/Widgets/FileWidget.jsx
Wagner3UB: Good point! I've separated the error messages into a dedicated <span  role="alert"> element, keeping it apart from the static aria-describedby status span. This way, errors are announced dynamically by screen readers when they appear, without triggering a re-read of the full status text (instructions, required state, filename). I've also switched to join('. ') so multiple errors are read with natural pauses between them.

wesleybl (approved):

Wagner3UB:
@giuliaghisini Hey! I need your approve on this one to merge. Thanks!

packages/volto/src/components/manage/Widgets/FileWidget.jsx
```diff
@@ -205,34 +229,52 @@ const FileWidget = (props) => {
...
+              </span>
+            )}
             <input
               {...getInputProps({
                 type: 'file',
                 style: { display: 'none' },
               })}
               id={`field-${id}`}
+              aria-required={props.required}
+              aria-invalid={props.error?.length > 0}
```
giuliaghisini: Only one question: if the field is not invalid, is the attribute still set?

If so, even if it has the value `false`, the browser will still consider it an invalid field.

The attribute should not be visible if the condition is false.

giuliaghisini (changes requested):

packages/volto/src/components/manage/Widgets/FileWidget.jsx
Wagner3UB: Very interesting indeed. Some screen readers announce aria-invalid="false" as 'invalid data, false', which can be very confusing. So it should be 'true' or just not be there. Thank you.

packages/volto/src/components/manage/Widgets/FileWidget.jsx
```diff
@@ -205,34 +229,52 @@ const FileWidget = (props) => {
...
+                {errorTextA11y}
+              </span>
+            )}
             <input
               {...getInputProps({
                 type: 'file',
                 style: { display: 'none' },
               })}
               id={`field-${id}`}
+              aria-required={props.required}
```
giuliaghisini: i think it's the same for aria-required @Wagner3UB

giuliaghisini (changes requested):

packages/volto/src/components/manage/Widgets/FileWidget.jsx
Wagner3UB: @giuliaghisini Done

giuliaghisini (approved):

PR #7986, opened 2026-03-10
Update to webpack 5.105.4

davisagli:
Replaces https://github.com/plone/volto/pull/7882

### DIFF
--- a/packages/volto-razzle/package.json
+++ b/packages/volto-razzle/package.json
@@ -35,7 +35,7 @@
     "dotenv": "^8.2.0",
     "dotenv-expand": "^5.1.0",
     "esbuild": "0.25.12",
-    "file-loader": "^4.3.0",
+    "file-loader": "^6.2.0",
     "fs-extra": "^9.1.0",
     "inquirer": "^7.3.3",
     "mri": "^1.1.4",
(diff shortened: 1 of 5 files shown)


packages/volto-razzle/package.json
```diff
@@ -57,8 +57,8 @@
     "terminate": "^2.1.2",
     "terser-webpack-plugin": "5.3.6",
     "tiny-async-pool": "^1.1.0",
-    "url-loader": "^2.3.0",
-    "webpack-manifest-plugin": "^3.1.1",
+    "url-loader": "^4.1.1",
+    "webpack-manifest-plugin": "^4.1.1",
```
wesleybl, line 61: Version 6 didn't work?

davisagli: It says it no longer supports webpack 4, which volto-razzle still claims to support. I'd like to drop support for Webpack 4 but I didn't spend time looking at that yet.

wesleybl: I understand your reasoning, but I wouldn't worry about it. Nobody should be using @plone/razzle outside of Volto.

davisagli: Probably true. Anyway, I will merge this and then we can finish removing support for Webpack 4 as a next step.

PR #7987, opened 2026-03-10
Docs: remove TanStack terminology and add httpexample

### LINKED ISSUE #7947: Improve `@plone/client` documentation
This issue pertains to `@plone/cllient`. It has several parts that can be worked on in separate pull requests. The affected files are primarily in https://github.com/plone/volto/tree/main/docs/source/client, but might be in other parts of the Volto documentation, too. If you find other places, please add them to this issue.

- [ ] Remove any reference to TanStack Query and its terminology, including mutators, useQuery, hooks, and any other of its terms. This task requires that the remaining narrative text is accurate, has correct English grammar, and makes sense to the reader.
- [ ] For HTTP API calls, use the lovely [`sphinxcontrib.httpexample`](https://sphinxcontrib-httpexample.readthedocs.io/en/latest/index.html) extension, thanks to @datakurre and @Hrittik20's work. This requires a configuration change to the Volto documentation's file `docs/source/conf.py`, adding it as a dependency to build the docs, and then implement HTTP requests following the examples from `sphinxcontrib.httpexample`'s documentation.
- [ ] Forward port the final result to the `seven` branch.

Hrittik20:
Fixes: #7947 

----
📚 Documentation preview 📚: https://volto--7987.org.readthedocs.build/

----
📚 Documentation preview 📚: https://plone-registry--7987.org.readthedocs.build/

### DIFF
--- a/docs/source/client/actions/content.md
+++ b/docs/source/client/actions/content.md
@@ -2,6 +2,15 @@
 
 Get the data for a specific content given its path.
 
+```{eval-rst}
+.. http:example:: curl wget httpie python-requests
+
+   GET /Plone/front-page HTTP/1.1
+   Host: localhost:8080
+   Accept: application/json
+   Authorization: Basic YWRtaW46c2VjcmV0
+```
+
 ## Get content
 
 ### Query options function
--- a/docs/source/client/future-improvements.md
+++ b/docs/source/client/future-improvements.md
@@ -2,11 +2,9 @@
 
 The following enhancements need to be considered for next iterations:
 
-## Separate out the `react-query` part from `@plone/client` into a separate package
+## Separate out the data‑fetching layer from `@plone/client` into a separate package
 
-This would make the integration cleaner.
-
-We could create wrapper over `@tanstack/react-query` that provides our enhanced `ploneClient` (just like `queryClient` in `@tanstack/react-query`) to the entire package via react-context.
+This would make the integration cleaner and allow the underlying data‑fetching implementation to evolve independently.
 
 ```jsx
 import { PloneClient, PloneClientProvider } from `@plone/client`;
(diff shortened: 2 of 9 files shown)


sneridagh:
@Hrittik20 all this documentation is kind of obsolete now and should be reviewed as a whole with the new status quo. I think we can take a look at it during the sprint next week. 

@stevepiercy what about the new improvements in the httpexample lib?

stevepiercy:
@sneridagh obsolete indeed. I discovered, for example, `createAliasesMutation` causes Sphinx to emit a warning of `volto/docs/source/client/actions/aliases.md:43: WARNING: duplicate function description of createAliasesMutation, other function in client/actions/aliases`, which was due to overloading the signature. This function appears not to exist in `plone/client` anymore.

Support for `plone/client` was added in [v1.2](https://github.com/collective/sphinxcontrib-httpexample/blob/main/CHANGELOG.md#12-2024-08-17), and improved in [v2.0](https://github.com/collective/sphinxcontrib-httpexample/blob/main/CHANGELOG.md#20-2026-02-23). Example at https://sphinxcontrib-httpexample.readthedocs.io/en/2.0/usage.html#http-with-slash-and-version.

The current `plone/client` docs also used a hack to get a suboptimal rendering using `{js:function}`. See http://127.0.0.1:8000/client/actions/aliases.html#add-aliases-for-many-pages. Instead we should use `sphinxcontrib-httpexample` syntax.

We should also consider using external files to use in tests, similar to what we do for `plone.restapi`, if that is possible. Example: https://sphinxcontrib-httpexample.readthedocs.io/en/latest/usage.html#http-get. What do you think?

BTW, Sphinx has at least one limitation with overloaded signatures.

- Shows only one docstring for all the functions, instead of a unique docstring per function: https://github.com/pybind/pybind11-stubgen/issues/221

Issue #7962, opened 2026-03-03
`cache-status` header not forwarded by images middleware


wesleybl:
**Describe the bug**

The images middleware (https://github.com/plone/volto/blob/main/packages/volto/src/express-middleware/images.js) uses an explicit allowlist of headers to forward from the upstream backend to the client. The standard `Cache-Status` header ([RFC 9211](https://www.rfc-editor.org/rfc/rfc9211)) is not included in this list and is silently dropped, making it impossible for reverse proxies (Nginx, Varnish, CDNs) and debugging tools to inspect the cache status of image responses.

The relevant code in [images.js](https://github.com/plone/volto/blob/dd35efb7a66c64b9f6e3e491145ed20e4670da90/packages/volto/src/express-middleware/files.js#L4-L13):

```js
const HEADERS = [
  'accept-ranges',
  'cache-control',
  'content-disposition',
  'content-range',
  'content-type',
  'x-sendfile',
  'x-accel-redirect',
  'x-robots-tag',
  // cache-status is missing here
];
```

Any stack that uses a caching layer compliant with RFC 9211 (e.g. Caddy + Souin) will have the `Cache-Status` header stripped when requests pass through Volto's image middleware, making cache debugging and monitoring effectively impossible for image routes (`@@images`).

**To Reproduce**

1. Set up a Volto + Plone stack with a RFC 9211-compliant cache between Volto and Plone (e.g. Caddy with the Souin plugin)
2. Make a request to any image route, e.g. `/path/to/content/@@images/image`
3. Inspect the response headers (e.g. `curl -I https://yoursite.com/path/@@images/image`)
4. Notice that `Cache-Status` header is absent, even though it is present when the same request is made directly to the Caddy/backend URL bypassing Volto

**Expected behavior**

The `Cache-Status` header should be forwarded to the client, as it is a standardized HTTP header (RFC 9211) and carries no sensitive information. Reverse proxies and monitoring tools rely on this header to determine whether a response was served from cache (`hit`) or fetched from origin (`miss`), along with the remaining TTL.

Expected response headers for image routes:

```
cache-status: Souin; hit; ttl=664; key=https://yoursite.com/@@images/image
cache-control: max-age=3600, public
content-type: image/jpeg
```

**Proposed fix**

Add `cache-status` to the `HEADERS` allowlist in `src/middleware/images.js`:

```js
const HEADERS = [
  'content-type',
  'content-disposition',
  'cache-control',
  'x-sendfile',
[...]

PR #7989, opened 2026-03-10
forward cache-status header in images middleware

Roshan-Pawar:
Closes # [7962](https://github.com/plone/volto/issues/7962)

Fix: Forward Cache-Status header in images middleware

The images middleware maintains an explicit allowlist of response headers forwarded from the backend.
The  Cache-Status header (RFC 9211) was missing from this list, causing it to be stripped from responses.

### DIFF
--- a/packages/volto/src/express-middleware/files.js
+++ b/packages/volto/src/express-middleware/files.js
@@ -10,6 +10,7 @@ const HEADERS = [
   'x-sendfile',
   'x-accel-redirect',
   'x-robots-tag',
+  'cache-status',
 ];
 
 function filesMiddlewareFn(req, res, next) {


davisagli:
From what I see in https://httpwg.org/specs/rfc9211.html#security, leaving out the Cache-Status header might be better from a security perspective.

wesleybl:
@davisagli I think the responsibility for sending or not sending the header lies with the Cache Server or the Web Server. This shouldn't be Volto's responsibility. It's frustrating when the Cache Server sends the header and Volto doesn't forward it. It makes debugging the cache more difficult.

wesleybl:
@Roshan-Pawar To contribute to Volto, you need to sign the Plone Contributor Agreement. See: https://github.com/plone/volto/pull/7989#issuecomment-4032278449

So I'm replacing this PR with: https://github.com/plone/volto/pull/8055

Issue #7993, opened 2026-03-11
Single day event with "open end" checked still needs a end date


TimoBroeskamp:
**Describe the bug**
When creating a single-day event with the "open end" option checked, the end date field is correctly hidden from the UI — but form validation still requires it, preventing submission. Since the end date selector is not visible, users have to uncheck the "open end" option to set a date that is after the start of the event just to check it again and save it.

**To Reproduce**

1. Create an event
2. Give the event a title
3. Set a starting date for the event
4. Check the "whole day" selector
5. Check the "open end" selector
6. Save the event

**Expected behavior**
I expect be able to save the event when I set it to "whole day" and "open end" and don't need to enter an end date.

**Screenshots**

<img width="458" height="532" alt="Image" src="https://github.com/user-attachments/assets/2c9817bb-cb46-4809-933f-f94f82590830" />

<img width="2336" height="1027" alt="Image" src="https://github.com/user-attachments/assets/54a07be9-1bb2-4938-847c-a1373e530982" />

**Software (please complete the following information):**

- OS: Ubuntu
- Browser: Chrome
- Volto Version: Volto 19.0.0-alpha.27
- Plone Version: 6.1.4
- Plone REST API Version: 9.15.4

**Additional context**

stevepiercy:
Confirmed with modified instructions:

3.b. Clear the Event Ends field.

The fix is to comply with the [iCalendar specification for VEVENT](https://datatracker.ietf.org/doc/html/rfc5545.html#section-3.6.1):

> For cases where a "VEVENT" calendar component specifies a "DTSTART" property with a DATE value type but no "DTEND" nor "DURATION" property, the event's duration is taken to be one day.  For cases where a "VEVENT" calendar component specifies a "DTSTART" property with a DATE-TIME value type but no "DTEND" property, the event ends on the same calendar date and time of day specified by the "DTSTART" property.

~~In other words, if someone clears the Event Ends field and checks Open End, then the same date or datetime should be populated in the Event Ends field before submission.~~

~~CORRECTION:~~

~~- If someone clears the Event Ends field and checks both Open End and Whole Day, then the Event Ends field should be populated with the Event Starts datetime plus 1 day.~~
~~- If someone clears the Event Ends field and checks only Open End and doesn't check Whole Day, then the same date or datetime should be populated in the Event Ends field before submission.~~

## Corrected correction

[...]

PR #7995, opened 2026-03-12
7993/fix end event date required with open end option checked

TimoBroeskamp:
> [!CAUTION]

-----

-----

Is partly closing #7993 

-----

### DIFF
--- a/packages/volto/src/components/manage/Widgets/DatetimeWidget.jsx
+++ b/packages/volto/src/components/manage/Widgets/DatetimeWidget.jsx
@@ -111,6 +111,10 @@ const DatetimeWidgetComponent = (props) => {
   if (id === 'end' && formData?.open_end) {
     return null;
   }
+  // If open_end is checked set formData.end to formData.start
+  if (formData?.open_end === true) {
+    formData.end = formData.start;
+  }
 
   const getInternalValue = () => {
     return parseDateTime(toBackendLang(lang), value, undefined, moment.default);
--- a/packages/volto/src/helpers/FormValidation/validators.ts
+++ b/packages/volto/src/helpers/FormValidation/validators.ts
@@ -161,7 +161,7 @@ export const startEventDateRangeValidator = ({
 }: Validator) => {
   const isValid =
     (value && formData.end && new Date(value) < new Date(formData.end)) ||
-    formData.open_end;
+    (formData.open_end && formData.start === formData.end);
   return !isValid
     ? formatMessage(messages.startEventRange, {
         endDateValueOrEndFieldName: formData.end || 'end',
@@ -177,7 +177,7 @@ export const endEventDateRangeValidator = ({
 }: Validator) => {
   const isValid =
     (value && formData.start && new Date(value) > new Date(formData.start)) ||
-    formData.open_end;
+    (formData.open_end && formData.start === formData.end);
   return !isValid
     ? formatMessage(messages.endEventRange, {
         startDateValueOrStartFieldName: formData.start || 'start',


stevepiercy:
Please edit the description to include `closes #7995` so that when this is merged, GitHub will automatically close the issue.

Also please add a change log entry per https://6.docs.plone.org/contributing/index.html#contributing-change-log-label

jackahl:
> Please edit the description to include `closes #7995` so that when this is merged, GitHub will automatically close the issue.
> 
> Also please add a change log entry per https://6.docs.plone.org/contributing/index.html#contributing-change-log-label

@stevepiercy I think this will most likely not fix https://github.com/plone/volto/pull/7993 as there are also backend changes to be made as noted in https://github.com/plone/plone.app.event/issues/97

TimoBroeskamp:
@stevepiercy @jackahl Okay now it's updated. When checking "open end" the formData.end is set to formData.start and the validation accepts that only when "open end" is checked too so otherwise you still need the end date to be at a later point.

jackahl:
@TimoBroeskamp this should work. But I just had another thought: This might become annoying if someone checks the box and then decides to use an end-date anyway. To avoid this we could make the change ob the request only, but not the actual formdata. Though I am unsure, if we have a good place to apply such transformations... 

@sneridagh @pnicolli is there a place where we can apply transformations on the formdata, before it is send? Or is this a bad idea in general, as then the database contents would differ from the formdata

stevepiercy:
Sorry, I misinterpreted the icalendar standard, per a further discussion at https://github.com/collective/icalendar/issues/1275#issuecomment-4050732782. Please see my revised comment at https://github.com/plone/volto/issues/7993#issuecomment-4042453088.

> This might become annoying if someone checks the box and then decides to use an end-date anyway.

@jackahl I don't understand why that's annoying. I can check the Open End box, then uncheck it and add an Event Ends date or datetime. Can you elaborate?

> Or is this a bad idea in general, as then the database contents would differ from the formdata

I'd transform the data in the client form before submittal and validate it both on the frontend and backend. Transformation on the backend has a funky smell to me.

packages/volto/src/components/manage/Widgets/DatetimeWidget.jsx
```diff
@@ -111,6 +111,10 @@ const DatetimeWidgetComponent = (props) => {
   if (id === 'end' && formData?.open_end) {
     return null;
   }
+  // If open_end is checked set formData.end to formData.start
+  if (formData?.open_end === true) {
+    formData.end = formData.start;
```
gyst, line 116: This deviates from the backend default, which is to set end to one hour after start.

https://github.com/plone/plone.app.event/blob/ffc29bbfe8a7461da0b72b8b7af844a2871be7b2/src/plone/app/event/dx/behaviors.py#L97

https://github.com/plone/plone.app.event/blob/ffc29bbfe8a7461da0b72b8b7af844a2871be7b2/src/plone/app/event/base.py#L37

https://github.com/plone/plone.app.event/blob/ffc29bbfe8a7461da0b72b8b7af844a2871be7b2/src/plone/app/event/base.py#L914

packages/volto/src/helpers/FormValidation/validators.ts
```diff
@@ -160,7 +160,8 @@ export const startEventDateRangeValidator = ({
   formatMessage,
 }: Validator) => {
   const isValid =
-    value && formData.end && new Date(value) < new Date(formData.end);
+    (value && formData.end && new Date(value) < new Date(formData.end)) ||
+    (formData.open_end && formData.start === formData.end);
```
gyst, line 164: This breaks with the backend defaults, see above. Better is to require that start <= end in case of open_end.

packages/volto/src/helpers/FormValidation/validators.ts
```diff
@@ -175,7 +176,8 @@ export const endEventDateRangeValidator = ({
   formatMessage,
 }: Validator) => {
   const isValid =
-    value && formData.start && new Date(value) > new Date(formData.start);
+    (value && formData.start && new Date(value) > new Date(formData.start)) ||
+    (formData.open_end && formData.start === formData.end);
```
gyst, line 180: Same as above.

stevepiercy:
I was in error yet again. Please see the amended note in https://github.com/plone/volto/issues/7993#issuecomment-4042453088.

TimoBroeskamp:
@stevepiercy Okay just to clarify so that I know I understood it right.
https://github.com/plone/volto/issues/7993#issuecomment-4042453088
When emptying the "Event Ends" field it should get the value Null or None and the same should happen when I check the "opne end" checkbox right? Therefore you get no entry in the DTEND property for the .ics file.

The third point about events without an end or recurring events is something that needs to be changed in the backend so its not related directly to my pr right?

stevepiercy:
> When emptying the "Event Ends" field it should get the value Null or None and the same should happen when I check the "opne end" checkbox right? Therefore you get no entry in the DTEND property for the .ics file.

Ummm... that begs the question, why do we have two inputs for the same purpose?

I added a comment to the original issue. Let's keep the discussion about how to proceed in https://github.com/plone/volto/issues/7993. I neglected to complete the draft of a PLIP, which I think is necessary because it touches multiple repositories, but I hope to finish it tonight.

PR #7996, opened 2026-03-12
Use a consistent version of jsdom

davisagli:
Get it from the catalog, and update to 28.1.0

### DIFF
--- a/packages/volto-slate/src/blocks/Text/__snapshots__/TextBlockEdit.test.jsx.snap
+++ b/packages/volto-slate/src/blocks/Text/__snapshots__/TextBlockEdit.test.jsx.snap
@@ -54,7 +54,7 @@ exports[`TextBlockEdit > renders w/o errors 1`] = `
     >
       <svg
         class="icon block-add-button"
-        style="height: 24px; width: auto; fill: currentColor;"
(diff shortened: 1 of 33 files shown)


davisagli:
Need to take a closer look at the unit test snapshots.

packages/volto-slate/package.json
```diff
@@ -23,7 +23,7 @@
     "image-extensions": "1.1.0",
     "is-hotkey": "0.2.0",
     "is-url": "1.2.4",
-    "jsdom": "^16.6.0",
+    "jsdom": "catalog:",
```
wesleybl: `volto-slate` does not use `jsdom`. It is a peer dependency of `vitest`:

https://github.com/vitest-dev/vitest/blob/main/packages/vitest/package.json#L144

Therefore, it should be moved to `devDependencies`.

wesleybl:
> Need to take a closer look at the unit test snapshots.

@davisagli Regarding the changes in the tests, I believe it's simply the new way `jsdom` displays CSS. I think it's safe to update the snapshots.

packages/volto-slate/package.json
davisagli: Yep, thanks

Issue #5121, opened 2023-08-24
a11y - key press focus - wrong behavior - cms ui


Wagner3UB:
**Describe the bug**
The "user" button, inside the toolbar, don't have the expected behavior when the key press action occurs. Instead of enter the newly opened panel remains, the focus remains inside the toolbar

**To Reproduce**
With the focus at the "user" button, inside the toolbar, after the enter key press, the focus remains inside the toolbar and makes impossible to select any option inside the new panel.

**Expected behavior**
After the action (enter key press), on the user button at the bottom of the toolbar, the focus must go to the first linkable item inside the new panel and do, at lest, a whole run through the option before exit. 

- If the panel doesn't have a close button, at some point, the `tab`  action must send the user out.
- If added the close button, the tab action 

**Screenshots**
https://github.com/plone/volto/assets/60133113/8b0fcf4e-eaab-4b92-9646-830d0a0d2e6c

**Software (please complete the following information):**

Volto Version 17.0.0-alpha.19
Plone Version 6.0.4

PR #8019, opened 2026-03-19
Personal tools flow - focus action - a11y

Wagner3UB:
Corrected the one way path through the personal tools container using assistive tech

issue: https://github.com/plone/volto/issues/5121

1. The "personal tools" button, when activated, moves focus to the first focusable element inside the newly opened panel.
2. After the focus is set, the screen reader announces the panel context: "Personal tools, menu opened".
3. Navigation is one-directional — the user tabs through the panel items without looping back.
4. When focus leaves the last item (onBlur), it is moved to the toolbar's expand/shrink button, and "Menu closed" is announced.
5. Simultaneously, the panel is closed and all related state is reset.

https://github.com/user-attachments/assets/11dd190f-099f-4896-87bd-e4f54bd5b60a


stevepiercy (approved):
Messages look good. Requires a tech review. Thank you!

packages/volto/src/components/manage/Toolbar/Toolbar.jsx
```diff
@@ -376,12 +431,44 @@ class Toolbar extends Component {
           <BodyClass
             className={expanded ? 'has-toolbar' : 'has-toolbar-collapsed'}
           />
+          <span
+            aria-live="assertive"
+            aria-atomic="true"
```
wesleybl, line 436: Add a `role="status"` to make it more semantically correct, since we have a status update here.

packages/volto/src/components/manage/Toolbar/Toolbar.jsx
```diff
@@ -376,12 +431,44 @@ class Toolbar extends Component {
...
+                  ?.querySelector('button.toolbar-handler-button')
+                  ?.focus();
+
+                // Close menu on blur only for personalTools — scoped to avoid
+                // impacting other toolbar flows or unrelated tests
+                const isPersonalTools =
+                  this.state.loadedComponents.includes('personalTools');
+                if (isPersonalTools) {
+                  this.closeMenu();
+                }
```
wesleybl: Why did you include an `if` statement to close the menu? When navigation is implemented in other panels, shouldn't they close on exit? Did any tests fail?

wesleybl (changes requested):

packages/volto/src/components/manage/Toolbar/Toolbar.jsx
wesleybl: Furthermore, it's possible to add panels via configuration. Therefore, the code cannot be fixed. Perhaps it could be detected if the panel has focusable elements.

Wagner3UB: > Why did you include an `if` statement to close the menu? When navigation is implemented in other panels, shouldn't they close on exit? Did any tests fail?

@wesleybl The if (isPersonalTools) guard was intentional, I scoped the menu close behavior only to personalTools because it was the failing case in my testing. Other panels can contain focusable elements (links, buttons) and closing on blur would break interaction with them. This if solved at least 5 cypress errors.

Wagner3UB: > Furthermore, it's possible to add panels via configuration. Therefore, the code cannot be fixed. Perhaps it could be detected if the panel has focusable elements.

Done. Instead of checking for a specific panel name, the code now detects if the panel contains focusable elements (a, button, input, etc.). If it does, the menu stays open so the user can navigate within it. If not, it closes on blur. This approach handles custom panels added via configuration automatically.

packages/volto/src/components/manage/Toolbar/Toolbar.jsx
```diff
@@ -305,10 +318,40 @@ class Toolbar extends Component {
...
-        showMenu: !state.showMenu,
-        menuStyle: { bottom: 0 },
-      }));
+      this.setState(
+        (state) => ({
+          showMenu: !state.showMenu,
+          menuStyle: { bottom: 0 },
+        }),
+        () => {
+          // Scoped only to personalTools — does not affect other toolbar flows
```
wesleybl, line 327: When opening other panels with the keyboard, focus doesn't go to the menu. So the menu remains open and we move to the next panel. For example, the add content panel (+). Is your intention to only address the Personal tools in this PR? Later we need a more generic solution.

packages/volto/src/components/manage/Toolbar/Toolbar.jsx
```diff
@@ -376,12 +431,45 @@ class Toolbar extends Component {
...
+              if (!this.toolbarWindow.current?.contains(e.relatedTarget)) {
+                this.toolbarRef.current
+                  ?.querySelector('button.toolbar-handler-button')
+                  ?.focus();
+
+                const hasFocusableElements =
+                  this.toolbarWindow.current?.querySelector(
+                    'a, button, input, select, textarea, [tabindex]:not([tabindex="-1"])',
+                  );
+                if (!hasFocusableElements) {
```
wesleybl: Here it should be:

```javascript
if (hasFocusableElements) {
```

As it is, the menu remains open even after we navigate away from it.

packages/volto/src/components/manage/Toolbar/Toolbar.jsx
```diff
@@ -376,12 +431,45 @@ class Toolbar extends Component {
...
+                if (this.announceRef.current) {
+                  this.announceRef.current.textContent = '';
+                  // Timeout to allow the screen reader to pick up the change in content after the menu is closed
+                  setTimeout(() => {
+                    if (this.announceRef.current) {
+                      this.announceRef.current.textContent =
+                        this.props.intl.formatMessage(messages.menuClosed);
+                    }
+                  }, 100);
+                }
```
wesleybl, line 464: The announcement must be inside the `if` statement above. Otherwise, situations may occur where the menu does not close, but the announcement still appears.

wesleybl (changes requested):

packages/volto/src/components/manage/Toolbar/Toolbar.jsx
Wagner3UB: @wesleybl Something went wrong yesterday, the behavior wasn't what I expected. I'll look into it.

PS: Yes, the focus is on personal tools for now. I need to take baby steps before moving further.

Wagner3UB:
@wesleybl After reviewing the feedback, I decided to simplify the approach, the onBlur handler now always calls closeMenu() when focus leaves the panel, regardless of which panel is open or whether it has focusable elements. This makes the behavior generic and handles panels added via configuration automatically. The announcement was also moved inside the same block so it only fires when the menu actually closes.

In practice, this close-on-blur behavior won't be exclusive to the toolbar, any panel that opens within this container will benefit from it automatically.

Thank you for your efforts! It's way more fun to work like this!

packages/volto/src/components/manage/Toolbar/Toolbar.jsx
wesleybl: It appears you removed the `role="status"`.

wesleybl (changes requested):

wesleybl (approved):

packages/volto/src/components/manage/Toolbar/Toolbar.jsx
```diff
@@ -376,12 +431,39 @@ class Toolbar extends Component {
...
+            className="visually-hidden"
+            ref={this.announceRef}
+          />
           <div
             style={this.state.menuStyle}
             className={
               this.state.showMenu ? 'toolbar-content show' : 'toolbar-content'
             }
             ref={this.toolbarWindow}
+            onBlur={(e) => {
```
iFlameing, line 447: @sneridagh This onBlur will be fired for all pluggable components, and if we are using react-aria-components Popover, it is not possible to develop any kind of dialog implementation using this. If we wanted to fix this closing thing when the focus moves away from the personal toolbar, we should have added some condition to particularly target that case, but this is a genuine problem in this toolbar component.

cc @wesleybl

wesleybl: @iFlameing We could add a condition, but in most cases, the action should be to close the menu. If we test the Volto menus, the added menus won't close. So, I'm not sure if this is a good idea.

Is there a marker on the Popover that we can check to see if the element that received focus is inside it?

PR #8023, opened 2026-03-19
PoC of /@blocks-schema

sneridagh:


packages/volto/src/config/Blocks.jsx
```diff
@@ -290,6 +304,23 @@ const blocksConfig = {
     mostUsed: true,
     sidebarTab: 1,
     getSizes: getImageBlockSizes,
+    llm: {
+      description:
+        'Displays an internal or external image with optional title, description, alt text, alignment, size, and link. Use for standalone images within page content.',
+      usage_notes:
+        "For INTERNAL Plone images: set 'url' to the Image content object path (e.g. /my-folder/my-image) and set 'image_field' to 'image'. For EXTERNAL images: set 'url' to the full image URL and omit 'image_field'. Do NOT use @@images URLs for internal images. Do not set 'align' or 'size' when nesting inside anpther block (e.g. gridBlock) since the parent block controls layout.",
```
davisagli: It is ok to use a full URL for internal images also. Either way, it gets converted into a UID-based path when it is stored.

```suggestion
        "For INTERNAL Plone images: set 'url' to the Image content object path (e.g. /my-folder/my-image) and set 'image_field' to 'image'. For EXTERNAL images: set 'url' to the full image URL and omit 'image_field'. Do NOT use @@images URLs for internal images. Do not set 'align' or 'size' when nesting inside another block (e.g. gridBlock) since the parent block controls layout.",
```

packages/volto/src/config/Blocks.jsx
```diff
@@ -338,6 +369,51 @@ const blocksConfig = {
...
+          sort_on: 'effective',
+          sort_order: 'descending',
+          b_size: '6',
+        },
+      },
+      field_hints: {
+        querystring:
+          "Required. Object with 'query' (required array of filter criteria), optional 'sort_on', 'sort_order', 'sort_order_boolean', 'b_size' (string, items per page), 'limit' (string, max total results). Multiple criteria are combined with AND logic. Each criterion requires 'i' (index name), 'o' (FULL operation string — see below), 'v' (value). " +
+          'Indexes: portal_type, path, Title, Description, SearchableText, Creator, getId, review_state, Subject, created, modified. ' +
+          'Operations (always use the FULL string): ' +
```
davisagli: The available indexes and operations are dynamic based on the configuration in the site's registry. This could refer to the API for getting them instead of including a hardcoded list.

packages/volto/src/config/Blocks.jsx
```diff
@@ -264,6 +264,13 @@ const blocksConfig = {
     mostUsed: false,
     blockHasOwnFocusManagement: true,
     sidebarTab: 0,
+    llm: {
```
davisagli: None of this is really specifically useful to LLMs. It's documentation that is also useful to humans. I'd suggest calling this key `docs` rather than `llms` (name it what it is, not based on one specific thing that will consume it).

packages/volto/src/config/Blocks.jsx
```diff
@@ -338,6 +369,51 @@ const blocksConfig = {
...
+          'plone.app.querystring.operation.selection.any/none/all — v: array of strings, e.g. ["News Item"]; ' +
+          'plone.app.querystring.operation.string.contains/is/isNot — v: string; ' +
+          'plone.app.querystring.operation.string.absolutePath — v: "UID::depth" (e.g. "abc123::2"); ' +
+          'plone.app.querystring.operation.string.relativePath — v: "./" (current) or "../" (parent); ' +
+          'plone.app.querystring.operation.string.currentUser — v: "" (matches logged-in user, use with Creator index); ' +
+          'plone.app.querystring.operation.date.largerThan/lessThan — v: "YYYY-MM-DD". ' +
+          "sort_on values: 'sortable_title', 'effective', 'created', 'modified', 'getObjPositionInParent', 'start', 'end', 'review_state'. " +
+          "sort_order: 'ascending' (default) or 'descending'.",
+        variation:
+          "Required. 'default' (simple list), 'summary' (list with images), 'imageGallery' (image carousel), 'eventCalendar' (calendar view), 'listDate' (list with date).",
```
davisagli: The available variations can be customized in Volto config.

packages/volto/src/express-middleware/blocks-schema.test.js
davisagli, line 1: I am trying to look here to see examples of the expected output, but the LLM info doesn't seem to be included in the test.

davisagli (commented):
Thoughts:
- We could potentially also use this info to auto-generate reference docs. So I would put it in a more generic key like `docs` instead of `llm`.
- I wonder if it would be more useful for an agent to produce Markdown output instead of structured json.
- Do we need to have a way to translate this info?
- Maybe we should have a function that returns this info instead of just an object in the config, so that it can take into account dynamic configuration.

sneridagh:
@davisagli @danalvrz I took a look at the current state, answering to your main questions:

> We could potentially also use this info to auto-generate reference docs. So I would put it in a more generic key like docs instead of llm.

+1 (done)

> I wonder if it would be more useful for an agent to produce Markdown output instead of structured json.

+1000 I think we reached a point that LLMs are more powerful with short markdown explanations. Also, it's instant docs for humans too. My take on this is that it feels wrong to have the structured docs in JSON... I see no reason for it, although some that I can't see now related to the MCP (which should not be the case, as I'd decouple the use case from the endpoint).


> Do we need to have a way to translate this info? 

Which would be the point? Machines talk any language, unless we'd like to generate docs out of this endpoint, which is unlikely... Anyways, you could define a .po with generic keys `button_docs` as `msgid`, with the correspondent `msgstr`.

> Maybe we should have a function that returns this info instead of just an object in the config, so that it can take into account dynamic configuration.

I am missing the point of this, if can we elaborate the use case it would be great.
In the case of the plain markdown, you would override the whole `docs` key, right?

Also, we need docs for this endpoint, since it's a whole new category (frontend endpoint), I guess the question is where in the docs structure.

sneridagh:
@davisagli @danalvrz I'd love to see this happening before the conf. Any feedback from my comment?

PR #8033, opened 2026-03-20
Text field alerts a11y

Wagner3UB:
https://github.com/plone/volto/issues/5619

Fix: screen readers were not announcing field errors or required state on form inputs.

- Added aria-live="polite" and aria-atomic="true" to the error region in FormFieldWrapper so error messages are announced when they appear in the DOM
- Added aria-required and aria-invalid attributes to the Input in TextWidget for proper field identification by assistive technologies

Screenshot 2026-03-20 at 15 32 55

Screenshot 2026-03-20 at 15 35 06


packages/volto/news/8033.bugfix
```diff
@@ -0,0 +1 @@
+Accessibility issues in form fields fixed. Announce errors via aria-live and expose required/invalid state on text inputs @Wagner3UB
```
stevepiercy:
```suggestion
Announce errors via `aria-live` and expose required and invalid states on text inputs to improve accessibility of form fields. @Wagner3UB
```

packages/volto/src/components/manage/Widgets/__snapshots__/IdWidget.test.jsx.snap
```diff
@@ -224,9 +241,14 @@ exports[`IdWidget > renders an id widget with invalid characters 1`] = `
...
           <div
-            class="ui red basic label form-error-label"
+            aria-atomic="true"
+            aria-live="polite"
           >
-            Only 7-bit bytes characters are allowed. Cannot contain uppercase letters, special characters: &lt;, &gt;, &, #, /, ?, or others that are illegal in URLs. Cannot start with: _, aq_, @@, ++. Cannot end with __. Cannot be: request,contributors, ., .., "". Cannot contain new lines.
+            <div
+              class="ui red basic label form-error-label"
+            >
+              Only 7-bit bytes characters are allowed. Cannot contain uppercase letters, special characters: &lt;, &gt;, &, #, /, ?, or others that are illegal in URLs. Cannot start with: _, aq_, @@, ++. Cannot end with __. Cannot be: request,contributors, ., .., "". Cannot contain new lines.
```
stevepiercy, line 250: I just noticed this. Does the ampersand render to the screen, or does it need to be encoded?

stevepiercy (changes requested):
News clarification and a question to check rendering missed on the original source.

packages/volto/src/components/manage/Widgets/__snapshots__/IdWidget.test.jsx.snap
pnicolli: This good question got me curious and I checked, since this is a snapshot test result and should (theoretically) represent the result of rendering.
This is the actual string that is being render, the `&lt;` was converted while the `&` was not, so I think it's safe to assume this is fine, at least for the sake of this PR.

https://github.com/plone/volto/blob/main/packages/volto/src/components/manage/Widgets/IdWidget.jsx#L30
https://github.com/plone/volto/blob/main/packages/volto/locales/en/LC_MESSAGES/volto.po#L2675

pnicolli (approved):

Wagner3UB:
@pnicolli Can we merge this one?

stevepiercy (approved):

davisagli (approved):
LGTM

sneridagh:
I found a regression regarding this one. (See https://volto.demo.plone.org/edit as long as is there, if not, just add an image, check the size widget).

See the size widget, it's not aligned anymore:

image

Because the markup changed with the change of this PR.

https://github.com/plone/volto/blob/d8cca3b242842a2322d1dd71b52b1d4e22387ff5/packages/volto/src/components/manage/Widgets/FormFieldWrapper.jsx#L64-L70

image

The grid now have another unaccounted element. No clue if that happens elsewhere...

Please, we need to have more care when we change markup, specially in grids/flex elements and check all the affected components. 

@Wagner3UB can we add some CSS to ignore that addition? Which is the best thing to proceed?

sneridagh:
@Wagner3UB I created https://github.com/plone/volto/issues/8317.

PR #8044, opened 2026-03-23
Input placeholder lacks sufficient contrast ratio - a11y

Wagner3UB:
### Description:
Fixes https://github.com/plone/volto/issues/8043

### What changed:
In themes/default/globals/site.variables:

`@inputColor` changed from `@textColor` to #000000, making it explicit and independent from the global text color

`@inputPlaceholderColor` changed from `lighten(@inputColor, 75)` to `lighten(@inputColor, 45)`, darker default placeholder for sufficient contrast

`@inputPlaceholderFocusColor` changed from `lighten(@inputColor, 45)` to `lighten(@inputColor, 75)`, lighter on focus, correctly making the placeholder less prominent once the field is active

### In themes/pastanaga/globals/site.variables:

Removed hardcoded `@inputPlaceholderColor: #B8C6C8` and the entire /* Input Text Color */ block, these values are now correctly inherited from the default theme

### Why:
The previous configuration had two problems:

The hardcoded `#B8C6C8` in Pastanaga failed WCAG 2.1 contrast requirements, not reaching even Level A
The focus/default placeholder logic was inverted, the focused state was darker than the default

### Visual impact:
This is a visual breaking change. All input placeholders using `@inputPlaceholderColor` and `@inputPlaceholderFocusColor` will have their colors inverted. Custom themes overriding these variables are not affected.

### Examples:

Share page input
without focus:
input-without-focus

with focus:
input-with-focus

Site search without and with focus inside:
site-search-without-focus
site-search-with-focus

### DIFF
--- a/packages/volto/news/8044.bugfix
+++ b/packages/volto/news/8044.bugfix
@@ -1,0 +1,1 @@
+Replace hardcoded color values with Semantic UI theme variables for input field - a11y @Wagner3UB
--- a/packages/volto/news/8044.bugfix
+++ b/packages/volto/news/8044.bugfix
@@ -1,1 +1,1 @@
-Replace hardcoded color values with Semantic UI theme variables for input field - a11y @Wagner3UB
+Replace hard-coded color values with Semantic UI theme variables for input field to improve accessibility. @Wagner3UB
--- a/packages/volto/theme/themes/default/globals/site.variables
+++ b/packages/volto/theme/themes/default/globals/site.variables
@@ -81,9 +81,9 @@
 @inputPadding           : @inputVerticalPadding @inputHorizontalPadding;
 
 /* Input Text Color */
-@inputColor: @textColor;
-@inputPlaceholderColor: lighten(@inputColor, 75);
-@inputPlaceholderFocusColor: lighten(@inputColor, 45);
+@inputColor: #000000;
+@inputPlaceholderColor: lighten(@inputColor, 45);
+@inputPlaceholderFocusColor: lighten(@inputColor, 75);
 
 /* Line Height Default For Inputs in Browser (Descenders are 17px at 14px base em) */
 @inputLineHeight: unit((17 / 14), em);
--- a/packages/volto/theme/themes/default/globals/site.variables
+++ b/packages/volto/theme/themes/default/globals/site.variables
@@ -6,34 +6,32 @@
        Fonts
 --------------------*/
 
-@fontName          : 'Lato';
-@fontSmoothing     : antialiased;
+@fontName: 'Lato';
+@fontSmoothing: antialiased;
 
-@headerFont        : @fontName, 'Helvetica Neue', Arial, Helvetica, sans-serif;
-@pageFont          : @fontName, 'Helvetica Neue', Arial, Helvetica, sans-serif;
+@headerFont: @fontName, 'Helvetica Neue', Arial, Helvetica, sans-serif;
+@pageFont: @fontName, 'Helvetica Neue', Arial, Helvetica, sans-serif;
 
-@googleFontName    : @fontName;
-@importGoogleFonts : true;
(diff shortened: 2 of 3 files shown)


packages/volto/news/8044.bugfix
```diff
@@ -0,0 +1 @@
+Replace hardcoded color values with Semantic UI theme variables for input field - a11y @Wagner3UB
```
stevepiercy:
```suggestion
Replace hard-coded color values with Semantic UI theme variables for input field to improve accessibility. @Wagner3UB
```

stevepiercy (changes requested):
With this minor change to the news, I approve. Thank you!

packages/volto/theme/themes/default/globals/site.variables
```diff
@@ -81,9 +81,9 @@
 @inputPadding           : @inputVerticalPadding @inputHorizontalPadding;
 
 /* Input Text Color */
-@inputColor: @textColor;
-@inputPlaceholderColor: lighten(@inputColor, 75);
-@inputPlaceholderFocusColor: lighten(@inputColor, 45);
+@inputColor: #000000;
```
wesleybl: Why did you decouple of `@textColor`? I think that breaks the theme's color relationship. Perhaps it would be worthwhile to calculate the contrast based on `@textColor`? Whoever is customizing the theme might be relying on that relationship. And if someone has a dark theme, leaving it black could be detrimental. Maybe the custom theme changed `@textColor` to prevent that from happening.

packages/volto/theme/themes/default/globals/site.variables
```diff
@@ -81,9 +81,9 @@
...
+@inputPlaceholderColor: lighten(@inputColor, 45);
+@inputPlaceholderFocusColor: lighten(@inputColor, 75);
```
wesleybl: I think that will result in low contrast. Would 65 be better?

wesleybl (changes requested):

packages/volto/theme/themes/default/globals/site.variables
Wagner3UB: It's my fault, sorry, it was a test and I forget to remove.

packages/volto/theme/themes/default/globals/site.variables
Wagner3UB: @wesleybl We don’t have much of an issue when it’s in focus, but we need a minimum contrast for the default placeholder. When in focus, the key is to show a state change. In our case, it’s not ideal because the placeholder just gets lighter and the bottom line changes color. I would have added an outline, but that would go against the design guidelines. To be safe, I’ve made the placeholder even darker and also darkened it for the focus state.

wesleybl (approved):

stevepiercy (approved):
I think this should be backported to 18, too.

One note, we've ripped out Semantic UI for Seven, so it would be good to check on the `seven` branch if this fix is needed, and if so, then adapt it to the new theming system. This would be a good sprint task. Maybe @pnicolli or @sneridagh can point y'all toward the proper CMSUI theming story. I found this: https://volto.readthedocs.io/seven/conceptual-guides/seven-vs-volto-differences.html#tailwind-css

Wagner3UB:
> I think this should be backported to 18, too.
> 
> One note, we've ripped out Semantic UI for Seven, so it would be good to check on the `seven` branch if this fix is needed, and if so, then adapt it to the new theming system. This would be a good sprint task. Maybe @pnicolli or @sneridagh can point y'all toward the proper CMSUI theming story. I found this: https://volto.readthedocs.io/seven/conceptual-guides/seven-vs-volto-differences.html#tailwind-css

@stevepiercy It will. Just to let you know, all our accessibility interventions will be present in both versions 18 and 19, without exception. Our goal at RedTurtle is to have these fixes ready for version 18 soon, followed by version 19 after its official release.

Issue #5141, opened 2023-08-28
a11y - site setup - Content Types - sidebar "add new content type" button  - cms ui


Wagner3UB:
**Describe the bug**
- The icon "+" (add new content type) inside de left sidebar doesn't receive a visible focus when select, either on hover or with the keyboard. After the click, a modal/pop-up appears, but the focus remains on the "+" button and it does not enter the modal.

**To Reproduce**

1. Go to "site setup" -> "Content types"
2. With a screen reader + keyboard, go inside the left toolbar on the + icon.
3. press enter to access the next behavior
4. a pop-up/modal will appear

**Expected behavior**
- The icon must have a visible focus, either on hover or with the keyboard, and the modal/pop-up must receive the focus after the keypress on the "+" button.
- The focus must remain in a loop inside the modal until the save or the cancel button is pressed.

**Screenshots**
<img width="1568" alt="Schermata 2023-08-28 alle 17 07 44" src="https://github.com/plone/volto/assets/60133113/204162b3-c434-4b66-9756-7e16d6c2ecbe">
<img width="1513" alt="Schermata 2023-08-28 alle 17 07 52" src="https://github.com/plone/volto/assets/60133113/63c7e9e9-f307-4af8-a398-4093e2c2aa51">

**Software (please complete the following information):
Volto Version 17.0.0-alpha.19
Plone Version 6.0.4

MayaSatishRao:
@Wagner3UB i want to work on this issue? can you assign this to me?

stevepiercy:
@MayaSatishRao please follow our contributing guidelines:

- https://6.docs.plone.org/contributing/index.html
- https://6.docs.plone.org/contributing/first-time.html

ShatilKhan:
Hi @Wagner3UB 
Can you assign me this issue?
I have worked on frontend contributions before.

Wagner3UB:
@ShatilKhan please follow https://6.docs.plone.org/contributing/index.html and https://6.docs.plone.org/contributing/first-time.html

Noobham:
Hi @Wagner3UB 

I am working on this issue. The goal is to set aria-hidden="false" on the root element and aria-hidden="true" on modal when a popup is displayed. When the popup is closed, these attributes should be toggled back to their original states.

facing issue in finding root element of Volto ?
Also can i know is my idea correct or should i find any other alternative.

PR #7685, opened 2025-12-10
Refactor ModalForm: Convert Class Component to Functional Component

Manik-Khajuria-5:
### **Issue number**

Contributes to —https://github.com/plone/volto/issues/4460

---

### Description 

This PR refactors the `ModalForm` component by converting it from a class-based React component to a modern functional component using React hooks.

---

## **Key Changes**

* Converted `ModalForm` from a **class-based component** to a **functional component**
* Replaced lifecycle methods (`componentDidMount`, `componentDidUpdate`, etc.) with appropriate **React hooks**
* Removed constructor and `this` bindings
* Simplified state management using **useState**
* Improved side-effect handling using **useEffect**
* Cleaned up unused code and improved overall readability
* Ensured the modal’s open/close behavior and form logic remain unchanged

-----

- [x] I signed and returned the [Plone Contributor Agreement](https://plone.org/foundation/contributors-agreement), and received and accepted an invitation to join a team in the Plone GitHub organization.
- [x] I verified there aren't other open [pull requests](https://github.com/plone/volto/pulls) for the same change.
- [x] I followed the guidelines in [Contributing to Volto](https://6.docs.plone.org/volto/contributing/index.html).
- [x] I successfully ran [code linting checks](https://6.docs.plone.org/volto/contributing/linting.html) on my changes locally.
- [x] I successfully ran [unit tests](https://6.docs.plone.org/volto/contributing/testing.html) on my changes locally.
- [x] I successfully ran [acceptance tests](https://6.docs.plone.org/volto/contributing/acceptance-tests.html) on my changes locally.
- [ ] If needed, I added new tests for my changes.
- [ ] If needed, I added [documentation](https://6.docs.plone.org/volto/contributing/documentation.html#narrative-documentation) for my changes, either in the Storybook or narrative documentation.
- [x] I included a [change log entry](https://6.docs.plone.org/contributing/index.html#contributing-change-log-label) in my commits.

PR #8058, opened 2026-03-27
Site setup content types add new - CMS UI - a11y

Wagner3UB:
Accessibility improvements for ModalForm and Content Types panel

Fixed several accessibility issues in the ModalForm component and the Content Types control panel:

- Added proper focus management: focus returns to the trigger button when a modal dialog is closed or submitted
- Added aria-live region announcements when a modal opens or closes (screen reader support)
- Added aria-modal support to the modal dialog
- Updated button ID from #toolbar-add to #toolbar-add-content-type for better semantic clarity
- Updated Cypress tests to reflect the new button ID and CSS classes
- Added translations for the new i18n strings across all supported locales

Issue: https://github.com/plone/volto/issues/5141


packages/volto/news/8058.bugfix
```diff
@@ -0,0 +1 @@
+Fixed accessibility issues in ModalForm: focus management, aria-live announcements and aria-modal support @Wagner3UB
```
stevepiercy:
```suggestion
Fixed accessibility issues in `ModalForm` with focus management, `aria-live` announcements, and `aria-modal` support. @Wagner3UB
```

stevepiercy (changes requested):
LGTM. Just the news. Also merge https://github.com/plone/volto/pull/8077 to fix linkcheck.

wesleybl:
@Wagner3UB Perhaps it would be better to convert `ModalForm` to a functional component before applying these changes? See:

https://github.com/plone/volto/pull/7685

wesleybl:
The bad thing about waiting for the conversion to a functional component is that the backport to Volto 18 will become more difficult.

@sneridagh said that we cannot do the backport of the conversion to functional components.

But the changes here can work for Volto 18.

davisagli:
@wesleybl From my perspective, the conversion to a functional component is a separate matter. We merge the one that is ready first, and then update the other one.

Wagner3UB:
> Perhaps it would be better to convert `ModalForm` to a functional component before applying these changes? See:

@plone/volto-team What do you think, folks? I can sustain these changes for Volto 18 and do the refactor that @wesleybl suggested. More work to do, but, can be beneficial.

wesleybl:
@Wagner3UB I think to make backporting to Volto 18 easier, you can continue here and then we'll do the conversion to functional.

packages/volto/src/components/manage/Form/ModalForm.jsx
```diff
@@ -54,6 +62,7 @@ const messages = defineMessages({
  * @extends Component
  */
 class ModalForm extends Component {
+  headerId = `modal-title-${Math.random().toString(36).slice(2)}`;
```
wesleybl: It's best to use a static counter to avoid collisions. And it should also be inside the constructor:

```javascript
static idCounter = 0;
constructor(props) {
  this.headerId = `modal-title-${++ModalForm.idCounter}`;
}
```

packages/volto/src/components/manage/Form/ModalForm.jsx
```diff
@@ -215,6 +226,22 @@ class ModalForm extends Component {
    * @param {Object} prevState
    */
   async componentDidUpdate(prevProps, prevState) {
```
wesleybl: There is no `await`. You can remove the `async`.

packages/volto/src/components/manage/Form/ModalForm.jsx
```diff
@@ -259,100 +286,133 @@ class ModalForm extends Component {
...
+          }}
+        />
+        <Modal
+          dimmer={this.props.dimmer}
+          open={this.props.open}
+          className={this.props.className}
+          aria-labelledby={this.headerId}
+          aria-modal="true"
+        >
+          <Header id={this.headerId} aria-live="assertive" aria-atomic="true">
```
wesleybl: There's already an `aria-live="assertive"` in the div above. It's best to remove it from here:


```suggestion
          <Header id={this.headerId}>
```

packages/volto/src/components/manage/Form/ModalForm.jsx
```diff
@@ -259,100 +286,133 @@ class ModalForm extends Component {
...
+              type="button"
+              onFocus={() => {
+                if (this.announceRef.current) {
+                  this.announceRef.current.textContent =
+                    this.props.intl.formatMessage(messages.dialogClosed);
+                }
+                onCancel();
+              }}
+              style={{ position: 'absolute', opacity: 0 }}
+            />
```
wesleybl: I find it strange to close the modal with a tab. I think there should be an explicit action to close it. But the way it exits the modal is also strange. I think it should stay in a loop inside the modal.

packages/volto/src/components/manage/Form/ModalForm.jsx
```diff
@@ -259,100 +286,133 @@ class ModalForm extends Component {
...
+          aria-atomic="true"
+          style={{
+            position: 'absolute',
+            width: '1px',
+            height: '1px',
+            overflow: 'hidden',
+            opacity: 0,
+          }}
+        />
+        <Modal
```
wesleybl, line 334: Add:

`role="dialog"`

wesleybl (changes requested):

packages/volto/src/components/manage/Form/ModalForm.jsx
Wagner3UB: > I find it strange to close the modal with a tab. I think there should be an explicit action to close it. But the way it exits the modal is also strange. I think it should stay in a loop inside the modal.

@wesleybl I can provide an loop there. No problem, I will remove the sentinel.

stevepiercy (approved):

packages/volto/src/components/manage/Controlpanels/ContentTypes.jsx
```diff
@@ -471,9 +475,14 @@ class ContentTypes extends Component {
                   </Link>
                   <Button
                     className="add"
-                    aria-label={this.props.intl.formatMessage(messages.add)}
+                    aria-label={this.props.intl.formatMessage(
+                      messages.addTypeButtonTitle,
+                    )}
                     tabIndex={0}
-                    id="toolbar-add"
+                    id="toolbar-add-content-type"
```
davisagli: Is it really necessary to change this id? It breaks pretty much any acceptance test that uses this id to click the button to add a content type. I don't understand the reason for the change, and it doesn't seem worth the effort required to update tests (also in add-ons) to me.

Wagner3UB: Hey @davisagli! 

The reason for the change is that both buttons coexist on the same page at the same time, the main Toolbar's add button (Toolbar.jsx) and the ContentTypes panel's add button both had id="toolbar-add". Duplicate ids in the same document are invalid HTML and can also confuse assistive technologies (screen readers, automated a11y tests).

That said, I understand the cost, updating tests in add-ons is significant churn. If you'd prefer, I can revert this specific change and address the duplicate id differently, for example by using a data-* attribute or a className instead of an id for the ref/focus-return logic, leaving the id="toolbar-add" untouched.

davisagli: @Wagner3UB ~~Sorry, I misunderstood the context here. I thought this was about the add button in the toolbar but it is actually about the one in the modal.~~

~~I agree we cannot have duplicate ids, and this is the easier one to change.~~

~~Just one request: let's take toolbar out of the name, since it is not in the toolbar.~~

davisagli: @Wagner3UB Uh, no, I'm still confused. This button replaces the normal "add content" button in the toolbar, from what I can see. They are not shown at the same time. So, I don't think there is any need to change the id.

wesleybl: > This button replaces the normal "add content" button in the toolbar, from what I can see. They are not shown at the same time. So, I don't think there is any need to change the id.

I agree. I also didn't see the need to change that id.

Wagner3UB: Hey @davisagli and @wesleybl, I reverted the id change. My intention was to avoid a potential scenario where both buttons could coexist, even intentionally, since they occupy the same region in the toolbar, and duplicate ids in that context could confuse assistive technologies and be semantic wrong.

wesleybl (approved):

PR #8087, opened 2026-04-08
Backport - Toast alert a11y - v18

Wagner3UB:
Backport from https://github.com/plone/volto/pull/8105

### Accessibility improvements for Toast notifications
This PR improves the accessibility of the Toast component for screen readers and general UX.

### Changes
**Toast component (Toast.jsx)**

- Icon is now hidden from screen readers (ariaHidden) since it is purely decorative
- Added a visually hidden label announcing the notification type (Error, Warning, Success, Information) so screen readers can identify the context before reading the content
- Labels are fully internationalized via react-intl (defineMessages + useIntl)

### Styles (main.less)

- Added .toast-visually-hidden scoped inside .toast-inner-content to hide the type label visually while keeping it accessible to screen readers

Date formatting in event validation (validators.ts)

- Date values in event range error messages were being displayed as raw ISO strings (e.g. 2026-04-08T15:00:00+00:00)
- Now formatted using Intl.DateTimeFormat respecting the user's browser locale (e.g. Apr 8, 2026, 3:00 PM)


stevepiercy (approved):

packages/volto/theme/themes/pastanaga/extras/main.less
```diff
@@ -359,7 +359,20 @@ button {
   .toast-inner-content {
     margin-left: 8px;
 
+    .toast-visually-hidden {
```
wesleybl, line 362: I think it's better to have a generic class for this, even if it's only for Volto 18. Otherwise, every time an element needs to be hidden, we'll have to write new CSS. For example:

https://github.com/plone/volto/pull/7551

wesleybl (changes requested):

packages/volto/theme/themes/pastanaga/extras/main.less
Wagner3UB: This cant be done, @wesleybl, take a look on these PR's, it's a breaking change that we can't manage at the moment, that's why the class need to be pointed every single time:
https://github.com/plone/volto/pull/7978
https://github.com/plone/volto/pull/6356#issuecomment-3444888562

wesleybl: What I meant was to have a generic class that doesn't conflict with framework classes. For example: `visually-hidden-v18`.

Wagner3UB: It's not a bad idea indeed.

@pnicolli @sneridagh To be honest, I have never thought about that before. Can we add a general rule that does exactly the same thing as the original but with a different name, as suggested by @wesleybl? 

The idea is to create a "visually-hidden-alias" class without compromising the ones already in use. This could help us from now on.

davisagli: @Wagner3UB That makes sense to me

wesleybl: We still have the `toast-visually-hidden` class.

wesleybl (changes requested):

packages/volto/theme/themes/pastanaga/extras/main.less
Wagner3UB: @wesleybl Yeah, I know, I'm struggling with the Cypress test that keep failing and I don't have a clue why yet...

wesleybl: I think if you update your branch, the errors will be fixed.

Wagner3UB: Class added to `mais.less` https://github.com/plone/volto/pull/8139

wesleybl (approved):

PR #8105, opened 2026-04-10
Toast alert a11y v19

Wagner3UB:
### Accessibility improvements for Toast notifications
This PR improves the accessibility of the Toast component for screen readers and general UX.

### Changes
**Toast component (Toast.jsx)**

- Icon is now hidden from screen readers (ariaHidden) since it is purely decorative
- Added a visually hidden label announcing the notification type (Error, Warning, Success, Information) so screen readers can identify the context before reading the content
- Labels are fully internationalized via react-intl (defineMessages + useIntl)

### Styles (main.less)

- Added .toast-visually-hidden scoped inside .toast-inner-content to hide the type label visually while keeping it accessible to screen readers

Date formatting in event validation (validators.ts)

- Date values in event range error messages were being displayed as raw ISO strings (e.g. 2026-04-08T15:00:00+00:00)
- Now formatted using Intl.DateTimeFormat respecting the user's browser locale (e.g. Apr 8, 2026, 3:00 PM)


Wagner3UB:
@sneridagh Hey Victor, could you take a look at the 'validations.js' file? I changed how the dates are displayed, but I'm not sure if it might cause issues since this file is used so frequently.

.claude/settings.local.json
stevepiercy, line 1: I don't think you meant to commit this file. Perhaps add it to your user .gitignore?

stevepiercy (changes requested):
LGTM, except for the local settings file.

stevepiercy (approved):

packages/volto/src/components/manage/Toast/Toast.jsx
```diff
@@ -22,12 +44,24 @@ const Toast = (props) => {
...
+
   const { title, content } = props;
+  const typeLabel = getTypeLabel(props);
 
   return (
     <>
       <Icon name={getIcon(props)} size="18px" />
       <div className="toast-inner-content">
+        {typeLabel && (
+          <span className="toast-visually-hidden">{typeLabel}</span>
```
wesleybl: We already have a `visually-hidden` class. Couldn't it be used here?

packages/volto/src/helpers/FormValidation/validators.ts
```diff
@@ -153,6 +153,15 @@ export const hasUniqueItemsValidator = ({
   return !isValid ? formatMessage(messages.uniqueItems) : null;
 };
 
+const formatDateValue = (isoString: string) => {
+  const date = new Date(isoString);
+  if (isNaN(date.getTime())) return isoString;
+  return new Intl.DateTimeFormat(undefined, {
```
wesleybl, line 159: The downside of `Intl.DateTimeFormat` is that it uses the browser's locale. So we might have a situation where the date is in one language and the website is in another. But we don't have the portal's locale here. So I think it's okay. It's better than it was before.

wesleybl (changes requested):

wesleybl:
I found the date validation message strange:

`Event start date must be on or before "date"`

It doesn't clearly show the relationship between the fields. It would be better to say:

`Event start date must be earlier than the event end date`

Having the date value there doesn't help much. But that was already the case before. So it's not related to this PR.

Wagner3UB:
> I found the date validation message strange:
> 
> `Event start date must be on or before "date"`
> 
> It doesn't clearly show the relationship between the fields. It would be better to say:
> 
> `Event start date must be earlier than the event end date`
> 
> Having the date value there doesn't help much. But that was already the case before. So it's not related to this PR.

@wesleybl You are absolutely right, I did realised that too, but the translations isn't the goal for this PR. I've booked a PR to better those phrases. Thank you for the note!

packages/volto/src/components/manage/Toast/Toast.jsx
Wagner3UB: Well observed! I first wrote it for Volto 18 exactly because of that detail, and when I put it here, I lost my focus and forgot it there! Will fix it right away!

wesleybl (approved):

PR #8106, opened 2026-04-10
Vite 8

sneridagh:


catalog.json
```diff
@@ -30,5 +30,5 @@
   "@tailwindcss/vite": "^4.1.4",
   "@vitejs/plugin-react": "^5.0.2",
```
pnicolli: This can also be upgraded to v6 iirc

davisagli:
I think this broke storybook (in prod build only).

https://plone-components.readthedocs.io/latest/ has this error: `Uncaught ReferenceError: __STORYBOOK_MODULE_CORE_EVENTS_PREVIEW_ERRORS__ is not defined`

Storybook 10 has in the release notes that it added support for Vite 8, so I don't think it was there in Storybook 9.

PR #8124, opened 2026-04-17
fix: Respect required and fixed block restrictions in the sidebar

### LINKED ISSUE #6481: Order blocks in sidebar doesn't respect layout restrictions: required and fixed position
[order-sidebar-issues.webm](https://github.com/user-attachments/assets/9690b5fb-bbd0-4d42-9d77-e75470e823ba)

**Steps to reproduce the behavior:**

1. Go to ControlPanel > Content Types
1. Add new content type: Book
1. Go to Layout for Book and enable editable blocks
1. Click on Title block and then on Settings in the Sidebar
1. Select "Required"
1. Select "Fixed position"
1. Click on the **Order** tab in Sidebar:
   *  You are able to move the Title block and remove it from here.

**Expected behavior**
1. If a block is required the `delete` button should not be present in the Sidebar Order Tab
2. If a block has `fixed position` the drag & drop button should not be present in the Sidebar Order Tab

**Software:**

- OS: All
- Browser All
- Volto Version 18.x.x
- Plone Version 6.x
- Plone REST API Version 9.x

**Additional context**

avoinea:
- fixes #6481

[order-respect-restrictions.webm](https://github.com/user-attachments/assets/590add35-ff61-4ac5-a508-ea1a3d4e240b)

### DIFF
--- a/packages/volto/src/components/manage/Blocks/Block/Order/Item.jsx
+++ b/packages/volto/src/components/manage/Blocks/Block/Order/Item.jsx
@@ -2,6 +2,7 @@ import React, { forwardRef } from 'react';
 import classNames from 'classnames';
 import { useDispatch, useSelector } from 'react-redux';
 import includes from 'lodash/includes';
+import isBoolean from 'lodash/isBoolean';
 import cx from 'classnames';
 import Icon from '@plone/volto/components/theme/Icon/Icon';
 import { setUIState } from '@plone/volto/actions/form/form';
@@ -43,6 +44,11 @@ export const Item = forwardRef(
     const icon =
       config.blocks.blocksConfig[data?.['@type']]?.icon ||
       config.blocks.blocksConfig.title?.icon;
+
+    const required = isBoolean(data?.required)
+      ? data.required
+      : includes(config.blocks.requiredBlocks, data?.['@type']);
+    const fixed = !!data?.fixed;
 
     return (
       <li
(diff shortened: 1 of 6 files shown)


packages/volto/src/components/manage/Blocks/Block/Order/Item.jsx
```diff
@@ -2,6 +2,7 @@ import React, { forwardRef } from 'react';
 import classNames from 'classnames';
 import { useDispatch, useSelector } from 'react-redux';
 import includes from 'lodash/includes';
+import isBoolean from 'lodash/isBoolean';
```
wesleybl: I don't see the need to use `lodash` here. It's better to use `typeof`.

wesleybl (changes requested):

packages/volto/src/components/manage/Blocks/Block/Order/Item.jsx
avoinea: @wesleybl Done 👍

wesleybl (approved):

wesleybl:
Within a grid block, in a block inside a column, when I check "Required" or "Fixed position", the checkbox is not visually selected, but the selection seems to work. But this is not part of this PR.

wesleybl:
@avoinea Thank you! This was really needed.

wesleybl:
An improvement to this functionality would be to prevent the insertion of blocks above a fixed block. Currently, if you place a fixed block below the title block, it's still possible to add blocks below the title block, and the fixed block can end up there instead of immediately after the title block.

davisagli:
@avoinea nice!

Issue #6736, opened 2025-02-09
Bug: Protected Pages Show Blank Screen Instead of Redirecting to Login


Shyam-Raghuwanshi:
**Describe the bug**
When visiting a protected page (e.g., control panel) without being logged in, the user sees a blank page and in volto sees settings instead of being redirected to the login page.

**To Reproduce**

1. > Open an incognito window or log out from the site.

2. > Visit a protected page like:
https://demo.plone.org/controlpanel/site
https://volto.demo.plone.org/controlpanel/

Observe that the page is blank instead of redirecting to the login screen.

**Expected behavior**
Users should be redirected to the login page when trying to access a protected page without authentication.

stevepiercy:
Screenshots below from a private browser window, unauthenticated. In both, the `Log in` link displays in the upper right.

- https://volto.demo.plone.org/controlpanel/ displays a partial list of control panels.

![Image](https://github.com/user-attachments/assets/0fd7d96b-c849-4c31-ac1e-9f761c5e47da)

- https://volto.demo.plone.org/controlpanel/site displays an empty template page.

![Image](https://github.com/user-attachments/assets/c717db74-78e7-4804-87e5-55b77768ac5b)

VarshanMaj1:
This issue should be closed ig

<img width="824" alt="Image" src="https://github.com/user-attachments/assets/878e70df-4e6a-462f-9d10-1e7af31ccb7b" />
<img width="828" alt="Image" src="https://github.com/user-attachments/assets/9f03ce6e-bb56-4c66-afdc-99c0aa72c430" />

stevepiercy:
We might need to move this issue over to the demo sites repos, or the distribution repos, for a deployment. Until that's done, let's leave this issue open.

However, I noticed that this issue affects plone.org. It receives a HTTP OK 200 response, renders the control panels for a moment, then hides them. See the network requests in order, and observe that the two 401 Forbiddens occur after the 200 OK.

https://plone.org/controlpanel/

![Image](https://github.com/user-attachments/assets/85dceebc-da3b-4e20-af60-20edce1a8654)

sagnik2001:
is this issue still open ?

wesleybl:
> is this issue still open ?

Yes, this is still a problem. When we access:

http://localhost:3000/controlpanel/site

We still see a blank page.

sagnik2001:
Can I work on it ?

stevepiercy:
[...]

PR #8126, opened 2026-04-18
Don't show login link in the Unauthorized component if the user is already authenticated

wesleybl:
Ref: https://github.com/plone/volto/issues/6736#issuecomment-4274263444


packages/volto/news/8126.bugfix
```diff
@@ -0,0 +1 @@
+Don't show login link in the Unauthorized component if the user is already authenticated. @wesleybl
```
stevepiercy:
```suggestion
Don't show login link in the `Unauthorized` component if the user is already authenticated. @wesleybl
```

stevepiercy (changes requested):
Just some MyST markup on the component name, and then it's OK to merge. Thank you!

davisagli (approved):

stevepiercy (approved):

sneridagh:
@davisagli @wesleybl if the user is already authenticated, in theory should receive a 403, right? then it will never get to see the Unauthorized page. Unless we have edge cases in RESTAPI that this is not observed, right?

nileshgulia1:
Probably not related to the PR but I sometimes get another issue when a user is logged in and a token get's expired. The frontend crashes with 
```
/Users/nilesh/work/sources/eea/wise/marine-frontend/node_modules/superagent/lib/node/index.js:696
        err = new Error(msg);
              ^
Unauthorized
    at Request.callback (/Users/nilesh/work/sources/eea/wise/marine-frontend/node_modules/superagent/lib/node/index.js:696:15)
    at /Users/nilesh/work/sources/eea/wise/marine-frontend/node_modules/superagent/lib/node/index.js:906:18
    at IncomingMessage.<anonymous> (/Users/nilesh/work/sources/eea/wise/marine-frontend/node_modules/superagent/lib/node/parsers/json.js:19:7)
    at IncomingMessage.emit (node:events:530:35)
    at IncomingMessage.emit (node:domain:489:12)
    
 ```
    
    
The solution is to delete the token manually and re-login. Happens in Volto 17 and Volto 18.

wesleybl:
> @davisagli @wesleybl if the user is already authenticated, in theory should receive a 403, right? then it will never get to see the Unauthorized page. Unless we have edge cases in RESTAPI that this is not observed, right?

@sneridagh When we access:

http://localhost:3000/++api++/@controlpanels

with an authenticated user but without permission, we receive a `401` error. Is this wrong?

stevepiercy:
@wesleybl compare these two response codes.

- https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/401
- https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/403

403 means they're logged in, but forbidden from accessing the resource.

401 means they're not logged in, and therefore can't access any protected resource.

So, yeah, we've been doing it wrong in Volto. Even Classic UI doesn't do it right.

I had assumed that since Plone and Pyramid share the same foundation, that, of course, Plone had got it right, but I was dead wrong to make that assumption. Even in Classic UI, it's wrong.

image

wesleybl:
I see there was an issue for this, but it was closed: https://github.com/plone/plone.restapi/issues/251

wesleybl:
@jensens FYI

wesleybl:
@sneridagh So should I revert this PR? Or should I leave it as is, at least for now?

davisagli:
@sneridagh @stevepiercy You're right, a 403 would be the correct response code in this situation. But Zope does not make a distinction between 401 and 403 and when we've looked in the past it looked quite difficult to change it to do so. Someone is welcome to take another look but I probably won't unless there's a client who wants to pay for my time.

@wesleybl Since Zope does return Unauthorized for both authenticated and unauthenticated users, I think this is a reasonable way to handle it on the frontend.

stevepiercy:
We could:

- Reopen https://github.com/plone/plone.restapi/issues/251
- [Create an issue in Nick](https://github.com/robgietema/nick/issues) to return the correct response code and headers

davisagli:
I don't think reopening the plone.restapi issue is helpful. There's nothing that can be done in plone.restapi unless Zope is changed to allow for distinguishing the two different scenarios where the user folder's validate function returns False (because no user was authenticated, or because the authenticated user doesn't have the necessary roles).

Nick can also be changed to do something better, but Volto would still not be able to conclude anything about whether the user is logged in from a 401 response, since it has to support all backends.

sneridagh:
@davisagli I forgot about the 401/403 limitation in Zope. Thanks for the clarification! I guess let's leave it like it is then.

PR #8131, opened 2026-04-19
Fix redirect after login on non-content routes

davisagli:
This is a follow-up on discussion in https://github.com/plone/volto/issues/6736#issuecomment-4274450790 and below.

If the user tries to access a non-content route and sees the Unauthorized component, the link to the login form should keep the full route as the base, so that the user is taken back to the correct place after logging in.

### DIFF
--- a/packages/volto/news/+login-redirect.bugfix
+++ b/packages/volto/news/+login-redirect.bugfix
@@ -1,1 +1,1 @@
-Fix redirect after logging in starting from the Unauthorized component on a non-content route. @davisagli
+Fixed redirect after logging in when starting from the `Unauthorized` component on a non-content route, such as control panels. @davisagli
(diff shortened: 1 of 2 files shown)


packages/volto/news/+login-redirect.bugfix
```diff
@@ -0,0 +1 @@
+Fix redirect after logging in starting from the Unauthorized component on a non-content route. @davisagli
```
stevepiercy:
```suggestion
Fixed redirect after logging in when starting from the `Unauthorized` component on a non-content route, such as control panels. @davisagli
```

stevepiercy (changes requested):

stevepiercy (approved):

PR #8158, opened 2026-04-28
Backend for comments and suggestions

sneridagh:


packages/plate/components/editor/block-editor-kit.tsx
```diff
@@ -34,7 +34,7 @@ import { ToggleKit } from './plugins/toggle-kit';
 import { SplitHotkeyPlugin } from './plugins/split-hotkey';
 
 export const BlockEditorKit = [
-  ...AIKit,
+  // ...AIKit,
```
davisagli, line 37: Unrelated change?

packages/plate/components/editor/plate-plugins-context.tsx
```diff
@@ -0,0 +1,38 @@
...
+import type { TDiscussion, TDiscussionUser } from './plugins/discussion-kit';
+
+type PlatePluginsContextValue = {
+  currentUser: TDiscussionUser | null;
+  currentUserId: string | null;
+  discussions: TDiscussion[];
+  setDiscussions: React.Dispatch<React.SetStateAction<TDiscussion[]>>;
+  setUsers: React.Dispatch<
+    React.SetStateAction<Record<string, TDiscussionUser>>
+  >;
```
davisagli: @sneridagh I don't think we have a use for setUsers. New comments are always made by the currentUser. The backend discards the `users` key during deserialization. It is only metadata added during serialization, so that the frontend has the necessary data to show users who previously made comments.

packages/plate/components/editor/block-editor-kit.tsx
sneridagh: This is fine, not needed for now, it was also using the suggestions/comments infrastructure and failing. 
Since we are not using it at all for now, let's unwire it.

packages/plate/components/editor/plate-plugins-context.tsx
sneridagh: Done, I removed it from the code since it's not needed.

PR #8159, opened 2026-04-28
Add ReactAriaRouterProvider to enable client-side navigation support …

iFlameing:
…for React Aria Link components.


sneridagh (approved):
@pnicolli not a big deal, but heads up!

PR #8163, opened 2026-04-28
change hotkey for strikethrough in Slate editor

wesleybl:
Backport of #4197 to Volto 18.


davisagli:
This was a breaking change, so it's not eligible for backporting. Users should be able to install a patch release without worrying about the UX changing like this.

Issue #3827, opened 2022-11-02
Create end-user Editor documentation


stevepiercy:
What if we had end-user Editor documentation for Volto? And then in the UI, we could place the universal help icon (❔) that links to the relevant Volto Editor page in the docs, based on the context. This would keep the UI clean and leave space for future development. PyCharm does that for its UI, and it is super helpful to me.

Additionally this would create something for directionless newbies to work on that is relatively easy. These docs would be perfect for newbies. We could tag an issue with "33 needs: docs" and "41: lvl easy" to give a good entry point for them.

_Originally posted by @stevepiercy in https://github.com/plone/volto/issues/3811#issuecomment-1298030528_

stevepiercy:
Hidden Editor gem: https://github.com/plone/volto/pull/3815#issuecomment-1301020379

stevepiercy:
Undo: https://github.com/plone/volto/pull/3815#issuecomment-1298005656

stevepiercy:
Add block: https://github.com/plone/volto/pull/3815#issue-1430794104

stevepiercy:
From Discord: https://discord.com/channels/786421998426521600/862642144401031188/1042085679091425363

Drag and drop file upload: You can drag an image from your desktop file browser (Windows Explorer, Finder, whatever) to a Volto slate text block, and it will upload and create an image block for you.

Copy and paste file upload: You can paste an image from your desktop clipboard to the slate text block. So, on my desktop, I open the screenshot tool, I click "Copy to clipboard", I paste inside an empty text box (using ctrl+v), and I get an image block.

spereverde:
@stevepiercy is there a folder foreseen for end-user docs?

stevepiercy:
Nothing has been decided yet, and it is up for discussion.

For Volto, I suggested a new directory and index, `/volto/docs/source/user-manual/index.md`, to get started. A volunteer stepped up in Discord, SonGoku#8071, in the #newbies-introduction channel, to start on this.

We also might need to consider whether this documentation belongs under Volto or the main documentation. For now, I think keeping it in Volto gives it a better chance of getting written and better integrated with Volto developer docs.

MAX-786:
Thanks, Steve for mentioning me here...😊 I think I should consider changing my discord username tho😂 -SonGoku

kenmanheimer:
[...]

PR #8166, opened 2026-04-30
Implement the "AutoSave" feature (`withSaveAsDraft`) as opt-in

sneridagh:
We found multiple edge cases which need to be addressed. For now, I think it is not usable yet and production ready.


sneridagh:
@plone/volto-team Also, I realised that we have very few documentation about it, specially user facing.

stevepiercy:
> @plone/volto-team Also, I realised that we have very few documentation about it, specially user facing.

Yup. https://github.com/plone/volto/issues/3827 😢

Issue #8128, opened 2026-04-19
Control panels are viewable by unauthenticated users


stevepiercy:
**Describe the bug**

The control panels at `/controlpanel` are viewable by unauthenticated users.

<img width="1675" height="1527" alt="Image" src="https://github.com/user-attachments/assets/fdab3d35-48e8-4421-b7b4-161938d02446" />

They can't do anything, but viewing them is not a good look.

**To Reproduce**

Visit https://demo.plone.org/controlpanel.

**Expected behavior**

- An unauthenticated and unauthorized user should be presented with the unauthorized screen, similar to that when attempting to view private _content_, with a link to login, as shown in the screenshot below.
- An authenticated but unauthorized user should be presented with the unauthorized screen, similar to that when attempting to view private _content_, but _without_ a link to login.

<img width="1675" height="1527" alt="Image" src="https://github.com/user-attachments/assets/e9433a88-db38-4357-8ceb-a9d551de9712" />

**Software (please complete the following information):**

- Volto 18.32.1
- Plone 6.1.4
- plone.restapi 9.15.4
- CMF 3.7
- Zope 5.13
- Python 3.12.12 (main, Jan 13 2026, 03:13:28) [GCC 12.2.0]
- PIL 11.3.0 (Pillow)

**Additional context**

Supersedes https://github.com/plone/volto/issues/6736.

@Shyam-Raghuwanshi started work on https://github.com/plone/volto/pull/6738 to resolve #6736, but there are some issues with it.

wesleybl:
On the main branch,

http://localhost:3000/controlpanel

is working, but:

http://localhost:3000/controlpanel/site

is not.

davisagli:
Each control panel route component is responsible for its own permission check(s).

PR #8184, opened 2026-05-08
Show Unauthorized page via SSR if an anonymous user accesses the user control panel

wesleybl:
If an anonymous user accesses the user control panel URL, they will likely do so via SSR. Therefore, we display the Unauthorized page via SSR.

Ref #8128

### DIFF
--- a/packages/volto/src/components/manage/Controlpanels/index.tsx
+++ b/packages/volto/src/components/manage/Controlpanels/index.tsx
@@ -2,7 +2,13 @@ import loadable from '@loadable/component';
 import {
   getSystemInformation,
   listControlpanels,
+  getControlpanel,
 } from '@plone/volto/actions/controlpanels/controlpanels';
+import { listRoles } from '@plone/volto/actions/roles/roles';
+import { listUsers, getUser } from '@plone/volto/actions/users/users';
+import { listGroups } from '@plone/volto/actions/groups/groups';
+import { getUserSchema } from '@plone/volto/actions/userschema/userschema';
+import jwtDecode from 'jwt-decode';
 import { asyncConnect } from '@plone/volto/helpers/AsyncConnect';
 
 // CONTROLPANELS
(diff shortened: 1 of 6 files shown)


stevepiercy:
@wesleybl does this need tests?

wesleybl:
@stevepiercy Yes. It would be good to have a test. I'll add one.

wesleybl:
@stevepiercy I added the tests.

stevepiercy:
@wesleybl would you please update your branch and check the failing acceptance tests?

wesleybl:
@stevepiercy The error in the test is not related to this PR. I fixed it in: https://github.com/plone/volto/pull/8190

wesleybl:
@stevepiercy I updated the branch.

stevepiercy (approved):
Looks good to me, but let's get a technical review.

packages/volto/src/components/manage/Controlpanels/index.tsx
```diff
@@ -66,13 +72,54 @@ export const ConfigureRuleControlpanel = loadable(
...
+      const token = state.userSession.token;
+      if (token) {
+        const userId = jwtDecode(token).sub;
+        if (userId) {
+          return dispatch(getUser(userId));
+        }
+      }
+    },
+  },
+])(LoadableUsersControlpanel);
```
davisagli: I think it'd be better to move both LoadableUsersControlpanel and the new asyncConnect-enabled UsersControlpanel into UsersControlpanel.jsx. (Then import and re-export it here, for backward-compatibility.)

index files are not a good place for significant code, because they are often overlooked.

wesleybl: @davisagli I followed the method that was being used with `Controlpanels`:

https://github.com/plone/volto/blob/a49084356a61a11005459016dc675a2ee32f13ac/packages/volto/src/components/manage/Controlpanels/index.tsx#L17

That is, using `asyncConnect` in the index. Do you think we should change this for the other components?

davisagli: @wesleybl Ok, I didn't see it was already that way for other control panels. My personal preference would be to change it for all of them, but it doesn't need to happen in this PR.

wesleybl: @davisagli If the trend is to remove `asyncConnect` from the index, I prefer to fix this now. I intend to fix this in other control panels as well.

wesleybl: > I think it'd be better to move both LoadableUsersControlpanel and the new asyncConnect-enabled UsersControlpanel into UsersControlpanel.jsx (Then import and re-export it here, for backward-compatibility.)

Can I do it without exporting and treat it as a breaking change? That way we can clean up the index little by little.

davisagli: @wesleybl That is ok with me for Volto 19.

wesleybl: @davisagli It's not possible to move the loadable to `UsersControlpanel.jsx` because it would lazily try to load the file it's actually in, thus losing the lazy loading capability. But at least I moved the `asyncConnect` to `UsersControlpanel.jsx`.

davisagli: @wesleybl Oh right, I forgot which part was being lazy-loaded. Looks ok to me.

davisagli (approved):

PR #8193, opened 2026-05-09
Fixes an intermittent error in the Cypress tests blocks-slate-backspace.js and blocks-slate-delete.js

wesleybl:
Backport of #8190 to Volto 18.

### DIFF
--- a/packages/volto/news/8190.internal
+++ b/packages/volto/news/8190.internal
@@ -1,0 +1,1 @@
+Fixes an intermittent error in the Cypress tests blocks-slate-backspace.js and blocks-slate-delete.js. @wesleybl
--- a/packages/volto/cypress/tests/core/blocks/blocks-slate-backspace.js
+++ b/packages/volto/cypress/tests/core/blocks/blocks-slate-backspace.js
@@ -19,7 +19,7 @@ describe('Slate Backspace Behavior', () => {
   it('Backspace at start of second block deletes it and merges content into first block', () => {
     cy.getSlateEditorAndType('First block text');
 
-    cy.addNewBlock('slate');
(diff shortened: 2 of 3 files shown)


packages/volto/news/8190.internal
```diff
@@ -0,0 +1 @@
+Fixes an intermittent error in the Cypress tests blocks-slate-backspace.js and blocks-slate-delete.js. @wesleybl
```
stevepiercy, line 1: File names should use the file role. There's no need to forward port to the original, already merged PR.

```suggestion
Fixes an intermittent error in the Cypress tests {file}`blocks-slate-backspace.js` and {file}`blocks-slate-delete.js`. @wesleybl
```

stevepiercy (changes requested):

packages/volto/news/8190.internal
davisagli: @stevepiercy I don't see the point of using the file role. It isn't rendered well when I view the changelog in github (It shows `{file}` literally), and it's not rendered any differently from normal backquoted literals on https://6.docs.plone.org/volto/release-notes/index.html.

stevepiercy: It renders semantic HTML markup when built in the docs. Files are not inline literals, which have a different semantic HTML markup.

davisagli: Where is the different semantic markup used to do something differently? If it isn't, it's just busywork.

stevepiercy: ## File

`<code class="file docutils literal notranslate"><span class="pre">src/config.js</span></code>`

## Inline literal

`<code class="docutils literal notranslate"><span class="pre">html_meta</span></code>`

PR #8202, opened 2026-05-12
Translate block titles in Order tab

### LINKED ISSUE #7348: Order tab does not translate blocks titles
Even when the rest of the Volto is translated, the Order tab will not translate the titles for available blocks.

Image

ericof:
## Summary

- The Order tab in the block sidebar was rendering `config.blocks.blocksConfig[…].title` (a plain string like `"Image"`, `"Listing"`) as-is, ignoring the locale catalog. Visible to anyone using Volto in a non-English locale: the rest of the sidebar localized correctly, the Order tab did not.
- Extract the existing `intl.formatMessage({ id: x, defaultMessage: x })` pattern — already used in `BlockChooser` and `BlockType` — into a shared `formatMessageWithFallback` helper at [packages/volto/src/helpers/I18n/I18n.ts](packages/volto/src/helpers/I18n/I18n.ts).
- Route the three call sites (`Order/Item.jsx`, `BlockChooser.jsx`, `BlockType.tsx`) through the helper, removing the local `getFormatMessage` from `BlockChooser`.
- Add unit tests for the helper (translated case, fallback case, falsy passthrough for `undefined`/`null`/`''`, no-op assertion for falsy inputs).

The capitalized block title strings (`"Title"`, `"Image"`, `"Listing"`, …) already exist in the locale catalogs (verified in `pt_BR.json`), so this fix activates them without any catalog churn.

## Test plan

- [x] `pnpm test:ci src/helpers/I18n` passes (6/6 green).
- [x] `pnpm test:ci` for the three affected component directories passes (23 files / 45 tests green; no snapshot churn).
- [X] In a non-English locale (e.g. `pt-BR`), opening the Order tab on a page with several block types shows translated titles (`"Imagem"`, `"Listagem"`, …) instead of the English defaults.
- [X] Block Chooser still filters and renders block titles in the user's locale.
- [x] The `/controlpanel/block-type/<id>` page still shows the translated block title in its heading.

Closes #7348

### DIFF
--- a/packages/volto/news/7348.bugfix
+++ b/packages/volto/news/7348.bugfix
@@ -1,0 +1,1 @@
+Translate block titles displayed in the Order tab of the sidebar. @ericof
--- a/packages/volto/src/components/manage/BlockChooser/BlockChooser.jsx
+++ b/packages/volto/src/components/manage/BlockChooser/BlockChooser.jsx
@@ -10,6 +10,7 @@ import { useIntl, defineMessages } from 'react-intl';
 import Icon from '@plone/volto/components/theme/Icon/Icon';
 import AnimateHeight from 'react-animate-height';
 import config from '@plone/volto/registry';
+import { formatMessageWithFallback } from '@plone/volto/helpers/I18n/I18n';
 import upSVG from '@plone/volto/icons/up-key.svg';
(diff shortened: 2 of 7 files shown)


sneridagh (approved):
LGTM

PR #8208, opened 2026-05-14
Update to cypress 15.13.1

wesleybl:
Backport of #8089 to Volto 18.

### DIFF
--- a/packages/volto-testing/news/+dep-cypress.internal
+++ b/packages/volto-testing/news/+dep-cypress.internal
@@ -1,0 +1,1 @@
+Update devDependency: cypress 15.13.1. @davisagli
--- a/packages/volto-testing/package.json
+++ b/packages/volto-testing/package.json
@@ -41,7 +41,7 @@
     "@testing-library/jest-dom": "6.4.2",
     "@testing-library/react": "12.1.5",
     "axe-core": "4.8.4",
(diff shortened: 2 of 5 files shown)


davisagli (commented):
That's a major version bump which is also forced on any project using Volto 18 that depends on the volto-testing package. Seems breaking to me.

wesleybl:
@davisagli If the update didn't break Volto's tests, it's unlikely to break tests for other packages.

Furthermore, if we check the migration guide:

https://docs.cypress.io/app/references/migration-guide#Migrating-to-Cypress-150

we'll see that most of the breaking changes are because Cypress stopped supporting package versions that Volto 18 no longer supports.

So I think it's safe to update.

davisagli:
@wesleybl I agree that it _mostly_ looks ok, but I see things in the migration guide that I think would be a bit surprising to get in a Volto patch release:
- dropped support for old Linux distributions and macOS 11
- dropped support for old browsers
- cy.origin() needs to be set explicitly for tests that visit more than one origin

We would need to add an entry to the upgrade guide linking to the Cypress one -- and if it needs an upgrade guide, it doesn't belong in a patch release (unless we can't avoid it).

wesleybl:
@davisagli Okay, I'll close this. Maybe I'll update to the latest version of 13.

PR #8216, opened 2026-05-14
[codex] Handle stale auth cookies on public requests

sneridagh:
## Summary

This PR fixes Seven's behavior when the `auth_seven` cookie contains a stale or invalid token and the user requests public content.

Previously, Seven initialized the Plone client with whatever token was present in the cookie and sent it as `Authorization: Bearer <token>` on content and asset requests. When that token was no longer valid, the backend returned `401`, and Seven surfaced that failure through the root `ErrorBoundary` even for public pages.

This change makes Seven recover gracefully instead of failing hard.

## What changed

- Added anonymous fallback behavior in `apps/seven/app/middleware.server.ts`.
- When a token-authenticated content or site request returns `401`, Seven now reinitializes the client without the token, retries the request anonymously, and continues rendering if anonymous access succeeds.
- When a stale cookie is detected and anonymous fallback succeeds, Seven now clears the `auth_seven` cookie on the final page response.
- Added the same stale-cookie recovery for proxied asset requests (`@@images`, `@@download`, `@@site-logo`, `@portrait`) so public assets still resolve when the cookie is invalid.
- Avoided sending `Authorization: Bearer undefined` on anonymous proxied asset requests.
- Decoupled `getUser()` lookup from the main content/site fetch so user lookup failures do not break otherwise public rendering.
- Centralized cookie-clearing helpers in `@plone/react-router` with shared helpers for clearing the auth cookie on a response or generating the clearing `Set-Cookie` header.
- Added regression tests for the anonymous retry flow, cookie clearing behavior, and the updated loader response shape.
- Added towncrier news entries for `apps/seven` and `packages/react-router`.

## Impact

- Public pages keep rendering even when the user's auth cookie has expired or become invalid.
- The stale auth cookie is removed automatically once Seven confirms anonymous access works.
- Shared auth-cookie cleanup behavior is available centrally from `@plone/react-router`.

## Validation

- `pnpm exec vitest run app/root.test.tsx app/middleware.server.test.ts`
- `pnpm --filter seven typecheck`
- `pnpm --filter @plone/react-router build`

### DIFF
--- a/apps/seven/app/middleware.server.ts
+++ b/apps/seven/app/middleware.server.ts
@@ -171,13 +171,13 @@ export const fetchPloneContent: Route.MiddlewareFunction = async (
   };
 
   try {
-    const [content, site] = await Promise.all([
+    const [content, site, user] = await Promise.all([
       cli.getContent({ path, expand }),
       cli.getSite(),
+      userId
+        ? Promise.resolve(cli.getUser({ id: userId })).catch(() => null)
+        : null,
     ]);
-    const user = userId
-      ? await cli.getUser({ id: userId }).catch(() => null)
-      : null;
 
     setPloneContext(content, site, user?.data ?? null);
   } catch (error: any) {
--- a/apps/seven/app/middleware.server.ts
+++ b/apps/seven/app/middleware.server.ts
@@ -174,9 +174,7 @@ export const fetchPloneContent: Route.MiddlewareFunction = async (
     const [content, site, user] = await Promise.all([
       cli.getContent({ path, expand }),
       cli.getSite(),
-      userId
-        ? Promise.resolve(cli.getUser({ id: userId })).catch(() => null)
-        : null,
+      userId ? cli.getUser({ id: userId }).catch(() => null) : null,
     ]);
 
     setPloneContext(content, site, user?.data ?? null);
--- a/apps/seven/app/middleware.server.ts
+++ b/apps/seven/app/middleware.server.ts
@@ -1,7 +1,7 @@
 import { jwtDecode } from 'jwt-decode';
 import { data, createContext } from 'react-router';
 import { flattenToAppURL } from '@plone/helpers';
-import { getAuthFromRequest } from '@plone/react-router';
+import { clearAuthOnResponse, getAuthFromRequest } from '@plone/react-router';
 import config from '@plone/registry';
 import type PloneClient from '@plone/client';
 import type { Route } from './+types/root';
@@ -16,6 +16,22 @@ export const ploneSiteContext =
 export const ploneUserContext = createContext<
   Awaited<ReturnType<PloneClient['getUser']>>['data'] | null
 >(null);
+export const ploneClearAuthCookieContext = createContext<boolean>(false);
+
+function getAuthorizedResourceHeaders(
+  request: Request,
+  token?: string,
+): HeadersInit {
+  const headers = new Headers(request.headers);
+
+  if (token) {
+    headers.set('Authorization', `Bearer ${token}`);
+  } else {
+    headers.delete('Authorization');
+  }
+
+  return headers;
+}
 
 export const installServerMiddleware: Route.MiddlewareFunction = async (
   { request, context },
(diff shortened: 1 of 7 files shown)


packages/react-router/src/index.ts
```diff
@@ -109,14 +109,33 @@ export async function setAuthOnResponse(
...
 }
 
+export async function clearAuthOnResponse(
+  response: Response,
+  options?: Parameters<typeof cookie.serialize>[1],
+) {
+  const header = await cookie.serialize('', {
+    maxAge: 0,
+    ...options,
+  });
```
pnicolli:
```suggestion
  const header = await getClearAuthCookieHeader();
```

Maybe this to unify logic and maintenance?

packages/react-router/src/index.ts
```diff
@@ -109,14 +109,33 @@ export async function setAuthOnResponse(
...
+  return await cookie.serialize('', {
+    maxAge: 0,
+    ...options,
+  });
+}
+
 export async function requireAuthCookie(request: Request) {
   const token = await getAuthFromRequest(request);
   if (!token) {
     throw redirect('/login', {
```
pnicolli:
```suggestion
    throw clearAuthCookieHeader(redirect('/login'));
```

This could be like this as well for consistency.. or the one below could be like this. In any case, I feel like using the same syntax in both functions is better for readability, otherwise the reader might think that there might be a difference.

We might even add a parameter to `redirectWithClearedCookie` and reuse it here, potentially.

apps/seven/app/middleware.server.ts
```diff
@@ -117,19 +158,63 @@ export const fetchPloneContent: Route.MiddlewareFunction = async (
...
   try {
-    const [content, site, user] = await Promise.all([
+    const [content, site] = await Promise.all([
       cli.getContent({ path, expand }),
       cli.getSite(),
-      userId ? cli.getUser({ id: userId }) : Promise.resolve(null),
     ]);
+    const user = userId
+      ? await cli.getUser({ id: userId }).catch(() => null)
+      : null;
```
pnicolli: This change serializes the third call. I would rather have all three queries fired in parallel and check them separately, for performance.
Something like:

```
const [contentPromise, sitePromise, userPromise] = [
  cli.getContent(...),
  cli.getSite(),
  userId ? cli.getUser(...) : Promise.resolve(null),
];
const [content, site] = await Promise.all([contentPromise, sitePromise]);
let user = null;
try {
  user = await userPromise;
} catch {}
```

Or, actually, also catching the user promise right away might work.

```
try {
    const [content, site, user] = await Promise.all([
      cli.getContent({ path, expand }),
      cli.getSite(),
      userId ? cli.getUser({ id: userId }).catch(() => null) : Promise.resolve(null),
    ]);
...
```

pnicolli (changes requested):

packages/react-router/src/index.ts
sneridagh: Done!

packages/react-router/src/index.ts
sneridagh: Done!

apps/seven/app/middleware.server.ts
sneridagh: Yep, I improved it.

apps/seven/app/middleware.server.ts
```diff
@@ -117,19 +158,63 @@ export const fetchPloneContent: Route.MiddlewareFunction = async (
...
+  };
+
   try {
     const [content, site, user] = await Promise.all([
       cli.getContent({ path, expand }),
       cli.getSite(),
-      userId ? cli.getUser({ id: userId }) : Promise.resolve(null),
+      userId
+        ? Promise.resolve(cli.getUser({ id: userId })).catch(() => null)
+        : null,
```
pnicolli:
```suggestion
      userId
        ? cli.getUser({ id: userId }).catch(() => null)
        : Promise.resolve(null),
```

I think this is the correct syntax, catch already returns a Promise and does not need the `.resolve`, while `null` itself is not a Promise... but `Promise.all` maybe works fine anyway, so maybe this would also be correct:

```
      userId
        ? cli.getUser({ id: userId }).catch(() => null)
        : null,
```

pnicolli (changes requested):

apps/seven/app/middleware.server.ts
sneridagh: Done!

pnicolli (approved):

PR #8221, opened 2026-05-15
Formtabs tab links

davisagli:
New PR to replace https://github.com/plone/volto/pull/7308 which was merged before the tests were passing.

@Wagner3UB Is there also a PR to make this change for the `main` branch? I want to avoid that we only fix something in 18.x.x and forget to do it on main.


davisagli:
@Wagner3UB The coresandbox acceptance test failures were not an existing problem with the coresandbox tests, but an actual regression caused by this change. Buttons default to `type="submit"` so the button for navigating to a tab was actually submitting the form (to the default form action which is the current URL, which was not set up to handle a POST, so it got a 404).

Fixed by adding `type="button"`.

Sorry that I took so long to take a closer look at this.

At this point I think the tests will pass. But please make and merge a PR to fix it on the main branch before you merge this one for 18.x.x.

davisagli:
Oh, I found the existing PR for main: https://github.com/plone/volto/pull/8219

PR #8226, opened 2026-05-18
Add nick as an embedded library workflow

sneridagh:


sneridagh:
Transferred: https://github.com/plone/aurora/pull/99

PR #8227, opened 2026-05-18
Sneridagh plone contents

sneridagh:


packages/contents/routes/contents.tsx
```diff
@@ -94,7 +97,17 @@ export default function Contents() {
...
+    const [sort_on, sort_order] = value.split('|');
+    const params = new URLSearchParams(window.location.search);
+
+    params.set('sort_on', sort_on);
+    params.set('sort_order', sort_order);
+    params.delete('page');
+
+    const querystring = params.size > 0 ? '?' + params.toString() : '';
+    navigate(`/@@contents${content['@id']}${querystring}`);
+  };
```
pnicolli, line 110: This was not the purpose of this function. The purpose of this in the previous implementation was to actually sort items on the backend, moving stuff around in the folder.
I know that Quanta also wants this feature though when clicking column names, I will think of a way to make both happen.

pnicolli (approved):

PR #8236, opened 2026-05-18
Replace moment with formatDate in theme display widgets (PR 1)

avoinea:
- Remove direct `import moment from 'moment'` from DateWidget and DatetimeWidget
- Use formatDate from @plone/volto/helpers/Utils/Date instead
- Add MOMENT_FORMAT_MAP to Date.js for backward-compatible moment token support
- formatDate now accepts both Intl.DateTimeFormat options and moment-style tokens
- Unknown string tokens fall back to short_date_format with a dev-mode console.warn
- Add tests for all moment-style format tokens (ll, lll, LLLL, L, LT)

See plone/aurora#94


packages/volto/news/6732.feature
```diff
@@ -0,0 +1 @@
+Replace moment.js with native Intl formatting in DateWidget and DatetimeWidget theme display widgets. The `formatDate` helper now accepts moment-style token strings ('ll', 'lll', 'LLLL', 'L', 'LT') for backward compatibility. @avoinea
```
wesleybl, line 1: Isn't this a breaking change? The way dates are displayed will change slightly.

packages/volto/src/helpers/Utils/Date.js
```diff
@@ -32,13 +47,23 @@ export const toDate = (d) =>
...
   includeTime, // true if short date format should include time
   formatToParts = false,
 }) {
   date = toDate(date);
+  // Resolve moment-style token strings to Intl format options
+  if (typeof format === 'string') {
+    if (process.env.NODE_ENV !== 'production' && !MOMENT_FORMAT_MAP[format]) {
+      // eslint-disable-next-line no-console
+      console.warn(
+        `[formatDate] Unknown format token "${format}", falling back to short_date_format`,
```
wesleybl: It's also best to display the supported formats so the user isn't left wondering what to do.

wesleybl (changes requested):

packages/volto/src/helpers/Utils/Date.js
avoinea: @wesleybl fixed.

packages/volto/news/6732.feature
avoinea: I tried to keep it backward compatible. If you look at the tests snapshots you will see the output is almost the same 😃

davisagli (approved):

wesleybl (approved):

wesleybl:
@davisagli Could we have this in Volto 18? The moment wasn't actually removed.

davisagli:
@wesleybl I would keep it for 19 only, since there are some minor differences in the formatting.

PR #8240, opened 2026-05-18
New Quanta components Modal, Dialog and Form

pnicolli:
Added new Modal and Form quanta components, exported already present Dialog component.
Will be needed to finish the Contents view, at least.


packages/components/src/components/Modal/Modal.quanta.stories.tsx
```diff
@@ -0,0 +1,52 @@
...
+import { Button } from '../Button/Button.quanta';
+
+import {
+  Dialog,
+  DialogTrigger,
+  Heading,
+  Input,
+  Label,
+  TextField,
+} from 'react-aria-components';
```
davisagli: We should make it possible to import more of these from `@plone/components`.

pnicolli: I updated the story to import all that we can from `@plone/components`. I had actually overlooked the quality of the story here and just copied over from the basic one but it definitely deserves a proper example on how to use `@plone/components` for a modal here. Now everything is from `@plone/components` except the couple pieces that are not available yet (we will need to figure out if actually everything will be re-exported from us, but that's a separate discussion)

davisagli:
@pnicolli Shouldn't there be some space between the input and button?

Screenshot 2026-05-19 at 11 12 50 AM

pnicolli:
> Shouldn't there be some space between the input and button?
> 
> Screenshot 2026-05-19 at 11 12 50 AM

@davisagli Yes, definitely. I added the Form quanta component with the proper styles.

davisagli (approved):

PR #8246, opened 2026-05-19
Integrate links with ObjectBrowser

sneridagh:
@pnicolli ready


packages/cmsui/components/BlockEditor/plugins/link-kit.tsx
```diff
@@ -0,0 +1,696 @@
...
+  useEditorPlugin,
+  useEditorRef,
+  useEditorSelection,
+  useFormInputProps,
+  usePluginOption,
+} from 'platejs/react';
+import { useFetcher } from 'react-router';
+import { Dialog, Heading } from 'react-aria-components';
+import { flattenToAppURL, isInternalURL } from '@plone/helpers';
+import { Modal } from '@plone/components';
```
pnicolli: There is a Modal component now in `@plone/components/quanta`

packages/cmsui/components/BlockEditor/plugins/link-kit.tsx
```diff
@@ -0,0 +1,696 @@
...
+    <ObjectBrowserProvider
+      config={{
+        mode: 'single',
+        title: 'Select content',
+        initialPath,
+        widgetOptions: {
+          pattern_options: {
+            maximumSelectionSize: 1,
+          },
+        },
```
pnicolli: This should not be needed anymore iirc. We merged a fix for the object browser widget today to actually work properly in single mode

sneridagh: Removed!

packages/cmsui/components/BlockEditor/plugins/link-kit.tsx
sneridagh: It can't be switched, because the quanta modal is assuming centered in the screen. 
I will check which are the assumptions of the objectbrowser when instantiated from the UI (not the widget)

sneridagh: At the end, I added a `ObjectBrowserModal` and enhanced it for not having to spawn your own Modal every single time that it's not used as a widget.

Issue #3753, opened 2022-10-12
Remove moment or luxon as dependency


tiberiuichim:
It's not ok to depend on 2 different libraries that do the same thing. In addition, moment is deprecated.

ksuess:
No luxon anymore since https://github.com/plone/volto/pull/3781

PR #8250, opened 2026-05-19
Replace moment.js in Comments and EventDatesInfo components (PR 2/4)

avoinea:
Part of the ongoing effort to remove moment.js from Volto core while keep it backward compatible - refs plone/aurora#94

### DIFF
--- a/packages/volto/news/6732.feature
+++ b/packages/volto/news/6732.feature
@@ -1,0 +1,1 @@
+Replace moment.js with native Intl formatting in Comments and EventDatesInfo components. @avoinea
--- a/packages/volto/news/6732.feature
+++ b/packages/volto/news/6732.feature
@@ -1,1 +1,1 @@
-Replace moment.js with native Intl formatting in Comments and EventDatesInfo components. @avoinea
+Replace `moment.js` with native `Intl` formatting in `Comments` and `EventDatesInfo` components. @avoinea
--- a/packages/volto/src/components/theme/View/EventDatesInfo.jsx
+++ b/packages/volto/src/components/theme/View/EventDatesInfo.jsx
@@ -7,7 +7,14 @@ import { injectLazyLibs } from '@plone/volto/helpers/Loadable/Loadable';
 import { useSelector } from 'react-redux';
 import { formatDate } from '@plone/volto/helpers/Utils/Date';
 
+/**
+ * @deprecated Use the native Date API directly. Will be removed in Volto 20.
+ */
 export const datesForDisplay = (start, end, moment) => {
+  // eslint-disable-next-line no-console
+  console.warn(
+    'datesForDisplay is deprecated and will be removed in Volto 20. Use the native Date API directly.',
+  );
   const mStart = moment(start);
   const mEnd = moment(end);
   if (!mStart.isValid() || !mEnd.isValid()) {
--- a/packages/volto/src/components/theme/View/EventDatesInfo.jsx
+++ b/packages/volto/src/components/theme/View/EventDatesInfo.jsx
@@ -3,9 +3,9 @@ import PropTypes from 'prop-types';
 import { List } from 'semantic-ui-react';
 import cx from 'classnames';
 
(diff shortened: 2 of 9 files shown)


packages/volto/news/6732.feature
```diff
@@ -0,0 +1 @@
+Replace moment.js with native Intl formatting in Comments and EventDatesInfo components. @avoinea
```
davisagli:
```suggestion
Replace `moment.js` with native `Intl` formatting in `Comments` and `EventDatesInfo` components. @avoinea
```

davisagli (approved):

packages/volto/src/components/theme/View/EventDatesInfo.jsx
```diff
@@ -3,9 +3,9 @@ import PropTypes from 'prop-types';
 import { List } from 'semantic-ui-react';
 import cx from 'classnames';
 
-import { toBackendLang } from '@plone/volto/helpers/Utils/Utils';
 import { injectLazyLibs } from '@plone/volto/helpers/Loadable/Loadable';
 import { useSelector } from 'react-redux';
+import { formatDate } from '@plone/volto/helpers/Utils/Date';
 
 export const datesForDisplay = (start, end, moment) => {
```
wesleybl, line 13: We still have this moment here. Will it be removed in the future? Should we note this down somewhere to remember to remove it? In an issue?

wesleybl (approved):

packages/volto/src/components/theme/View/EventDatesInfo.jsx
davisagli: @wesleybl I think @avoinea is still working on it this week.

avoinea: @wesleybl I deprecated `datesForDisplay` as it's not used in Volto Core anymore but maybe some add-on is using it. Thus we can safely remove it in the next major release, Volto 20.

wesleybl: @avoinea Is this enough to remind us that removal is necessary? Wouldn't it be better to open an issue?

avoinea: @wesleybl moment js is deprecated. See also #3753

avoinea: @wesleybl @davisagli 

@robgietema  mentioned that if we can't completely replace moment js with vanilla JS we can replace it with https://day.js.org/.

wesleybl: > @robgietema mentioned that if we can't completely replace moment js with vanilla JS we can replace it with https://day.js.org/.

@avoinea @robgietema Could you explain that in more detail?

avoinea: @wesleybl Not for this PR. 

This one is clean enough and backward compatible and can safely be merged.

But it seems that I am hitting a wall in the DatetimeWidget / TimeWidget.

stevepiercy: I'd also question why to use _any_ moment-like library, per https://github.com/plone/Products.CMFPlone/issues/4115#issuecomment-2665112972, https://github.com/plone/Products.CMFPlone/issues/4115#issuecomment-2665112991, and https://momentjs.com/docs/#/-project-status/. It doesn't make sense to me to force confusion unto the end user, as we experience on every ploneconf.org website that used it. Specifically, times of events would display in the user's timezone, not in that of the location.

Sorry to belabor the point, but I have a keen interest in all things date and time related.

pnicolli: Also remember we already have `@internationalized/date` as a dependency because of `react-aria-components`, it helps in some situations.
https://react-aria.adobe.com/internationalized/date/

PR #8257, opened 2026-05-20
[Seven] Component component

arybakov05:
Closes plone/aurora#69


sneridagh (changes requested):
It can't go to @plone/components, sorry my bad specifying it in the issue. Let's put it in @plone/layout. Thanks!

pnpm-lock.yaml
pnicolli, line 1: Please double check this file, try to run `pnpm install` again and see if it is still needed, it feels like it should not be here since the dependencies didn't change

arybakov05: It didn't have the latest commits from yesterday. I just updated it.

pnicolli (approved):

PR #8259, opened 2026-05-20
fix(Subpath): dispatch async actions for controlpanels on /controlpanels

### LINKED ISSUE #8258: volto 19(Subpath feature): ssr calls fails in /controlpanel
If you go to https://demo.plone.org/next/controlpanel , and refresh you see some of the controlpanels are missing. This is happening due to stripping out prefix paths in ssr asyncConnect calls. 

The fix is probably: https://github.com/eea/volto-prefixpath/pull/33

nileshgulia1:
closes https://github.com/plone/volto/issues/8258

### DIFF
--- a/packages/volto/news/8258.bugfix
+++ b/packages/volto/news/8258.bugfix
@@ -1,0 +1,1 @@
+fix(Subpath): dispatch async actions for controlpanels on /controlpanels @nileshgulia1
--- a/packages/volto/news/8258.bugfix
+++ b/packages/volto/news/8258.bugfix
@@ -1,1 +1,1 @@
-fix(Subpath): dispatch async actions for controlpanels on /controlpanels @nileshgulia1
+Fixed dispatched async actions for controlpanels on /controlpanels in subpath deployments @nileshgulia1
--- a/packages/volto/src/helpers/Site/index.js
+++ b/packages/volto/src/helpers/Site/index.js
@@ -1,5 +1,12 @@
 import { GET_SITE } from '@plone/volto/constants/ActionTypes';
 import { getSite } from '@plone/volto/actions/site/site';
+import {
+  listControlpanels,
+  getSystemInformation,
+} from '@plone/volto/actions/controlpanels/controlpanels';
+import config from '@plone/volto/registry';
+
+const prefixPath = config.settings.subpathPrefix || '';
 
 const getSiteAsyncPropExtender = {
   path: '/',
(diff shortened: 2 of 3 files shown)


nileshgulia1:
@wesleybl Does this ever happened with any of your prefixed site as well?

packages/volto/news/8258.bugfix
```diff
@@ -0,0 +1 @@
+fix(Subpath): dispatch async actions for controlpanels on /controlpanels @nileshgulia1
```
pnicolli:
```suggestion
Fixed dispatched async actions for controlpanels on /controlpanels in subpath deployments @nileshgulia1
```

Don't take my version as final, just want to recommend a human readable log line since it goes in the changelog without being transformed

packages/volto/src/helpers/Site/index.js
```diff
@@ -18,4 +25,44 @@ const getSiteAsyncPropExtender = {
...
+        key: 'controlpanels',
+        promise: ({ location, store: { dispatch } }) =>
+          __SERVER__ && dispatch(listControlpanels()),
+      });
+    }
+    return dispatchActions;
+  },
+};
+
+const SystemInfoAsyncPropExtender = {
```
pnicolli: I would recommend consistent naming and casing for these variable names, so at least lowercase `s` or maybe starting with `get`. I know that not all start with `get` but at least all start with a verb like `get` or `list`.

pnicolli (changes requested):

packages/volto/src/helpers/Site/index.js
nileshgulia1: Agree. Done in [2a2755f](https://github.com/plone/volto/pull/8259/commits/2a2755f95a9cb97c8c132f87a581e7b489989666)

pnicolli (approved):

pnicolli:
Looks good to me but I don't have good examples to test the bugfix on, so I would recommend also waiting for @wesleybl to review

wesleybl:
@nileshgulia1 I don't understand why this is necessary. These actions are already added here:

https://github.com/plone/volto/blob/e0698d5363ea80d4fd84600d39f538fe5e7c10b1/packages/volto/src/components/manage/Controlpanels/index.tsx#L17-L28

I couldn't reproduce the problem locally. Even at https://demo.plone.org/controlpanel, I managed to reproduce it once, but then I couldn't anymore. Something is strange.

PR #8265, opened 2026-05-21
Add DialogTrigger component

jnptk:


davisagli (commented):
I would put these in the same file as the existing Dialog components, since it's closely related. Then it's possible to import them both in one line.

Also we should re-export them from https://github.com/plone/volto/blob/main/packages/components/src/index.ts like we do for the other components.

davisagli (approved):

pnicolli:
@sneridagh should this be targeted at the main branch or the seven branch?

davisagli:
@pnicolli probably both. We need it for a volto 19 add-on

sneridagh:
@pnicolli Both, I guess. @jnptk can you please take care of the forwardport PR?

jnptk:
@sneridagh forwardport pr: https://github.com/plone/volto/pull/8270

PR #8268, opened 2026-05-21
[Seven] Slots Manager View

### LINKED ISSUE #8264: [Aurora] Slots Manager View
The viewlet manager of Plone's Classic UI  (see https://classic.demo.plone.org/en/manage-viewlets) can be ported to Seven, providing a structured overview over existing slots, as well as their subslots.

PR #8268 implements a proof of concept (which will need refinement later) of a such slots manager view. It adds the access protected `@@slots` view, which offers two different representations.

The first one is a flat representation of all slots as can be read from `config.slots` that have or had components associated with them from component registration.
Image

The second one is more complex and renders the actual hierarchy of slots. Based on the use of the `SlotRenderer` inside components, a tree can recursively be created by following the rendered slots and the components registered for these slots. The result looks like this:
Image

In a later version, some more features of the viewlet manager can be added to the slots manager view, such as unregistering ("hiding") components or rendering a preview of them.

arybakov05:
see #8264

### DIFF
--- a/packages/registry/bin/slot-tree.js
+++ b/packages/registry/bin/slot-tree.js
@@ -1,0 +1,89 @@
+import fs from 'fs';
+import path from 'path';
+
+// Uses the Vite module graph rather than file scanning so only truly-loaded
+// modules are included and addons in node_modules are covered naturally.
+async function findSlotRendererImporters(server) {
+  const resolved = await server.pluginContainer.resolveId(
+    '@plone/layout/slots/SlotRenderer',
+    undefined,
+    { ssr: true },
+  );
+  if (!resolved) return [];
+
+  const mod = server.moduleGraph.getModuleById(resolved.id);
+  if (!mod) return [];
+
+  return [...mod.importers].filter((m) => m.file);
+}
+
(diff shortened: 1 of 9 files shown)


packages/registry/bin/slot-tree.js
```diff
@@ -0,0 +1,89 @@
+import fs from 'fs';
```
sneridagh, line 1: @arybakov05 Why do you have to do this static analysis of the build? Wouldn't be enough that the route reads the config (`config.slots`) and builds it from there?

sneridagh:
> [!important]
> Hello! Thanks for contributing to Plone Aurora! FYI, since June 5th, Plone Aurora lives in its own repository, thus, the `seven` branch does not accept any merges. If you want to merge your PR, please follow instructions: https://github.com/plone/volto/issues/8318
> Sorry for the inconvenience!

packages/registry/bin/slot-tree.js
arybakov05: @sneridagh The issue with `config.slots` is that it only captures the registered slots in a flat hierarchy. The static analysis also captures (empty) SlotRenderers and is able to reconstruct the hierarchy as nested in the implementation.

PR #8269, opened 2026-05-22
Fix Nick usage of babel/core, add bcrypt and node-gyp to the externals

sneridagh:


apps/seven/.gitignore
```diff
@@ -8,3 +8,4 @@ node_modules
 /packages/
 public/locales/
 /stats*
+var/
```
pnicolli, line 11: Is it a var folder always at the root, for backend data or something? I would recommend `/var/` if that is the case, to narrow and be specific

sneridagh: no, it's a var that is spawned just here... (only in nick as a embedded library setups). Maybe we can improve it later, but now it's a leftover that the bootstrap generates.

PR #8271, opened 2026-05-22
[Seven] Login layout implementation

arybakov05:
Since login/logout was already implemented, this PR only implements the layout for the login page. See https://www.figma.com/design/YpaRSKjFcFaAkVYnTSVGuz/Volto-UI-%E2%80%93-Quanta?node-id=1771-11364&t=x1EZjffFClXU9xsd-0

Closes plone/aurora#24


pnicolli (approved):
LGTM. @sneridagh what do you think about the implementation with the slots? Do you have another opinion of how to make it customizable? I think this should work fine

sneridagh:
I think this is fine for now. I haven't tried it out. But you told me this is the direct implementation of the figma's login form, right? I think Albert's design in there went a bit in a more daring and "let's see what happens" direction, not set in stone kind of thing and should be taken with a ponch of salt. 

That being said, we will need to change it to a big standalone Plone logo by default, not Volto's one (again I can't see the outcome, I am on mobile).

But yeah, I like the flexibility it gave us.

arybakov05:
Yeah it implements Figma's login form, including the Volto logo. I will change it to the Plone logo

arybakov05:
I swapped the small Volto logo with the Plone logo. The existing Plone SVG was white, however I changed it to use `currentcolor` instead. But that's only the small logo. Changing the big hero would require more effort. I also updated the return link to use the new `asButton` prop, so the classes don't have to be added manually.

Here's a preview of what it looks like on desktop:
image

and on mobile:
image

The "my cool site" is the value for the registry entry `plone.site_title`.

stevepiercy:
The Volto design still appears on the right half on desktop, and the color is not Plone-ish.

See:
- https://plone.org/foundation/copyright-licensing-logo/logoguidelines.pdf/@@download/file
- https://plone.org/foundation/copyright-licensing-logo/plone_logo_pack_v2.zip/@@download/file

For some inspiration, maybe https://2026.ploneconf.org/ ?

arybakov05:
> The Volto design still appears on the right half on desktop, and the color is not Plone-ish.
> 
> See:
> 
> * https://plone.org/foundation/copyright-licensing-logo/logoguidelines.pdf/@@download/file
> * https://plone.org/foundation/copyright-licensing-logo/plone_logo_pack_v2.zip/@@download/file
> 
> For some inspiration, maybe https://2026.ploneconf.org/ ?

Yes, I edited my message, saying only the left logo is swapped. I also left the orange since it matches the section on the right. The right section is exported straight from the Figma design, so I would need someone to redesign it, unless we want to dump this design and go for something simpler

ionlizarazu:
If we are not going to have any header or footer, what about the actual site branding?
I mean, are we going to have (by default) the same login page for all of our "seven" sites, and have to replace it site by site?
I would add the site logo replacing the small one and, if needed, add some "seven" image in the right side.
My two cents.

sneridagh:
> If we are not going to have any header or footer, what about the actual site branding? I mean, are we going to have (by default) the same login page for all of our "seven" sites, and have to replace it site by site? I would add the site logo replacing the small one and, if needed, add some "seven" image in the right side. My two cents.

I agree that we can add some default branding, eg. if the logo is customized, add it instead of the default (Plone) one. We can improve and complete the customization story later.

I added some documentation too.

sneridagh:
@stevepiercy I see your logo concerns, but I think we can improve/fix later. These were the mocks for Quanta styling, let's implement them, and discuss later if they are appropiate or not.

sneridagh:
@stevepiercy it's not the first PR merged for @arybakov05 and the puppy with the rainbow shows on every single time we merge 😅 any idea?

arybakov05:
@sneridagh It's because I haven't committed to main branch yet, only to seven

sneridagh:
@arybakov05 😱

pnicolli:
@sneridagh Yeah, we checked this briefly in Bonn and I guess that oh well @arybakov05 is just getting a whole lot of puppies :)

PR #8273, opened 2026-05-23
Remove unused postcss-overrides

wesleybl:
Backport of #8137 to Volto 18.

### DIFF
--- a/packages/volto/news/+dep-postcss-overrides.internal
+++ b/packages/volto/news/+dep-postcss-overrides.internal
@@ -1,0 +1,1 @@
+Remove unused devDependency `postcss-overrides`. @davisagli
--- a/packages/volto/package.json
+++ b/packages/volto/package.json
@@ -359,7 +359,6 @@
     "postcss-less": "6.0.0",
     "postcss-load-config": "^6.0.1",
     "postcss-loader": "^8.2.1",
-    "postcss-overrides": "3.1.4",
     "postcss-scss": "4.0.9",
     "razzle": "4.2.18",
     "razzle-dev-utils": "4.2.18",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1038,9 +1038,6 @@ importers:
       postcss-loader:
         specifier: ^8.2.1
         version: 8.2.1(postcss@8.5.15)(typescript@5.9.3)(webpack@5.105.4)
-      postcss-overrides:
-        specifier: 3.1.4
-        version: 3.1.4
       postcss-scss:
         specifier: 4.0.9
         version: 4.0.9(postcss@8.5.15)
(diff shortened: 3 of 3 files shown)


sneridagh:
@wesleybl as far as I remember, this one is needed. I created it for IDEs to be aware of scss in .overrides (semanticui) files. It does not hurt to continue having it in there.

davisagli:
@sneridagh How do IDEs know to use it? I don't see it imported, used, or documented anywhere. That's why we already removed it on the main branch some time ago.

sneridagh:
@davisagli it's convention. vscode knows which syntax to apply (and the linter in stylelint) by looking into the postcss plugins available (I know, it's horrible). I created the package myself back in the day and pushed it to npm. 

Nowadays probably literaly nobody will edit/create .overrides files, but anyways...

davisagli:
@sneridagh But what makes it an available postcss plugin? Just the fact that it's in node_modules with a name starting with `postcss-`?

sneridagh:
@davisagli Yes, as far as I remember...

sneridagh:
@davisagli @wesleybl maybe the current extensions allow the detection using another method... I tried now in main and both the syntax and the stylelint integration in the IDE seems ok. What it does not work is `pnpm stylelint`... Maybe that was the problem. Anyways, nobody is going to fix a .overrides of the semantic UI theme at this point.

I am ok in merging it.

wesleybl:
>  I tried now in main and both the syntax and the stylelint integration in the IDE seems ok

@sneridagh My VS Code doesn't detect less syntax in .overrides files, whether this package is included or not. What needs to be done?

sneridagh:
> > I tried now in main and both the syntax and the stylelint integration in the IDE seems ok
> 
> @sneridagh My VS Code doesn't detect less syntax in .overrides files, whether this package is included or not. What needs to be done?

I think it's:
```
	"files.associations": {
		"*.zcml": "zcml",
		"*.overrides": "less",
		"*.jsx": "javascriptreact",
		"*.js": "javascriptreact",
		"*.pt": "xml",
	},
```

wesleybl:
> 	"files.associations": {
		"*.zcml": "zcml",
		"*.overrides": "less",
		"*.jsx": "javascriptreact",
		"*.js": "javascriptreact",
		"*.pt": "xml",
	},

@sneridagh I thought detection would be automatic if `postcss-overrides` was present. But that's not the case. So I think `postcss-overrides` isn't related to applying less syntax to `.overrides` files.

> But what makes it an available postcss plugin? Just the fact that it's in node_modules with a name starting with postcss-?

@sneridagh @davisagli The configuration should supposedly be here:

https://github.com/plone/volto/blob/main/packages/volto/package.json#L87-L92

But it's not. That is, we're using `postcss-less` for `.overrides` files and it works fine that way.

So it's safe to remove `postcss-overrides`.

davisagli (approved):

PR #8275, opened 2026-05-23
[Seven] Added link view redirects and handled redirects from the backend

pnicolli:
This adds two unrelated but similar features:

- when fetching content items of type Link, a new middleware checks user permissions and redirects to the remoteUrl if the user cannot edit the item
- when fetching content from the backend, `@plone/client` returns the location header from 3xx responses and the app middleware redirects if that happens

### DIFF
--- a/apps/seven/app/middleware.server.test.ts
+++ b/apps/seven/app/middleware.server.test.ts
@@ -1005,8 +1005,8 @@ describe('middleware', () => {
 
       await initializePloneClientContext(request, context);
 
-      try {
-        await fetchPloneContent(
+      await expect(() =>
+        fetchPloneContent(
           {
             request,
             params: {},
@@ -1015,10 +1015,8 @@ describe('middleware', () => {
             unstable_url: new URL(request.url),
           },
           nextMock,
-        );
-      } catch (err: any) {
-        expect(err.init.status).toEqual(301);
-      }
+        ),
+      ).rejects.toMatchObject({ init: { status: 301 } });
     });
   });
 
--- a/apps/seven/app/middleware.server.test.ts
+++ b/apps/seven/app/middleware.server.test.ts
@@ -7,6 +7,7 @@ import {
   fetchPloneContent,
   getAPIResourceWithAuth,
   installServerMiddleware,
+  linkMiddleware,
   ploneClearAuthCookieContext,
   PloneClientMiddleware,
   otherResources,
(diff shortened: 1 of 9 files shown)


apps/seven/app/middleware.server.test.ts
```diff
@@ -957,5 +958,134 @@ describe('middleware', () => {
       expect(context.get(ploneUserContext)).toBeNull();
       expect(context.get(ploneClearAuthCookieContext)).toBe(true);
     });
+
+    it('redirects when getContent throws a 3xx error with location', async () => {
+      const getContentMock = vi
+        .fn()
+        .mockRejectedValue({ status: 301, location: '/new-path' });
```
davisagli, line 965: Does the client already take care of turning an internal hostname from the backend into a frontend URL?

apps/seven/app/middleware.server.test.ts
```diff
@@ -957,5 +958,134 @@ describe('middleware', () => {
...
+            request,
+            params: {},
+            context,
+            unstable_pattern: '/',
+            unstable_url: new URL(request.url),
+          },
+          nextMock,
+        );
+      } catch (err: any) {
+        expect(err.init.status).toEqual(301);
```
davisagli: Be careful asserting about exceptions this way. If the exception doesn't occur, we never run the expect to check if it occurred.

apps/seven/app/middleware.server.ts
```diff
@@ -179,6 +180,11 @@ export const fetchPloneContent: Route.MiddlewareFunction = async (
 
     setPloneContext(content, site, user?.data ?? null);
   } catch (error: any) {
+    if (error.status >= 300 && error.status < 400 && error.location) {
+      return redirect(error.location, {
```
davisagli, line 184: We should probably match the type of redirect (i.e. serve a 302 or 301 depending what came from the backend)

apps/seven/app/middleware.server.ts
```diff
@@ -218,3 +224,19 @@ export const fetchPloneContent: Route.MiddlewareFunction = async (
...
 };
+
+export const linkMiddleware: Route.MiddlewareFunction = async (
+  { context },
+  next,
+) => {
+  const content = context.get(ploneContentContext);
+
+  if (
+    content['@type'] === 'Link' &&
```
davisagli, line 235: I would do this for any content type with a remoteUrl field

sneridagh (approved):
LGTM!!

packages/client/src/api.ts
```diff
@@ -35,20 +36,37 @@ export function getBackendURL(
...
+    debug(error);
+    const status = error.status ?? 0;
+    return Promise.reject({
+      status: error.status ?? error.code,
+      data: error.response?.data,
+      location:
+        status >= 300 && status < 400
+          ? error.response?.headers.location
+              .replaceAll(
+                `${config.apiPath}${config.apiSuffix ?? APISUFFIX}`,
```
davisagli, line 50: Isn't there more needed to turn an internal URL into an external one? It's quite likely the API is being accessed at a different hostname.

apps/seven/app/middleware.server.test.ts
pnicolli: Yes, it is part of the implementation in this PR (api.ts at line 48)

apps/seven/app/middleware.server.ts
pnicolli: Yea I gave this some thought, another idea would be to have a list of content types in the config, I thought that this would be a discussion to be had separately

packages/client/src/api.ts
davisagli: Oh I see, you're replacing the API path and `++api++` suffix with an empty string.

That's going to need some adjustment to support serving seven on a subpath, but ok.

apps/seven/app/middleware.server.ts
davisagli: Yeah, this is fine for now

apps/seven/app/middleware.server.test.ts
pnicolli: This looked to me like the easiest way, since the test does not have other `expect` calls it would fail if this isn't called, but I understand it's a side effect. I'll think of another way of testing this

apps/seven/app/middleware.server.ts
pnicolli: It is done in the next line of code where I passed `status: error.status` to the `redirect` options

packages/client/src/api.ts
pnicolli: Exactly. This currently reflects what is done when creating the backend URL at line 34

apps/seven/app/middleware.server.test.ts
davisagli: https://vitest.dev/api/expect.html#tothrow

pnicolli: With toThrow I couldn't check the status, but I can with toMatchObject. Done.

davisagli (approved):

PR #8279, opened 2026-05-26
Move packages that are used only in testing to devDependencies: react-test-renderer and redux-mock-store. Remove unnecessary dependency: full-icu

wesleybl:


davisagli (commented):
I think we can remove `full-icu` entirely. https://www.npmjs.com/package/full-icu says: "Since Node.js 13, full ICU data has been the default for Node.js. So this module is only helpful for older Node.js versions, or custom builds which specify small-icu."

wesleybl:
> I think we can remove full-icu entirely

@davisagli done.

davisagli (approved):

sneridagh:
@wesleybl @davisagli turns out that no, we cannot do this, because it had a reason. Really, this nonesense micromanagement led to a 1h me puzzled about why this was happening:

https://github.com/kitconcept/volto-light-theme/actions/runs/26586573125/job/78335255963?pr=873

We need to retract this one and make a --prod when building things to have this catched up.

sneridagh:
@wesleybl @davisagli 

reason:
https://github.com/plone/volto/blob/main/packages/volto-slate/src/widgets/HtmlSlateWidget.jsx#L7

It's a hacky way of accomplish it, but it did the trick all these years. Thus, the actual code needs it.

sneridagh:
Weird, `@plone/volto-slate` declares it as a direct dep, but the last `pnpm install --prod` obliterates it. Investigating.

Fosten:
Thanks for looking into this @sneridagh 
I also noticed the same issue and was on the same troubleshooting mission today
``Error: Cannot find module 'redux-mock-store’``
Wondered what was happening since @plone/volto-slate declares it as a direct dependency.

sneridagh:
Fixed in: https://github.com/plone/volto/pull/8299

sneridagh:
This was the official machine verdict about the reasoning behind:

```
• It fails because the runtime require no longer belongs to @plone/volto-slate after the build.


  Because redux-mock-store is present in @plone/volto during the initial full install as a devDependency, webpack-node-externals externalizes it instead of bundling it. So the built server keeps a bare runtime call like:

  require("redux-mock-store")

  Then pnpm i --prod prunes @plone/volto devDependencies. At runtime, Node resolves that require from:

  packages/volto/build/server.js
  packages/volto/node_modules
  ...

  It does not search sideways into:

  packages/volto-slate/node_modules/redux-mock-store

  So @plone/volto-slate declaring it is technically correct for the package, but ineffective for this bundled server artifact because the import has been externalized into @plone/volto’s runtime context.

  The clean fix is to remove redux-mock-store from runtime code in HtmlSlateWidget; it is a test utility being used in production code. The pragmatic workaround is to keep redux-mock-store as an @plone/volto production dependency, or adjust the server
  bundling so it does not externalize that import.
  ```

davisagli:
@sneridagh Sorry about that, and thanks for taking care of it.

PR #8296, opened 2026-05-28
Add ReactAriaRouterProvider to enable client-side navigation support for React Aria Link components

wesleybl:
Backport of #8159 to Volto 18.


davisagli (commented):
Volto 18 does not depend on `@plone/components`, so I would rather not add the dependency on react-aria-components here.

sneridagh:
@wesleybl @davisagli We just can't. `@plone/components` (and thus, RAC) is not officially core in 18. Only from 19 onwards.

PR #8297, opened 2026-05-28
increase icon size to comply with WCAG 2.2

polyester:
According to WCAG 2.2 criterion 2.5.8, a target for navigation should be 24 css pixels minimum. 

Screenshot 2026-05-29 at 9 58 52 AM


packages/volto/news/8297.bugfix
```diff
@@ -0,0 +1 @@
+increase home icon size to comply with WCAG 2.2 accessibility @polyester
```
davisagli:
```suggestion
Increase home icon size to comply with WCAG 2.2 accessibility. @polyester
```

Wagner3UB (approved):

PR #8299, opened 2026-05-28
Fixed redux-mock-store in production

sneridagh:


wesleybl:
Strange. `redux-mock-store` was still a dependency of volto-slate. And the acceptance tests passed. The good thing is that the "hack" was removed.

PR #8308, opened 2026-06-02
[Seven] Add `renderEmptyState` to `Table` component

jnptk:
https://react-aria.adobe.com/Table#empty-state

### DIFF
--- a/packages/components/src/components/Table/Table.tsx
+++ b/packages/components/src/components/Table/Table.tsx
@@ -1,6 +1,7 @@
 import React, { type ReactNode, type ComponentProps } from 'react';
 import {
   type TableProps as RACTableProps,
+  type TableBodyRenderProps,
   ResizableTableContainer,
   ColumnResizer,
   Table as RACTable,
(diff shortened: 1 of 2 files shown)


packages/components/src/components/Table/Table.tsx
```diff
@@ -29,6 +30,7 @@ interface TableProps<C, R> extends RACTableProps {
   rows?: R[];
   resizableColumns?: boolean;
   dragColumnHeader?: ComponentProps<typeof TableHeader>['dragColumnHeader'];
+  renderEmptyState?: (props: TableBodyRenderProps) => ReactNode;
```
pnicolli: I would try to type it like this, I find it to be the simplest and more future proof way whenever possible:


```suggestion
  renderEmptyState?: ComponentsProps<typeof TableBody>['renderEmptyState'];
```

exactly like the `dragColumnHeader` type in the previous line

pnicolli (changes requested):

pnicolli (approved):

PR #8247, opened 2026-05-19
Fix issue where label for the checkbox doesn't show.

frapell:
@sneridagh Fixed the issue of the label not showing in the checkbox, and also fixed it to show the proper style (I believe it is `denim` correct?)

PR #8237, opened 2026-05-18
Create a wrapper for Boolean widget for showing labels correctly.

iRohitSingh:
ScreenShot: 

Screenshot 2026-05-18 at 8 13 45 PM

PR #8309, opened 2026-06-02
BooleanWidget and clarify widgets definitions in documentation

sneridagh:
Supersedes: #8237 and #8247


sneridagh:
@pnicolli damn, I kept thinking about it and went through all the current widgets that we have, and it's harder than I thought... it definitelly needs more love. Only alone the use case for public forms is a thing we have to address.

docs/conceptual-guides/form-fields-controls-and-widgets.md
```diff
@@ -0,0 +1,187 @@
...
+    "property=og:title": "Form fields, controls, and widgets"
+    "keywords": "Seven, forms, fields, controls, widgets, @tanstack/form, @plone/cmsui"
+---
+
+(form-fields-controls-and-widgets-label)=
+
+# Form fields, controls, and widgets
+
+Seven forms are schema-driven.
+A content schema describes metadata fields, and the CMS UI turns those fields into interactive form elements.
```
pnicolli: I am wondering if "CMS UI" (or CMSUI, or anything similar) should be a glossary term in the docs

docs/conceptual-guides/form-fields-controls-and-widgets.md
```diff
@@ -0,0 +1,187 @@
...
+This process is intentionally split into several concepts: fields, controls, widgets, and adapters.
+The distinction matters because not every visual input component can be registered directly as a CMS widget.
+
+## Field
+
+A field is the form-level representation of one piece of content data.
+It has a name, a current value, validation state, and schema metadata.
+In a content form, a field usually comes from a Plone schema property.
+For example, `title`, `description`, `effective`, and `is_folderish` are fields when the form renders them as editable metadata.
```
pnicolli: I'm not sure we should include `is_folderish` in this example

docs/conceptual-guides/form-fields-controls-and-widgets.md
```diff
@@ -0,0 +1,187 @@
...
+An image widget may combine an object browser, upload button, link input, preview, clear action, and validation message.
+Both are widgets if they satisfy the same field contract from the form generator.
+
+## Adapter
+
+An adapter is a wrapper that turns a control into a widget.
+It maps the form field contract to the control contract.
+This is the right place to translate names, normalize values, and convert validation metadata into the shape expected by the control.
+
+For example, a boolean widget can adapt a checkbox control by mapping `value` to `isSelected` and by mapping `onChange(value)` to the checkbox's selected state callback.
```
pnicolli, line 70: I don't understand how this is different from the Widget definition.

docs/conceptual-guides/form-fields-controls-and-widgets.md
```diff
@@ -0,0 +1,187 @@
...
+```ts
+import type { FormWidget, FormWidgetProps } from '@plone/types';
+```
+
+`FormWidgetProps` describes the props that a widget receives from the form generator.
+Use it when you implement a widget or adapter.
+
+```tsx
+function BooleanWidget(props: FormWidgetProps<boolean>) {
+  return null;
```
pnicolli: If what you wanted here is to avoid focusing on what is the code of the function, I would just write the function signature, like 

```tsx
function BooleanWidget(props: FormWidgetProps<boolean>)
```

or something like this

```tsx
function BooleanWidget(props: FormWidgetProps<boolean>) {
  // ...
}
```

with `return null` I was like "wth" for a moment and I can only assume I won't be the only one

docs/conceptual-guides/form-fields-controls-and-widgets.md
```diff
@@ -0,0 +1,187 @@
...
+
+Raw controls often use APIs that are correct for their interaction pattern but incomplete for CMS forms.
+A checkbox control may use `isSelected` instead of `value`.
+A date picker may emit a date object while the content API expects an ISO string.
+A select control may know about option labels but not about Plone vocabularies.
+An object browser may need content context, selectable type constraints, and a normalized relation value.
+
+Registering those controls directly makes the form generator depend on many incompatible prop conventions.
+That weakens the registry because the registry can no longer promise what a widget receives or returns.
+Adapters keep the boundary explicit.
```
pnicolli, line 147: Why? But my question might come from the fact that I didn't understand adapters, see my other question above

docs/conceptual-guides/form-fields-controls-and-widgets.md
```diff
@@ -0,0 +1,187 @@
...
+`Adapter`
+:   A wrapper that maps between the two.
+
+A component can be both a control and a widget only when its prop API already matches the form field contract.
+Text inputs often come close to this shape.
+Checkboxes, date pickers, object browsers, image pickers, and rich editors usually need adapters.
+
+## Design implications
+
+The form generator should stay boring.
```
pnicolli: "Boring" is a weird definition. What does "boring" mean to the readers? Simple?

pnicolli (changes requested):
Yeah it needs more love, but I think starting with a good documentation of what we want to achieve is a good starting point to actually achieve consistent behavior in widgets and better code quality. It might be challenging and we might end up adapting these docs to actually match what we need, but I think it's a good starting point.

docs/conceptual-guides/form-fields-controls-and-widgets.md
stevepiercy: It is, but y'all need to decided whether it has a space in it, and be consistent. I think the last time we had the discussion, no space was settled. https://6.docs.plone.org/glossary.html#term-CMSUI

pnicolli: Ah, it's already there, my bad sorry. Then definitely apply the term here properly without space.

stevepiercy:
```suggestion
A content schema describes metadata fields, and the {term}`CMSUI` turns those fields into interactive form elements.
```

docs/conceptual-guides/form-fields-controls-and-widgets.md
```diff
@@ -0,0 +1,187 @@
...
+---
+
+(form-fields-controls-and-widgets-label)=
+
+# Form fields, controls, and widgets
+
+Seven forms are schema-driven.
+A content schema describes metadata fields, and the CMS UI turns those fields into interactive form elements.
+This process is intentionally split into several concepts: fields, controls, widgets, and adapters.
+The distinction matters because not every visual input component can be registered directly as a CMS widget.
```
stevepiercy: CMS or CMSUI?

docs/conceptual-guides/form-fields-controls-and-widgets.md
```diff
@@ -0,0 +1,187 @@
...
+
+A field is the form-level representation of one piece of content data.
+It has a name, a current value, validation state, and schema metadata.
+In a content form, a field usually comes from a Plone schema property.
+For example, `title`, `description`, `effective`, and `is_folderish` are fields when the form renders them as editable metadata.
+
+A field belongs to the form system.
+It participates in form state, validation, submission, and synchronization with the content object.
+In Seven, `@tanstack/react-form` owns this part of the model.
+The CMS UI field component connects TanStack Form state to the widget registry and to any additional state that the editing UI needs.
```
stevepiercy:
```suggestion
The CMSUI field component connects TanStack Form state to the widget registry and to any additional state that the editing UI needs.
```

docs/conceptual-guides/form-fields-controls-and-widgets.md
```diff
@@ -0,0 +1,187 @@
...
+Adapters keep the boundary explicit.
+
+## Field resolution
+
+The widget registry answers a different question than the form state system.
+The form state system asks, "What is the value and validation state of this field?"
+The widget registry asks, "Which widget should render this field?"
+
+Seven can resolve a widget from several schema hints.
+The field id, explicit widget name, choices, vocabulary, factory, and type can all influence the result.
```
stevepiercy: Microsoft Writing Style Guide uses ID in narrative text.

```suggestion
The field ID, explicit widget name, choices, vocabulary, factory, and type can all influence the result.
```

docs/conceptual-guides/form-fields-controls-and-widgets.md
stevepiercy:
```suggestion
The form generator should be basic.
```

stevepiercy (changes requested):

docs/conceptual-guides/form-fields-controls-and-widgets.md
sneridagh: I rephrased it.

docs/conceptual-guides/form-fields-controls-and-widgets.md
sneridagh: I thought that would make sense, but if we standardize and further fully specify what is a widget-able component and what is not, we might get away only with one. It also came from the fact of having "half way threre" components in `@plone/components` that might need an adapter... but it's true that we could make them behave fully with our canonical widget definition specs, and that's it.

sneridagh:
Moving it to https://github.com/plone/aurora/pull/97

PR #8315, opened 2026-06-04
fix(a11y): hide clear button on required select and array widgets

### LINKED ISSUE #5140: a11y - site setup - Date and Time - cancel/delete buttons input fields - cms ui
**Describe the bug**
- None of the 'x' buttons work to cancel an item when using the keyboard. Additionally, not all of them are reachable using the keyboard.
- Along with this issue, it's necessary to pay attention to other accessibility points, such as set the correct labels.

**To Reproduce**

1. Go to "site setup" -> "Date and Time"
2. With a screen reader + keyboard, go to any input and try to add and delete the items

→ to access the "Available timezones" item and the close button for a single item:
1. go inside the input
2. with one or more items previously set, press the "arrow left" on the keyboard. This button doesn't work with the keyboard.

→ the small red button on the right (clear field), inside all the inputs, it is not reachable by keyboard

**Expected behavior**
- The item close button must delete the selection of the current item.
- The small red button on the right needs to work exclusively for 'Available timezones'. It does not serve a proper purpose for the first and third input. Those are required fields.

**Screenshots**
If applic
Schermata 2023-08-28 alle 15 54 50

Volto Version 17.0.0-alpha.19
Plone Version 6.0.4

Wagner3UB:
Partially fixes #5140

## What

Improves keyboard accessibility for the clear button and hides it on required fields in `SelectWidget` and `ArrayWidget`.

## Changes

- **`SelectWidget` / `ArrayWidget`**: `isClearable` now respects the `required` prop — the clear button is hidden on required fields.
- **`SelectStyling` — `ClearIndicator`**: overrides react-select's default `aria-hidden="true"` on the clear button, adding `role="button"`, `tabIndex=0`, and an `onKeyDown` handler so the button is reachable via Tab and activatable with Enter or Space.

## Behavior after fix

- Required fields no longer show the clear button (e.g. "Portal default timezone", "First weekday" in the Date and Time control panel).
- Non-required fields: the clear button is now reachable via Tab and can be activated with Enter or Space.


packages/volto/news/8315.bugfix
```diff
@@ -0,0 +1 @@
+Hide the clear button in `SelectWidget` and `ArrayWidget` when the field is marked as required, and make it keyboard accessible (Tab, Enter, Space) on non-required fields. @Wagner3UB
```
stevepiercy: A sprinkling of MyST fair dust.

```suggestion
Hide the clear button in `SelectWidget` and `ArrayWidget` when the field is marked as required, and make it keyboard accessible ({kbd}`Tab`, {kbd}`Enter`, {kbd}`Space`) on non-required fields. @Wagner3UB
```

packages/volto/src/components/manage/Widgets/SelectStyling.jsx
```diff
@@ -119,7 +119,21 @@ export const DropdownIndicator = injectLazyLibs('reactSelect')((props) => {
...
   const { ClearIndicator } = props.reactSelect.components;
   return (
-    <ClearIndicator {...props}>
+    <ClearIndicator
+      {...props}
+      innerProps={{
+        ...props.innerProps,
+        'aria-hidden': false,
+        role: 'button',
+        tabIndex: 0,
```
stevepiercy, line 142: Just to verify the desired behavior, if there are multiple controls each with a clear indicator of `tabindex=0`, then first the controls would be tab-able in the order they appear, followed by the clear buttons in the order they appear.

If it were me, instead I'd like to have the clear button tab-able immediately after its control.

stevepiercy (changes requested):
A minor tweak and a question.

packages/volto/src/components/manage/Widgets/SelectStyling.jsx
wesleybl: That makes sense.

packages/volto/src/components/manage/Widgets/SelectStyling.jsx
```diff
@@ -119,7 +119,21 @@ export const DropdownIndicator = injectLazyLibs('reactSelect')((props) => {
 export const ClearIndicator = injectLazyLibs('reactSelect')((props) => {
   const { ClearIndicator } = props.reactSelect.components;
   return (
-    <ClearIndicator {...props}>
+    <ClearIndicator
+      {...props}
+      innerProps={{
+        ...props.innerProps,
+        'aria-hidden': false,
```
wesleybl, line 137: We need an `aria-label` stating that it is a cleaning action.

packages/volto/src/components/manage/Widgets/SelectStyling.jsx
```diff
@@ -119,7 +119,21 @@ export const DropdownIndicator = injectLazyLibs('reactSelect')((props) => {
...
+        role: 'button',
+        tabIndex: 0,
+        onKeyDown: (e) => {
+          if (e.key === 'Enter' || e.key === ' ') {
```
wesleybl: It would be good to also add the condition `e.key === 'Spacebar'`

packages/volto/src/components/manage/Widgets/SelectStyling.jsx
```diff
@@ -119,7 +119,21 @@ export const DropdownIndicator = injectLazyLibs('reactSelect')((props) => {
...
+            e.preventDefault();
```
wesleybl, line 145: Add `e.stopPropagation()` as well.

wesleybl (changes requested):

packages/volto/src/components/manage/Widgets/SelectStyling.jsx
Wagner3UB: The tab order is already correct with the current implementation. Since the `ClearIndicator` is rendered **inside** the same react-select container as the input, the natural DOM order ensures that each clear button immediately follows its own control — not after all controls:

```
Field A input → Field A clear → Field B input → Field B clear
```

I recorded a short video demonstrating this behavior with two multi-selection fields in sequence. *(attach video)*

https://github.com/user-attachments/assets/d0c05d57-ea5a-47f1-b7b0-799f5a3d7441

Tell me if this is what you thought, please
@stevepiercy @wesleybl

stevepiercy: Upload didn't come through. 😢 

But I trust that this implementation has the behavior that I mentioned.

Wagner3UB: @stevepiercy done, idk what happened

stevepiercy: OK, I see now. The punctuation and extra spaces is strange in some of the messages. See at 0:10 and 0:15.

- `Select is focused ,type`
- `beginning of text., search`

There's also the use of "press left" and "down", instead of "* arrow". But that could be outside the scope of this PR.

stevepiercy: As for the `tabindex`, that looks good.

Wagner3UB: > OK, I see now. The punctuation and extra spaces is strange in some of the messages. See at 0:10 and 0:15.
> 
> * `Select is focused ,type`
> * `beginning of text., search`
> 
> There's also the use of "press left" and "down", instead of "* arrow". But that could be outside the scope of this PR.

@stevepiercy Those lines are from react-select, not ours. Unfortunately, they chose to use it without "arrow" when keys are pressed. The lack of organization is one of the most common mistakes from old libraries. Not much we can do about it.

stevepiercy: Is it no longer maintained or is it pinned to an old version? If neither, we could at least open an issue, or create a pull request.

https://github.com/search?q=repo%3AJedWatson%2Freact-select%20%22press%20left%22&type=code

wesleybl (approved):

packages/volto/src/components/manage/Widgets/SelectStyling.jsx
Wagner3UB: @stevepiercy It's still maintained, not actively, the latest version is v5.10.2 (last commit July 2025). Volto is pinned to v4.3.1, a major version behind, so upgrading would be a breaking change. I believe no one ever considered this relevant before. We could fix it directly in Volto by overriding the messages via react-select's `ariaLiveMessages` prop, but that would be out of scope for this PR.

stevepiercy: @Wagner3UB thanks for the explanation. I'm not surprised it wasn't considered relevant before, because a11y, like docs, is not considered by developers as much as we would like.

I'll leave it to you to make the decision about whether to create an issue, since I had no idea that it was possible. It could establish a good pattern as well. What do you think?

Wagner3UB: @stevepiercy https://github.com/plone/volto/issues/8335

Wagner3UB:
@stevepiercy I need your ok here

packages/volto/src/components/manage/Widgets/SelectStyling.jsx
stevepiercy: Excellent!

stevepiercy (approved):

PR #8314, opened 2026-06-04
fix(a11y): announce field name and current value in select widgets via screen reader

Wagner3UB:
Fixes #5139

## What

Improves screen reader accessibility for `SelectWidget` and `ArrayWidget` (react-select based fields).

## Changes

- **`SelectWidget`**: replaced `aria-labelledby` with a dynamic `aria-label` that includes the field title and the current selected value(s). Falls back to "No value" (translated) when the field is empty. Separated `inputId` from `id` so the HTML label correctly points to the react-select inner input.
- **`ArrayWidget`**: same `aria-label` pattern applied for multi-select fields.

## Behavior after fix

- Screen reader announces field name + current value(s) on focus (e.g. "Available timezones: Europe/Berlin, UTC")
- Empty fields announce "Available timezones: No value"

PR #8316, opened 2026-06-04
fix(a11y): improve screen reader announcements for multi-value chips in select widgets

### LINKED ISSUE #5140: a11y - site setup - Date and Time - cancel/delete buttons input fields - cms ui

Wagner3UB:
Partially fixes #5140

## What

Improves screen reader accessibility for multi-value chips in `ArrayWidget` and `SelectWidget`.

## Changes

- **`SelectStyling` — `SortableMultiValue`**: added `aria-label` with field name and chip value, and `aria-describedby` pointing to a hidden element with keyboard instructions — overriding the dnd-kit drag instructions that were being announced instead.
- **`SelectWidget` / `ArrayWidget`**: suppressed the misleading "press left to focus selected values" announcement via `ariaLiveMessages`.

## Behavior after fix

- Screen reader announces chip value on focus (e.g. "Available timezones: Africa/Asmera")
- Screen reader announces keyboard instructions via description: "Press Space to add items. Press Delete or Backspace to remove."
- The misleading "press left/right to navigate" message is no longer announced


packages/volto/src/components/manage/Widgets/ArrayWidget.jsx
```diff
@@ -402,6 +402,14 @@ class ArrayWidget extends Component {
               }
               isClearable
               isMulti
+              ariaLiveMessages={{
+                onFocus: (props) => {
+                  if (props.context === 'input') {
+                    return `${props['aria-label'] || 'Select'} is focused${props.isSearchable ? ', type to refine list' : ''}, press Down to open the menu.`;
```
stevepiercy: I think these are better as distinct sentences as commands than a run-on comma-separated list.

Are they translatable?

```suggestion
                    return `${props['aria-label'] || 'Select'} is focused${props.isSearchable ? '. Type to refine list' : ''}. Press Down to open the menu.`;
```

Out of curiosity, is "Down" preferred over "Down-arrow" in a11y? I typically see "Down-arrow" in narrative text documentation.

packages/volto/src/components/manage/Widgets/SelectStyling.jsx
```diff
@@ -10,6 +11,16 @@ import checkSVG from '@plone/volto/icons/check.svg';
 import checkBlankSVG from '@plone/volto/icons/check-blank.svg';
 import clearSVG from '@plone/volto/icons/clear.svg';
 
+const messages = defineMessages({
+  chip_instructions: {
+    id: 'Press Space to add items. Press Delete or Backspace to remove.',
```
stevepiercy, line 16: Yes! Like this! Separate actions or commands as separate sentences.

packages/volto/src/components/manage/Widgets/SelectWidget.jsx
```diff
@@ -305,6 +305,14 @@ class SelectWidget extends Component {
             );
           }}
           isClearable={this.props.isClearable}
+          ariaLiveMessages={{
+            onFocus: (props) => {
+              if (props.context === 'input') {
+                return `${props['aria-label'] || 'Select'} is focused${props.isSearchable ? ', type to refine list' : ''}, press Down to open the menu.`;
```
stevepiercy: Same as above.

```suggestion
                return `${props['aria-label'] || 'Select'} is focused${props.isSearchable ? '. Type to refine list' : ''}. Press Down to open the menu.`;
```

stevepiercy (changes requested):
Please let me know what you think. Thank you!

packages/volto/src/components/manage/Widgets/ArrayWidget.jsx
Wagner3UB: @stevepiercy 

> I think these are better as distinct sentences as commands than a run-on comma-separated list.

You are right, I'll update them.

> Are they translatable?

They should be. I missed them in my review, they slipped through hardcoded and still need to be translated. I'll fix that.

> Out of curiosity, is "Down" preferred over "Down Arrow" in a11y? I typically see "Down Arrow" in narrative text documentation.

I had the same doubt about all the key names (Down, Left, Backspace, Delete... etc.), whether it was better to keep them in English as a kind of "universal language" for PC interfaces, or translate them into each possible language, for example.

The answer will always be: it depends. It depends on the scope, the complexity of the topic, the target audience, and whether the language has an established dictionary translation for the key name.

There is no formal standard defining key names in `aria-live` messages. The two most relevant (for us, in this scope) references diverge:

- The **WAI-ARIA APG (W3C)** consistently uses **"Down Arrow"** (two words, both capitalized) in keyboard interaction documentation.
- **react-select** itself uses the shorter **"Down"** in its default `ariaLiveMessages`.

The strongest argument for **"Down Arrow"** is consistency with screen readers themselves, NVDA and VoiceOver both announce the key as "Down Arrow" when pressed, so matching that phrasing reduces cognitive load for the user.

What do you think? We can follow "Down" as react-select does, or follow the "bible", the APG, with "Down Arrow". We could also add a note somewhere to document this kind of decision.

stevepiercy: I think "Down Arrow" makes the most sense. "Down" alone would sound like "Press Down", and I'm left wondering, "Press down on _what_?"

Looking at a small sample of my keyboards, there are no text labels for the arrow keys, just indicators pointing in four different directions.

stevepiercy (approved):
This is much more understandable. Thank you!

Let's get a second technical review.

packages/volto/src/components/manage/Widgets/ArrayWidget.jsx
Wagner3UB: > I think "Down Arrow" makes the most sense. "Down" alone would sound like "Press Down", and I'm left wondering, "Press down on _what_?"
> 
> Looking at a small sample of my keyboards, there are no text labels for the arrow keys, just indicators pointing in four different directions.

Sorry @stevepiercy, not the physical keyboard, but the action, you can see in the eventListeners

Screenshot 2026-06-05 at 16 39 08</br>

Screenshot 2026-06-05 at 16 42 35

stevepiercy: @Wagner3UB thanks for clarifying. "Arrow *" is good.

packages/volto/src/components/manage/Widgets/SelectStyling.jsx
```diff
@@ -46,7 +58,12 @@ export const SortableMultiValue = injectLazyLibs([
       }}
       {...attributes}
       {...listeners}
+      aria-label={`${props.selectProps.fieldTitle || ''}: ${props.data.label}`}
+      aria-describedby={CHIP_INSTRUCTIONS_ID}
     >
+      <span id={CHIP_INSTRUCTIONS_ID} style={{ display: 'none' }}>
```
wesleybl: If there is more than one widget on the page, the `id` will be repeated.

wesleybl: I think the "display none" option will cause the screen reader to ignore the content.

packages/volto/src/components/manage/Widgets/SelectStyling.jsx
```diff
@@ -46,7 +58,12 @@ export const SortableMultiValue = injectLazyLibs([
       }}
       {...attributes}
       {...listeners}
+      aria-label={`${props.selectProps.fieldTitle || ''}: ${props.data.label}`}
```
wesleybl, line 66: Where does `fieldTitle` come from? I don't think anyone passes that around.

wesleybl (changes requested):

packages/volto/src/components/manage/Widgets/SelectStyling.jsx
wesleybl: https://github.com/plone/volto/pull/8314 does that, right?

wesleybl: Yes, I can see that now after merging #8314 and updating this branch.

PR #8319, opened 2026-06-08
fix(a11y): use visually-hidden on empty aria-live region in FormFieldWrapper

### LINKED ISSUE #8317: Regression in widgets (SizeWidget)
I found a regression regarding https://github.com/plone/volto/pull/8033. (See https://volto.demo.plone.org/edit as long as is there, if not, just add an image, check the size widget).

See the size widget, it's not aligned anymore:

image

Because the markup changed with the change of this PR.

https://github.com/plone/volto/blob/d8cca3b242842a2322d1dd71b52b1d4e22387ff5/packages/volto/src/components/manage/Widgets/FormFieldWrapper.jsx#L64-L70

image

The grid now have another unaccounted element. No clue if that happens elsewhere...

Please, we need to have more care when we change markup, specially in grids/flex elements and check all the affected components. 

@Wagner3UB can we add some CSS to ignore that addition? Which is the best thing to proceed?

Wagner3UB:
Fix [#8317](https://github.com/plone/volto/issues/8317)

Regression introduced by #8033: the `div[aria-live]` added to `FormFieldWrapper` was acting as an extra flex/grid item inside the widget column, breaking the layout of widgets like `SizeWidget` (S/M/L buttons misaligned).

## Changes

- `FormFieldWrapper.jsx`: apply `visually-hidden` class to the `aria-live` container when there are no errors. This removes it from the flex/grid flow via `position: absolute` while keeping it in the accessibility tree so the live region continues to work correctly for screen readers.
- Snapshots updated to reflect the new class on the error container.

### DIFF
--- a/packages/volto/news/8319.bugfix
+++ b/packages/volto/news/8319.bugfix
@@ -1,0 +1,1 @@
+Fix layout regression in `FormFieldWrapper` where the empty `aria-live` container was acting as an extra flex item, breaking widget layouts like `SizeWidget`. @Wagner3UB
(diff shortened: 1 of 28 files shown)


sneridagh (approved):

sneridagh:
@Wagner3UB thanks for fixing this!

PR #8320, opened 2026-06-09
fix(a11y): FileWidget and RegistryImageWidget a11y improvements - backport

Wagner3UB:
Backport of https://github.com/plone/volto/pull/7982 from Volto 19 to Volto 18, with an additional fix for the screen reader announcement when a file is already present.

## Changes

- **FileWidget**: added `role="button"`, `aria-labelledby`, and `aria-describedby` to the dropzone container for proper keyboard and screen reader support
- **FileWidget**: label is now `aria-hidden` with `stopPropagation` on click to prevent double file dialog opening
- **FileWidget**: hidden status span announces upload instructions, required state, and current filename to screen readers
- **FileWidget**: distinguishes between "open file browser" and "replace existing file" in the screen reader announcement depending on whether a file is already uploaded (improvement over the original v19 PR)
- **FileWidget**: `aria-required` and `aria-invalid` added to the file input
- **FileWidget**: download link now has a descriptive `aria-label` with the filename
- **FileWidget**: new `fieldSet` prop to correctly build the `aria-labelledby` reference
- **RegistryImageWidget**: updated `editFile` message to match new wording
- **main.less**: added `padding-bottom: 1rem` to `.file-widget-dropzone`
- **locales**: translated all new/updated strings across 65 languages


packages/volto/news/8320.bugfix
```diff
@@ -0,0 +1 @@
+Improved accessibility of FileWidget and RegistryImageWidget: added proper ARIA roles, labels, and screen reader announcements for the dropzone area, required state, and file download link. @Wagner3UB
```
stevepiercy:
```suggestion
Improved accessibility of `FileWidget` and `RegistryImageWidget`: added proper ARIA roles, labels, and screen reader announcements for the dropzone area, required state, and file download link. @Wagner3UB
```

packages/volto/locales/volto.pot
```diff
@@ -2843,11 +2862,6 @@ msgstr ""
 msgid "Preferences"
 msgstr ""
 
-#. Default: "Press Enter to browse files from your computer."
-#: components/manage/Widgets/FileWidget
-msgid "Press Enter to browse files from your computer."
-msgstr ""
-
```
stevepiercy: IIRC, changing or removing the `msgid` is a breaking change.

packages/volto/src/components/manage/Widgets/FileWidget.jsx
```diff
@@ -60,8 +60,21 @@ const messages = defineMessages({
     defaultMessage: 'File is not of the accepted type {accept}',
   },
   dragAndDropActionA11y: {
-    id: 'Press Enter to browse files from your computer.',
-    defaultMessage: 'Press Enter to browse files from your computer.',
+    id: 'File upload area. Press Enter to open the file browser',
+    defaultMessage: 'File upload area. Press Enter to open the file browser',
```
stevepiercy: IIRC, changing or replacing the `msgid` is a breaking change.

stevepiercy (changes requested):
A sprinkling of MyST markup, and a concern to address.

This needs a technical review as well.

packages/volto/locales/volto.pot
Wagner3UB: Not good! Or... not ideal! I will think on a solution! Thanks @stevepiercy

Wagner3UB: @stevepiercy Add new ones is a problem too?

stevepiercy: Adding new ones is OK. The core issue is that add-ons might depend on a translation's `msgid`, and if you delete or change it, then that could break their add-on.

stevepiercy (approved):
Yay! `msgid`s preserved. LGTM. Thank you!

packages/volto/src/components/manage/Widgets/FileWidget.jsx
```diff
@@ -36,11 +36,13 @@ const messages = defineMessages({
...
     id: 'Drop file here to replace the existing file',
-    defaultMessage: 'Drop file here to replace the existing file',
+    defaultMessage:
+      'File upload area. Press Enter or click to replace the existing file',
   },
   fileDrag: {
     id: 'Drop file here to upload a new file',
-    defaultMessage: 'Drop file here to upload a new file',
+    defaultMessage:
+      'File upload area. Press Enter or click to open the file browser',
```
wesleybl: Why is the message different from the main branch? If so, it would be good to change it in the main branch as well.

wesleybl (changes requested):
I noticed that the text has changed since the PR on the main branch. I think that, for consistency and to facilitate future backports, the text should be the same. If that's the case, we need to update the text on the main branch.

Wagner3UB:
> I noticed that the text has changed since the PR on the main branch. I think that, for consistency and to facilitate future backports, the text should be the same. If that's the case, we need to update the text on the main branch.

@wesleybl You are right. But the problem isn't here, its in the 19 version: https://github.com/plone/volto/pull/8334

stevepiercy (approved):

Wagner3UB:
> I noticed that the text has changed since the PR on the main branch. I think that, for consistency and to facilitate future backports, the text should be the same. If that's the case, we need to update the text on the main branch.

@wesleybl Here we are good to go. I did a fix on the main PR to be like this one. We can continue.

wesleybl (approved):

PR #7981, opened 2026-03-09
aria properties added to formfieldwrapper - a11y

Wagner3UB:
FormFieldWrapper

This component renders most of the text/text-like fields.
When the component is of type Input and has the required prop set to true, the corresponding ARIA attributes are added to the element.

Issue related: https://github.com/plone/volto/issues/5619

PR #8321, opened 2026-06-09
fix(a11y): FormFieldWrapper aria-required and aria-invalid - backport

Wagner3UB:
Backport of https://github.com/plone/volto/pull/7981 from Volto 19 to Volto 18, with an additional fix to omit `aria-invalid` when false.

## Changes

- **FormFieldWrapper**: when a child is a Semantic UI `Input` and the field is `required`, `aria-required="true"` is automatically injected via `cloneElement`
- **FormFieldWrapper**: `aria-invalid="true"` is added only when the field has errors — omitted entirely when false (improvement over the original v19 PR)
- **TextWidget.test**: added two new tests covering `aria-required` and `aria-invalid` attribute injection

## Unrelated fix included

The `blocks-table.js` Cypress test was failing in CI (headless mode) with all 3 retry attempts. Root cause: the test was clicking `body` at coordinates `(0, 0)` to reset Slate focus before typing in cells, but this caused a race condition in headless Electron where the first cell click only activated the block without setting Slate's internal selection. Additionally, the outline check was clicking the same cell (`.first()`) that was the first typing target, causing Slate to receive a second click on an already-focused element and lose its selection.

Fix: move the outline check to `.eq(1)` (second cell) so the first typing target always receives a fresh click, and add `.should('be.focused')` before each `.type()` as a synchronization point.

### DIFF
--- a/packages/volto/news/8321.bugfix
+++ b/packages/volto/news/8321.bugfix
@@ -1,0 +1,1 @@
+Add `aria-required` and `aria-invalid` attributes to Input fields inside FormFieldWrapper when the field is required or has errors, respectively. @Wagner3UB
--- a/packages/volto/news/8321.bugfix
+++ b/packages/volto/news/8321.bugfix
@@ -1,1 +1,1 @@
-Add `aria-required` and `aria-invalid` attributes to `Input` fields inside `FormFieldWrapper` when the field is required or has errors, respectively. @Wagner3UB
+Add `aria-required` and `aria-invalid` attributes to `Input` fields inside `FormFieldWrapper` when the field is required or has errors, respectively. Also fixes the `blocks-table` Cypress test which was failing in headless/CI mode due to a fragile focus reset strategy. @Wagner3UB
--- a/packages/volto/news/8321.bugfix
+++ b/packages/volto/news/8321.bugfix
@@ -1,1 +1,1 @@
-Add `aria-required` and `aria-invalid` attributes to Input fields inside FormFieldWrapper when the field is required or has errors, respectively. @Wagner3UB
+Add `aria-required` and `aria-invalid` attributes to `Input` fields inside `FormFieldWrapper` when the field is required or has errors, respectively. @Wagner3UB
--- a/packages/volto/src/components/manage/Widgets/FormFieldWrapper.jsx
+++ b/packages/volto/src/components/manage/Widgets/FormFieldWrapper.jsx
@@ -112,7 +112,7 @@ class FormFieldWrapper extends Component {
           if (isValidElement(child) && required && child.type === Input) {
             return cloneElement(child, {
               'aria-required': true,
-              'aria-invalid': !!(error && error.length > 0),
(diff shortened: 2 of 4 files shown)


packages/volto/news/8321.bugfix
```diff
@@ -0,0 +1 @@
+Add `aria-required` and `aria-invalid` attributes to Input fields inside FormFieldWrapper when the field is required or has errors, respectively. @Wagner3UB
```
stevepiercy:
```suggestion
Add `aria-required` and `aria-invalid` attributes to `Input` fields inside `FormFieldWrapper` when the field is required or has errors, respectively. @Wagner3UB
```

stevepiercy (changes requested):
A sprinkling of MyST on the news fragment.

Needs a technical review.

Thank you!

packages/volto/src/components/manage/Widgets/FormFieldWrapper.jsx
```diff
@@ -103,7 +108,15 @@ class FormFieldWrapper extends Component {
 
     const wdg = (
       <>
-        {this.props.children}
+        {Children.map(this.props.children, (child) => {
+          if (isValidElement(child) && required && child.type === Input) {
+            return cloneElement(child, {
+              'aria-required': true,
+              ...(error && error.length > 0 ? { 'aria-invalid': true } : {}),
```
wesleybl: Why has the testing method changed compared to the main branch? I think it's better to leave it as it is in the main branch.

wesleybl (changes requested):

packages/volto/src/components/manage/Widgets/FormFieldWrapper.jsx
Wagner3UB: I understand your concern because using Children.map and cloneElement adds overhead on every render just to inject two attributes in specific cases.

The motivation was to centralize the ARIA logic in FormFieldWrapper since it already holds both required and error and uses them for visual rendering like the label asterisk and error messages. It felt like the right place to also propagate that information to screen readers without touching every individual widget.

That said, I see the tradeoff. What approach would you consider more viable here? If we choose to change the children to add the two attributes, we will need to modify a lot of widgets. @wesleybl

Wagner3UB: @wesleybl We need to pay attention to what we will decide here, this change was made in the v19 https://github.com/plone/volto/blob/1e9c18b890c24fc3fc94b070a4965052306f313b/packages/volto/src/components/manage/Widgets/FormFieldWrapper.jsx#L60

Wagner3UB: Or I'm overthinking and you mean only back to `'aria-invalid': !!(error && error.length > 0),`

wesleybl (approved):

stevepiercy (approved):

PR #8330, opened 2026-06-12
fix: prevent form submission when adding rows/columns in table block

ilariamassa96:
Adding a row or column inside a table block embedded in the text editor was unexpectedly triggering a form submit.

Added preventDefault() to table block to stop this behaviour


packages/volto-slate/src/blocks/Table/TableBlockEdit.jsx
```diff
@@ -229,7 +229,9 @@ const Edit = (props) => {
     [data, block, onChangeBlock],
   );
 
-  const onInsertRowBefore = useCallback(() => {
+  const onInsertRowBefore = useCallback((ev) => {
+    ev.preventDefault();
```
wesleybl: I think the problem is that the `Button` components lack the `type` attribute, so the default action ends up being the submission of the parent form. Does `volto-blocks-widget` place the table component inside a form? Could you please test adding the `type="button"` property to the `Button` components?

wesleybl (dismissed):

packages/volto-slate/src/blocks/Table/TableBlockEdit.jsx
pnicolli: Agree with wesley here, it's worth trying

ilariamassa96: @wesleybl I followed your suggestion and added `type="button"` to the Button components and the form submission is no longer triggered. Thanks!

pnicolli (approved):

pnicolli (approved):

PR #8382, opened 2026-07-14
fix/minor: wrong skiplink class inside focus.less style file

Wagner3UB:
It fix a class error for https://github.com/plone/volto/pull/7850


packages/volto/news/8382.bugfix
```diff
@@ -0,0 +1 @@
+Fix wrong skiplink selector in focus.less so skiplinks focus styles apply correctly. @Wagner3UB
```
stevepiercy:
```suggestion
Renamed the selector in {file}`focus.less` so that skiplinks focus styles apply correctly. @Wagner3UB
```

stevepiercy (changes requested):
Tweak on the news. Otherwise LGTM.

stevepiercy (approved):

PR #8386, opened 2026-07-16
Listing block: add an offset to skip the first N results

### LINKED ISSUE #8385: Listing block: add an offset to skip the first N results
## Problem

A Listing block can limit how many results it shows, but it cannot skip results. There is no way to say "start from the 4th item".

This makes a very common editorial layout impossible to build without duplication: a page that features the latest N items in one prominent block (a grid or a carousel, say), and then lists **the rest** below in a second block. Today the second block re-renders the same items already shown above, and the only workarounds are to hand-maintain an exclusion list, write a custom variation, or drop to a bespoke block.

The building block for this already exists on the backend — `plone.restapi`'s `@querystring-search` accepts `b_start` — but the Listing block never exposes it as an editable setting.

## Proposal

Add an `offset` setting to the Listing block's querystring: skip the first N results of the query, and paginate over the remainder. With `offset: 3`, the block starts at the 4th result and page 2 continues from there.

Two blocks sharing a query then compose cleanly:

| block | settings | shows |
|---|---|---|
| featured | `limit: 3` | items 1–3 |
| rest | `offset: 3` | items 4…N, paginated |

`offset` defaults to `0`, which is exactly today's behaviour, so this is backwards compatible.

## What needs to change

**1. Expose the field** — `QuerystringWidget` defines the querystring fieldset as `['query', 'sort_on', 'sort_order_boolean', 'limit', 'b_size']`. Add `offset` alongside `limit`, as an integer field defaulting to `0`.

**2. Pass it through the HOC** — `withQuerystringResults` builds the request data for the block. It currently forwards `b_size` and the copied fields but has no notion of `offset`; it needs to forward it so the action can act on it.

**3. Translate it into `b_start` in the action** — `getQueryStringResults` derives `b_start` from the current page. It needs to add the offset:

```
b_start = offset + (page - 1) * b_size
```

**4. Raise the `limit` by the offset** — when a query has both a `limit` and an `offset`, the limit must be increased by the offset, otherwise the backend returns `limit` items and the block drops the first `offset` of them, showing fewer than asked. `limit: 0` must keep meaning "no limit" — that matches `plone.app.querystring`'s querybuilder, which defaults `limit=0` and only applies it when truthy.

**5. Do the same on the SSR path** — `getAsyncData` issues the same query during server-side rendering. If it is not updated in step with the HOC, the first paint and the client re-render disagree.

**6. Exclude skipped items from the counts** — this is the part that is easy to miss. `items_total` from the backend is the length of the **whole** result set; `b_start` does not reduce it (`HypermediaBatch` derives it from the sequence length). So the page count and the result counter must both be computed from `items_total - offset`, not `items_total`.

Without this, a query of 23 results with `offset: 3` and `b_size: 10` advertises 3 pages when only 20 items are reachable. Page 3 requests `b_start: 23` and comes back empty. The result counter has the same problem, announcing items the listing will never show.

## Implementation notes

Worth knowing for whoever picks this up:

- **The settings arrive as strings.** `offset`, `limit` and `b_size` are stored through JSON schema `number` fields, so they can reach the code as strings or floats rather than integers. They need coercing before any arithmetic, or `b_start` silently becomes `"30"` or `NaN`.

- **`b_start` should only be set when there is a page.** The action is also called from contexts that pass no `page` at all (the query widget's own preview, for instance). Computing `page - 1` there yields `NaN`, which serialises to `b_start: null`. Guard it.

- **A set limit caps `actual_result_count`.** The querybuilder caps it at `offset + limit`, so `items_total - offset` is the right visible total both with and without a limit.

- **Folder listings are a different path.** With no query the block takes the `getContent` path, where the offset does not apply. It should be left alone.

## Acceptance criteria

- [ ] `offset` is editable on the Listing block and defaults to `0`
- [ ] Two blocks with the same query, one `limit: N` and one `offset: N`, show no duplicates between them
- [ ] Pagination over an offset listing has no empty trailing page
- [ ] The result counter reports only the items the listing can show
- [ ] SSR and the client render the same items
- [ ] `offset: 0` behaves exactly as today
- [ ] Unit tests cover the `b_start` / `limit` / total arithmetic, including string inputs and `limit: 0`

ericof:
## What

Adds an `offset` setting to the Listing block's querystring: skip the first N results of the query and paginate over the remainder.

Two blocks sharing a query now compose without duplicates:

| block | settings | shows |
|---|---|---|
| featured | `limit: 3` | items 1–3 |
| rest | `offset: 3` | items 4…N, paginated |

`offset` defaults to `0`, which is exactly today's behaviour.

## How

The arithmetic lives in one new helper, `helpers/Pagination/Pagination.ts`, and is applied in the action. Both the client HOC and the SSR path go through that action, so the first paint and the client re-render cannot disagree by construction.

- **`b_start`** = `offset + (page - 1) * b_size`
- **`limit`** is raised by the offset when set, otherwise the backend returns `limit` items and the block drops the first `offset` of them. `limit: 0` keeps meaning "no limit", matching `plone.app.querystring`'s querybuilder.
- **The counts** come from `items_total - offset`. `HypermediaBatch` derives `items_total` from the whole sequence, so `b_start` never reduces it. Without this, a query of 23 results with `offset: 3` and `b_size: 10` advertises 3 pages when only 20 items are reachable, and page 3 comes back empty.

The settings are stored through JSON schema `number` fields, so they can arrive as strings or floats. Every value is coerced before any arithmetic rather than assumed to be an integer.

## Notes for reviewers

- **`getAsyncData` needed no arithmetic.** It already spreads `data.querystring`, so `offset` reaches the action for free. It only gained an explicit `offset` default, to shape the SSR request exactly like the client's. Computing `b_start` there would apply the offset twice.
- **The `Offset` field is on the shared querystring widget**, so it also appears in the Search block's editor — exactly as `limit` and `b_size` already do.
- **Folder listings are untouched**: with no query the block takes the `getContent` path, where the offset does not apply.
- **Backwards compatibility**: with `offset: 0` the request is byte-identical to today's. A caller passing neither a page nor an offset still gets no `b_start` key at all, and a `limit` that needs no raising is forwarded untouched. There is a test pinning this.

## Tests

Unit tests cover the `b_start` / `limit` / total arithmetic including string inputs, floats, negatives, and `limit: 0` — 44 tests on the helper plus the action's own. Full Volto suite passes (1523 passed, 1 skipped).

Not exercised live against a backend; the same logic is in production in a Volto Light Theme derivative.

## Acceptance criteria

- [x] `offset` is editable on the Listing block and defaults to `0`
- [x] Two blocks with the same query, one `limit: N` and one `offset: N`, show no duplicates between them
- [x] Pagination over an offset listing has no empty trailing page
- [x] The result counter reports only the items the listing can show
- [x] SSR and the client render the same items
- [x] `offset: 0` behaves exactly as today
- [x] Unit tests cover the `b_start` / `limit` / total arithmetic, including string inputs and `limit: 0`

Closes #8385


sneridagh (approved):

pnicolli (approved):
I like this new option and this looks good, thanks!
My personal feedback for the future: please only keep js->ts migrations in dedicated PRs, otherwise the PR changes become much more difficult to read.

PR #8388, opened 2026-07-16
Skip config entries that are not blocks in the Block Types control panel

### LINKED ISSUE #8384: BlockTypesControlPanel crashes on Blocksconfig objects without an id
Volto Light Theme contains a bug that registers the volto-accordion-block incomplete, with a missing id key, if the volto-accordion-block is not installed and registered so it does its own applyConfig: 

https://github.com/kitconcept/volto-light-theme/issues/903

This crashes the BlocktypesControlPanel, because the id is missing in one of the returned blocks, crashing another compoment. 

Should we be more defensive in  the BlockTypes Control Panel?
Or should Volto code  in general  assume Blocks always have an id? 

@sneridagh  @jnptk

Credits to @Fosten for finding the relation with installed extra Bloxcks and the missing ID in the Control Panel view when we chatted about it: https://github.com/plone/next.plone.org/pull/39

ericof:
## Problem

An add-on can configure a block that was never registered — typically by enhancing a block that ships with another add-on that is not installed. `applyConfig` happily writes to `blocksConfig[key]`, and what it leaves behind is not a block: it carries only what the add-on set, with no `id` or no `title`.

The Block Types control panel listed those entries anyway, and a row with no `id` crashed the table.

The case in #8384 is Volto Light Theme configuring `volto-accordion-block` when it is not installed (kitconcept/volto-light-theme#903).

## Fix

Skip them. They are not blocks, so listing them would link to a block that does not exist.

The filter reads the configuration **before** the title falls back to the id:

```js
Object.entries(blocksConfig)
  .filter(([, blockConfig]) => blockConfig.id && blockConfig.title)
  .map(...)
```

Order matters here. Filtering after the map does not work: the fallback (`blockConfig.title ? … : blockConfig.id ?? key`) hands a title to an entry that has none, so an entry carrying an id but no title survives the `block.title` check and is listed. With the filter first, id and title are guaranteed and the map no longer needs the ternary or the key.

## Notes for reviewers

- **`sortConfigBlocks` is now exported**, so the behaviour can be tested directly. `BlockTypes.test.ts` is the first test for this component: 10 tests covering the three malformed shapes (no id + no title, id only, title only), the sort order, the title translation, and the invariant the table's row key depends on — every listed block has an id.
- **The tests were verified to fail.** Against the previous logic exactly one goes red — `drops an entry with an id but no title` — and the other nine stay green.
- **A rejected alternative is pinned by a test.** Falling back to the config key for a missing id (`id: blockConfig.id ?? key`) looks tempting, since Volto's convention is that the key *is* the id. It is wrong here: it resurrects the phantom entry and lists a block that does not exist. `does not fall back to the config key for a missing id` guards against reintroducing it.

## The open question

This answers the first half of #8384 — the control panel is now defensive. It does not answer the second half: *should Volto code in general assume blocks always have an id?*

Right now the types say yes. `BlockConfigBase` declares both `id: string` and `title: string` as required, so TypeScript believes this bug is impossible, and the test fixtures have to cast around the type to describe what the configuration really holds. If the answer is "no, do not assume", the type is where that belongs, and this fix is a workaround for a contract that is not enforced.

Worth deciding before other call sites grow the same guard.

Closes #8384
Ref: https://github.com/kitconcept/volto-light-theme/issues/903

### DIFF
--- a/packages/volto/news/8384.bugfix
+++ b/packages/volto/news/8384.bugfix
@@ -1,0 +1,1 @@
+Stopped the Block Types control panel from crashing when an add-on configures a block that was never registered, which leaves behind an entry with no id or no title. Those entries are not blocks, and are no longer listed. @ericof
--- a/packages/volto/src/components/manage/Controlpanels/BlockTypes.test.ts
+++ b/packages/volto/src/components/manage/Controlpanels/BlockTypes.test.ts
@@ -1,0 +1,132 @@
+import { createIntl, createIntlCache } from 'react-intl';
+import type { BlocksConfigData } from '@plone/types';
+import { sortConfigBlocks } from './BlockTypes';
+
+const intl = createIntl(
+  { locale: 'en', messages: { Teaser: 'Destaque' } },
+  createIntlCache(),
+);
+
+// `BlockConfigBase` types id and title as required, so a malformed entry
+// cannot be expressed through it. That mismatch between the type and the
+// configuration is the bug these tests cover, so the fixtures go around it.
+const asBlocksConfig = (config: Record<string, unknown>) =>
+  config as unknown as BlocksConfigData;
+
+describe('sortConfigBlocks', () => {
+  it('lists a registered block', () => {
+    const blocks = sortConfigBlocks(
+      asBlocksConfig({ image: { id: 'image', title: 'Image' } }),
+      intl,
+    );
+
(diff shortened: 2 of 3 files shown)


sneridagh (approved):
LGTM. 

A proper blocks API is needed. I started this many times but never got to a completion. My idea: make it work for Aurora, export it to Volto. I will try again to make it happen during summer.

PR #8390, opened 2026-07-20
Fix portrait passthrough API traversal

sneridagh:
## What

Move the `/@portrait/*` server passthrough out of the general image middleware into a dedicated portrait middleware.

The dedicated helper preserves the backend traversal choice while using the development backend URL:

- Plone uses `/++api++/@portrait/...`, so the portrait endpoint is served through the API traversal layer.
- Backends configured with legacy traversal use `/@portrait/...` without the `++api++` segment.

Other image resources (`@@images` and `@@site-logo`) retain their existing passthrough behavior.

## Why

This addresses a regression introduced by [#5429](https://github.com/plone/volto/pull/5429). That change intentionally omitted `++api++` for development passthroughs to support the Nick backend, but it also made the Plone `@portrait` passthrough bypass the API traversal layer.

## Compatibility note

This may break portraits in development for Nick installations that have not enabled legacy traversal. Configure the development environment with **`RAZZLE_LEGACY_TRAVERSE`** to keep the portrait request on the non-`++api++` route.

## Tests

- `pnpm --filter @plone/volto test --run src/helpers/Api/PloneBackendAPIResourceWithAuth.test.js src/express-middleware/images.test.js`

### DIFF
--- a/packages/volto/news/8390.bugfix
+++ b/packages/volto/news/8390.bugfix
@@ -1,0 +1,1 @@
+Fix development-mode portrait passthrough for Plone by routing requests through the `++api++` traversal layer. @sneridagh
--- a/packages/volto/src/config/server.js
+++ b/packages/volto/src/config/server.js
@@ -1,5 +1,6 @@
 import imagesMiddleware from '@plone/volto/express-middleware/images';
 import filesMiddleware from '@plone/volto/express-middleware/files';
+import portraitsMiddleware from '@plone/volto/express-middleware/portraits';
 import robotstxtMiddleware from '@plone/volto/express-middleware/robotstxt';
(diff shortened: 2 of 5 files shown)


sneridagh:
@robgietema I've added this PR for fixing a bug introduced in https://github.com/plone/volto/pull/5429 

This is a "middle-ground" workaround, since it would work in production for both Plone Backend and Nick, this would break now in Nick for `@portrait` in `__DEVELOPMENT__` :( 

The workaround is to run with `RAZZLE_LEGACY_TRAVERSE` in dev mode too for Nick development too. 

In the long run: I would design something like what we did in Aurora and `@plone/client`. The `APISuffix` solution is more satisfactory than using the ugly env var.

PR #8391, opened 2026-07-21
Add configurable API suffix

sneridagh:
## What changed

Adds a configurable API traversal suffix for Volto. This is resolved like we did for `@plone/client` initialize method, using `apiSuffix`. In Volto, we go for:

- Introduces `config.settings.apiSuffix`, with `RAZZLE_API_SUFFIX` as its environment-variable counterpart.
- Centralizes suffix resolution in `getApiSuffix()` and uses it across API URL formatting, backend URL expansion, authenticated resource requests, and sitemap requests.
- Keeps existing behavior by default: Volto continues to use `/++api++`.
- Gives an explicitly configured `apiSuffix` precedence over `legacyTraverse`.
- Supports an empty suffix (`RAZZLE_API_SUFFIX=''` or `apiSuffix: ''`) to intentionally omit traversal.
- Hydrates `RAZZLE_API_SUFFIX` from `window.env` so seamless-mode runtime configuration stays consistent between SSR and browser requests.
- Adds a feature Towncrier entry and documents the new environment variable.

## Why

Volto currently chooses only between the fixed `/++api++` suffix and no suffix through `legacyTraverse`.
Deployments that expose the Plone API under another traversal suffix need a supported configuration option without relying on the legacy traversal switch.
Deployments for Plone Nick will also need to setup this to empty string.

## Impact

Integrators can configure a nonstandard API traversal suffix through `config.settings.apiSuffix` or `RAZZLE_API_SUFFIX`.
The setting overrides `legacyTraverse` when present, while existing installations retain their current behavior when it is absent.

## Validation

- `pnpm --filter @plone/volto test --run src/helpers/Url/Url.test.js src/helpers/Api/Api.plone.rest.test.js src/helpers/Api/PloneBackendAPIResourceWithAuth.test.js`
  - 109 tests passed.
- Prettier check passed for the changed source, type, documentation, and test files.
- Commit hooks passed, including package lint/format checks and the i18n check.
- `pnpm --filter @plone/volto check:ts` still reports existing unrelated typing errors in widget and Storybook files; no errors are associated with this change.

### DIFF
--- a/docs/source/configuration/environmentvariables.md
+++ b/docs/source/configuration/environmentvariables.md
@@ -66,6 +66,12 @@ You can also generate builds on your continuous integration, then deploy them an
 
 ```{glossary}
 :sorted:
+`RAZZLE_API_SUFFIX`
+    Overrides the API traversal suffix that Volto appends to API URLs.
+    The default suffix is `/++api++`.
+    Set this variable to use a different suffix, or set `config.settings.apiSuffix` directly in your Volto configuration.
+    This setting takes precedence over `RAZZLE_LEGACY_TRAVERSE`.
+
 `RAZZLE_LEGACY_TRAVERSE`
     If `true`, Volto will construct API URLs without the `/++api++` prefix.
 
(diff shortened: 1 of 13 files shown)


sneridagh:
Superseded by #8392, which uses the corrected `configurable-api-suffix` branch name and includes the missing `@plone/types` Towncrier entry.

PR #8396, opened 2026-08-04
Forces the use of webpack 5.105.4

wesleybl:
This avoids compatibility errors with the version required by Storybook. It prevents the error:

`TypeError: Cannot use 'in' operator to search for 'start' in undefined`

Two incompatible versions of webpack were installed. Objects generated by one version were being processed by the other, causing the error. This enforces the use of a single version.

See the error in: https://github.com/plone/volto/actions/runs/30935387825/job/92080103014


nileshgulia1 (approved):
Tested with local yeoman generator. The project scaffolds with webpack :5.105.4 in resolutions.

PR #8400, opened 2026-08-14
Fix broken links and redirects

stevepiercy:
- nodejs.com now resolves, removed from ignore
- added stackoverflow.com to ignore

A change log entry is not needed for a chore such as this.


nileshgulia1 (approved):

stevepiercy:
Thanks for the review, @nileshgulia1!

PR #8401, opened 2026-08-14
Support chore type for Towncrier

stevepiercy:
This PR adds support for the `chore` type in Towncrier. Change log entries with this file extension will satisfy the change log entry requirement in pull requests, but won't display in the changelog.

Without this, we have to override the merge requirements, like in #8400.


nileshgulia1:
@stevepiercy should we need to mention in the developer docs somewhere that .`chore` should mean "this change has no observable effect on anyone consuming the package and the contents might be hidden in final changelog"?  
 I don't know if we even have such developer docs for changelogs types...
 Anyway there are also some caveats that it may still show content for orphan-fragments: https://towncrier.readthedocs.io/en/latest/configuration.html#use-toml-tables-alphabetical-order

stevepiercy:
@nileshgulia1 I updated the docs in https://github.com/plone/documentation/pull/2099. Would you review that one as well?

nileshgulia1 (approved):

PR #5273, opened 2023-10-04
onFocus function to handle block selection

Wagner3UB:
Having focus on the block now is equivalent to clicking on it. Now the block becomes active when focused

To solve this issue: https://github.com/plone/volto/issues/5204

https://github.com/plone/volto/assets/60133113/5410aa64-c4e9-42b0-bb09-190704477322

PR #8402, opened 2026-08-18
Fixes the cursor position when clicking on a Slate block

wesleybl:
On pages with many blocks, clicking on a Slate block would cause the click position to be lost, with the cursor jumping to the beginning of the text. Now, we save the position and set it on the Slate block after it is selected.

### DIFF
--- a/packages/volto-slate/news/8402.bugfix
+++ b/packages/volto-slate/news/8402.bugfix
@@ -1,0 +1,1 @@
+Fixed cursor position lost when clicking on a Slate block in edit, causing the cursor to jump to the start of the text. @wesleybl
--- a/packages/volto-slate/src/editor/SlateEditor.jsx
+++ b/packages/volto-slate/src/editor/SlateEditor.jsx
@@ -54,6 +54,7 @@ class SlateEditor extends Component {
     this.createEditor = this.createEditor.bind(this);
     this.multiDecorator = this.multiDecorator.bind(this);
     this.handleChange = this.handleChange.bind(this);
+    this.handlePointerDown = this.handlePointerDown.bind(this);
     this.getSavedSelection = this.getSavedSelection.bind(this);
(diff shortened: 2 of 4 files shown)


wesleybl:
I noticed that this is happening even when there are few blocks. Further investigation is needed. This didn't happen before.

wesleybl:
Upon closer inspection, I noticed the issue affects only a few blocks, and specifically within the `/edit` —not when creating new content. This problem arose after: #5273

However, the fix remains essentially the same; I’ve made some improvements and added a Cypress test.

@davisagli @sneridagh @Wagner3UB Could you take a look at this? This bug is really annoying.

davisagli:
@wesleybl How can I reproduce the issue? It seems to be working okay for me on https://demo.plone.org/edit in Firefox.

nileshgulia1:
This issue happens when you go to /edit on any page and click on a Slate block. You'll notice the cursor always jumps to the front of the block instead of where you clicked.

I reproduced this on the main branch, and also on a custom Volto 18 site, where it happens inconsistently, sometimes it occurs, sometimes it doesn't.

On main, I confirmed that this branch fixes the issue.

nileshgulia1 (approved):
Lgtm from my side. Would be great to have other reviews as well.

wesleybl:
> @wesleybl How can I reproduce the issue? It seems to be working okay for me on https://demo.plone.org/edit in Firefox.

@davisagli I couldn't reproduce the issue on https://demo.plone.org. Perhaps because it uses a different theme?

But I was able to reproduce it with vanilla Volto installed locally.

wesleybl:
I removed the modification in Edit.jsx. The fix works without it. This makes the solution simpler and preserves the previous behavior.

Wagner3UB (approved):
All good. Thank you.

PR #8403, opened 2026-08-26
Fix invalid production builds from Terser 5.51.0

sneridagh:
## Summary

- pin Terser to 5.46.0 in `@plone/razzle`
- prevent Terser 5.51.0 from generating invalid production server bundles
- add an internal Towncrier news item

Terser 5.51.0 can assign the same mangled identifier to a `const` and a function declaration in the same scope, causing Node to fail with `SyntaxError: Identifier has already been declared`. Cookieplone generated projects resolve dependencies without Volto’s root lockfile, so they picked up the newly released version.

## Validation

- `pnpm --filter @plone/razzle test`
- `pnpm --filter @plone/volto build`
- reproduced a clean Cookieplone workspace and confirmed its production build succeeds
- `node --check core/packages/volto/build/server.js` in the generated workspace
- `git diff --check`


wesleybl:
@sneridagh I think the best place to do this is in the catalog, since this package isn't a direct dependency of volto-razzle.

wesleybl:
Hmm, the catalog wouldn't work, because it would have to become a dependency anyway. Maybe it would have to be done via `pnpm overrides` in `cookieplone`.

sneridagh:
@wesleybl works there. The catalog won't enforce it, and I want to avoid overrides that we can't manage in a centralized way, otherwise we will need to ask people to add them to all projects to fix this :(

PR #8405, opened 2026-08-27
Extract an overridable UserAvatar component from the Toolbar

### LINKED ISSUE #8404: Toolbar: extract an overridable UserAvatar component
## Problem

The signed-in user is drawn in two places, and in both of them the markup is inlined into a component that cannot be reached by an add-on without replacing the whole thing.

**`Toolbar`** — `src/components/manage/Toolbar/Toolbar.jsx`, the `#toolbar-personal` button:

```jsx
<button className="user" ... id="toolbar-personal">
  <Icon
    name={userSVG}
    size="30px"
    title={this.props.intl.formatMessage(messages.personalTools)}
  />
</button>
```

**`PersonalTools`** — `src/components/manage/Toolbar/PersonalTools.jsx`, the avatar block:

```jsx
<div className={cx('avatar', { default: !user.portrait })}>
  {user.portrait ? (
    <Image src={expandToBackendURL(user.portrait)} alt={...} />
  ) : (
    <Icon name={cameraSVG} size="96px" />
  )}
</div>
```

Both draw the same picture for everybody who has not uploaded a portrait — a generic person glyph in the toolbar, a camera icon in the menu. Neither has an extension point: `toolbar-personal` is a DOM id on a button, not a pluggable, and `toolbar-user-menu` sits at the end of the menu *list*, below the avatar block. An add-on that wants to change what the toolbar button shows has to shadow `Toolbar.jsx` — a ~700-line class component — for two lines of JSX, and then carry that copy across every Volto upgrade.

## Proposal

Extract a `UserAvatar` component and use it in both places:

```jsx
// src/components/manage/Toolbar/UserAvatar.jsx
const UserAvatar = ({ size }) => { ... };
```

- `Toolbar` renders `<UserAvatar size="30px" />` in the `#toolbar-personal` button.
- `PersonalTools` renders `<UserAvatar size="96px" />` in the avatar block.
- Default behaviour stays exactly what it is today, so nothing changes for anybody who does not override it.

An add-on then shadows one small, purpose-built component instead of two large ones, and the override survives upgrades to the components around it.

If a full component feels like too much, a pluggable in the button (`toolbar-personal-avatar`, say) would also solve it — but a component is the smaller surface and matches how the rest of the toolbar is built.

## Why we want it

We maintain [`pas.plugins.identity`](https://github.com/collective/pas-plugins-identity), where a user's fields — including their picture — can live on a content object rather than in `portal_memberdata`. Our frontend draws a per-user avatar: their portrait when they have one, and otherwise their initials on a colour derived from their user id, so two people in a list are told apart at a glance rather than sharing one glyph.

To put that in the toolbar we currently ship a verbatim copy of Volto's `Toolbar.jsx` with a two-line patch, purely to reach that `<Icon>`. That copy is a liability at every Volto upgrade, and it is the only Volto component we have not been able to rewrite as our own — precisely because the change we need is so small.

The general case is broader than us: gravatars, org-directory photos, presence indicators and initials are all the same one-component change, and all of them are currently a fork of the toolbar.

Happy to open the PR.

ericof:
## What

Extracts a `UserAvatar` component and uses it for the signed-in user's avatar in the toolbar's `#toolbar-personal` button.

Nothing changes for anybody who does not override it — the same user glyph, at the same 30px. The `Toolbar` snapshots are untouched, which is the proof.

## Why

An add-on that wants to change what the toolbar button shows currently has to shadow `Toolbar.jsx` — a ~700-line class component — for two lines of JSX, and then carry that copy across every Volto upgrade. `#toolbar-personal` is a DOM id on a button, not a pluggable, so there is no smaller seam to reach for.

We hit this in [`pas.plugins.identity`](https://github.com/collective/pas-plugins-identity), where a user's picture can live on a content object rather than in `portal_memberdata`, and we draw a per-user avatar — their portrait when they have one, otherwise their initials on a colour derived from their user id. That was the only Volto component we had not been able to rewrite as our own, precisely because the change we needed was so small.

The general case is broader: gravatars, org-directory photos, presence indicators and initials are all the same one-component change, and all of them are currently a fork of the toolbar.

## How

`packages/volto/src/components/manage/Toolbar/UserAvatar.tsx`:

```tsx
export type UserAvatarProps = {
  /** Title of the rendered icon (a11y). */
  title?: string;
  /** Icon to render in place of the default user glyph. */
  icon?: typeof userSVG;
  /** Size of the avatar (in px). Defaults to the toolbar button size. */
  size?: string;
};

const UserAvatar = ({ title, icon, size = '30px' }: UserAvatarProps) => {
  return <Icon name={icon || userSVG} size={size} title={title} />;
};
```

- **`UserAvatarProps` is exported**, so an add-on shadowing the component can type against it rather than re-deriving the shape.
- **`size` defaults to `'30px'`.** `Icon`'s own `defaultSize` is `36px`, so passing `size` straight through when undefined would have silently grown the toolbar button by 6px. `Toolbar` also passes `size="30px"` explicitly, to document intent at the call site.
- **`icon` is typed `typeof userSVG`**, which the ambient `declare module '*.svg'` resolves to `string`. Any `import x from '….svg'` is therefore assignable without the `as any` cast that `ButtonsWidget` needs.

## Notes for reviewers

- **`PersonalTools` is deliberately left alone.** The original issue proposed using `UserAvatar` there too, at 96px. That is not a pure extraction: the two sites do not render the same thing today. The toolbar shows a bare `userSVG` glyph with no portrait support, while `PersonalTools` reads `state.users.user` and shows `user.portrait` when set and `cameraSVG` otherwise, wrapped in a `div.avatar` whose `default` class carries CSS. Sharing one component there means deciding whether the toolbar button starts showing portraits — a user-visible behaviour change that deserves its own PR and its own discussion. This PR is the seam only.
- **The `icon` prop is currently unused by Volto itself.** It is there for callers that want to vary the glyph without shadowing. Happy to drop it if reviewers would rather keep the surface minimal.

## Tests

New `UserAvatar.test.tsx`, 8 tests, following the `AlignWidget.test.tsx` pattern: default-icon snapshot, default and explicit sizing, the a11y `<title>` present and absent, a custom `icon` overriding the default, and `icon` + `size` composing.

- All 21 tests across the 7 `Toolbar` test files pass.
- Full `volto` suite passes.
- `yarn check:ts` goes from 14 to 12 errors on this branch — the two removed are exactly `UserAvatar`'s former implicit-`any` props. The remaining 12 are pre-existing in `Widgets/*.stories.tsx` and `Widgets/index.tsx` and are untouched here.
- `eslint --max-warnings=0` and prettier clean on the changed files.

## Unrelated observation

While establishing the typecheck baseline: `packages/volto/tsconfig.json` excludes test and stories files using brace expansion —

```json
"exclude": ["src/**/*.test.{js,jsx,ts,tsx}", "src/**/*.stories.{js,jsx,ts,tsx}"]
```

TypeScript's `include`/`exclude` do not support brace expansion, so these patterns match nothing and **422 test and stories files are currently in the typecheck program**. That is why most of the pre-existing `check:ts` errors sit in `.stories.tsx` files. Verified by running two otherwise-identical tsconfigs, one with braces and one with the patterns written out: 10 test files in the program versus 0.

Not fixed here, since it would change the error baseline for everyone. Happy to open a separate issue.

## Acceptance criteria

- [x] The toolbar's user avatar is rendered by a dedicated, shadowable component
- [x] Default rendering is unchanged (same glyph, same size, snapshots untouched)
- [x] The component is typed, and its props type is exported for add-ons
- [x] `size` is configurable, and defaults to the current toolbar size
- [x] Unit tests cover the default, the sizing, the a11y title, and the icon override

Closes #8404


sneridagh (approved):
LGTM!

Issue #8407, opened 2026-08-28
Cannot copy/cut/paste blocks that are inside container blocks (Grid, Group, Columns)


nileshgulia1:
**Describe the bug**

Blocks that live *inside* a container block cannot be copied, cut, or pasted. This affects all container-style blocks, e.g. the core Grid block, Group blocks, and Columns blocks (including third-party implementations). When a block is selected inside a container, the copy/cut/paste options never show up in the toolbar, so it is impossible to duplicate or move an inner block.

**To Reproduce**

1. Go to the Edit view of a page
2. Add a container block, e.g. Grid block (`gridBlock`) or if you have EEA columns block
3. Add a block inside it (e.g. a teaser or image in one of the grid columns)
4. Select the inner block (also try multi-selecting with `Shift`-click, like you would for top-level blocks)
5. Look at the toolbar at the bottom

→ Expected: copy/cut/paste options appear, so the inner block can be duplicated or moved to another container/page.

→ Actual: no copy/cut/paste options appear for any block nested inside a container. Copy/cut/paste only works for the page's top-level blocks.

**Software**

- Volto version: 18.x.x (tested on 18.35.1)
- Also reproducible on https://demo.plone.org (Plone 6)

**Related issues/Discussion**

- [#6103](https://github.com/plone/volto/issues/6103) – Copy paste container blocks do not generate new ids for contained blocks (bug, fixed by [#6112](https://github.com/plone/volto/pull/6112)): about inner blocks keeping the same ids when copying a container. Related, but distinct from the missing copy/cut/paste UI inside containers described here.

PR #8410, opened 2026-09-02
fix: enable copy/cut/paste of blocks inside container blocks 

nileshgulia1:
This PR addresses the issues that prevent copy/cut/paste from working reliably on blocks nested inside container blocks (Grid, Group, Columns, Tabs, Accordion), and fixes related selection and clipboard-persistence bugs.

Fixes: https://github.com/plone/volto/issues/8407

### Changes
1. Selection guard in Edit.jsx
Added isNestedFormEvent() method that detects when a click/mouse/focus event originates inside a nested block form (.blocks-form). Updated onMouseDown, onClick, and onFocus handlers to skip parent container selection logic when the event targets a nested form. This allows proper selection and interaction with child blocks inside containers.
2. Multi-selection fix in Form.jsx
Fixed the top-level multi-selection logic (onSelectBlock) that was reading an undefined this.state.selected instead of this.props.uiState.selected, which caused broken Ctrl/Meta-click multi-selection and ghost null entries in the selection.
3. Clipboard persistence in BlocksToolbar.jsx
loadFromStorage() now reads the clipboard from localStorage using the split keys (blocksClipboard.cut / blocksClipboard.copy) that Volto 18's persistentReducers actually writes, instead of the monolithic blocksClipboard key that is never written. This fixes cross-tab clipboard sync and clipboard restore on remount. Also fixed a storage listener leak on unmount.


wesleybl:
@nileshgulia1 I created a text block and, right below it, an empty Grid block. Then I selected and copied the text block. Next, I clicked on an empty column within the Grid block and pasted it. The text block didn't move into the column. Shouldn't that happen? Or does this PR address a different issue?

It was also impossible to select a text block located inside a Grid block.

wesleybl:
@nileshgulia1 This is a fix for #8407, right?

wesleybl:
Wouldn't it be better to make a fix for the `main` branch first?

nileshgulia1:
> @nileshgulia1 I created a text block and, right below it, an empty Grid block. Then I selected and copied the text block. Next, I clicked on an empty column within the Grid block and pasted it. The text block didn't move into the column. Shouldn't that happen? Or does this PR address a different issue?
> 
> It was also impossible to select a text block located inside a Grid block.

@wesleybl This PR used to fix the multiselection and the blocksClipbaord bug in 18.x.x. 

I just noticed this happens in 18.x.x that you cannot multiselect and copy blocks anymore? Can you verify this also? 

The fix: BlocksToolbar.jsx now reads the clipboard from localStorage using the split keys (blocksClipboard.cut / blocksClipboard.copy) that Volto 18's persistentReducers actually writes, instead of the never-written monolithic blocksClipboard key. This restores cross-tab sync and clipboard state on remount. 


Also after [2cbe11a](https://github.com/plone/volto/pull/8410/commits/2cbe11ab09febcb05770a02a85d04988f8eb8956) You should be able to Copy/cut from/to Grid block as you tested in the comment. I plan to keep that as a separate PR but it can be shipped with this as well.

PR #8416, opened 2026-09-02
Move table row/column insertion controls to a floating menu (#8409)

### LINKED ISSUE #8409: Move table row/column insertion controls to a floating menu

Yes. The table block currently displays the row and column insertion controls (such as "Insert row before", "Insert row after", etc.) in a fixed position at the top of the table.

This becomes inconvenient when working with tables containing many rows. If the user is editing a row near the bottom of a long table, they have to scroll back to the top of the table to access these controls, perform the action, and then scroll back to the row they were working on.

The row and column insertion controls should be displayed in a floating menu positioned above the currently selected row.

For example, when a row is selected, the menu containing options such as "Insert row before", "Insert row after", and other table operations should appear above that row. This would keep the controls close to the user's current context, regardless of the row's position within the table.

The menu should move with the currently selected row when the selection changes.

Keeping the controls fixed at the top of the table is the current implementation, but this requires unnecessary scrolling when working with long tables.

Another alternative would be to make the controls sticky while scrolling through the table, but positioning the menu relative to the currently selected row would provide a more contextual and direct interaction.

**Additional context**

This is particularly noticeable with tables containing a large number of rows. The current behavior makes inserting or modifying rows near the bottom of the table unnecessarily cumbersome.

A floating contextual menu above the selected row would make table editing more efficient and consistent with the user's current editing context.

r4-rahul123:

-----

Closes #8409

### Summary of changes
- Updated the Table block edit component (`TableBlockEdit.jsx`) so that the row and column insertion/deletion controls float contextually above the currently selected row rather than staying fixed at the top of the table.
- Reuses the existing toolbar buttons and icons as guided in https://github.com/plone/volto/issues/8409#issuecomment-5512117592.
- Added smooth transition when navigating between table rows.
- Added Towncrier news entry (`8409.feature`).

### DIFF
--- a/packages/volto-slate/src/blocks/Table/TableBlockEdit.jsx
+++ b/packages/volto-slate/src/blocks/Table/TableBlockEdit.jsx
@@ -1,4 +1,4 @@
-import { useState, useEffect, useCallback, useMemo } from 'react';
+import { useState, useEffect, useCallback, useMemo, useRef } from 'react';
 import PropTypes from 'prop-types';
 import isEmpty from 'lodash/isEmpty';
 import map from 'lodash/map';
@@ -182,12 +182,33 @@ const Edit = (props) => {
     row: 0,
     cell: 0,
   });
+  const tableRef = useRef(null);
+  const [toolbarTop, setToolbarTop] = useState(-32);
+
   // If selected prop is unset, update the state
   useEffect(() => {
     if (!selected) {
       setSelectedCell(null);
     }
   }, [selected]);
+
+  useEffect(() => {
+    if (!selected || !selectedCell || !tableRef.current) {
+      setToolbarTop(-32);
+      return;
+    }
+    let rowEl;
+    if (!data.table?.hideHeaders && selectedCell.row === 0) {
+      rowEl = tableRef.current.querySelector('thead tr');
+    } else {
+      const bodyRows = tableRef.current.querySelectorAll('tbody tr');
+      const rowIndex = selectedCell.row > 0 ? selectedCell.row - 1 : 0;
+      rowEl = bodyRows[rowIndex];
+    }
+    if (rowEl) {
+      setToolbarTop(rowEl.offsetTop - 34);
+    }
+  }, [selected, selectedCell, data.table]);
 
   useEffect(() => {
     if (!data.table || isEmpty(data.table)) {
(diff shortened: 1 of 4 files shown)


packages/volto-slate/src/blocks/Table/TableBlockEdit.jsx
```diff
@@ -182,13 +182,34 @@ const Edit = (props) => {
     row: 0,
     cell: 0,
   });
+  const tableRef = useRef(null);
+  const [toolbarTop, setToolbarTop] = useState(-32);
```
wesleybl: -32 and -34 are fixed numbers. If the CSS changes the button height, it will cause issues. The offset can be calculated using only CSS, based on the toolbar's actual height. Something like this:

```css
.block.table {
  --toolbar-gap: 2px;
 .toolbar {
    top: 0;
    transform: translate(-50%, calc(-100% - var(--toolbar-gap)));
    transition: top 0.15s ease-in-out;
  }
}
```

This way, `toolbarTop` is used solely to define the top position (which line the toolbar should align with).

packages/volto-slate/src/blocks/Table/TableBlockEdit.jsx
```diff
@@ -182,13 +182,34 @@ const Edit = (props) => {
...
+      rowEl = tableRef.current.querySelector('thead tr');
+    } else {
+      const bodyRows = tableRef.current.querySelectorAll('tbody tr');
+      const rowIndex = selectedCell.row > 0 ? selectedCell.row - 1 : 0;
+      rowEl = bodyRows[rowIndex];
+    }
+    if (rowEl) {
+      setToolbarTop(rowEl.offsetTop - 34);
+    }
+  }, [selected, selectedCell, data.table]);
```
wesleybl: We can rely solely on:

selectedCell.row, selected, and data.table?.hideHeaders

to avoid unnecessary re-renders.

wesleybl (changes requested):

r4-rahul123:
Hi @wesleybl,

Thank you for the review and helpful suggestions! 
I have updated the code to:
1. Handle toolbar height and offset purely via CSS using `transform: translate(-50%, calc(-100% - var(--toolbar-gap)))` and `transition: top 0.15s ease-in-out`.
2. Use `rowEl.offsetTop` directly for `toolbarTop`.
3. Optimize the `useEffect` dependencies to `[selected, selectedCell?.row, data.table?.hideHeaders]`.

All lint and unit tests are passing and the branch is up to date with `main`. Please let me know if any further adjustments are needed!

wesleybl (approved):

wesleybl:
@r4-rahul123 Thanks! Could you please backport this to the 18.x.x branch? Usually, a cherry-pick suffices, but that won't work here because the components modified here are class-based in that branch.

PR #8418, opened 2026-09-04
Allow .cjs and .mjs files to be shadowed

kunalKumar-13:
Backport of #8413 to `18.x.x`, as requested.

Cherry-picked from `799e67d81` with `-x`; it applied without conflicts, and the
glob line it changes is identical on this branch.

`packages/registry` is green here: 4 test files, 83 tests. I also checked the
backported test is doing its job on `18.x.x` rather than passing regardless —
reverting just the glob change fails `addon-registry.test.js` on the `.cjs`
alias assertion.


wesleybl (approved):

wesleybl:
Thanks!

Issue #572, opened 2018-11-12
Stop Pyup-Bot usage


jensens:
May you please not use @pyup-bot ? It really spams the list of open PRs in Plone orga here. https://github.com/pulls?q=is%3Aopen+is%3Apr+user%3Aplone+archived%3Afalse+sort%3Aupdated-desc

At least, if you can not stop using it, then merge or close those PRs very very fast. 

And I am pretty sure that the way it updates the requirements.txt is the wrong approach for Plone. 

Thanks

jensens:
FYI: I blocked @pyup-bot for whole Plone for now to prevent more spam. 
Please delete/close all the PRs thanks.

PR #8429, opened 2026-09-17
Fix diff of text block word by word

iFlameing:
**Summary**

Fixes the History → Diff view (/page/diff) highlighting an entire paragraph as changed when only a single word changed inside it.

**Root cause:**  the diff algorithm was fed the whole page as one giant token blob, and even within one changed field it diffed the entire block's text word-by-word in one pass. With enough repetitive words, the diff library often found it "cheaper" to mark the whole paragraph as removed+added rather than pinpointing the one changed word — which is why single-line fields (e.g. Title) always worked, but multi-line paragraphs didn't.

screenshot

diffscreenshot


reekitconcept (commented):
Thanks for taking this on, Alok. I tested the PR against Volto 19.3.1 with two small vitest cases in `DiffField.test.jsx`, and I think the diagnosis in the description does not hold, which is why the fix does not reach the reported bug. Details below.

**What I tested**

| Code | Case A: plain multi-sentence text, one word changed (the PR's own test) | Case B: the same text preceded by an `` (byline avatar, image block with caption, mention pill) |
| --- | --- | --- |
| unpatched 19.3.1 | passes: only "old" / "new" marked | fails: the whole text after the image is one deletion + one addition |
| this PR | passes | fails identically |
| this PR + the `splitWords` fix below | passes | passes: only "Porttitor" / "Lorem ipsum" marked; the existing DiffField tests stay green |

Case B:

```js
const before = '<img src="/avatar.png" alt="Admin"> Nisl natoque vehicula condimentum conubia. Porttitor nisl netus bibendum.';
const after  = '<img src="/avatar.png" alt="Admin"> Nisl natoque vehicula condimentum conubia. Lorem ipsum nisl netus bibendum.';
// expected: deletions ['Porttitor'], additions ['Lorem', 'ipsum']
// received (before and after this PR): the whole text as one deletion and one addition
```

**1. The root cause is in the tokeniser, not in the diff algorithm.** jsdiff's `diffArrays` is a Myers diff and always finds the minimal edit script; on word tokens it never "prefers" a whole-paragraph replacement. That is why case A passes on unpatched code: the new test does not exercise the bug, and the sentence-unit two-pass diff changes nothing observable.

The actual cause was found by Tisha on GitLab #573 in August: `splitWords` sets `insideSpecialTag` on `` and only clears it on a closing tag. `` is a void element, so the flag stays on for the rest of the string, spaces are no longer split, and everything after the first image becomes a single token. Any single-word change then shows as a whole-paragraph change. Per-block rendering happens to hide this for text blocks that contain no image (the byline avatar is no longer in the same string), but every block with an image before text still degrades.

Fix, three lines in `splitWords`:

```js
if (tagNameMatch && specialTags.includes(tagNameMatch[1])) {
  // Only container tags (svg) span until their closing tag; void
  // elements such as <img> are complete with the opening tag.
  insideSpecialTag =
    tagNameMatch[1] !== 'img' &&
    !tagNameMatch[0].startsWith('</') &&
    !tagBuffer.endsWith('/>');
```

plus case B as a test. With that in place I would drop `groupIntoUnits` and the two-pass diff, or keep them only with a test that fails without them.

**2. Per-block diffing in `Diff.jsx` is a good direction**, and it matches where we want to go (block-level diff, see GitLab #636). Two gaps in this version:

- `changedBlockIds` only contains blocks whose data changed, so a layout-only change (a block moved) now produces no diff at all; before, any layout change produced one.
- The field title "Blocks" is rendered once per changed block.

Added or removed blocks are still only coloured through their text tokens, so #571 is unchanged by this PR, which is fine, just noting it.

**3. GitLab #573** (description shown twice: description field plus description block) is not addressed here.

**Verdict: request changes.** Needed before this can claim #572: the `splitWords` void-tag fix with an image-prefixed test, handling of layout-only changes, and either a failing test that justifies the sentence grouping or its removal.

Alok, do you want to continue with this yourself, or shall I take it over and continue next week? Either is fine with me, just say which.

iFlameing:
@reekitconcept please take it over

PR #8430, opened 2026-09-18
Honor the frontendOptions schema hint when rendering a field in views

### LINKED ISSUE #8428: frontendOptions is not used to select the view widget
Explicitly setting the widget in the backend schema with

```python
    directives.widget(
        "social_links",
        frontendOptions={
            "widget": "social_media_object_list",
            "widgetProps": {"schemaName": "socialMedia"},
        },
    )
```

has no effect in generated views (for content types that do not have blocks) because `getWidget` ignores the backend information.

ericof:
## What

A field can ask for a specific frontend widget from the backend schema:

```python
directives.widget(
    "social_links",
    frontendOptions={
        "widget": "social_media_object_list",
        "widgetProps": {"schemaName": "socialMedia"},
    },
)
```

The edit form honors that. The view never looked at it, so for a content type without blocks the field rendered with the default text widget, whatever the schema asked for. This makes the view honor it too, including `widgetProps`.

Closes #8428

## Why the fix is not where the issue says it is

The issue points at `getWidget`, but the schema shows `frontendOptions` lives under `widgetOptions`, not on the field root — and the function that actually picks the component is `getWidgetView`. Resolving `social_links` against the current code returns `"json"` from `getWidget`, and `getWidgetView` then walks its whole chain down to `default`, even when `social_media_object_list` *is* registered. The lookup simply never happens.

### One normalization step instead of a second resolution step

`Form/Field.jsx` implements the convention as a resolution step; `FormValidation.jsx` implements it as field normalization, inline:

```js
field = { ...field, ...field.widgetOptions?.frontendOptions?.widgetProps };
const hasSpecificValidator = field.widgetOptions?.frontendOptions?.format || field.format;
const widgetName = field.widgetOptions?.frontendOptions?.widget || field.widget || '';
```

That third key, `format`, is the argument for normalization: a resolution step can only ever carry `widget`, so every new tagged value needs a new step in a new chain.

`helpers/Widget/taggedValues.ts` makes that shape official. `applyTaggedValues` merges the hints onto the field before any resolution runs: the hinted widget replaces the declared one, `widgetProps` and every other tagged value are merged in, and the displaced widget is kept as `_widget` so resolution can fall back to it when the hinted widget is not registered — which is what happens when the add-on providing it is absent.

It is exported from `helpers/index.js`. Adopting it in `Field.jsx` and `FormValidation.jsx` is deliberately left for a follow-up: `Field.jsx` is the highest-traffic component in Volto and its `frontendOptions` coverage is two snapshot tests.

### Two gaps in the views registry

`getWidget` was papering over them:

- no entries for `title` and `description` by field id;
- no `factory` registry at all, and no step to consult one.

Both are filled, and `getWidgetView` gained a `factory` step. The view chain now has the same steps as the edit chain, against the `views` registry, with one deliberate difference in order: in views, the factory is consulted **before** `choices` and vocabularies.

```
id → widget (hint) → _widget → factory → choices → vocabulary → vocabulary hint → type → default
```

That difference exists for relation fields. A `Relation Choice` carries a catalog vocabulary and no widget, so in edit it reaches the `choices` step and you pick the related item with a select. Placing the factory where edit has it made views render the same select instead of a link — Coresandbox's *Test Relationship Field* caught it. `getWidget` had been giving the factory that precedence implicitly, by mapping it to a widget name before anything else looked at the field. A widget named by `frontendOptions` or by the schema still wins over the factory.

With that, nothing in Volto needs `getWidget`. It stays exported and untouched for add-ons that import it, marked `@deprecated`, with its eleven tests still covering its behavior.

### Widget props reach the view widget

`FieldView` hands the field definition to the widget as props, the way the edit side does, with `value` applied last so nothing the schema carries can shadow it. Every widget in `theme/Widgets/` destructures its props explicitly and none spreads onto a DOM node, so the extra keys are inert for existing widgets.

### A `FieldsetView` component

The default content view and the default block view each carried their own copy of the same schema-driven rendering, and the two had drifted. They now share `components/theme/FieldsetView`, which removes 58 lines of duplicated JSX and gives the widget decision a single home.

Where the two copies disagreed, they converge on the better behavior: the content view gains the error boundaries the block view already had, the block view gains the `title` attribute on field labels, and the title field is covered by the boundary in both.

## Behavior changes worth review

- Fields named `title` and `description` now resolve through `views.id`. Same widgets as before, reached differently.
- `views.factory` is new. A field whose `factory` is `Choice`, `File`, `Image`, `Relation Choice` or `Relation List` resolves through it — previously it depended on `getWidget` mapping the factory to a widget name.
- A field whose declared `widget` is registered now renders with that widget even when its factory is in `views.factory`. Before, `getWidget` let the factory win over the declared widget.
- View widgets now receive the whole field definition as props, not only `value`.
- `WidgetsConfigViews` gains a **required** `getWidget` (it was declared but commented out) and an **optional** `factory`. Anyone constructing a `WidgetsConfigViews` object literal in TypeScript will need `getWidget`.

## Tests

85 tests across the touched modules, on top of the existing suite.

Each piece of the fix was reverted in turn to confirm the suite catches it:

| Reverted | Tests failing |
| --- | --- |
| hint not applied at all | 7 |
| displaced widget not kept | 2 |
| no `_widget` fallback step | 1 |
| no factory step | 2 |
| `widgetProps` never reach the widget | 1 |
| `title` missing from `views.id` | 1 |
| factory consulted after `choices` again | 4 |

The `social_links` case from the issue is covered end to end against the real registry, both with the add-on widget registered and without it. So is the relation case from Coresandbox: a `Relation Choice` and a `Relation List` carrying a catalog vocabulary, resolved against the real registry, render as relations and not as selects.

## Verification

- `yarn vitest run --project volto`: 339 files, 1606 passed, 1 skipped, 0 failures.
- `yarn check:ts`: 12 errors, the unchanged pre-existing baseline, none from the touched files.
- `yarn eslint --max-warnings=0` and `yarn prettier --check`: clean.
- `make docs-html`: build succeeded, 9 warnings, all pre-existing and none from the changed page.

## Documentation

`docs/source/development/widget.md` gains a section on the widget used to view a field: where to register it, that the edit and view registries are separate, and the resolution order as a table, including where it differs from edit and why.


sneridagh (approved):
LGTM! Great work!

